PR #10, opened 2018-02-19
KPBMS-685: Add basic input component for BLKN project

jaroslav-kubicek:


blkn/src/Input/index.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import * as React from "react";
+import Typography from "./../Typography";
+
+type Props = {
+  label?: string,
+  error?: string,
+  value?: string,
```
oreqizer: I don't think `value` and `onChange` should be optional

blkn/src/Input/index.js
```diff
@@ -0,0 +1,46 @@
...
+  onChange?: (SyntethicInputEvent<HTMLInputElement>) => any,
+};
+
+const Input = (props: Props) => (
+  <div className="root">
```
oreqizer: why a class if it is not used in the CSS?

blkn/src/Input/index.js
```diff
@@ -0,0 +1,46 @@
...
+      <input value={props.value} onChange={props.onChange} />
+    </label>
+    {props.error && (
+      <div>
+        <Typography size="small" type="error">
+          {props.error}
+        </Typography>
+      </div>
+    )}
+    <style jsx>{`
```
oreqizer: shouldn't some of the values be somehow centralized, such as `font-size`, colors, `line-height` etc?

blkn/src/Typography/index.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+
+import * as React from "react";
+
+const getSize = size => {
```
oreqizer: wouldn't objects make more sense, especially if you require the `size` prop?

blkn/src/Typography/index.js
```diff
@@ -0,0 +1,50 @@
...
+      return "#171B1E";
+    case "error":
+      return "#D02228";
+    default:
+      return "#46515D";
+  }
+};
+
+type Props = {
+  size?: "large" | "small" | "normal",
```
oreqizer: a required prop with a default value makes more sense imo

blkn/src/Input/index.js
jaroslav-kubicek: that's definitely worth a thought, I'm just not decided yet

jakubzaba: Will be optimized when it's needed - with more components.

PR #11, opened 2018-02-20
Improve testing environment for whole repo

jakubzaba:
Add circleci
Move enzyme into root package.json
Add testing scripts into root package.json
Turn off some eslint rules


.eslintrc
```diff
@@ -9,10 +13,24 @@
   ],
   "plugins": [
     "flowtype",
-    "prettier"
+    "prettier",
+    "jest"
   ],
   "rules": {
-    "prettier/prettier": "error",
+    "prettier/prettier": [
```
oreqizer: nope, use `.prettierrc`. also, the only rules that make sense configuring are the ones already present

.eslintrc
```diff
@@ -9,10 +13,24 @@
...
+      "error",
+      {
+        "bracketSpacing": true,
+        "printWidth": 80,
+        "singleQuote": false,
+        "tabWidth": 2,
+        "trailingComma": "all"
+      }
+    ],
+    "import/no-extraneous-dependencies": "off",
```
oreqizer: nope

.eslintrc
```diff
@@ -9,10 +13,24 @@
...
+    "react/jsx-filename-extension": "off",
```
oreqizer: why?

.eslintrc
```diff
@@ -9,10 +13,24 @@
...
+    "react/require-default-props": "off",
```
oreqizer: why?

.eslintrc
```diff
@@ -9,10 +13,24 @@
...
+    "jsx-a11y/label-has-for": "off",
```
oreqizer: why?

.eslintrc
Fallion: I agree with @oreqizer . If this is causing issues in tests then use a glob pattern to fix it. https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

.eslintrc
jaroslav-kubicek: I would agree with turning this off, I don't see any reason to make difference whether file has some JSX or not, it's like prefixing all string variables with `s`...

.eslintrc
jakubzaba: It's not working for yarn workspaces. Globs didn't helped for all cases.
https://github.com/benmosher/eslint-plugin-import/issues/458#issuecomment-361352578

.eslintrc
jakubzaba: Changed to `.js` only. I don't see benefit to use `.jsx`.

.eslintrc
jakubzaba: removed

.eslintrc
jakubzaba: Added comment. Label is wrapping input, no need for `for`.

.eslintrc
jakubzaba: Reverted

jakubzaba:
@oreqizer pls take a look on changes.

jakubzaba:
To push this forward I suggest to merge it. All comments were fixed or answered.
@Fallion @oreqizer

.eslintrc
```diff
@@ -9,10 +13,14 @@
   ],
   "plugins": [
     "flowtype",
-    "prettier"
+    "prettier",
+    "jest"
   ],
   "rules": {
     "prettier/prettier": "error",
+    "import/no-extraneous-dependencies": "off",
```
oreqizer: no

.eslintrc
oreqizer: well JSX is not a valid JS syntax, that's a pretty simple reason. whatever, you guys are the majority

.eslintrc
oreqizer: so turn off where needed

oreqizer:
sorry for the delay, guys. yesterday was hectic

jakubzaba:
I added `devDependencies` for `import/no-extraneous-dependencies` and `react/require-default-props` because workaround for Flow would be insane.

oreqizer:
The point about working around Flow is invalid IMO, but OK, requiring all props to be defaulted would be daunting I guess

PR #28, opened 2018-02-26
Kiwi-ui Icons

Fallion:

This Pull Request adds Icons to Kiwi-ui. Including the build process and publishing settings.


VaclavSir:
Note that I removed airlines logos from the set, because they need to be handled differently.

Fallion:
@VaclavSir That's fine. Icons will be adjusted in the future according to UX.

PR #31, opened 2018-02-27
Add Select component

### LINKED ISSUE #17: Select box
- [x] Select box can be optionally preceded with a label on top.
- [x] Optional placeholder prop, which sets the select to blank.
- [x] Optional function passed as props to be called `onChange`.
- [x] Each option in the select can have a `disabled` attribute -> Doesn't render it or greys it out.
- [x] Error reporting http://zpl.io/2yBmG3o
  - [x] Red border around box
  - [x] Error message red font below box. Smaller font. 
- [x] Select label has DarkGray color if no value is selected, otherwise LightGray
- [x] Select fontColor is MiddleGray color if no value is selected, otherwise DarkGray.
- [x] Replace native dropdown icon with `@kiwicom/icons/ChevronUp` icon.
- [x] Add tests.
  - [x] Test `onChange` callback.
  - [x] Test rendering via jest snapshots.  

Example:
image

Error reporting:
image

git-toni:
Closes #17

### DIFF
--- a/blkn/src/Select/index.js
+++ b/blkn/src/Select/index.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from "react";
 import css from "styled-jsx/css";
-import { ChevronUp } from "@kiwicom/icons";
+import ChevronDown from "react-icons/lib/fa/chevron-down";
 import classnames from "classnames";
 import Typography from "./../Typography";
 
@@ -132,7 +132,7 @@ class Select extends React.Component<Props, State> {
             {options.map(this.renderOption)}
           </select>
           <span className="dropdown-icon">
-            <ChevronUp fill={DARKGRAY_COLOR} height={20} />
+            <ChevronDown fill={DARKGRAY_COLOR} height={20} />
           </span>
         </label>
         {this.feedbackLine()}
--- a/blkn/src/Select/index.js
+++ b/blkn/src/Select/index.js
@@ -4,11 +4,8 @@ import css from "styled-jsx/css";
 import ChevronDown from "react-icons/lib/fa/chevron-down";
 import classnames from "classnames";
 import Typography from "./../Typography";
+import { fontColors } from "../constants";
 
-const ERROR_COLOR = "#d21c1c";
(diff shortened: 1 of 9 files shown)


blkn/src/Select/index.js
```diff
@@ -0,0 +1,145 @@
...
+  }
+  select {
+    font-size: 16px;
+    border-radius: 3px;
+    background-color: ${WHITE_COLOR};
+    border: solid 1px ${LIGHTGRAY_COLOR};
+    line-height: 1.25;
+    padding: 12px 16px;
+    color: ${DARKGRAY_COLOR};
+    -moz-appearance: none;
```
jaroslav-kubicek: these prefix variants are not done automatically by `styled-jsx`?

blkn/src/Typography/index.js
```diff
@@ -19,6 +19,7 @@ type Props = {
   size: "large" | "small" | "normal",
   type: "primary" | "secondary" | "attention" | "error",
   children: React.Node,
+  color?: string,
```
jaroslav-kubicek: is it desired to introduce colour? I thought all possible states should be achievable by combination of `type` and `variant` https://zpl.io/VxBkGZW

jaroslav-kubicek: I just created PR to fix colors 🙂

git-toni: How do you make use of `variant` with the `Typography` component?
I think it's good to have a handy set of colors `primary`, `secondary`, etc and maximize its use. But imho having an optional `color` prop makes the `Typography` component much more re-usable at a very low cost. (*End of salespitch*) :smile:

blkn/src/Select/index.js
git-toni: I couldn't find it in the documentation, but just confirmed it does by trying it. :+1:

blkn/src/Typography/index.js
Fallion: We'll be bundling Theme support in the future. #35

jaroslav-kubicek:
there is conflict, cc @git-toni

PR #37, opened 2018-02-27
BLKN: button 

prztrz:


blkn/src/Button/__tests__/icon.test.js
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+import Enzyme, { mount } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import { Money } from "@kiwicom/icons";
+import Icon from "../Icon";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe("Icon", () => {
+  //   const icon = jest.fn();
```
FilipMessa: Why is this code commented? If we don't need it delete it.

blkn/src/Button/__tests__/index.test.js
```diff
@@ -0,0 +1,38 @@
...
+    const button = component.find("button");
+    expect(button.render().text()).toBe(title);
+  });
+
+  describe("Rendered with icon", () => {
+    const icon = jest.fn();
+    component = shallow(<Button title={title} icon={icon} />);
+    const button = component.find("button");
+    it("Should contain an Icon", () => {
+      //   expect(button.find("Icon").exists()).toBe(true);
```
FilipMessa: Why is this code commented? If we don't need it delete it.

blkn/src/Button/__tests__/index.test.js
```diff
@@ -0,0 +1,38 @@
+import * as React from "react";
+import Enzyme, { shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import Button from "../";
+
+Enzyme.configure({ adapter: new Adapter() });
```
FilipMessa: Just for my information what is this `Adapter`  for? I never used it before thanks.

blkn/src/Button/index.js
```diff
@@ -1,22 +1,55 @@
 // @flow
 import * as React from "react";
+import Icon from "./Icon";
 
 type Props = {
   title: string,
-  onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
+  onClick?: (e: SyntheticEvent<HTMLButtonElement>) => void,
```
FilipMessa: I think `onClick` is not optional you always want from button be clickable.

blkn/src/Button/index.js
```diff
@@ -1,22 +1,55 @@
...
+  large: "16px",
+  small: "14px",
+};
+
+const bgColors = {
+  primary: "#00a991",
+  secondary: "#e8edf1",
+};
+
+const colors = {
```
FilipMessa: there is already `constants.js` file in master so you should use it instead of hardcoded.  Same for bgColors this colors should be in `constants.js`

blkn/src/Button/__tests__/index.test.js
prztrz: Well it turned out it was configured already in separated file  `jest.json` and it is not necessary. Removed it.

prztrz:
I am closing this due to [https://github.com/kiwicom/react-kiwi-ui/pull/49](49)

PR #49, opened 2018-03-01
BLKN: 37 button

prztrz:

Closes #37

### DIFF
--- a/blkn/src/Button/IconWrapper.js
+++ b/blkn/src/Button/IconWrapper.js
@@ -1,0 +1,40 @@
+// @flow
+import * as React from "react";
+import { colors } from "../constants";
+
+const sizes = {
+  normal: "20px",
+  small: "20px",
+  large: "26px",
+};
+
+const fillColors = {
+  primary: colors.white,
+  secondary: colors.shuttle,
+};
+
+type Props = {
+  size: $Keys<typeof sizes>,
+  type: $Keys<typeof fillColors>,
+  icon: React.StatelessFunctionalComponent<*>,
+};
+
+const IconWrapper = (props: Props) => {
+  const { icon: Icon, type, size } = props;
+
+  return (
+    <span>
+      <Icon fill={fillColors[type]} height={sizes[size]} />
+      <style xml>{`
+        svg {
+          position: absolute;
+          top: 50%;
+          left: 12px;
+          transform: translateY(-50%);
+        }
+      `}</style>
+    </span>
+  );
+};
+
+export default IconWrapper;
--- a/blkn/src/Button/__tests__/index.test.js
+++ b/blkn/src/Button/__tests__/index.test.js
@@ -1,0 +1,41 @@
+import * as React from "react";
+import { shallow } from "enzyme";
+import Button from "../";
+
+describe("Button", () => {
+  const title = "Title";
+  const component = shallow(<Button title={title} />);
+  const button = component.find("button");
+  it("Should contain a title ", () => {
+    expect(button.render().text()).toBe(title);
+  });
+});
+
+describe("When button is clicked", () => {
+  const onClick = jest.fn();
+  const title = "title";
+  const component = shallow(<Button title={title} onClick={onClick} />);
+  const button = component.find("button");
+<<<<<<< HEAD
+  beforeEach(() => {
+    button.simulate("click");
+  });
+  it("should execute onClick method", () => {
+=======
+
+  it("should execute onClick method", () => {
+    button.simulate("click");
+>>>>>>> 2464380... test cleared
+    expect(onClick).toHaveBeenCalled();
+  });
+});
+
+describe("Rendered with icon", () => {
+  const title = "title";
+  const icon = jest.fn();
+  const component = shallow(<Button title={title} icon={icon} />);
+  const button = component.find("button");
+  it("Should contain IconWrapper", () => {
+    expect(button.find("IconWrapper").exists()).toBe(true);
+  });
+});
--- a/blkn/src/Button/index.js
+++ b/blkn/src/Button/index.js
@@ -1,22 +1,50 @@
 // @flow
 import * as React from "react";
+import IconWrapper from "./IconWrapper";
+import { colors, fontSizes } from "../constants";
 
 type Props = {
   title: string,
   onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
+  isDisabled?: boolean,
+  type: "primary" | "secondary",
+  size: "normal" | "large" | "small",
+  icon?: React.StatelessFunctionalComponent<*>,
 };
 
-const Button = ({ title, onClick }: Props) => (
-  <button onClick={onClick}>
-    {title}
+const fontColors = {
+  primary: colors.white,
+  secondary: colors.shuttle,
+};
+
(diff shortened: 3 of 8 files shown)


blkn/src/index.js
```diff
@@ -1,10 +1,9 @@
 // @flow
 
 export { default as Button } from "./Button";
+export { default as IconWrapper } from "./Button/IconWrapper";
```
FilipMessa: Don't export this one, it is component just for Button.

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import * as React from "react";
+import { colors } from "../constants";
+
+type Props = {
+  size: "normal" | "small" | "large",
+  type: "primary" | "secondary",
+  icon: Function,
```
oreqizer: nope, too vague. be specific

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+  large: "26px",
+};
+
+const fillColors = {
+  primary: colors.white,
+  secondary: colors.shuttle,
+};
+
+const IconWrapper = (props: Props) => {
+  const { icon: Icon, type, size } = props;
```
oreqizer: if `Icon` is a class and not an element, make the prop itself PascalCase

blkn/src/Button/__tests__/index.test.js
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+import { shallow } from "enzyme";
+import Button from "../";
+
+describe("Button", () => {
```
oreqizer: this whole thing is unnecessarily complicated.. use local variables instead of reassigning, remove the useless before each, don't nest describes so much

blkn/src/Button/index.js
```diff
@@ -1,22 +1,50 @@
...
+import IconWrapper from "./IconWrapper";
+import { colors, fontSizes } from "../constants";
 
 type Props = {
   title: string,
   onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
+  isDisabled?: boolean,
+  type: "primary" | "secondary",
+  size: "normal" | "large" | "small",
+  icon?: Function,
```
oreqizer: same thing as with the above `icon` prop

blkn/src/constants.js
```diff
@@ -30,3 +30,9 @@ export const fontColors = {
   error: colors.red,
   input: colors.casper,
 };
+
+export const fontSizes = {
```
oreqizer: 😄 at least sort it by size or something

blkn/src/Button/IconWrapper.js
VaclavSir: The right type here would be something like `React.ComponentType<{ height: number, fill: string }>`. But I'm not so sure if we want to use this pattern (`<SomeComponent icon={SomeIcon} />`), it seems unusual.

[In the Header component](https://github.com/kiwicom/react-kiwi-ui/blob/master/blkn/src/Container/Header.js), we're passing a `React.Node` (like `<SomeComponent icon={<SomeIcon height={32} />} />`), but that's not applicable here, because the size and color are concerns of the button.

Maybe we can use a render prop for this icon (like `<SomeComponent icon={(size, color) => <SomeIcon height={size} fill={color}) />`), what do you think @oreqizer about this?

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import * as React from "react";
+import { colors } from "../constants";
+
+type Props = {
+  size: "normal" | "small" | "large",
+  type: "primary" | "secondary",
```
VaclavSir: I would use the `$Keys<T>` type for this (see [Flow utility types](https://flow.org/en/docs/types/utilities/)):

```
size: $Keys<typeof sizes>,
type: $Keys<typeof fillColors>,
```

blkn/src/constants.js
VaclavSir: This was copied from an older version of the Typography component. :-)

I would rather see the Typography component used in the Button.

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+  primary: colors.white,
+  secondary: colors.shuttle,
+};
+
+const IconWrapper = (props: Props) => {
+  const { icon: Icon, type, size } = props;
+
+  return (
+    <span>
+      <Icon fill={fillColors[type]} height={sizes[size]} width={sizes[size]} />
```
VaclavSir: It's enough to set just the `fill` and `height`, no need to set the `width`.

blkn/src/Button/IconWrapper.js
prztrz: It does not work when prop is capitalized.

blkn/src/constants.js
prztrz: Well first i was instructed to not import other components, so i could create the copy of typography component just for this one purpose but I see no reason tbh

FilipMessa: It must be misunderstanding, I didn't instruct you don't import other components.I meant don't reuse IconWrapper as the universal component. Because in my opinion, IconWrapper is just helper component for Button. If you try It makes so universal it will be quickly overwhelmed.

blkn/src/Button/IconWrapper.js
FilipMessa: I also stopped at this 'unusual' patter, I was considering render props and compound component pattern. But I like it because of it less verbose than others.

blkn/src/constants.js
prztrz: ok . I will use ```Typography``` then :)

blkn/src/Button/__tests__/index.test.js
prztrz: I simplified the tests.

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+
+const fillColors = {
+  primary: colors.white,
+  secondary: colors.shuttle,
+};
+
+type Props = {
+  size: $Keys<typeof sizes>,
+  type: $Keys<typeof fillColors>,
+  icon: React.StatelessFunctionalComponent<*>,
```
oreqizer: `Icon: React.ComponentType<any>`

blkn/src/Button/IconWrapper.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const IconWrapper = (props: Props) => {
+  const { icon: Icon, type, size } = props;
+
+  return (
+    <span>
```
oreqizer: is the `<span>` actually useful? couldn't it be simply `<React.Fragment>`?

blkn/src/Button/IconWrapper.js
prztrz: Tried that already before. Doesn't work...

oreqizer: plz elaborate. `React.StatelessFunctionalComponent<any>` is a subtype of `React.ComponentType<any>`, it _has_ to work

PR #59, opened 2018-03-02
Icons: Add readme and preview

VaclavSir:
Important changes are in `readme.md` and `build.js` only, everything else is generated.

---


git-toni:
What about adding a preview for the actual React components, using Storybook?

VaclavSir:
I think that if we add a storybook to the icons package, nobody will ever run it. And I don't think we want to add icons to the storybook of balkan-ui package, because it is a different package.

git-toni:
@VaclavSir Fair enough

PR #63, opened 2018-03-05
relocate components stories for the storybook.

### LINKED ISSUE #15: Use .stories instead of /stories
Minor folder structure fix in order to keep components and all their files together.

FilipMessa:
Close #15


blkn/src/Button/Button.stories.js
```diff
@@ -0,0 +1,80 @@
...
+      "Type",
+      {
+        primary: "primary",
+        secondary: "secondary",
+      },
+      "primary",
+    );
+    const label = text("Label", "Hello Button");
+    return {
+      info: "Examples of all possible buttons, you can eazy change button type and state in knobs.",
```
Fallion: `you can easily change`

PR #67, opened 2018-03-07
Inline feedback line

### LINKED ISSUE #27: General inline feedback line
As raised in [this discussion](https://github.com/kiwicom/react-kiwi-ui/pull/22/files/5d2c8202c97cbe286d8776abf693c7837af6a51a#diff-952da1b6189f08337eae8f7e0bc74ade), having a feedback component for form fields would avoid code duplication. This would be useful for components such as `Input`, `Select`, `Checkbox`, etc.
We should decide on a name and path for the component. Bootstrap [uses](https://v4-alpha.getbootstrap.com/components/forms/#validation) `form-control-feedback` classes for these. 
Suggestions: 

- :one: `src/Feedback/FieldFeedback.js`
- :two: `src/Hints/FieldHint.js`

Any other ideas?

image

image

image

git-toni:
Closes #27 

* [X] Tests have been added/adjusted for my new feature
* [X] New Components are registered in index.js of my project


blkn/src/FieldFeedback/index.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import * as React from "react";
+import Typography from "./../Typography";
+
+type Props = {
+  error?: string,
+  help?: string,
```
VaclavSir: I would use `React.Node` here. According to the design, the text may be formatted: https://zpl.io/VDDzogO

git-toni: That's a great way to solve it, thanks.

Accoding to zeplin mockups `error`s are not formatted in any case. But maybe we could also set it as `React.Node` to be on the safe side?

VaclavSir: Agree, let's make it both React.Node to make it consistent.

PR #99, opened 2018-03-19
add screenshot testing of Storybook

### LINKED ISSUE #75: Add storybook-chrome-screenshot

From discussion on #68. It has been decided to use this library instead to test that all stories keep working.

Fallion:
This PR should create a discussion, because it's apparent that it is time consuming and it looks like we will have to create stories specifically for it.

What works:
- Screenshots (duh)
- Can be extended to test specific screen sizes (this is great)

What doesn't:
- Stories that have scrolling are not fully screenshot

What is annoying:
- Having to run a time consuming script before commiting (This would have to be a pre-commit hook)
- Having to add a configuration to each story.

For base level check of rendering I feel like we should revisit Storyshots. 

Closes #75


Fallion:
Closing this. This solution is not good. We will test other solutions.

Fallion:
Relevant discussion https://skypicker.slack.com/archives/C9B1H0ACW/p1521477906000163

PetrSnobelt:
Closing an issue with a link to the relevant issue will be definitely more helpful and friendly then without it 😉  

I don't think CI should commit images, they only should compare it with previous and if it didn't match do not allow to merge - similar to jest snapshots.

Fallion:
> I don't think CI should commit images, they only should compare it with previous and if it didn't match do not allow to merge - similar to jest snapshots.

For this you need either a pre-commit hook or the CI to add them.

PetrSnobelt:
I imagine it like this:
initially
- you run `yarn screenshots`, which can take a while and generate all screenshots
- You commit screenshots to repo
- ci run `yarn screenshots:check` and compare screenshots, it should be same so it allows merge

then
- you change something 
- commit without screenshots
- ci runs `yarn screenshots:check` and compare screenshots, if it is different it do not allow merge

so if anything breaks you need to run `yarn screenshots`, compare screenshots and decide if it is OK - commit changed screenshots or not - you have to fix the code

PR #104, opened 2018-03-20
Added value prop to checkbox

### LINKED ISSUE #103: Add prop value to Checkbox component

prztrz:
closes #103 

### DIFF
--- a/src/Checkbox/index.js
+++ b/src/Checkbox/index.js
@@ -8,6 +8,7 @@ import Icon from "./Icon";
 type Props = {
   label: string,
   checked?: boolean,
+  value?: string,
   onChange: (SyntheticInputEvent<HTMLInputElement>) => any,
 };
 
@@ -27,6 +28,7 @@ const Checkbox = (props: Props) => (
         type="checkbox"
         checked={props.checked}
         onChange={props.onChange}
+        value={props.value}
       />
     </label>
     <style jsx>{`


git-toni:
Shouldn't we add a story to `Checkbox.stories.js` to be able to see the effects of the new `value` prop? Also I realized the component doesn't have a unit test? :)

prztrz:
@git-toni `value` is to faciliate managing parent form component `state` but there won't be any effect visible for users. 

This is the idea:

```
// parent component state
this.state = {
    checkedOptions:  []
}

// now we can use value of Checkbox:
handleCheck = e => {

    if (this.state.checkedOptions.indexOf(e.target.value > -1) {

        const index = this.state.checkedOptions.indexOf(e.target.value)
        const checkedOptions = [...this.state.checkedOptions.filter(option => option !== e.target.value)];
        this.setState({ checkedOptions })
    }  else {
        const checkedOptions = [...this.state.checkedOptions.slice(), e.target.value];
        this.setState({ checkedOptions })
    }
}

// render Checkbox

<Checkbox 
    onChange={this.handleCheck} 
    checked={this.state.checkedOptions.indexOf("something"} > -1} 
    value="something" 
/>
```

therefore I did not modify the story.

VaclavSir:
I created another issue for the unit test: https://github.com/kiwicom/orbit-components/issues/108

VaclavSir:
I don't like this idea. The `value` on a checkbox doesn't do anything, it's just using the DOM as a store of information. Something like a `data-` attribute. They were handy back then in jQuery times, but I wouldn't use them in React.

If you need a button or checkbox to pass a value on click, you should wrap it in another component that does that:
```js
type Props = {
  value: string,
  checked: boolean,
  onChange: (value: string, checked: boolean) => void,
};

class Option extends React.Component<Props> {
  handleChange = (event) => {
    this.props.onChange(this.props.value, event.target.checked);
  }

  render() {
    return <Checkbox
      onChange={this.handleChange}
      checked={this.props.checked}
    />
  }
}

// And then use:
const value = 'something';
return <Option
  value={value}
  checked={this.state.checkedOptions.includes(value)}
  onChange={(value, checked) => { console.log(value, checked); }}
/>
```

What does @jaroslav-kubicek think?

prztrz:
What is a benefit of using additional component?

VaclavSir:
React documentation recommends using arrow function or bind: https://reactjs.org/docs/handling-events.html#passing-arguments-to-event-handlers

But if you do that, it will create a new function on every render and it can cause performance issues. The additional component prevents that.

michalsanger:
I recommend arrow functions in class props https://michalzalecki.com/react-components-and-class-properties/#bounded-methods

michalsanger:
I don't see the point. `value` is valid prop of the HTML checkbox, no need to force wrapping it with other component. The `onChange` prop also makes sense to me to be a part of checkbox.

jaroslav-kubicek:
LGTM, I could also imagine some little tweak in future, e.g. having `onChange` with signature `onChange(e: SyntheticInputEvent<HTMLInputElement>, value: null | string)` so you don't need to `e.target.value` every damn time

VaclavSir:
I still don't like the idea of storing data in the DOM. I have a different idea - let's add this functionality to the Checkbox component itself, so you don't have to wrap it: https://github.com/kiwicom/orbit-components/compare/pass-data-to-event?diff=unified&name=pass-data-to-event

Wouldn't this be better? We can extend all our inputs of this functionality.

prztrz:
I really do not get how is it better than using `value` on `checkbox`, and in your solution is it even necessary to pass `event` to `onChange` method?

VaclavSir:
The event is passed because it holds the information whether it was checked or unchecked. We can pass just that (`onChange(checked: boolean, Data) => void`), but it would be a breaking change.

michalsanger:
@VaclavSir Is the wrapper solving any real problem? I'd prefer to KISS.

VaclavSir:
I think using DOM for model data is conceptually wrong. I googled it and there are people who agree with me and people who disagree. So if you're all OK with that, I'll keep my opinion on this subject to myself. 🤷‍♂️ 

But at least please use `name` attribute instead of `value` - it's more an identifier than a value and we can use the same approach with other inputs (where the `value` attribute is used for the input value).

michalsanger:
@VaclavSir can't we just go with HTML spec? There is `name` and `value` prop for the checkbox both have it's meaning.

VaclavSir:
OK, whatever.

PR #111, opened 2018-03-28
Theming using theming package

PetrSnobelt:

Sample component which is used for storybook tests is not exported


src/Theming/ThemeSwitcher.js
```diff
@@ -0,0 +1,53 @@
...
+  children: any,
+  mainTheme?: typeof defaultTheme,
+  themes: any[],
+};
+
+type ThemeState = {
+  themeIndex: number,
+};
+
+export default class ThemeSwitherSample extends React.Component<ThemeProps, ThemeState> {
```
jaroslav-kubicek: I would suggest to have the same name for default export as for the filename

src/Theming/ThemeSwitcher.js
```diff
@@ -0,0 +1,53 @@
+// @flow
+import * as React from "react";
+import defaultTheme from "orbit-design-token";
+
+import ThemeProvider from "./ThemeProvider";
+
+type ThemeProps = {
+  children: any,
```
jaroslav-kubicek: children is not definitely any :)

infi-pc: nice catch. just hint for better code reviews, if you know the solution it is good to give some hint. Something like:
Don't use `any` for children, React children have type `Node` see https://flow.org/en/docs/react/children/

```
import type { Node } from "react"
...
children: Node,
...
```

src/Theming/ThemeSwitcher.js
PetrSnobelt: OK

src/Theming/ThemeSwitcher.js
PetrSnobelt: OK, Fixed

tomashapl:
@PetrSnobelt please can you strict define, `defaultTheme` object to use it in flow type annotation

src/Theming/ThemeSwitcher.js
```diff
@@ -0,0 +1,53 @@
+// @flow
+import * as React from "react";
+import defaultTheme from "orbit-design-token";
+
+import ThemeProvider from "./ThemeProvider";
+
+type Props = {
+  children: React.Node,
+  mainTheme?: typeof defaultTheme,
+  themes: any[],
```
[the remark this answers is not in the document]
PetrSnobelt: 👍

PR #112, opened 2018-03-28
Update button

williamkolmacka:

Updated Button component with new types of props, added tokens to constants.js, replace old icons for the new ones.


config/build.js
```diff
@@ -53,8 +49,9 @@ names.forEach(async ({ inputFileName, outputComponentFileName, functionName }) =
     import * as React from "react";
  
     export default function ${functionName}(props) {
+      const { color, size } = props;
       return (
-        ${optimizedSvg.replace(/<svg(\s[^>]*)>/, "<svg$1 {...props}>")}
+        ${optimizedSvg.replace(/<svg(\s[^>]*)>/, "<svg$1 height={size} fill={color}>")}
```
Fallion: Why this limitation?

src/Button/index.js
```diff
@@ -2,54 +2,140 @@
 import * as React from "react";
 
 import IconWrapper from "./IconWrapper";
-import { colors, fontSizes } from "../constants";
+import Loading from "../icons/Loading";
+import { darken } from 'polished';
```
Fallion: Shouldn't these colours be part of the design token system without requiring an extra library?

config/build.js
```diff
@@ -38,13 +38,9 @@ const svgo = new SVGO();
 
 names.forEach(async ({ inputFileName, outputComponentFileName, functionName }) => {
   const dom = await JSDOM.fromFile(inputFileName);
-  const shapes = dom.window.document.querySelector("defs").innerHTML;
+  const shapes = dom.window.document.querySelector("svg").innerHTML;
```
Fallion: Not sure this is a good move. This is there for the structure that Sketch exports.

config/build.js
```diff
@@ -38,13 +38,9 @@ const svgo = new SVGO();
...
 
-  const svg = `<svg viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg" height="24" fill="#bac7d5">
-      <g>
-        ${shapes}
-      </g>
-    </svg>`;
+  const svg = `<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">${shapes}</svg>`;
```
Fallion: Same as above.

src/Button/index.js
williamkolmacka: I understand exactly what you mean, but it's easier to keep it like this. In future it can be helpful, f.e. We don't need to add a new colors for hover and active states whenever is added a new whitelabel.

config/build.js
williamkolmacka: Because we don't need to change more props on icon. We need only size of the icons and their colors, but if you think that's better pass whole props object, we can do it like that.

config/build.js
williamkolmacka: It's alright, because @darkwindcz is preparing every svg icon to the same structure as it's now. Sketch always export very messy svg with lot of unusable things there. That's the reason why I changed this file.

config/build.js
williamkolmacka: Same reason.

config/build.js
Fallion: It assigns any props that are passed. If you want to do specific ones then it would be better to prepare Flow types to be autogenerated with this information as well since it's a generated file.

PR #134, opened 2018-04-16
Add new heading component

williamkolmacka:

Add new heading component.


src/Heading/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+const sizeHeading = {
+  display: tokens.fontSizeHeadingDisplay,
+  large: tokens.fontSizeHeadingLarge,
+  medium: tokens.fontSizeHeadingMedium,
+  small: tokens.fontSizeHeadingSmall,
+};
+
+type Props = {
+  element: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "div",
```
jaroslav-kubicek: does it make sense to have both properties, `element` and `size`?  I mean, you could infer `element` from `size`, `<Heading size="small" element="h1">Heading</Heading>` is quite weird combination.

williamkolmacka: You're right, this case looks weird, but sometimes we want to use h2 element in large size or if you want to use div instead of h1, h2 etc. you just want to be able to choose the size anyway.

PR #139, opened 2018-04-17
Add new TextLink component

### LINKED ISSUE #92: Create <TextLink /> component
Description: Link is used usually in the paragraph text to show that something navigates you to another location.

## Parameters

Parameter | Values | Default
-- | -- | --
Style | Primary, Secondary | Primary
External | [boolean] | false

Note: When external is true, it opens links in new tab/window.

## Design tokens

**For all configurations:**
` font-weight: $fontWeightLinks`
  
**Based on selected configuration:**

Parameter:Value | property: $token
-- | --
Style:Primary | `color: $colorLinkPrimary`
Style:Secondary | `color: $colorLinkSecondary`<br>`text-decoration: underline` // token not needed, just visual change

## Hover behavior

There are two things that will happen when the user hovers any link:
 - Underline appears
 - The text color will change to the darker version of `$colorLinkPrimary` color (for both styles, actually). For darkening percentage use design token `$modifierDarkenHover`

## Other things to mention
-  If it’s possible, it should be used only inside <Text> or <Headline> components and it should inherit Size from parent text.
- We are showing pointer cursor for every link
- For the visited state of links, please use original link color (at least for now, it’s possible that this will change in future).
- For focus state please keep outline there, until we decide what to do next with visited and focus state.

## Some inspiration from other’s React components for heading component:
https://polaris.shopify.com/components/navigation/link
https://instacart.github.io/Snacks/#link-1
http://grommet.io/docs/anchor
http://design-system.pluralsight.com/components/link/
http://www.jetbrains.org/ring-ui/link.html

wEprous:

Add new TextLink component.

closes #92

### DIFF
--- a/src/TextLink/index.js
+++ b/src/TextLink/index.js
@@ -1,0 +1,64 @@
+// @flow
+import * as React from "react";
+import { darken } from "polished";
+import * as tokens from "orbit-design-token";
+
+type Props = {
+  title: string,
+  url: string,
+  onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
+  newTab: boolean,
+  type: "primary" | "secondary",
+  size: "small" | "normal" | "large",
+};
+
+const colorTextLink = {
+  primary: tokens.colorLinkPrimary,
+  secondary: tokens.colorLinkSecondary,
+};
+const fontSizeTextLink = {
+  small: tokens.fontSizeTextSmall,
+  normal: tokens.fontSizeTextNormal,
+  large: tokens.fontSizeTextLarge,
+};
+
+const TextLink = (props: Props) => {
+  const { title, url, onClick, newTab, type, size } = props;
+
+  return (
+    <a href={url} target={newTab ? "_blank" : undefined} onClick={onClick}>
+      {title}
+      <style jsx>{`
+        a {
+          font-size: ${fontSizeTextLink[size]};
+          color: ${colorTextLink[type]};
+          font-weight: ${tokens.fontWeightLinks};
+          text-decoration: ${type === "secondary"
+            ? "underline" // replace for token textDecorationLinkSecondary
+            : "none"}; // replace for token textDecorationPrimary
+          cursor: pointer;
+        }
+        a:hover {
+          text-decoration: ${type === "secondary"
+            ? "underline" // replace for token textDecorationLinkSecondaryHover
+            : "underline"}; // replace for token textDecorationLinkPrimaryHover
+          color: ${darken(tokens.modifierDarkenHover, tokens.colorLinkPrimary)};
+        }
+        a:visited {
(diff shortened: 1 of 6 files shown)


src/TextLink/index.js
```diff
@@ -0,0 +1,64 @@
+// @flow
+import * as React from "react";
+import { darken } from "polished";
+import * as tokens from "orbit-design-token";
+
+type Props = {
+  title: string,
+  url: string,
+  onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
+  newTab: boolean,
```
jaroslav-kubicek: `type`, `size` and `newTab` should be declared as optional props

src/TextLink/index.js
```diff
@@ -0,0 +1,64 @@
+// @flow
+import * as React from "react";
+import { darken } from "polished";
+import * as tokens from "orbit-design-token";
+
+type Props = {
+  title: string,
+  url: string,
+  onClick: (e: SyntheticEvent<HTMLButtonElement>) => void,
```
jaroslav-kubicek: accepting `e` param should be optional

src/TextLink/index.js
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <a href={url} target={newTab ? "_blank" : undefined} onClick={onClick}>
+      {title}
+      <style jsx>{`
+        a {
+          font-size: ${fontSizeTextLink[size]};
+          color: ${colorTextLink[type]};
+          font-weight: ${tokens.fontWeightLinks};
+          text-decoration: ${type === "secondary"
+            ? "underline" // replace for token textDecorationLinkSecondary
```
jaroslav-kubicek: is this some kind of forgotten TODO?

src/TextLink/index.js
wEprous: Changing `type` and `size` will results in flow error `Computed property cannot be accessed with undefined` because these props are used as keys, e.g. `${fontSizeTextLink[size]}`

oreqizer: @jaroslav-kubicek they should not, see https://flow.org/en/docs/react/components/#toc-using-default-props-for-functional-components

src/TextLink/index.js
oreqizer: `(SyntheticEvent<HTMLButtonElement>) => any`

src/TextLink/index.js
wEprous: Will fix all TODO tokens.

src/TextLink/index.js
wEprous: OK

src/TextLink/index.js
wEprous: Leaving as it is.

Issue #91, opened 2018-03-17
Create <Text /> component


darkwindcz:
**Description**: The most basic component for rendering text blocks.

### Visual look:
![body-text](https://user-images.githubusercontent.com/6851247/37559090-76129e4c-2a28-11e8-9531-0c615dc705e7.png)

We should be able to render all of these styles with these parameters:

Parameter | Values | Default
-- | -- | --
Style | Primary, Secondary, Attention | Primary
Size | Large, Normal, Small | Normal
Weight | bold, normal | normal
Italic | [boolean] | false
Uppercase | [boolean] | false
Element | p, span, div, (anything else?) | p

## Design tokens

**For all configurations:**
  font-family: $fontFamily
  line-height: $lineHeightText

**Based on selected configuration:**

Parameter:Value | property: $token
-- | --
Style:Primary [optional] | color: $colorTextPrimary
Style:Secondary [optional] | color: $colorTextSecondary
Style:Attention [optional] | color: $colorTextAttention
Size:Large [optional] | font-size: $fontSizeTextLarge
Size:Normal [optional] | font-size: $fontSizeTextNormal
Size:Small [optional] | font-size: $fontSizeTextSmall
Weight:normal [optional] | font-weight: $fontWeightNormal
Weight:bold [optional] | font-weight: $fontWeightBold

Italic and uppercase don't need tokens as they aren’t expressed with numeric values.

**Possible examples of configuration:**

    <Text>A) <strong>We're here for you.</strong> First, let's narrow down your request.</Text>
    <Text Element="div" Italic>B) We're here for you. First, let's narrow down your request.</Text>

**Rendered HTML on those configurations:**

    <p>A) <strong>We're here for you.</strong> First, let's narrow down your request.</p>
    <div><em>B) We're here for you.</strong> First, let's narrow down your request.</em></div>

Note: Please take this just as my component interpretation for better understanding how I meant which param. I am totally not sure about `<strong>` though, but it should be possible to make just part of the text bold (or link in other cases). I don’t have to include “real” code, just give me a feedback how you want to specify it and I can do it your way. I won’t be able to add these code examples to more complicated components anyway.

## Some inspiration from others React components for solving text component:
We probably don't need to reinvent the wheel, a lot of other teams solved these basics already. I have a good collection of links with some inspiration from other design systems so I can include for every common component so we have something to study and inspire from 🙂​Hope it helps a little! I think that it should be possible to check their Github repositories, some of them can have them public.

[...]

PR #143, opened 2018-04-18
Add new Text component

williamkolmacka:

Closes: #91

### DIFF
--- a/src/Text/Text.stories.js
+++ b/src/Text/Text.stories.js
@@ -80,8 +80,8 @@ storiesOf("Text", module)
                   size={size}
                   weight={weight}
                   align={align}
-                  isUppercase={boolean("Uppercase", false)}
-                  isItalic={boolean("Italic", false)}
+                  uppercase={boolean("Uppercase", false)}
+                  italic={boolean("Italic", false)}
                 >
                   {customText}
                 </Text>
@@ -152,8 +152,8 @@ storiesOf("Text", module)
                   size={size}
                   weight={weight}
                   align={align}
-                  isUppercase={boolean("Uppercase", false)}
-                  isItalic={boolean("Italic", false)}
+                  uppercase={boolean("Uppercase", false)}
+                  italic={boolean("Italic", false)}
                 >
                   {customText}
                 </Text>
@@ -224,8 +224,8 @@ storiesOf("Text", module)
                   size={size}
                   weight={weight}
                   align={align}
-                  isUppercase={boolean("Uppercase", false)}
-                  isItalic={boolean("Italic", false)}
+                  uppercase={boolean("Uppercase", false)}
+                  italic={boolean("Italic", false)}
                 >
                   {customText}
                 </Text>
--- a/src/Text/index.js
+++ b/src/Text/index.js
@@ -24,11 +24,11 @@ type Props = {
   size: $Keys<typeof sizeText>,
   weight: $Keys<typeof weightText>,
   align: "left" | "center" | "right",
-  italic?: boolean,
-  uppercase?: boolean,
+  italic: boolean,
+  uppercase: boolean,
   element: "p" | "span" | "div",
-  className?: string,
   children: React.Node,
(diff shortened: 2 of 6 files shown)


src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+type Props = {
+  type: $Keys<typeof colorText>,
+  size: $Keys<typeof sizeText>,
+  weight: "bold" | "regular",
+  align: "left" | "center" | "right",
+  isItalic?: boolean,
+  isUppercase?: boolean,
+  element: "p" | "span" | "div",
+  className?: string | void,
```
tomashapl: className as function ?

src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+type Props = {
+  type: $Keys<typeof colorText>,
+  size: $Keys<typeof sizeText>,
+  weight: "bold" | "regular",
+  align: "left" | "center" | "right",
+  isItalic?: boolean,
+  isUppercase?: boolean,
+  element: "p" | "span" | "div",
```
tomashapl: is element in this component necessary ?

src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+const sizeText = {
+  large: tokens.fontSizeTextLarge,
+  normal: tokens.fontSizeTextNormal,
+  small: tokens.fontSizeTextSmall,
+};
+
+type Props = {
+  type: $Keys<typeof colorText>,
+  size: $Keys<typeof sizeText>,
+  weight: "bold" | "regular",
```
tomashapl: Why did you define above via `$Keys` and now you listed it all ?

please choose one way

src/Text/Text.stories.js
```diff
@@ -0,0 +1,239 @@
+// @flow
```
tomashapl: This file is too long, but OK

src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+  align: "left" | "center" | "right",
+  isItalic?: boolean,
```
tomashapl: please use only `italic` or `uppercase` it looks better if you pass 

```
<Text italic />
```

instead of

```
<Text isItalic />
```

src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+  isUppercase?: boolean,
```
tomashapl: same as above

tomashapl:
and one more thing: I don't see component registering in `index.js` @williamkolmacka

src/Text/index.js
williamkolmacka: I forgot to change the weight prop via `$Keys`, I will fix this.

src/Text/index.js
williamkolmacka: Will fix this.

src/Text/index.js
williamkolmacka: same there.

src/Text/index.js
williamkolmacka: It's written in the specification of Text component -> https://github.com/kiwicom/orbit-components/issues/91

src/Text/index.js
williamkolmacka: Will fix this to string only

jaroslav-kubicek:
> I don't see component registering in index.js

yes please, don't forget to add this 🙂

src/Text/index.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const sizeText = {
+  large: tokens.fontSizeTextLarge,
+  normal: tokens.fontSizeTextNormal,
+  small: tokens.fontSizeTextSmall,
+};
+
+type Props = {
+  type: $Keys<typeof colorText>,
```
jaroslav-kubicek: `type`, `size`, `weight` & `align` props should be defined as optional as well

PR #146, opened 2018-04-19
Update imported Roboto font and orbit-design-tokens version

wEprous:
There is a new version of orbit-design-tokens available: https://github.com/kiwicom/orbit-design-tokens/releases/tag/v0.0.2

We need also to import more Roboto weights and styles, because we use them through tokens.


.storybook/preview-head.html
```diff
@@ -1 +1 @@
-<link href="https://fonts.googleapis.com/css?family=Roboto:300,500" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css?family=Roboto:400,400i,500,500i,700" rel="stylesheet">
```
tomashapl: Why you updated storybook fonts ?

wEprous: Changes to this file were reverted.

PR #148, opened 2018-04-19
Update Radio component

### LINKED ISSUE #121: Create <Radio /> component
## Description 
Use a list of radio buttons when using a list with one selectable option. If you want to allow select more than one option, use instead.

## Note
Check #122 (Checkbox specs) before starting on this, it’s highly connected. 

## Visual style (for radio and checkboxes)
radio-button--checkbox

Zeplin: https://zpl.io/bPdzpPA

### Interactions
radio-button--checkbox-interactive

Different states **don't affect label** of radio buttons.

When **radio buttons are hovered**, we use darken() function for making border of the radio button little darker, e.g. darken([original-color], $modifierDarkenHover). This affects only radio button, not its label.

When **radio buttons are active**, we use darken() function for making border of the radio button little darker, e.g. darken([original-color], $modifierDarkenActive). Also, when button is active, we scale it's size down, with the css function transform: scale($modifierScaleButtonActive)

A **disabled state** should use opacity $opacityRadioButtonDisabled, hover styles should be disabled too so it doesn’t look like there is some interaction possible.

**Focus state** should use default browser styling (outline). It’s possible that we'll change it in the future with our own defined box-shadow, but we need to discuss this first.

**Error state**: We didn't discuss validation yet but this component should only reflect when a group of radios needs to select some option. In that case, it's pretty simple - radio symbol will turn red (as in image). Validation message should be shown near radio group but it'll probably need another component for it (eg. form-field, ..)

### Additional information
- Radio button symbol and label are vertically centered to each other
- Help text is 2px below the label, so we’ll probably need new $spaceXXXSmall token
- There should be a transition for changing states (for border color and scale properties)
- We agreed that it may be good idea to use same tokens for visual style as we use for InputField because the look is currently the same (and it’s planned to be the same by design)

## Functional specs
- All radio buttons should disable all the default browser styling so it doesn’t mess with our own visual style
- I can’t imagine a situation where the help text below the label could be clickable… however, we should probably let the help text open in same way we have it open in inputs and other form elements. For sake of consistency. 
- We discussed that we will have maybe one component <ChoiceList /> which will use object to map all the choices inside and then just use radio button & checkbox components for rendering. 

## Storybook chapters
One chapter for basic radio button (is it default?), second one for radio button with help text, playground. More after the knobs are defined.

## Some inspiration
http://elemental-ui.com/forms (scroll down to Radio buttons)
https://mineral-ui.com/components/radio
https://polaris.shopify.com/components/forms/radio-button#navigation

# Deprecated specs
This part of the specs was here before but we decide to not to define things in this kind of detail (because it always change). But we kept it here for references.

## Parameters

Parameter | Values | Default
-- | -- | --
Label | [string] | _empty_
Checked | [boolean] | false
Disabled | [boolean] | false
Error | [boolean] | false

### For all configurations
```
**For radio button*
background-color: $backgroundInput
height: $heighRadioButton
width: $widthRadioButton
border-radius: $borderRadiusCircle

**For label**
font-family: $fontFamily
font-size: $fontSizeLabelForm
color: $colorLabelForm
margin-left: $spaceXSmall (distance from radio button)
```

### Based on selected configuration:

Parameter:Value | property: $token
-- | --
Checked:false | **For label**<br>font-weight: `$fontWeightNormal`
Checked:true | **For label**<br>font-weight: `$fontWeightMedium`
Error:false | **For radio button**<br>border-color: `$borderColorInput`
Error:true | **For radio button**<br>border-color: `$borderColorInputError`

wEprous:

Added new RadioButton component.

closes #121


src/RadioButton/index.js
```diff
@@ -0,0 +1,121 @@
...
+import * as React from "react";
+import { darken } from "polished";
+import * as tokens from "orbit-design-token";
+
+type Props = {
+  name: string,
+  label: string,
+  value: string,
+  error?: boolean,
+  disabled: boolean,
```
jaroslav-kubicek: disabled is requred, but checked is defined as optional and both have defaults => I would suggest to unify this and define them as required with defaults

src/RadioButton/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+import * as React from "react";
+import { darken } from "polished";
+import * as tokens from "orbit-design-token";
+
+type Props = {
+  name: string,
+  label: string,
+  value: string,
+  error?: boolean,
```
jaroslav-kubicek: `hasError` would sound better IMHO, without looking on types, I would expect `error` is some string or Error object

wEprous: Good point, I'll use `hasError`.

src/RadioButton/index.js
wEprous: I'm going to define all props as required as you suggested.

jaroslav-kubicek:
ping @wEprous

wEprous:
I'm rewriting RadioButton and CheckBox to styled-components and implementing Theming. Both components are in Milestone 0.2 so it's not necessary to merge them this week.
cc @jaroslav-kubicek

wEprous:
@tomashapl @jaroslav-kubicek plz CR one more

PR #163, opened 2018-05-03
Fix exports

Fallion:
Small adjustment of exports.


jaroslav-kubicek:
pls rebase, cc @Fallion

Issue #154, opened 2018-04-24
Create <Alert /> component


darkwindcz:
## Description
Alerts have a purpose to inform the user about what is going on.

## Visual style
![specs](https://user-images.githubusercontent.com/6851247/39183170-72303e60-47bf-11e8-8bc2-c200e8c71392.png)

## Additional information
- Close / dismiss should be a link button, not only a clickable icon. Icon color is based on message type.
- It should be possible to add any text inside the <Alert />. Text color is based on message type.

## Parameters, tokens, etc.
@williamkolmacka will cover this

darkwindcz:
Let's implement Alert Button when it's ready for this component as well.

PR #171, opened 2018-05-10
Add new Alert component

williamkolmacka:
Added a new Alert component

Closes: #154


williamkolmacka:
Completed. You can check it.

src/Alert/Alert.stories.js
```diff
@@ -0,0 +1,177 @@
...
+            },
+          ],
+        },
+      ],
+    };
+  })
+  .addWithChapters("Info alert", () => {
+    const title = text("Title", "");
+    const message = text("Message", "The quick, brown fox jumps over a lazy dog.");
+    const noIcon = boolean("No icon", false);
```
tomashapl: Why ? I think is better to specify Icon instead of `noIcon` if Icon isn't specified just don't render it 💪

src/Alert/Alert.stories.js
```diff
@@ -0,0 +1,177 @@
...
+      info:
+        "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
+      chapters: [
+        {
+          sections: [
+            {
+              sectionFn: () => (
+                <Alert
+                  type={type}
+                  Icon={Icons[source]}
```
tomashapl: Why this prop is with capital letter in the start ? 🤔

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
+// @flow
+import * as React from "react";
+import * as tokens from "@kiwicom/orbit-design-tokens";
+
+import * as Icons from "../icons";
+import { Button } from "../index";
+
+const Icon = ({ type, Icon: CustomIcon, customColor }) => {
```
tomashapl: What is doing `CustomIcon` ?

tomashapl: Also same namespaces for component and prop isn't good pattern

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
...
+  info: tokens.colorAlertIconInfo,
+  success: tokens.colorAlertIconSuccess,
+  warning: tokens.colorAlertIconWarning,
+  critical: tokens.colorAlertIconCritical,
+};
+
+type Props = {
+  type: $Keys<typeof backgroundAlert | typeof colorTextAlert>,
+  title?: string,
+  Icon?: React.ComponentType<*>,
```
tomashapl: Is Icon component receiving any props in this use case ? If no, please rather use React.Node

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
...
+  children: React.Node,
+};
+
+const Alert = (props: Props) => {
+  const { type, title, closable, noIcon, onClose, children } = props;
+  return (
+    <div className="alert">
+      <div className="alertIcon">
+        {!noIcon &&
+          (Icon && <Icon type={type} Icon={props.Icon} customColor={colorIconAlert[type]} />)}
```
tomashapl:
```
!noIcon && Icon
```
 
please use only one variable as prop

src/Alert/Alert.stories.js
williamkolmacka: Because when you don't fill the `Icon` prop, there should be rendered a default icon according to the type of Alert. If you will use `noIcon` prop, there will be Alert without the icon. Do you think I can solve it in a better way?

src/Alert/Alert.stories.js
wEprous: It's the only case when it needs to be capitalized.
```
const { Icon,... } = props;

return (
   <Icon ... />
)
```
If we use `icon` with a small letter on the start, we get an error:
`If you meant to render a React component, start its name with an uppercase letter.`

src/Alert/index.js
williamkolmacka: `CustomIcon` is there for a case if you want to choose the custom icon, different from default one.

src/Alert/Alert.stories.js
tomashapl: just pass in icon prop `null`

src/Alert/Alert.stories.js
```diff
@@ -0,0 +1,194 @@
...
+              ),
+              options,
+            },
+          ],
+        },
+      ],
+    };
+  })
+  .addWithChapters("Playground", () => {
+    const typeOptions = {
```
tomashapl: Can we move this to component and export it ? I want to maintain this in one place

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
...
+  Check,
+  Alert as AlertTriangle,
+  AlertCircle,
+  CloseCircle,
+} from "../icons";
+import { Button } from "../index";
+
+const Icon = ({ icon, type }) => {
+  let renderIcon;
+  if (typeof icon === "boolean") {
```
tomashapl: icon as boolean, really ?

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
...
+    if (type === "info") {
+      renderIcon = <InformationCircle />;
```
tomashapl: just do return here

```
return <InformationCircle />
```

src/Alert/Alert.stories.js
```diff
@@ -0,0 +1,194 @@
...
+import ThemeProvider from "../Theming/ThemeProvider";
+import * as Icons from "../icons";
+
+import Alert from "./index";
+
+setAddon(chaptersAddon);
+
+const getIcons = defaultIcon => select("Icon", [undefined, ...Object.keys(Icons)], defaultIcon);
+const getIcon = source => Icons[source];
+
```
tomashapl: we can add `renderTitle` fci here

```
const renderTitle = (title) => title !== "" && title
```

to prevent repeating code

then use:

```
<Alert title={renderTitle(title)} />
```

src/Alert/index.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+  } else {
+    return icon;
+  }
+  return renderIcon;
+};
+
+type Props = {
+  type: "info" | "success" | "warning" | "critical",
+  title?: string,
```
tomashapl: title could be string or boolean if user pass

```
title={false}
```

src/Alert/index.js
williamkolmacka: It shouldn't be a problem because when you type `<Alert type="info" icon />` it returns you `true`, without the icon, it returns you `false`. When you want to pass there some specific icon, you can do it like `<Alert type="info" icon="<Airplane />" />` and this returns you a `React.Element`. So you can define it in flow like `icon?: React.Element<any> | boolean`

src/Alert/index.js
williamkolmacka: Will fix it.

src/Alert/Alert.stories.js
williamkolmacka: Will fix it.

src/Alert/index.js
wEprous: @tomashapl plz explain your suggest more, I don't think there is no need to define title as boolean also

PR #172, opened 2018-05-10
Update Text component

wEprous:
Update stories
Fix margin

### DIFF
--- a/src/Text/Text.stories.js
+++ b/src/Text/Text.stories.js
@@ -21,22 +21,21 @@ storiesOf("Text", module)
       padding: "20px",
     }),
   )
-  .addWithChapters("Default text", () => {
+  .addWithChapters("Primary text", () => {
     const customText = text(
       "Text",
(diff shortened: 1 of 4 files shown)


wEprous:
You can do CR on this PR. It's complete.

src/Text/Text.stories.js
```diff
@@ -166,6 +66,37 @@ storiesOf("Text", module)
...
+    const type = select(
+      "Type",
+      {
+        primary: "primary",
+        secondary: "secondary",
+        attention: "attention",
+      },
+      "attention",
+    );
     const element = select(
```
tomashapl:
```
<p> <div> <span>
```

may have different shared behavior from browser, like `display: block` please always override it, or is this on purpose ?

wEprous: Talked with @darkwindcz and we can't decide now. We want to leave it on browser for now.

PR #174, opened 2018-05-10
Created deprecated folder

tomashapl:
Created `_deprecated` folder and refactored deprecated components

- Container
- FieldFeedback
- Input
- InputText
- InputTextarea
- Radio
- Checkbox
- Select
- SystemMessage
- Typography


Fallion:
Please rebase this.

src/index.js
```diff
@@ -3,22 +3,13 @@ import defaultTheme from "@kiwicom/orbit-design-tokens";
...
-export { default as Section } from "./Container/Section";
-export { default as Select } from "./Select";
-export { default as SystemMessage } from "./SystemMessage";
-export { default as FieldFeedback } from "./FieldFeedback";
 export { default as TextLink } from "./TextLink";
-export { default as Typography } from "./Typography";
-export { default as Radio } from "./Radio";
 export { default as ThemeProvider } from "./Theming/ThemeProvider";
-export { Icons, defaultTheme, withTheme };
+export { Icons, DeprecatedComponents, defaultTheme, withTheme };
```
jaroslav-kubicek: This is breaking change ⚠️ 

I mean - I'm totally ok to move these components to deprecated folder, but it's nonwelcomed for me as a user of orbit to change imports all over the app if I want to bump the version.

If I currently use `Typography`, the goal is to get rid of using it in my code over time and replace it with new components => There might be some time when I use both deprecated and new components in code

With this change, I would be stuck with `0.0.0-rc7` because I don't want to change imports for components I plan to remove anyway.

If we intend to hint users these components should not be used, we should put `console.warn` or `console.error` into them.

TL;DR: Don't change how components can be imported.

src/_deprecated/Checkbox/index.js
```diff
@@ -4,6 +4,9 @@ import * as React from "react";
 
 import Typography from "../Typography";
 import Icon from "./Icon";
+import deprecationWarning from "../../../config/deprecationWarning";
+
+deprecationWarning("Checkbox is deprecated and will be removed in next major release");
```
jaroslav-kubicek: deprecations warnings need to be in `render` function, this results in console warnings no matter what specifically is imported, since it's executed when file is imported, which is exactly what is happening in index:

```
import { Typography } from '...';
```

tomashapl: Yes, but if you add it to render then it warns when every render

jaroslav-kubicek: yes, IMHO this is more desired behaviour, early adoptions should be rewarded

No usage of deprecated components => no warning
Many usages of deprecated components => many warnings

tomashapl: Imagine `InputText` component, it rerenders everytime when you type into => sooo many warings, but one usage

also in index there isn't importing, it warns due to `*.stories.js` files

jaroslav-kubicek: > also in index there isn't importing, it warns due to *.stories.js files 

https://github.com/kiwicom/orbit-components/pull/174/files#diff-1fdf421c05c1140f6d71444ea2b27638R7

jaroslav-kubicek: It would be batched together in such case

tomashapl: OK, refactored, according to this i had to create new file inside components folder that includes pure component with export without HOC due to tests, because i couldn't get into component via `shallow` ¯\_(ツ)_/¯

Also left `index.js` file with HOC export, so it shouldn't break anything

config/DeprecatedComponent.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import React from "react";
+
+const DeprecatedComponent = (Component: React$ComponentType<any>, text: string) => {
+  let warned = false;
+  return (props: Object) => {
```
jaroslav-kubicek: cool, just for future, it's possible to parametrize the types, I think something like:

```
const DeprecatedComponent = <Props>(Component: React$ComponentType<any>, text: string) => {
```

and then inside the function like `return (props: Props) => {`

tomashapl: Ok, thanks 😊

PR #175, opened 2018-05-10
Update Button support Theming

wEprous:
Update stories of Button
Fixed disabled states
Fixed loading icon


wEprous:
You can do CR on this PR. It's complete.

src/Button/Button.stories.js
```diff
@@ -157,25 +137,31 @@ storiesOf("Button", module)
...
         large: "large",
       },
-      "normal",
+      "small",
     );
-    const width = number("Width", 0);
-
+    const showIcon = boolean("Show Icon", true);
+    const source = select("Icon", Object.keys(Icons), "PlusCircle");
+    const onlyIcon = boolean("onlyIcon", false);
```
tomashapl: I think, if somebody does not want to show text just doesn't fill `title` prop is this really necessary ?, please make it simple as possible

src/Button/Button.stories.js
```diff
@@ -157,25 +137,31 @@ storiesOf("Button", module)
...
         small: "small",
         normal: "normal",
         large: "large",
       },
-      "normal",
+      "small",
     );
-    const width = number("Width", 0);
-
+    const showIcon = boolean("Show Icon", true);
```
tomashapl: just don't define icon 🤔

wEprous: OK

src/Button/Button.stories.js
```diff
@@ -184,15 +171,13 @@ storiesOf("Button", module)
...
-                  loading={loading}
-                  block={block}
                   variation="filled"
-                  type="facebook"
+                  type={type}
                   size={size}
-                  width={width}
-                  Icon={Icons.Facebook}
-                />
+                  Icon={Icons[source]}
```
tomashapl: Please define component class above and then pass it as node

```
const icon = Icons[source];

<Button ... icon={icon} />
```

src/Button/index.js
```diff
@@ -277,21 +277,22 @@ const Button = (props: Props) => {
   const sizeIcon = size === "large" || size === "normal" ? "medium" : size;
 
   // allow onlyIcon only when Icon is passed
-  const allowOnlyIcon = !!(onlyIcon && Icon);
+  const allowOnlyIcon = !children && !!Icon;
```
tomashapl: Just remove it and evaluate the same

```
if (icon)
```

jaroslav-kubicek:
there are conflicts, cc @wEprous or @tomashapl

package.json
```diff
@@ -50,6 +50,8 @@
     "prop-types": "^15.6.1",
     "react-icon-base": "^2.1.2",
     "react-icons": "^2.2.7",
+    "styled-components": "^3.2.6",
+    "styled-jsx": "^2.2.6",
```
jaroslav-kubicek: `styled-jsx` is already in peer & dev dependencies, pls don't add it here

wEprous: Fixed.

src/Button/index.js
```diff
@@ -1,365 +1,332 @@
...
+        info: theme.borderColorButtonInfoBorderedHover,
+        success: theme.borderColorButtonSuccessBorderedHover,
+        warning: theme.borderColorButtonWarningBorderedHover,
+        critical: theme.borderColorButtonCriticalBorderedHover,
+        facebook: theme.borderColorButtonFacebookBorderedHover,
+        google: theme.borderColorButtonGoogleBorderedHover,
+      },
+    },
+    borderColorButtonActive: {
+      bordered: {
```
tomashapl: I think it shouldn't be nested right ?

src/Button/index.js
```diff
@@ -1,365 +1,332 @@
...
+      small: theme.paddingButtonSmallWithIcon,
+    },
+    marginRightIcon: {
+      small: theme.marginButtonIconSmall,
+      normal: theme.marginButtonIconNormal,
+      large: theme.marginButtonIconLarge,
+    },
+    backgroundButton: {
+      filled: {
+        primary: theme.backgroundButtonPrimary,
```
tomashapl: Can we please create constants for this ? like in `<Alert />`

wEprous: OK

src/Button/index.js
wEprous: This `tokens` concept is from the old version where was `variation` as prop. I can rewrite this as it was in the very first version, e.g.:
`color: ${ bordered ? colorBordered : color }`

src/Button/index.js
```diff
@@ -1,365 +1,308 @@
...
-  // allow onlyIcon only when Icon is passed
-  const allowOnlyIcon = !!(onlyIcon && Icon);
+  &:disabled {
+    cursor: default;
+    opacity: ${({ theme }) => theme.opacityButtonDisabled};
+  }
+  &:enabled:hover {
+    background: ${({ bordered, tokens, type }) =>
+      bordered ? tokens.backgroundButtonBorderedHover : tokens.backgroundButtonHover[type]};
+    border-color: ${({ bordered, tokens, type }) =>
```
tomashapl: `border: 0` I think this doesn't change anything  use `box-shadow`

src/Button/__tests__/index.test.js
```diff
@@ -1,46 +1,58 @@
...
-  });
-});
-
-describe("Button with icon", () => {
-  const component = mount(
-    <Button variation="link" size="normal" title={title} Icon={Airplane} onClick={onClick} />,
-  );
-  const button = component.find("button");
-  it("Should contain SVG", () => {
+  // it("should execute onClick method", () => {
```
jaroslav-kubicek: I would delete it, you can retrieve it back from history if needed, but my experience is that it will be commented out forever :)

wEprous: We have a solution for these broken tests already, so I hope not :)

Issue #119, opened 2018-04-10
Create <Card /> component


darkwindcz:
## Description
Card component is a simple container for grouping some relevant information. In some cases, it can contain title and description. When you need to show content inside, wrap it in the `<CardSection />` component, or use a Sectioned parameter.

## Visual style
![card](https://user-images.githubusercontent.com/6851247/38561109-7c0a3d98-3cd7-11e8-8384-6caf32000093.png)

## Parameters
Parameter | Values | Default
-- | -- | --
Title | [string] | _empty_ 
TitleElement | h1, h2, h3, div | h2
Description | [string] | _empty_
Sectioned | true/false | false
Closable | true/false | false

`Sectioned` param is for cards with a simple layout, it auto-wraps content in a card with padding from left and right side. If `Sectioned` is false, it's possible to insert full-width sections inside.

### For all configurations
```
border-radius: $borderRadiusNormal
background-color: $backgroundSection
border-width: $borderWidthSection
border-style: $borderStyleSection
border-color: $borderColorSection
```
### Additional information
- For icon use medium size
- Use `$paddingCardSection` token for setting padding for inner section of sectioned card
- For title use `<Heading />` component, description should be wrapped in `<p></p>`
- Margin bottom after title should use `$spaceMedium` token

## Inspiration
[Card component in Shopify Polaris](https://polaris.shopify.com/components/structure/card)

Issue #120, opened 2018-04-10
Create <CardSection /> component


darkwindcz:
## Description
Card section create a space for sections and adds necessary paddings and delimiters to them for a clear hierarchy. 

## How it looks
![cartsection](https://user-images.githubusercontent.com/6851247/38565087-b29c0df6-3ce0-11e8-9a50-829ac7a08e9b.png)

## Parameters
_none_

## Visual design & design tokens
```
padding: $paddingCardSection
```

### Additional information
- If there is more then one `<CardSection />`, it should show a **line between sections**. For line style use these tokens: `border-width: $borderWidthSection; border-style: $borderStyleSection; border-color: $borderColorSection`
- The content that should go inside section shouldn't set any padding. Section padding is set globally and content is just displayed inside: 
![cardsection-paddings](https://user-images.githubusercontent.com/6851247/38565288-36a9bab2-3ce1-11e8-8d52-ed0321553a00.png)

wEprous:
added in #180

PR #180, opened 2018-05-23
Add new Card component

williamkolmacka:

Add a new Card component.

Closes: #119 and #120

### DIFF
--- a/src/Card/Card.stories.js
+++ b/src/Card/Card.stories.js
@@ -1,0 +1,165 @@
+// @flow
+import * as React from "react";
+import { storiesOf, setAddon } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import styles from "@sambego/storybook-styles";
+import chaptersAddon from "react-storybook-addon-chapters";
+import { withKnobs, text, boolean, select } from "@storybook/addon-knobs/react";
+
+import * as Icons from "../icons";
+import Heading from "../Heading/Heading";
+import Text from "../Text/Text";
+import Card from "./Card";
+import CardHeader from "./CardHeader/CardHeader";
+import CardContent from "./CardContent/CardContent";
+import CardSection from "./CardSection/CardSection";
+
+setAddon(chaptersAddon);
+
+const getIcons = defaultIcon => select("Icon", Object.keys(Icons), defaultIcon);
+const getIcon = source => Icons[source];
+
+storiesOf("Card", module)
+  .addDecorator(withKnobs)
+  .addDecorator(
+    styles({
+      padding: "20px",
+    }),
+  )
+  .addWithChapters("Default", () => {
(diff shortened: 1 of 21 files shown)


jaroslav-kubicek:
I might be missing something, but how I'm supposed to use `CardSection` in my code?

I would expect something like
```
<Card>
  <Card.Section />
</Card>
```

but section seems to be not exported

src/Card/index.js
```diff
@@ -0,0 +1,109 @@
...
+  position: relative;
+  width: 100%;
+  padding: ${props => props.theme.spaceLarge};
+  padding-right: ${props => (props.closable ? props.theme.spaceXXLarge : props.theme.spaceLarge)};
+  box-sizing: border-box;
+`;
+
+const StyledCardBody = styled.div`
+  width: 100%;
+  padding: ${props =>
```
jaroslav-kubicek: this function is not too readable, would be nicer to define it as normal named function and avoid nested ternary operator

williamkolmacka: Fixed.

src/Card/Card.stories.js
```diff
@@ -0,0 +1,194 @@
...
+        {
+          sections: [
+            {
+              sectionFn: () => (
+                <ThemeProvider>
+                  <Card
+                    icon={<Icons.Airplane color="attention" />}
+                    title={title}
+                    description={description}
+                    sectioned
```
wEprous: Can we get rid of this `sectioned` prop?
We can check if children contains `Section` component - something like this:
```js
let sectioned = false;
children.map(child => {
  if (child.type === Section) {
    sectioned = true;
  }
  return sectioned;
});
```

wEprous: Let me know @williamkolmacka

src/Card/CardSection.js.flow
```diff
@@ -0,0 +1,15 @@
+// @flow
+import defaultTokens from "@kiwicom/orbit-design-tokens"
+
+export type Props = {|
+  children: React$Node,
+  theme?: typeof defaultTokens
+|}
+
+export type styledCardSection = {
```
wEprous: Is it necessary to export these types?

src/index.js
```diff
@@ -18,6 +18,13 @@ export { default as Button } from "./Button";
 export { default as ButtonLink } from "./ButtonLink";
 export { default as CarrierLogo } from "./CarrierLogo";
 export { default as Checkbox } from "./Checkbox";
+
+// Card
+export { default as Card } from "./Card";
+export { default as CardHeader } from "./Card/CardHeaderWithTheme";
```
wEprous: As we talked, move subcomponents into separated folders, eg. `Card/CardHeader/`

PR #182, opened 2018-05-24
Update Text support Theming

williamkolmacka:
Added support for theming
Rewrited to styled components
Added flow strict type
Added constants for props


src/Text/__tests__/index.test.js
```diff
@@ -2,15 +2,34 @@
...
+          {children}
+        </Text>
+      </ThemeProvider>,
+    );
+    expect(
+      component
+        .children()
+        .children()
+        .exists(),
+    ).toBe(true);
```
jaroslav-kubicek: Pls check that it contains `Children text`

src/Text/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -1,27 +1,375 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Rendered with children Should contain text 1`] = `
-<React.Fragment>
-  <p
-    className="jsx-1052758011 "
+exports[`Text should match snapshot 1`] = `
+<ThemeProvider
```
jaroslav-kubicek: this snapshot doesn't make much sense, please keep the old one - it actually shouldn't be affected by changes in this PR (so it the test, do something like `expect(wrapper.find('Text')).toMatchSnapshot()`)

src/Text/__tests__/index.test.js
williamkolmacka: Fixed.

src/Text/__tests__/__snapshots__/index.test.js.snap
williamkolmacka: Fixed.

src/Text/index.js
```diff
@@ -1,79 +1,67 @@
...
-          text-align: ${align};
-          ${uppercase ? `text-transform: uppercase` : ""};
-          ${italic ? `font-style: italic` : ""};
-          margin: 0;
-        `}
-      </style>
-    </scope>,
-  );
+const StyledText = styled(TextElement)`
+  font-family: ${props => props.theme.fontFamily};
```
wEprous: Use object destructuring

src/Text/index.js
```diff
@@ -1,79 +1,67 @@
...
+  text-transform: ${props => props.uppercase && `uppercase`};
+  font-style: ${props => props.italic && `italic`};
+  margin: 0;
+`;
 
+const Text = (props: Props) => {
+  const { children, theme } = props;
+  const tokens = {
+    colorText: {
+      primary: theme.colorTextPrimary,
```
wEprous: Define constants for type, size and weight

src/Text/index.js
```diff
@@ -1,79 +1,67 @@
...
+import { withTheme } from "theming";
+import createComponentFromTagProp from "react-create-component-from-tag-prop";
 
 type Props = {
-  type: $Keys<typeof colorText>,
-  size: $Keys<typeof sizeText>,
-  weight: $Keys<typeof weightText>,
+  type: "primary" | "secondary" | "attention",
+  size: "small" | "normal" | "large",
+  weight: "regular" | "bold",
```
wEprous: Why is here defined `regular` and token from orbit-design-tokens is called `fontWeightNormal` ?

wEprous: Fixed.

src/Text/index.js
wEprous: Fixed.

src/Text/index.js
wEprous: Fixed.

PR #183, opened 2018-05-24
Update TextLink support Theming

### LINKED ISSUE #197: TextLink href prop
`<TextLink>` has `url` props. I suggest to have `href`.
In Next.js you can generate routes between pages with `<Link>`. It passes url into `<a href=...>` in `render`
```jsx
<Link href="/bookings">
  <a>link</a>
</Link>
```
It could even pass `href` from `<Link>` to `<TextLink>` by `<Link passHref href="/booking">` but only when `<TextLink>` has `href` not `url`.

Second reason for `href` is that `<a href="mailto:xxx@xxx.com">` is not url, but it's still **H**ypertext **REF**erence.

----

williamkolmacka:
Added support for theming
Rewrited to styled components
Changed prop url to href
Updated tests
Added flow strict type
Added constants for type prop

Closes #197


package.json
```diff
@@ -6,7 +6,7 @@
     "build": "yarn clean && yarn build:icons && yarn build:lib && yarn build:flow && yarn build:module",
     "build:lib": "babel --out-dir lib --ignore *.stories.js,*.test.js src",
     "build:flow": "flow-copy-source -v -i '**/*.?(stories|test).js' src lib",
-    "build:module": "babel --presets react --plugins styled-jsx/babel,babel-plugin-styled-components,transform-class-properties --no-babelrc --out-dir es --ignore *.stories.js,*.test.js src",
+    "build:module": "babel --presets react --plugins styled-jsx/babel,babel-plugin-styled-components,transform-class-properties,transform-object-rest-spread --no-babelrc --out-dir es --ignore *.stories.js,*.test.js src",
```
jaroslav-kubicek: topic for another PR - create `.babelrc` for module, this is already too weird

src/TextLink/index.js
```diff
@@ -1,63 +1,77 @@
...
+    | "search"
+    | "tag",
+  theme: typeof defaultTokens,
 };
 
+const StyledTextLink = styled(({ tokens, theme, type, url, ...props }) => (
+  <a {...props}>{props.children}</a>
+))`
+      color: ${props => props.tokens.colorTextLink[props.type]};
+      font-weight: ${props => props.theme.fontWeightLinks};
```
jaroslav-kubicek: what about object destructuring? `props => props.theme.fontWeightLinks` => `({ theme }) => theme.fontWeightLinks`

wEprous: I like it. But in some components, we need more props to destructuring.
`height: ${({ tokens, size }) => tokens.heightButton[size]};`

So what do you like more? cc @jaroslav-kubicek @tomashapl @williamkolmacka

williamkolmacka: I like style that @jaroslav-kubicek mentioned. From my point of view, it's more readable and typing only props you want is much better.

wEprous: OK, so we need to use it everywhere then.

src/TextLink/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -1,63 +1,3 @@
...
-        "underline",
-        "underline",
-        "#00a991",
-      ]
-    }
-    styleId="211793505"
-  />
-</a>
-`;
+exports[`TextLink should match snapshot 1`] = `null`;
```
wEprous: the test probably cannot find anything and snap it, can you check it @williamkolmacka ?

src/TextLink/index.js
```diff
@@ -1,63 +1,84 @@
...
+      outline-width: 3px;
+    }        
+  `;
+
 const TextLink = (props: Props) => {
-  const { children, url, onClick, newTab, type, size } = props;
+  const { children, url, external, theme, rel, onClick } = props;
+  const tokens = {
+    colorTextLink: {
+      primary: theme.colorLinkPrimary,
```
wEprous: Define constant for type

tomashapl:
@wEprous plz rebase 🙏

PR #186, opened 2018-05-28
Add new ButtonLink component

### LINKED ISSUE #181: "url" prop for TextLink needs to be optional
Most of the time, all login is handled by `onClick` in react apps and url isn't relevant

wEprous:

closes #181


src/ButtonLink/index.js
```diff
@@ -0,0 +1,210 @@
+// @flow
+import * as React from "react";
+import defaultTokens from "@kiwicom/orbit-design-tokens";
+import styled from "styled-components";
+import { withTheme } from "theming";
+
+import { iconSizes } from "../Icon";
+
+type Props = {
+  children?: React.Node,
```
jaroslav-kubicek: if the children is optional, I would add such example into stories

wEprous: OK, good point

src/ButtonLink/index.js
```diff
@@ -0,0 +1,210 @@
...
+    theme,
+    type,
+    icon,
+    sizeIcon,
+    width,
+    disabled,
+    url,
+    children,
+    ...props
+  }) => <a {...props}>{children}</a>,
```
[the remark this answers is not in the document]
tomashapl: @darkwindcz We make decision tomorrow right ?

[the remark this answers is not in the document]
darkwindcz: yes, I'll try to find something in our calendars but it looks already pretty busy day :/

src/ButtonLink/ButtonLink.js
```diff
@@ -0,0 +1,212 @@
...
+  disabled: boolean,
+  block: boolean,
+  external: boolean,
+  type: $Values<typeof TYPES>,
+  size: $Values<typeof SIZES>,
+  href?: string,
+  width: number,
+  icon?: React.Node,
+  theme: typeof defaultTokens,
+  transparent?: boolean,
```
wEprous: We can define `transparent` as required and set default to `false`

src/Alert/Alert.js
```diff
@@ -0,0 +1,159 @@
...
+
+const Content = styled.div`
+  display: flex;
+  align-items: center;
+  min-height: ${({ theme, title }) => !title && theme.heightIconMedium};
+  margin-bottom: ${({ theme, title }) => title && theme.spaceXXSmall};
+`;
+
+const CloseContainer = styled.div`
+  position: absolute;
```
wEprous: We can use here `flex` as well, right? Personally, I don't like mixing flex with `position`

tomashapl: I don't want to affect layout by close icon, I think `absolute` is better

src/ButtonLink/ButtonLink.js
tomashapl: Yes, but  i don't see any benefits, the same we can do with other props, but why ?

wEprous: Currently, you can see `transparent={false}` in the preview source code.
Also, if someone will copy this code, there is no need to pass `anyProp={false}` to a component.

wEprous: It'll probably results in warning/error I think.

src/Alert/Alert.js
wEprous: OK, as you think, but I rather use only `flex`.
On the **line 90** you have hardcoded `-16px`, it needs to be token.

PR #198, opened 2018-05-31
Change <TextLink> url prop to href

### LINKED ISSUE #197: TextLink href prop

jakubzaba:
Closes #197 

If you are aware of BC break, I can add `url` prop back for some time and add warning for those who uses it, but it's still RC, so it shouldn't be big deal.

-----

### DIFF
--- a/src/TextLink/TextLink.stories.js
+++ b/src/TextLink/TextLink.stories.js
@@ -24,7 +24,7 @@ storiesOf("TextLink", module)
     }),
   )
   .addWithChapters("Primary link", () => {
-    const url = text("URL", "https://kiwi.com");
+    const href = text("href", "https://kiwi.com");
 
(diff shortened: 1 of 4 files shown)


wEprous:
The `url` prop will be replaced for `href` in #183

PR #199, opened 2018-05-31
DEV: created withTheme decorator

tomashapl:
Created withTheme decorator for storybook

It wraps stories via `<ThemeProvider />` 

Simplify usage and doesn't render `<ThemeProvider>` in storybook Source section

image


tomashapl:
Maybe we can move this to somehing like `decorators` and prevent babel to build this ?

wEprous:
We're going to write pure components and components withTheme, so decorator for entire storybook is not needed.

PR #205, opened 2018-06-07
Give a value to textarea via prop

alexedev:
Fixes the `Warning: Use the `defaultValue` or `value` props instead of setting children on <textarea>`


alexedev:
well, need to fix tests...

PR #206, opened 2018-06-07
NEW: InputField component

### LINKED ISSUE #153: Create <InputField /> component
## Description
Input field is a form input that we use when we need from users to fill some simple data.

## Visual style
inputfield

Zeplin: https://zpl.io/2pdrdzQ

### Interaction states
inputfield-interactions

When an input is **hovered**, border colors should use :hover version of their color. Same behavior is for the error input. Tokens should be defined for this.

When input is **focused**, we show a small shadow (from `$boxShadowInputFocus` token) and specific border color (`$borderColorInputFocus`). For focused error input use tokens `$boxShadowInputErrorFocus` and `$borderColorInputErrorFocus`

**Hover & focus states** are same for all input sizes and types.

**Disabled state** should change background and text color (`background-color: $backgroundInputDisabled`
`color: $colorTextInputDisabled`), **hover styles should be disabled** too so it doesn’t look like there is some interaction with it. However, **when disabled is hovered**, we should change the cursor to "not-allowed".

### Additional information
- Everything inside the text field is vertically centered
- Input width is 100%
- The space between prefix and value/placeholder should be $spaceXSmall

## Functional specs
- If the user fills some value into an input, the label color needs to change to a lighter shade (an exact color is defined as a token)
- The prefix should accept text or other components like icon, country flag, ...
- Width of prefix should push input value to the right (for example, you can have currency $ or CZK)  
- Suffix should accept other components like action icon, service logo, country flag, …
- Help text should be defined as text with a possibility of React component, as it can have a custom action attached to part of it (for example, click on „John Doe“ will fill the value with „John Doe")

## Storybook chapters
This will be specified after we know which props we have. For now, start with playground and default. It’s very possible that there will be inputs with a prefix (icon or text), suffix (action icon or service logo) 

## Some inspiration
https://polaris.shopify.com/components/forms/text-field#navigation
https://www.material-ui.com/#/components/text-field
http://seek-oss.github.io/seek-style-guide/textfield/

# Deprecated specs
This is from old specs, it’s probable that’s it not applicable anymore. But it shouldn’t hurt to have it here.

## Proposed props
Note: feel free to update it anyhow you need or like

Prop | Values | Default
-- | -- | --
Size | Small, Normal | Normal
Type | text,number,email,password | text
Name | string | _empty_
Label | string _or_ boolean | _empty_
Value | string | _empty_
Placeholder | string | _empty_
Prefix | component | _empty_
Suffix | component | _empty_
Help text | string | _empty_
Error | boolean | false
Error text | string | _empty_
Disabled | boolean | false
Max | number | _empty_
Min | number | _empty_
MaxLength | number | _empty_
MinLength | number | _empty_

## Visual design & design tokens
**Note: this is not updated, please refer to Design tokens page at orbit.kiwi**

### For input
```
border-radius: $borderRadiusNormal
border: $borderStyleInput $borderWidthInput $borderColorInput;
background-color: $backgroundInput
color: $colorTextInput

for placeholder: 
color: $colorPlaceholderInput
```
* prefix text should have same color & size as a value in the input

### For label
```
font-size: $fontSizeLabelForm
color: $colorLabelForm

when filled with value:
color: $colorLabelFormFilled
```

### For message (error & help)
```
font-size: $fontSizeMessageForm

for error message:
color: $colorTextError
font-weight: $fontWeightNormal

for info message:
color: $colorTextInfo

```
### For icons (front & action)
```
color: $colorIconInput
```

wEprous:

Closes #153


src/InputField/InputField.js
```diff
@@ -0,0 +1,235 @@
+// @flow
+import * as React from "react";
+import defaultTokens from "@kiwicom/orbit-design-tokens";
+import styled from "styled-components";
+
+import FormFeedback from "../FormFeedback/FormFeedback";
+import FormLabel from "../FormLabel/FormLabel";
+
+export const TYPE_OPTIONS = {
```
tomashapl: Please move to `consts.js`

src/InputField/InputField.js
```diff
@@ -0,0 +1,235 @@
...
+  TEXT: "text",
+  NUMBER: "number",
+  EMAIL: "email",
+  PASSWORD: "password",
+};
+
+export const SIZE_OPTIONS = {
```
tomashapl: Also here

src/InputField/InputField.js
```diff
@@ -0,0 +1,235 @@
...
+
+  &:focus {
+    outline: none;
+  }
+  &::placeholder {
+    color: ${({ theme }) => theme.colorPlaceholderInput};
+  }
+`;
+
+type Props = {|
```
tomashapl: Move this to `InputField.js.flow` 🙏

src/InputField/InputField.js
```diff
@@ -0,0 +1,235 @@
...
+  justify-content: center;
+  cursor: pointer;
+  pointer-events: ${({ disabled }) => disabled && "none"};
+
+  & svg {
+    color: ${({ theme }) => theme.colorIconSecondary};
+  }
+`;
+
+const Input = styled(({ theme, tokens, size, prefix, suffix, error, help, ...props }) => (
```
tomashapl: defined default props in destructuring

src/InputField/InputField.js
```diff
@@ -0,0 +1,235 @@
...
+          error={props.error}
+        />
+      </InputContainer>
+      {props.help && !props.error && <FormFeedback type="help">{props.help}</FormFeedback>}
+      {props.error && <FormFeedback type="error">{props.error}</FormFeedback>}
+    </Field>
+  );
+};
+
+InputField.defaultProps = {
```
tomashapl: get rid off

src/InputField/InputField.js
wEprous: Done.

src/TextLink/TextLink.js.flow
```diff
@@ -7,8 +7,8 @@ export type Props = {|
   children: React$Node,
   href?: string,
   onClick?: (SyntheticEvent<HTMLLinkElement>) => void,
-  external: boolean,
-  type: Type,
+  external?: boolean,
```
wEprous: I also had to change `external` and `type` to be optional. Can be merged in this PR or we should create another one.

tomashapl: It is OK to be merged with this 💪

tomashapl:
🙏 Can we unify `prefix` and `suffix` styles which is used in Select, or is it completely different ? cc @wEprous

wEprous:
We can unify something and make shared components. But in case of the Select, we need to extend them with some styles e.g. `position`. And more thing is that Select prefix is accepting `<Icon color="secondary" />`, but in Inputfield, color is defined in Prefix styles and you just pass `<Icon />`.

PR #213, opened 2018-06-14
Separate withTheme from Heading component

wEprous:
Separated withTheme from Heading component.
Deleted ThemeProvider and unnecessary options from stories.
Updated test.
Added constants for element and type options.

**No more changes**

### DIFF
--- a/src/Heading/Heading.js
+++ b/src/Heading/Heading.js
@@ -1,0 +1,72 @@
+// @flow
+import * as React from "react";
+import styled from "styled-components";
+import defaultTokens from "@kiwicom/orbit-design-tokens";
+import createComponentFromTagProp from "react-create-component-from-tag-prop";
+
+export const ELEMENT_OPTIONS = {
+  H1: "h1",
+  H2: "h2",
+  H3: "h3",
+  H4: "h4",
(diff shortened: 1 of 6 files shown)


src/Heading/Heading.js
```diff
@@ -0,0 +1,72 @@
...
+    },
+    sizeHeading: {
+      [TYPE_OPTIONS.DISPLAY]: theme.fontSizeHeadingDisplay,
+      [TYPE_OPTIONS.TITLE1]: theme.fontSizeHeadingTitle1,
+      [TYPE_OPTIONS.TITLE2]: theme.fontSizeHeadingTitle2,
+      [TYPE_OPTIONS.TITLE3]: theme.fontSizeHeadingTitle3,
+    },
+  };
+  return (
+    <StyledHeading tokens={tokens} {...props}>
```
tomashapl: I think there is avaible only `type` and `theme` right ? please pass is directly is more readable

wEprous: `tokens` are defined above, line 45

wEprous: Fixed.

PR #214, opened 2018-06-14
Separate withTheme from Button component

wEprous:
Separated withTheme from Button component.
Deleted ThemeProvider and unnecessary options from stories.
Updated test.

**No more changes**


tomashapl:
@darkwindcz Should `<Button />` render <a> like in `<ButtonLink />` ?

darkwindcz:
@tomashapl yes,  both `<Button />` and `<ButtonLink />` should be able to render `a` or `button`.

In my opinion, when developer use prop `href`, it should render `button`. I am not sure if it's needed to be able to tell component what type it should render manually.

tomashapl:
@darkwindcz Thanks.

@wEprous Can you handle this ? It should be the same like in ButtonLink component i think.

wEprous:
@tomashapl Yeah, sure :)

tomashapl:
@wEprous 

You forgot delete old code from `Button/index.js` right ? please check your code before pushing it to review. 🙏

src/Button/Button.js
```diff
@@ -0,0 +1,339 @@
...
+  disabled: boolean,
+  block: boolean,
+  type: $Values<typeof TYPE_OPTIONS>,
+  size: $Values<typeof SIZE_OPTIONS>,
+  width: number,
+  icon?: React.Node,
+  theme: typeof defaultTokens,
+|};
+
+const IconContainer = styled(({ theme, tokens, bordered, sizeIcon, type, onlyIcon, ...props }) => (
```
tomashapl: I think there isn't available any other props

why there you need it ?

```
<IconContainer
          size={size}
          onlyIcon={onlyIcon}
          bordered={bordered}
          sizeIcon={sizeIcon}
          tokens={tokens}
          type={type}
          theme={theme}
        >
```


just render `<div />` without props

wEprous: We need to pass at least `children` and `className` to this div.

tomashapl: OK, so i think this is more readable, and you shouldn't destruct unused props

image

PR #219, opened 2018-06-19
Add new Select component

williamkolmacka:

Add a new Select component. There are still some values in CSS that are not in tokens.


src/Select/Select.js
```diff
@@ -0,0 +1,188 @@
...
+  state = {
+    value: this.props.value,
+  };
+
+  handleChange = (e: SyntheticInputEvent<HTMLSelectElement>) => {
+    this.setState({ value: e.target.value });
+    this.props.onChange(e);
+  };
+
+  renderOption = ({ value, label, visible = true }: Option) => {
```
wEprous: You defined `disabled` in Option type but you are not rendering it anywhere.

src/Select/Select.js
```diff
@@ -0,0 +1,188 @@
...
+  &::-ms-expand {
+    display: none;
+  }
+  &:disabled {
+    color: ${({ theme }) => theme.colorTextInputDisabled};
+    background: ${({ theme }) => theme.backgroundInputDisabled};
+    cursor: default;
+  }
+  /* placeholder */
+  &:invalid {
```
wEprous: Please for `placeholder` style use another way. Required attr will probably render native browser error. Do this like:
```js
const isPlaceholder = this.props.placeholder && this.state.value === ""
```
You then pass this const to `StyledSelect` and it'll decide which color should apply.

src/Select/Select.js
```diff
@@ -0,0 +1,188 @@
...
+export const SIZE_OPTIONS = {
+  SMALL: "small",
+  NORMAL: "normal",
+};
+
+type Option = {
+  value: string | number,
+  label?: string,
+  disabled?: boolean,
+  visible?: boolean, // eslint-disable-line react/no-unused-prop-types
```
wEprous: The is no logical reason to use `visible` more, please delete it everywhere.

src/Select/Select.js
```diff
@@ -0,0 +1,188 @@
...
+    }
+    return null;
+  };
+
+  render() {
+    const { theme, prefix, label, size, disabled, placeholder, help, error, options } = this.props;
+    const { value } = this.state;
+    return (
+      <React.Fragment>
+        {label && <Text>{label}</Text>}
```
wEprous: I'm completely missing `label` knob in stories. Don't forget that label's color is changing on the selected value.

src/Select/__tests__/Select.test.js
```diff
@@ -14,15 +16,30 @@ const objectOptions = [
   { disabled: true, value: "disabled-six", label: "Disabled Six" },
 ];
 
-const mockChange = jest.fn();
 describe("Select", () => {
   const component = shallow(<Select value="1" options={objectOptions} onChange={mockChange} />);
+  const componentWithPlaceholder = shallow(
```
wEprous: You can merge these two shallow wrappers into one.

src/Select/Select.js
```diff
@@ -0,0 +1,218 @@
...
+  display: inline-block;
+  font-family: ${({ theme }) => theme.fontFamily};
+  font-size: ${({ theme }) => theme.fontSizeLabelForm};
+  color: ${({ theme, value }) =>
+    value === "" ? theme.colorLabelForm : theme.colorLabelFormFilled};
+  line-height: ${({ theme }) => theme.lineHeightText};
+  margin-bottom: ${({ theme }) => theme.spaceXXSmall};
+`;
+
+const SelectContainer = styled(({ theme, ...props }) => <div {...props} />)`
```
wEprous: Omit `error` or:
```js
const SelectContainer = styled(({ children, className }) => (
  <div className={className}>{children}</div>
))`
```

src/Select/Select.js
```diff
@@ -0,0 +1,218 @@
...
+    ${({ theme, error }) => (error ? theme.borderColorInputError : theme.borderColorInput)};
+  border-color: ${({ theme, active, error }) =>
+    active && (error ? theme.borderColorInputError : theme.borderColorInputFocus)};
+  box-shadow: ${({ theme, active, error }) =>
+    active && (error ? theme.boxShadowInputErrorFocus : theme.boxShadowInputFocus)};
+  width: 100%;
+  box-sizing: border-box;
+  cursor: pointer;
+  &:hover {
+    border-color: ${({ theme, active, error }) =>
```
wEprous: Disable `hover` when select is `disabled`

src/Select/Select.js
```diff
@@ -0,0 +1,218 @@
...
+
+  render() {
+    const { theme, prefix, label, size, disabled, placeholder, help, error, options } = this.props;
+    const { value, active } = this.state;
+    const isPlaceholder = placeholder && value === "";
+
+    return (
+      <React.Fragment>
+        {label && (
+          <SelectLabel theme={theme} value={value}>
```
wEprous: Should be `select` focused and opened after clicking on `label`?

williamkolmacka: Can make this when we will have custom select ready. For now, it is not so much important.

src/Select/Select.js
```diff
@@ -0,0 +1,207 @@
...
+  label?: string,
+  disabled?: boolean,
+};
+
+type Props = {
+  size: $Values<typeof SIZE_OPTIONS>,
+  label?: string,
+  placeholder?: string,
+  value: string,
+  disabled?: boolean,
```
wEprous: When select is disabled, styles are not changed.

src/Select/Select.stories.js
```diff
@@ -0,0 +1,186 @@
...
+          },
+        ],
+      },
+    ],
+  }))
+  .addWithChapters("Playground", () => {
+    const placeholder = text("Placeholder", "Select value from list");
+    const size = select("Size", Object.values(SIZE_OPTIONS), SIZE_OPTIONS.NORMAL);
+    const disabled = boolean("Disabled", false);
+    const option = object("Options", objectOptions);
```
wEprous: I would add `value` knob to playground.

src/Select/__tests__/Select.test.js
```diff
@@ -5,6 +5,8 @@ import { shallow } from "enzyme";
 
 import Select from "../Select";
 
+const mockChange = jest.fn();
+const placeholder = "Default placeholder";
 const objectOptions = [
```
wEprous: `visible` is not needed in this object more.

src/Select/Select.js
```diff
@@ -0,0 +1,207 @@
...
+
+  &::-ms-expand {
+    display: none;
+  }
+
+  /* With prefix */
+  padding-left: ${({ prefix }) => prefix && "48px"}; // TODO: Create token
+
+  /* Based on state of select */
+  border-color: ${({ theme, error }) =>
```
wEprous:
```css
border: 1px solid
```
and delete line 44

src/Select/Select.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  &:disabled {
+    color: ${({ theme }) => theme.colorTextInputDisabled};
+    background: ${({ theme }) => theme.backgroundInputDisabled};
+    cursor: default;
+  }
+`;
+
+const SelectContainer = styled.div`
```
wEprous: `disabled` attr is render for this `div`

src/Select/Select.js
```diff
@@ -0,0 +1,207 @@
...
+  <div className={className}>{children}</div>
+))`
+  display: flex;
+  align-items: center;
+  position: absolute;
+  padding: ${({ theme }) => `0 ${theme.spaceSmall}`};
+  pointer-events: none;
+`;
+
+const SelectSuffix = styled.div`
```
wEprous: `disabled` attr is render for this `div`

src/Select/Select.js
tomashapl: Good catch 👏

src/Select/Select.js
tomashapl: It defines basic style of border, so I want to let it

src/Select/Select.js
tomashapl: Fixed

src/Select/__tests__/Select.test.js
tomashapl: Fixed

src/Select/Select.stories.js
tomashapl: Added

src/Select/Select.js
wEprous: Ok :)

PR #223, opened 2018-06-20
NEW: Textarea component

### LINKED ISSUE #194: Create <Textarea /> component
## Description
Textarea is a form field that we use when we need from users to insert a longer chunk of text.

## Visual style
textarea

Zeplin: https://zpl.io/bAvkD6x

### Interactions
Behavior is same as in InputField, described in issue #153

### Additional information
- Textarea width is 100%
- We agreed that it may be a good idea to use same tokens for visual style as we use for InputField because the look is currently the same (and it’s planned to be the same by design)

## Functional specs
- If the user insert some value, the label color needs to change to a lighter shade (an exact color is defined as a token)
- It shouldn’t be possible to change the width of textarea with browser build-in features
- Default height of the input should be 2 rows, but it should be possible to affect height with props

## Storybook chapters
Probably just default and playground, but we will see after the knobs are defined. It’s possible that we will have one with help text.

## Some inspiration
https://polaris.shopify.com/components/forms/text-field#navigation (they have it only as props for text field which will change it to textarea)
http://designsystem.morningstar.com/components/forms.html#text-area-1
http://seek-oss.github.io/seek-style-guide/textarea
https://v0.material-ui.com/#/components/text-field (similar as Polaris, they have it just as the prop, but their textareas looks always like input and they stretch when user

wEprous:

Closes #194


tomashapl:
@wEprous please can you update tokens ?

src/Textarea/README.md
```diff
@@ -0,0 +1,39 @@
+# Textarea
+To implement Textarea component into your project you'll need to add the import:
+```jsx
+import Textarea from "@kiwicom/orbit-components/lib/Textarea";
+```
+After adding import into your project you can use it simply like:
+```jsx
+<InputField />
```
tomashapl: please change 😄

src/Textarea/Textarea.js.flow
```diff
@@ -0,0 +1,20 @@
...
+  +size?: "small" | "normal",
+  +name?: string,
+  +label?: string,
+  +children?: string,
+  +placeholder?: string,
+  +help?: React$Node,
+  +error?: React$Node,
+  +disabled?: boolean,
+  +maxLength?: number,
+  +onChange?: (ev: SyntheticEvent<HTMLInputElement>) => void,
```
tomashapl: please allow here return Promise

src/Textarea/Textarea.js.flow
```diff
@@ -0,0 +1,20 @@
...
+  +onFocus?: (ev: SyntheticEvent<HTMLInputElement>) => void,
```
tomashapl: also here

src/Textarea/Textarea.js.flow
```diff
@@ -0,0 +1,20 @@
...
+  +onBlur?: (ev: SyntheticEvent<HTMLInputElement>) => void,
```
tomashapl: and here

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,111 @@
...
+      {props.error && (
+        <FormFeedback type="error" theme={theme}>
+          {props.error}
+        </FormFeedback>
+      )}
+    </Field>
+  );
+};
+
+Textarea.defaultProps = {
```
tomashapl: Remove this defaultProps, are defined in constructor

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,111 @@
...
+    !disabled && (error ? theme.boxShadowInputErrorFocus : theme.boxShadowInputFocus)};
+  border-color: ${({ theme, error, disabled }) =>
+    !disabled && (error ? theme.borderColorInputErrorFocus : theme.borderColorInputFocus)};
+  };
+    outline: none;
+  }
+`;
+
+const Textarea = (props: Props) => {
+  const { size = SIZE_OPTIONS.NORMAL, theme = defaultTokens } = props;
```
tomashapl: There is missing prop `disabled`

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,111 @@
...
+    },
+  };
+  return (
+    <Field theme={theme}>
+      {props.label && (
+        <FormLabel theme={theme} filled={!!props.children} disabled={props.disabled}>
+          {props.label}
+        </FormLabel>
+      )}
+      <Area
```
tomashapl: Can we keep conventions of naming ? `StyledTextarea`

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,111 @@
...
+    !disabled && (error ? theme.borderColorInputErrorFocus : theme.borderColorInputFocus)};
+  };
+    outline: none;
+  }
+`;
+
+const Textarea = (props: Props) => {
+  const { size = SIZE_OPTIONS.NORMAL, theme = defaultTokens } = props;
+
+  const tokens = {
```
tomashapl: In future is better create fcis something like `getSize` and `getPadding` that receives needed arguments to calculate and return correct values and handle it in styled-component, for now is OK, but in future i want to refactor it

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+    paddingInput: {
+      [SIZE_OPTIONS.SMALL]: `${theme.spaceXSmall} ${theme.paddingInputSmall}`, // TODO create token "8px 12px"
+      [SIZE_OPTIONS.NORMAL]: `${theme.paddingInputSmall}`, // TODO create token "12px"
+    },
+  };
+  return (
+    <Field theme={theme}>
+      {props.label && (
+        <FormLabel theme={theme} filled={!!props.children} disabled={props.disabled}>
```
tomashapl: disabled is destructed pls use `disabled={disabled}`

src/Textarea/Textarea.js
```diff
@@ -0,0 +1,107 @@
...
+    !disabled && (error ? theme.boxShadowInputErrorFocus : theme.boxShadowInputFocus)};
+  border-color: ${({ theme, error, disabled }) =>
+    !disabled && (error ? theme.borderColorInputErrorFocus : theme.borderColorInputFocus)};
+  };
+    outline: none;
+  }
+`;
+
+const Textarea = (props: Props) => {
+  const { size = SIZE_OPTIONS.NORMAL, theme = defaultTokens, disabled } = props;
```
tomashapl: in default props was `disabled = false` and there isn't fallback `disabled = false` is it OK ?

wEprous: I think so, evaluation of `undefined` is `false`

PR #251, opened 2018-07-03
DEV: Add readonly for types

lspdv:
I am adding readonly for types and adding few missing strict types. Without readonly types users of orbit has to declare also those props which are optional even they don t use them. Adding readonly will fix it for users.


src/Theming/ThemedSample.js
```diff
@@ -3,23 +3,21 @@ import * as React from "react";
...
+import type { Props } from "./ThemedSample";
 
-const ComponentWithTheme = ({ theme }: ComponentProps) => (
+const ComponentWithTheme = ({ theme }: Props) => (
   <div
     style={{
-      color: theme.colorTextPrimary,
-      fontFamily: theme.fontFamily,
-      lineHeight: theme.lineHeightText,
+      color: theme && theme.colorTextPrimary,
```
lspdv: I wanted do simply `const ComponentWithTheme = ({ theme }: Props) => { theme && (<div style={{...` but eslint rules are too strict. If you rather want me add if(theme) and returns instead of checking each defined, let me know.

tomashapl:
@lspdv Please resolve conflicts 🙂

wEprous:
I will solve it cc @tomashapl @lspdv

lspdv:
Ok I am on it.

PR #323, opened 2018-08-22
DEV: Cleaning dependencies

tomashapl:


wEprous:
Also `faker`, `react-icons` and `react-create-component-from-tag-prop`

PR #328, opened 2018-08-28
Add: Support React Node for Alert title

tomashapl:


tomashapl:
@wEprous Updated, thanks for catch 💪

PR #332, opened 2018-08-31
FIX: Radio and CheckBox label type

### LINKED ISSUE #331: <Radio /> pass JSX to the label
I would like to be able to pass JSX to the label of the Radio component so that I can nest another component inside the label. See the screenshot below for an example use case.

To be able to pass JSX as a prop to the label.

Cannot think of any.

**Additional context**

image

wEprous:
`label` prop now accepts `React.Node`

Closes #331


tomashapl:
@wEprous what about InputField, we have to keep this unified

PR #337, opened 2018-09-05
UPDATE: ContactInput, NEW: Asterisk at label

tomashapl:

### DIFF
--- a/src/FormLabel/index.js.flow
+++ b/src/FormLabel/index.js.flow
@@ -4,6 +4,7 @@ export type Props = {
   children: React$Node,
   filled?: boolean,
   disabled?: boolean,
+  required?: boolean,
 };
 
 declare export default React$ComponentType<Props>;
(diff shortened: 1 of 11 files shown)


src/FormLabel/index.js.flow
```diff
@@ -4,6 +4,7 @@ export type Props = {
   children: React$Node,
   filled?: boolean,
   disabled?: boolean,
+  required?: boolean,
 };
```
wEprous: Should be typing for `FormLabel` strict too?

wEprous: Because when someone will import `FormLabel` when needed to use `compact InputField`, then it's necessary to be  strict in flow type, otherwise the person will be able to pass eg. `className`.

tomashapl: I think if someone want to use FormLabel they should read documentation firstly how to use. If there will be more questions about this, we can update, so for now it is OK.

PR #346, opened 2018-09-10
UPDATE: inlineLabel in InputField

tomashapl:
InputField should receive `inlineLabel` prop which renders label on the left side of input


src/InputField/index.js
```diff
@@ -192,17 +205,25 @@ const InputField = (props: Props) => {
...
   return (
     <Field>
-      {props.label && (
-        <FormLabel filled={!!props.value} required={required}>
-          {props.label}
-        </FormLabel>
-      )}
+      {props.label &&
+        !props.inlineLabel && (
+          <FormLabel filled={!!props.value} required={required}>
```
wEprous: You can only create stateless component for `FormLable`
```jsx
const InputLabel = (value, required, label) =>
  <FormLabel filled={!!value} required={required}>
    {label}
  </FormLabel>;
```
and use it here and on the line 222

PR #350, opened 2018-09-13
FIX: Add focus to Button

williamkolmacka:

### DIFF
--- a/src/Button/index.js
+++ b/src/Button/index.js
@@ -139,6 +139,12 @@ export const StyledButton = styled(
     }
   }
 
+  &:enabled:focus {
+    box-shadow: ${({ tokens, bordered, type }) =>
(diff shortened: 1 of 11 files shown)


src/Button/index.js
```diff
@@ -139,6 +139,12 @@ export const StyledButton = styled(
     }
   }
 
+  &:focus {
+    box-shadow: ${({ tokens, bordered, type }) =>
+        bordered && `inset 0 0 0 1px ${tokens.borderColorButton[type]},`}
```
wEprous: Please use a token for borderWidth (everwhere).

williamkolmacka: We don't have created token `borderWidthButton` yet. Which token did you mean?

PR #387, opened 2018-09-26
UPDATE: Add circled to Button, ButtonLink

wEprous:


tomashapl:
@wEprous Can we also add it ButtonLink component please ? I want to prevent future request 😄

wEprous:
@tomashapl We discussed this with @darkwindcz and it's not necessary. ButtonLink is not designed to be `circled` in Sketch.

PR #402, opened 2018-10-01
FIX: Illustration ThemeProvider

tomashapl:
Fixed Illustration getting theme from ThemeProvider

and adjusted tests for more coverage

### DIFF
--- a/src/Illustration/index.js
+++ b/src/Illustration/index.js
@@ -7,8 +7,26 @@ import { SIZE_OPTIONS, baseURL } from "./consts";
 
 import type { Props } from "./index";
 
-export const StyledImage = styled.img`
-  height: ${({ tokens, size }) => tokens.height[size]};
+const getHeightToken = ({ theme, size }) => {
+  const tokens = {
+    [SIZE_OPTIONS.SMALL]: theme.orbit.heightIllustrationSmall,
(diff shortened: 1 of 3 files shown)


src/Illustration/index.js
```diff
@@ -7,8 +7,27 @@ import { SIZE_OPTIONS, baseURL } from "./consts";
...
+
+const getURL = (retina = false) => ({ theme, size, name }) => {
+  const height = parseInt(getHeightToken({ theme, size }), 10);
+  return retina
+    ? `${baseURL}/illustrations/0x${height * 2}/${name}.png 2x`
+    : `${baseURL}/illustrations/0x${height}/${name}.png`;
+};
+
+export const StyledImage = styled.img.attrs({
+  name: false, // Prevent rendering name in DOM
```
wEprous: Maybe we can use this:
```jsx
export const StyledImage = styled(props => (
  <img
    src={getURL()(props)}
    srcSet={getURL(true)(props)}
    className={props.className}
    data-test={props.dataTest}
    alt={props.name}
  />
))
```
The name is still rendering to the DOM and I would stick to one solution - either `attrs` or custom arrow fuction.

wEprous: Or if you wanna stick to this solution, use `name: undefined`

PR #405, opened 2018-10-01
FIX: Checkbox ThemeProvider

tomashapl:
Fixed Checkbox getting theme from ThemeProvider

### DIFF
--- a/src/Checkbox/index.js
+++ b/src/Checkbox/index.js
@@ -3,9 +3,24 @@ import * as React from "react";
 import styled from "styled-components";
 
 import defaultTokens from "../defaultTokens";
+import TOKENS from "./consts";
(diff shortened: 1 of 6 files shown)


src/Checkbox/index.js
```diff
@@ -127,32 +146,21 @@ const Checkbox = (props: Props) => {
...
       <Input
         value={value}
         type="checkbox"
         disabled={disabled}
         checked={checked}
         onChange={onChange}
       />
-      <IconContainer tokens={tokens}>
-        <Check customColor={tokens.iconColor} />
+      <IconContainer disabled={disabled}>
```
wEprous: Unnecessary, it also renders to the DOM.

PR #423, opened 2018-10-02
FIX: TextLink Icon DOM element

### LINKED ISSUE #420: TextLink DOM nesting validation error
React's `validateDOMNesting` method throws a warning:
`index.js:2177 Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`

## Expected Behavior
It should not throw a warning

## Current Behavior
It throws a warning

## Possible Solution
Do not use `div` in TextLink's IconContainer:
```
const IconContainer = styled(({ children, className }) => React.createElement(
  "div", // here
  { className: className },
  children
)).withConfig({
  displayName: "TextLink__IconContainer"
})(["color:", ";transition:color ", " ease-in-out;& svg{height:16px;width:16px;}"], ({ tokens, type }) => tokens.colorTextLink[type], ({ theme }) => theme.orbit.durationFast);
```

## Steps to Reproduce
Nest a TextLink inside a p tag

## Context (Environment)
Trying to put an external link in a paragraph - the link has an icon near it by design

wEprous:
Closes #420


src/TextLink/index.js
```diff
@@ -17,14 +17,14 @@ const getColor = ({ theme, type }) => {
 };
 
 const IconContainer = styled(({ children, className }) => (
-  <div className={className}>{children}</div>
+  <span className={className}>{children}</span>
```
tomashapl: Are we ready to change from block element to inline element ?

PR #424, opened 2018-10-02
Basic storybook setup for react-native

RobinCsl:
This is a proposition for basic setup.

You can run a react-native specific storybook if you go in the folder `storybooknative` and run `yarn`, then `yarn storybook` with `yarn ios` in a different terminal. 

I think components should be add to `src/` inside of `storybooknative` for now; we will need to migrate them later.

An example of component using `styled-components` as well as `flow` types is under `storybooknative/src/Welcome/index.js`.


storybooknative/ios/storybooknative-tvOS/Info.plist
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tbergq: We can delete everything related to tvOS right @ferrannp?

storybooknative/storybook/stories/CenterView/index.js
```diff
@@ -0,0 +1,20 @@
...
+import * as React from "react";
+import { View } from "react-native";
+import styled from "styled-components";
+
+const Main = styled(View)`
+  flex-grow: 1;
+  justify-content: center;
+  align-items: center;
+  background-color: #f5fcff;
+`;
```
tbergq: Also, we could wrap styled like

```
import styled from 'styled-components';

export default (Component: React.Node, style: Object | (() => Object)) =>
  styled(Component)([], style);
```

then we can pass it objects and functions instead of a string

```
import myStyled from './myStyled';

const LolText = myStyled(Text, {
  fontSize: 16,
  color: "pink",
});

const LolText2 = myStyled(Text, props => ({
  color: props.color,
  fontSize: 22,
}));
```

RobinCsl: True and it looks closer to React Native StyleSheets. The question is: which approach is "better"? I chose to stay as close as possible to the existing standards in this codebase. But this is up for discussion.

storybooknative/storybook/stories/Button/index.ios.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import React from "react";
```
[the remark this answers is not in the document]
RobinCsl: Will move the dummy components generated by storybook to `src` folder under `storbybooknative`.

storybooknative/storybook/stories/CenterView/index.js
tbergq: Well, the type of the style could be our StylePropType instead of Object. That way you know you pass it only valid styles. There is nothing in styled components that let's you know if you typed fnotSize instead of font-size e.g

RobinCsl: That makes perfect sense. I agree.

RobinCsl:
@tbergq Can we make the styling a separate issue and merge this, so that people can start developing/extracting stuff?

tbergq:
sure

PR #438, opened 2018-10-08
Wrap styled and use flow to validate styles passed to styled

tbergq:


RobinCsl:
Maybe we should add this to a README.

tbergq:
> Maybe we should add this to a README

Yes, we could do that 😄

PR #444, opened 2018-10-09
UPDATE: FLOW: Restricted Translation in some components

### LINKED ISSUE #331: <Radio /> pass JSX to the label

### LINKED ISSUE #343: Support React Component in Select label prop
I have a `Select` component and want to add a label. The label should use component `Text` from nitro for translations. But `label` prop accepts only string.

Extend `label` prop to accept also `React.Node`.

Is this valid solution? If yes, I would create a PR.

### LINKED ISSUE #371: Allow React.Element as title prop in CardHeader
Currently only **String** is allowed as `title` prop to `<CardHeader/>`
Please allow React.Element / React.Component  as well.

### LINKED ISSUE #711: InputField - Flow typing of `label` should be `React.node`
Flow typing of `label` prop in InputField should be a `React.Node`, now its `string` which is too limiting. Especially when I want to supply translation component.

(It works fine - just the flow type is potentionally wrong).

Orbit 0.23.1 (todays master)

https://github.com/kiwicom/orbit-components/blob/70d55e70eaf46ef285742ea61b15f868f78ce9e7/src/InputField/index.js.flow#L12

tomashapl:
Added flow restriction to some props which can receive only translation via component or string

Closes #371
Closes #343 
Closes #331 
Closes #711 <br/><br/><br/><url>LiveURL: https://orbit-components-restrict-flow-to-translation.surge.sh</url>

### DIFF
--- a/src/InputField/index.js.flow
+++ b/src/InputField/index.js.flow
@@ -1,7 +1,7 @@
 // @flow
 import type { ReactComponentStyled } from "styled-components";
 
-import type { Globals, Ref } from "../common/common.js.flow";
+import type { Globals, Ref, Translation } from "../common/common.js.flow";
 
 export type Size = "small" | "normal";
 
@@ -9,10 +9,10 @@ export type Props = {|
   +size?: Size,
   +type?: "text" | "number" | "email" | "password" | "passportid",
   +name?: string,
-  +label?: string,
+  +label?: Translation,
   +inlineLabel?: boolean,
   +value?: string | number,
-  +placeholder?: string,
+  +placeholder?: Translation,
   +prefix?: React$Node,
   +suffix?: React$Node,
   +help?: React$Node,
--- a/src/InputFile/index.js.flow
+++ b/src/InputFile/index.js.flow
@@ -1,12 +1,12 @@
 // @flow
-import type { Globals, Ref } from "../common/common.js.flow";
+import type { Globals, Translation, Ref } from "../common/common.js.flow";
 
 export type Props = {|
   ...Globals,
-  +label?: string,
-  +title?: string,
+  +label?: Translation,
+  +title?: React$Node,
   +name?: string,
-  +placeholder?: string,
+  +placeholder?: Translation,
   +fileName?: string,
   +allowedFileTypes?: string | string[],
   +help?: React$Node,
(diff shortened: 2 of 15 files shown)


src/InputField/index.js.flow
```diff
@@ -1,14 +1,14 @@
...
 export type Props = {|
   +size?: "small" | "normal",
   +type?: "text" | "number" | "email" | "password",
   +name?: string,
-  +label?: string,
+  +label?: Translation,
   +inlineLabel?: boolean,
   +value?: string,
-  +placeholder?: string,
+  +placeholder?: Translation,
```
wEprous: `placeholder` should only `string` type, right? It's rendered into HTML.

src/Textarea/index.js.flow
```diff
@@ -1,12 +1,12 @@
...
+import type { Globals, Translation } from "../common/common.js.flow";
 
 export type Props = {|
   +size?: "small" | "normal",
   +name?: string,
-  +label?: string,
+  +label?: Translation,
   +value?: string,
-  +placeholder?: string,
+  +placeholder?: Translation,
```
wEprous: The same, `string` probably.

src/common/common.js.flow
```diff
@@ -2,3 +2,5 @@
 export type Globals = {|
   +dataTest?: string,
 |};
+
+export type Translation = React$Element<React$ComponentType<any>> | string;
```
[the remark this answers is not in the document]
wEprous: It makes sense, but probably we don't want to allow passing `ChildrenArray` etc. Those types of props should be strings, even if we need to support translations.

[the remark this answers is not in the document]
oreqizer: simplicity is strong with this one. just do `React.Node`

src/InputFile/index.js.flow
```diff
@@ -1,10 +1,10 @@
...
+import type { Globals, Translation, Ref } from "../common/common.js.flow";
 
 export type Props = {|
   ...Globals,
-  +label?: string,
-  +title?: string,
+  +label?: Translation,
+  +title?: React$Node,
   +name?: string,
   +placeholder?: string,
```
wEprous: Either we should type placeholders with `string` or `Translation`

wEprous: TLDR: choose one unified approach for this prop

PR #471, opened 2018-10-10
Text component

denbalogh:
Set up testing.
lineHeight commented for now.
Updated snapshot and added test-update script
Flow type changed


package.json
```diff
@@ -13,7 +13,7 @@
     "eslint": "eslint .",
     "flow:check": "flow check",
     "test": "jest --config=jest.json",
-    "test-ci": "cd storybooknative; yarn install; cd ..; yarn flow:check && yarn eslint && yarn test --ci",
+    "test-ci": "cd storybooknative; yarn install; cd ..; yarn flow:check && yarn eslint && yarn test --ci; cd storybooknative; yarn test",
```
RobinCsl: I added this; this is hacky for now and because `jest` needs `babel.config.js` when the rest is happy with `.babelrc`

storybooknative/babel.config.js.bak
```diff
@@ -0,0 +1,10 @@
+// @flow
```
RobinCsl: This is necessary as becomes clear when looking at `storybooknative/package.json`

storybooknative/src/Text/__tests__/index.test.js
```diff
@@ -0,0 +1,20 @@
...
+import * as React from "react";
+import renderer from "react-test-renderer";
+
+import Text from "../index";
+
+describe("Text", () => {
+  const text = "Children text";
+  const component = renderer
+    .create(
+      <Text type="primary" size="small">
```
tbergq: since this is a component library, it makes sense to add tests for all different combinations of props

storybooknative/src/Text/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,791 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Text test1 should match snapshot 1`] = `
+<Text
+  align="center"
+  className="sc-bdVaJa clsgSE"
```
tbergq: why is it rendering className?

RobinCsl: Maybe because we would need to realise two sets of tests, or rather run them twice, one time for each environment.

RobinCsl: From https://stackoverflow.com/questions/41482306/how-to-run-jest-test-for-react-native-and-web

>Assuming you using react-native-web to get your HTML output, you should know that Jest won't actually run your react-native code through webpack and thus it won't handle the react-native -> react-native-web alias, so you'll never see the actual HTML in your snapshot.

in which case, it does not really matter that this className is given to us, but it will be of importance for the web.

RobinCsl:
@tbergq Can you review this? If approved, it can be merged in my opinion

storybooknative/package.json
```diff
@@ -4,7 +4,10 @@
   "private": true,
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
-    "test": "jest",
+    "test-hack-setup": "mv .babelrc .babelrc.bak; mv babel.config.js.bak babel.config.js;",
+    "test-hack-teardown": "mv .babelrc.bak .babelrc; mv babel.config.js babel.config.js.bak;",
+    "test": "yarn test-hack-setup; jest; yarn test-hack-teardown",
+    "test-update": "yarn test-hack-setup; jest -u; yarn test-hack-teardown",
```
tbergq: could all of this be part of a node-script instead, so we don't need to pollute the package.json file?

storybooknative/src/Text/index.js
```diff
@@ -0,0 +1,95 @@
...
+  lineHeight: props.theme.orbit.lineHeightText,
+  textAlign: props.align,
+  textTransform: props.uppercase ? `uppercase` : `none`,
+  fontStyle: props.italic ? `italic` : `normal`,
+  margin: 0,
+}));
+
+StyledText.defaultProps = {
+  theme: {
+    orbit: {
```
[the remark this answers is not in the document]
wEprous: Yeah, I see.

Looking into the test of this component, the only thing which is tested is props passing.
I would rather reduce the test only to one component (renderer), instead of a ton of different combinations, and would use only one snapshot with full `theme` prop.

Also, is there any chance to test visual styles with `jest-styled-component` and `expect(component).toHaveStyleRule("property", "value")`? We just start to use it and we wanna have it in all tests in the future.

See [this](https://github.com/kiwicom/orbit-components/blob/b90e88dbdec8d83db4259065121b553872917f33/src/Icon/__tests__/index.test.js#L21).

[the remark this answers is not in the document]
denbalogh: Oops, I fixed it already. How do you revert?

storybooknative/src/Text/__tests__/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+  Object.keys(TYPE_OPTIONS).forEach(type => {
+    Object.keys(SIZE_OPTIONS).forEach(size => {
+      Object.keys(WEIGHT_OPTIONS).forEach(weight => {
+        Object.keys(ALIGN_OPTIONS).forEach(alignment => {
+          test(`${type}-${size}-${weight}-${alignment} should match the snapshot`, () => {
+            expect(
+              renderer
+                .create(
+                  <Text type={type} size={size} weight={weight} align={alignment}>
```
wEprous: @denbalogh, you're passing the `key` to the component and not the `value`. I'm curious why the Flow doesn't throw an error? cc @RobinCsl

RobinCsl: Actually, I wrote this and you're right, it should not be the key but the corresponding value. That's really odd.

storybooknative/src/Text/index.js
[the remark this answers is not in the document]
RobinCsl: It's fine, leave as it is for now. We need to discuss of how we want to move forward.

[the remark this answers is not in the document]
RobinCsl: We will try to incorporate `jest-styled-component` testing with `toHaveStyleRule` in a different PR. It's already becoming too big, and not just about the Text component.

RobinCsl:
@tbergq @wEprous Is it mergeable now?

tbergq:
please fix merge conflicts

PR #493, opened 2018-10-15
FIX: InputGroup without label

wEprous:
bug
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-inputgroup-without-label.surge.sh</url>


tomashapl:
Should we create some story in storybook with this case for visual tests ? 🙏

PR #498, opened 2018-10-16
Block prop + flow fixes

denbalogh:
Solution for block prop.
Fixed flow errors.

### DIFF
--- a/storybooknative/src/Touchable.js
+++ b/storybooknative/src/Touchable.js
@@ -18,6 +18,7 @@ type Props = {|
   // Should the ripple render outside of the view bounds?
   +borderlessRipple?: boolean,
   +rippleColor?: string,
+  +width?: string | number,
   ...AccessibilityProps,
 |};
 
@@ -33,6 +34,7 @@ export default class Touchable extends React.Component<Props> {
     rippleColor: "rgba(0, 0, 0, .32)",
     disabled: false,
     accessibilityRole: "button",
+    width: "auto",
   };
 
   /**
@@ -63,7 +65,7 @@ export default class Touchable extends React.Component<Props> {
       return (
         <TouchableNativeFeedback
           {...this.props}
-          style={null}
+          style={{ width: this.props.width }}
           useForeground={useForeground}
           background={TouchableNativeFeedback.Ripple(
             this.props.rippleColor,
@@ -79,6 +81,7 @@ export default class Touchable extends React.Component<Props> {
       <TouchableOpacity
         activeOpacity={0.5}
         disabled={this.props.disabled}
+        style={{ width: this.props.width }}
         accessibilityRole="button"
       >
         {children}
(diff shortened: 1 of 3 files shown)


storybooknative/src/Touchable.js
```diff
@@ -18,6 +18,7 @@ type Props = {|
   // Should the ripple render outside of the view bounds?
   +borderlessRipple?: boolean,
   +rippleColor?: string,
+  +width?: string | number,
```
RobinCsl: I don't really like mixed types like this. @tbergq can you think of a better way?

storybooknative/src/Touchable.js
```diff
@@ -63,7 +65,7 @@ export default class Touchable extends React.Component<Props> {
       return (
         <TouchableNativeFeedback
           {...this.props}
-          style={null}
+          style={{ width: this.props.width }}
```
RobinCsl: For styles, we usually want to use:
```
const styles = StyleSheet.create({
   customWidth: {
      width: this.props.width,
   }
};
```
and then use as 
```
<Component style={styles.customWidth} />
```

RobinCsl: The reason is that `StyleSheet.create` makes it more performant.

storybooknative/src/Touchable.js
tbergq: Why not use flex instead?

storybooknative/src/Touchable.js
```diff
@@ -51,6 +53,11 @@ export default class Touchable extends React.Component<Props> {
     // but since we are building for both platforms it is here as well
     // to discover this mistake as soon as possible
     const children = React.Children.only(this.props.children);
+    const styles = StyleSheet.create({
```
tbergq: Also, I think we should stick to styled components to handle the styling.

RobinCsl: Absolutely, good catch

storybooknative/src/Touchable.js
denbalogh: Do you mean alignSelf: "stretch"?

tbergq: I mean flex: 1, so it will take full width. If we want it to take any less, we can wrap it with another view, and it will by default take the full width of that view

RobinCsl: `flex: 1` should flex only in the flexDirection

tbergq: But I think our buttons have a fixed height, so it should respect that, then grow only in the width

denbalogh: I tried your solution, but in emulator it actually stretches the button in its height and not width. 
I don't know what is happening or where I made an error.

tbergq: Yes, you are right, it was a bad advice, sorry 🙈

PR #500, opened 2018-10-16
Bug android ripple border radius

denbalogh:
It seems to work, but it's an ugly solution. Would appreciate some suggestions to improve.

### DIFF
--- a/storybooknative/src/Accessibility.js
+++ b/storybooknative/src/Accessibility.js
@@ -18,7 +18,7 @@ export type AccessibilityProps = {|
   accessible?: boolean,
   accessibilityLabel?: string,
   accessibilityHint?: string,
-  accessibilityRole?: AccessibilityRoleValues,
+  accessibilityRole?: AccessibilityRoleValues | $ReadOnlyArray<AccessibilityRoleValues>,
   accessibilityState?: "selected" | "disabled",
   accessibilityViewIsModal?: boolean, // iOS only
   accessibilityIgnoresInvertColors?: boolean, // iOS only
--- a/storybooknative/src/Accessibility.js
+++ b/storybooknative/src/Accessibility.js
@@ -18,7 +18,7 @@ export type AccessibilityProps = {|
   accessible?: boolean,
   accessibilityLabel?: string,
   accessibilityHint?: string,
-  accessibilityRole?: AccessibilityRoleValues | $ReadOnlyArray<AccessibilityRoleValues>,
+  accessibilityRole?: AccessibilityRoleValues,
   accessibilityState?: "selected" | "disabled",
   accessibilityViewIsModal?: boolean, // iOS only
   accessibilityIgnoresInvertColors?: boolean, // iOS only
--- a/storybooknative/src/Button/index.js
+++ b/storybooknative/src/Button/index.js
@@ -7,9 +7,9 @@ import styled from "../styled";
 import defaultTokens from "../defaultTokens";
 import { TYPE_OPTIONS, SIZE_OPTIONS, TOKENS } from "./consts";
 
-import type { Props } from "./index";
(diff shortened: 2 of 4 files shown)


storybooknative/src/Touchable.js
```diff
@@ -2,9 +2,13 @@
 
 import * as React from "react";
 // eslint-disable-next-line no-restricted-imports
-import { Platform, TouchableNativeFeedback, TouchableOpacity } from "react-native";
+import { Platform, TouchableNativeFeedback, TouchableOpacity, View } from "react-native";
 
 import type { AccessibilityProps } from "./Accessibility";
+import styled from "./styled";
+import { TOKENS } from "./Button/consts";
```
RobinCsl: I don't think I would add anything related to `Button` inside this file. This is meant to be a quite general building block for other components.
Also, since this is specifically for Android, maybe you could just create a wrapper for the Button component for Android only, using index.android.js or something using `Platform` helper from `react-native`.

storybooknative/src/Button/index.js
```diff
@@ -9,7 +9,7 @@ import { TYPE_OPTIONS, SIZE_OPTIONS, TOKENS } from "./consts";
 
 import type { Props } from "./index";
 
-const getSizeToken = (name, { theme, size }) => {
+export const getSizeToken = (name: string, { theme, size }) => {
```
RobinCsl: Do you need the `export`?

RobinCsl: And why is `name` of type `string`? Shouldn't it be more precise? Like just the possibilities from `TOKENS.heightButton`, or something like this.

storybooknative/src/Button/index.js
```diff
@@ -29,7 +29,7 @@ const getSizeToken = (name, { theme, size }) => {
   };
   return tokens[name][size];
 };
-const getTypeToken = (name, { theme, type }) => {
+const getTypeToken = (name: string, { theme, type }) => {
```
RobinCsl: Same question here about `name` being typed as a `string`.

storybooknative/src/Button/index.js
```diff
@@ -132,6 +144,24 @@ const Button = (props: Props) => {
     width = props.width || "auto";
   }
 
+  if (Platform.OS === "android") {
```
RobinCsl: That's much better. 
However, are you sure that all the components need access to `...rest`?
Maybe to improve even further, this could be in a different file `index.android.js`, so that the bundle size for other platform is not bloated with non-necessary code. @tbergq What do you think?

tbergq: I think that the .android and .ios suffixes should be used with care and only when utmost needed. Like the datepicker component. 

I doubt the effect on the bundle size is that significant.

storybooknative/src/Accessibility.js
```diff
@@ -18,7 +18,7 @@ export type AccessibilityProps = {|
   accessible?: boolean,
   accessibilityLabel?: string,
   accessibilityHint?: string,
-  accessibilityRole?: AccessibilityRoleValues | $ReadOnlyArray<AccessibilityRoleValues>,
+  accessibilityRole?: AccessibilityRoleValues,
```
RobinCsl: Why do you remove this?

storybooknative/src/Button/index.js
```diff
@@ -9,7 +9,9 @@ import { TYPE_OPTIONS, SIZE_OPTIONS, TOKENS } from "./consts";
 
 import type { Props } from "./index";
 
-const getSizeToken = (name, { theme, size }) => {
+type Tokens = $Keys<typeof TOKENS>;
```
RobinCsl: I believe all the types should live in the `index.js.flow` file.

PR #507, opened 2018-10-18
React-native - Button component

denbalogh:


storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,181 @@
...
+
+const Button = (props: Props) => {
+  const {
+    onPress,
+    children,
+    size = SIZE_OPTIONS.NORMAL,
+    type = TYPE_OPTIONS.PRIMARY,
+    ...rest
+  } = props;
+  const onPressHandler = onPress || (() => {});
```
RobinCsl: Maybe you could instead do:
```
const {
  onPress = () => {},
  children, 
 ...
};
```
and remove that line.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,181 @@
...
+        width={props.width}
+        circled={props.circled}
+      >
+        <Touchable
+          disabled={props.disabled || !props.onPress}
+          onPress={onPressHandler}
+          width={width}
+        >
+          <StyledButton size={size} type={type} {...rest}>
+            <StyledText size={size} type={type} bordered={props.bordered}>
```
RobinCsl: I am also wondering here if it would be possible to reuse the `Text` Orbit component.

denbalogh: They don't use the same tokens for text color.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,176 @@
...
+export const StyledButton = styled(View, props => {
+  const { bordered, theme, block, width, circled, disabled } = props;
+  return {
+    justifyContent: "center",
+    alignItems: "center",
+    height: getSizeToken(TOKENS.heightButton, props),
+    backgroundColor: bordered
+      ? theme.orbit.backgroundButtonBordered
+      : getTypeToken(TOKENS.backgroundButton, props),
+    width: block ? "100%" : width || "auto",
```
RobinCsl:
```suggestion
    width: block ? "100%" : `${width}px` || "auto",
```
as it does not work in `storybook-web` otherwise.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,176 @@
...
+  };
+});
+
+StyledText.defaultProps = {
+  theme: defaultTokens,
+};
+
+const StyledTouchableWrapper = styled(View, props => ({
+  overflow: "hidden",
+  width: props.block ? "100%" : props.width || "auto",
```
RobinCsl:
```suggestion
  width: props.block ? "100%" : `${props.width}px` || "auto",
```
probably here too.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,176 @@
...
+    children,
+    size = SIZE_OPTIONS.NORMAL,
+    type = TYPE_OPTIONS.PRIMARY,
+    ...rest
+  } = props;
+  let width;
+  if (props.block) {
+    width = "100%";
+  } else {
+    width = props.width || "auto";
```
RobinCsl:
```suggestion
    width = props.width ? `${props.width}px` : "auto";
```
and here too.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,180 @@
...
+  } else {
+    width = props.width ? `${props.width}px` : "auto";
+  }
+
+  if (Platform.OS === "android") {
+    return (
+      <StyledTouchableWrapper
+        size={size}
+        block={props.block}
+        width={props.width}
```
RobinCsl: Here you are using `props.width`, when in the component below we are using `width`. Could you fix that?

storybooknative/src/Button/index.js.flow
```diff
@@ -0,0 +1,36 @@
...
+  +circled?: boolean,
+  +bordered?: boolean,
+  +disabled?: boolean,
+  +block?: boolean,
+  +type?: Type,
+  +size?: Size,
+  +width?: number,
+|};
+
+declare export var StyledButton: React$ComponentType<Props>;
```
RobinCsl: Is it necessary to `declare export var`? Maybe just a 
```suggestion
export type StyledButton = React$ComponentType<Props>;
```
would be enough.

storybooknative/src/Button/index.js.flow
```diff
@@ -0,0 +1,36 @@
...
+
+declare export default React$ComponentType<Props>;
```
RobinCsl: Same question here? Maybe just remove it entirely.

storybooknative/src/Button/index.js.flow
RobinCsl: Okay, maybe ignore this comment since they do
```
declare export var StyledButton: React$ComponentType<Props>;

declare export default React$ComponentType<Props>;
```
in the web Orbit Button component.

storybooknative/src/Accessibility.js
```diff
@@ -0,0 +1,30 @@
+// @flow strict
```
RobinCsl: This file should probably be renamed `accessibility.flow.js` or similar, as one would believe this is a component called Accessibility.

storybooknative/src/Button/index.js
```diff
@@ -1,13 +1,171 @@
...
+    ...rest
+  } = props;
+  let width;
+  if (props.block) {
+    width = "100%";
+  } else {
+    width = props.width ? `${props.width}px` : "auto";
+  }
+
+  if (Platform.OS === "android") {
```
RobinCsl: Should this live in a different _platform-specific_ file? The button component will be used everywhere, so I don't know if that makes sense that this `if` statement is performed every time.

RobinCsl: Okay, let's say for now this is a possible improvement which does not need to be implemented yet.

storybooknative/src/Button/index.js.flow
```diff
@@ -0,0 +1,32 @@
...
+  | "critical"
+  | "facebook"
+  | "google";
+
+export type Size = "small" | "normal" | "large";
+
+export type Tokens = $Keys<typeof TOKENS>;
+
+export type Props = {|
+  +children?: React.Element<Text> | string,
```
tbergq: What if we pass it our Translation component?

RobinCsl: So it probably should be `React.Node` to be more permissive. Or can't we create some `abstract` type, like `Text`, and then leave the implementation/description of this type to the people using the library?

denbalogh: How would you do it syntactically?

tbergq: Why limit it to text, I could pass a new inside a button as well. For me, React.Node is fine.

storybooknative/src/Button/index.js.flow
RobinCsl: This was actually useful!

package.json
```diff
@@ -13,7 +13,7 @@
     "eslint": "eslint .",
     "flow:check": "flow check",
     "test": "jest --config=jest.json",
-    "test-ci": "cd storybooknative; yarn install; cd ..; yarn flow:check && yarn eslint && yarn test --ci; cd storybooknative; yarn test",
+    "test-ci": "cd storybooknative; yarn install; cd ..; (yarn flow:check && yarn eslint && yarn test --ci) || exit $?; cd storybooknative; yarn test",
```
RobinCsl: This should be a non-issue once we merge into `master`.

storybooknative/src/CenterView/index.js
```diff
@@ -4,13 +4,16 @@ import * as React from "react";
 import { View } from "react-native";
 
 import styled from "../styled";
+import type { StylePropType } from "../types/StylesPropType";
 
-const Main = styled(View, {
+const style: StylePropType = {
```
tbergq: if we need to do this, we should export StylePropType from same file as styled, since they will often be used together

storybooknative/src/CenterView/index.js
```diff
@@ -4,13 +4,16 @@ import * as React from "react";
...
   flexGrow: 1,
   justifyContent: "center",
   alignItems: "center",
   backgroundColor: "#f5fcff",
-});
+};
+
+const Main = styled(View, style);
```
tbergq: also it could be used like: 
```
const Main = styled(
  View,
  ({
    flexGrow: 1,
    justifyContent: "center",
    alignItems: "center",
    backgroundColor: "#f5fcff",
  }: StylePropType),
);
```

both is fine by me

storybooknative/src/Touchable/index.js
```diff
@@ -0,0 +1,83 @@
...
+import * as React from "react";
+// eslint-disable-next-line no-restricted-imports
+import { Platform, TouchableNativeFeedback, TouchableOpacity } from "react-native";
+
+import styled from "../styled";
+import type { StylePropType } from "../types/StylesPropType";
+
+import type { Props } from "./index";
+
+const StyledTouchableOpacity = styled(TouchableOpacity, (props: Props): StylePropType => ({
```
tbergq: `(props: Props)` this is not correct, it can only be passed width.

storybooknative/src/Touchable/index.js.flow
```diff
@@ -0,0 +1,51 @@
...
+  accessibilityIgnoresInvertColors?: boolean, // iOS only
+  accessibilityElementsHidden?: boolean, // iOS only
+  onAccessibilityTap?: () => void, // iOS only
+  onMagicTap?: () => void, // iOS only
+  accessibilityLiveRegion?: "none" | "polite" | "assertive", // Android only
+  importantForAccessibility?: "auto" | "yes" | "no" | "no-hide-descendants", // Android only
+|};
+
+export type Props = {|
+  +children: React.Node,
```
tbergq: Shouldn't react be imported int this file?

storybooknative/src/types/StylesPropType.js
```diff
@@ -6,7 +6,11 @@
...
 
 import {
-  type ____DangerouslyImpreciseStyleProp_Internal, // eslint-disable-line camelcase
+  type ____DangerouslyImpreciseStyle_Internal, // eslint-disable-line camelcase
 } from "react-native/Libraries/StyleSheet/StyleSheetTypes";
+import AnimatedNode from "react-native/Libraries/Animated/src/nodes/AnimatedNode";
 
-export type StylePropType = ____DangerouslyImpreciseStyleProp_Internal; // eslint-disable-line camelcase
+export type StylePropType = {
+  ...____DangerouslyImpreciseStyle_Internal, // eslint-disable-line camelcase
```
tbergq: maybe just `/* eslint-disable camelcase */` at the top of the file?

storybooknative/src/types/StylesPropType.js
```diff
@@ -6,7 +6,11 @@
...
+  borderWidth?: number | AnimatedNode | string,
+};
```
tbergq:
```
type ExactStyles = $Exact<____DangerouslyImpreciseStyle_Internal>;
type ShapedStyles = $Shape<ExactStyles>;

export type StylePropType = {|
  ...ShapedStyles,
  +borderWidth?: number | AnimatedNode | string,
|};
```
now we have exact and optional styles.

storybooknative/src/CenterView/index.js
RobinCsl: I find this
```
const Main = styled(
  View,
  ({
    flexGrow: 1,
    justifyContent: "center",
    alignItems: "center",
    backgroundColor: "#f5fcff",
  }: StylePropType),
);
``` 
less readable, so I will stick with first option.

storybooknative/src/types/StylesPropType.js
RobinCsl: However:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ storybooknative/src/types/Style.js:14:6

Inexact ____DangerouslyImpreciseStyle_Internal [1] is incompatible with exact Style [2].

     11│ import AnimatedNode from "react-native/Libraries/Animated/src/nodes/AnimatedNode";
     12│
 [2] 13│ export type Style = {|
 [1] 14│   ...____DangerouslyImpreciseStyle_Internal,
     15│   +borderWidth?: number | AnimatedNode | string,
     16│ |};
     17│

```

storybooknative/src/Touchable/index.js
RobinCsl: I am not sure this is true actually as we have below:

```
      return (
        <StyledTouchableNativeFeedback
          {...this.props}
          useForeground={useForeground}
          background={TouchableNativeFeedback.Ripple(
            this.props.rippleColor,
            this.props.borderlessRipple,
          )}
        >
          {children}
        </StyledTouchableNativeFeedback>
      );
    }

    return (
      <StyledTouchableOpacity activeOpacity={0.5} accessibilityRole="button" {...this.props}>
        {children}
      </StyledTouchableOpacity>
    );
```

so we are passing them all props.

tbergq: Right, that is true, I forgot how it works 😄

storybooknative/src/types/StylesPropType.js
tbergq: I think you need to stop and start flow server

RobinCsl: I read your comment too fast and forgot the first part 🥇

PR #508, opened 2018-10-18
React-native - TextInput code extracted from kiwi/mobile

denbalogh:


storybooknative/src/InputField/index.js
```diff
@@ -0,0 +1,32 @@
...
+|};
+
+const StyledLabel = styled(Text, props => {});
+const StyledTextInput = styled(TextInput, props => {});
+
+const InputField = (props: Props) => {
+  const { label, ...rest } = props;
+  return (
+    <React.Fragment>
+      {label != null && <StyledLabel>{label}</StyledLabel>}
```
RobinCsl: I am just wondering whether you should be reusing the `Text` Orbit component here.

storybooknative/src/InputField/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+import * as React from "react";
+import { TextInput } from "react-native"; // eslint-disable-line no-restricted-imports
+
+import Text from "../Text";
+import styled from "../styled";
+
+type Props = {|
+  +placeholder?: React.Element<typeof Text>,
+  +label?: React.Element<typeof Text>,
```
RobinCsl: If the label is already of type `Text`, then it probably is not necessary to wrap it in a `Text` component too.

storybooknative/src/InputField/index.js
```diff
@@ -0,0 +1,32 @@
...
+  +maxLength?: number,
+|};
+
+const StyledTextInput = styled(TextInput, props => {});
+
+const InputField = (props: Props) => {
+  const { label, ...rest } = props;
+  return (
+    <React.Fragment>
+      {label && <Text>{label}</Text>}
```
RobinCsl: I think it was better to keep the label as being of type `Text`, because then we could use `Translate` component and the flow types would match.

storybooknative/src/InputField/index.js
```diff
@@ -0,0 +1,32 @@
+// @flow strict
+
+import * as React from "react";
+import { TextInput } from "react-native"; // eslint-disable-line no-restricted-imports
+
+import Text from "../Text";
+import styled from "../styled";
+
+type Props = {|
+  +placeholder?: string,
```
RobinCsl: The placeholder should also have been left to be of type Text I believe.

storybooknative/src/InputField/index.js
RobinCsl: And so now you can remove the Text tags I guess.

storybooknative/src/InputField/index.js
denbalogh: In docs, placeholder should be of type string. I thought that I would leave it like that for now, and change it when I make placeholder as a separate Text component, if it comes to that.

PR #514, opened 2018-10-22
Basic Input structure

denbalogh:

### DIFF
--- a/storybooknative/src/InputField/InputField.stories.js
+++ b/storybooknative/src/InputField/InputField.stories.js
@@ -1,0 +1,27 @@
+// @flow
+
+import React from "react";
+import { action } from "@storybook/addon-actions";
+import { text } from "@storybook/addon-knobs";
+
+import { storiesOf } from "../../helpers/storiesOf";
+import CenterView from "../../src/CenterView";
+import Text from "../Text";
+
+import InputField from "./index";
+
+storiesOf("InputField", module)
+  .addDecorator(getStory => <CenterView>{getStory()}</CenterView>)
+  .add("with text", () => {
+    const label = <Text>Label</Text>;
+    const value = text("Value", "");
+    const placeholder = text("Placeholder", "Placeholder");
+    return (
+      <InputField
+        label={label}
+        value={value}
+        placeholder={placeholder}
+        onChange={action("change")}
+      />
+    );
+  });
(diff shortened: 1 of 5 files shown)


storybooknative/src/InputField/InputField.stories.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+/* eslint-disable jsx-a11y/accessible-emoji, import/no-unresolved, import/extensions */
```
tbergq: Why do we need this eslint-disables?

storybooknative/src/InputField/index.js
```diff
@@ -1,30 +1,113 @@
...
-  const { label, ...rest } = props;
+  const {
+    size = SIZE_OPTIONS.NORMAL,
+    type = TYPE_OPTIONS.TEXT,
+    label,
+    help,
+    error,
+    inlineLabel,
+    prefix,
+    sufix,
```
tbergq: suffix?

storybooknative/src/InputField/index.js.flow
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import * as React from "react";
+
+import Text from "../Text";
+
+export type Props = {|
+  +size?: "small" | "normal",
+  +type?: "text" | "number" | "email" | "password",
+  +defaultValue?: string,
```
wEprous: What's the reason to have `defaultValue` and `value` at one time? 🙂

wEprous: Thanks @tbergq :) I thought that it has some use case 👍

storybooknative/src/InputField/index.js
```diff
@@ -1,30 +1,113 @@
 // @flow strict
 
 import * as React from "react";
-import { TextInput } from "react-native"; // eslint-disable-line no-restricted-imports
+import { TextInput, View } from "react-native"; // eslint-disable-line no-restricted-imports
```
tbergq: I don't think you have restricted imports for this project. You can delete this eslint-disable

storybooknative/src/InputField/index.js
```diff
@@ -1,30 +1,113 @@
 // @flow strict
```
tbergq: Also, there is no flow strict setup for orbit, so `@flow` will suffice.

PR #519, opened 2018-10-23
Label styles + inline label prop

denbalogh:

### DIFF
--- a/storybooknative/src/CenterView/index.js
+++ b/storybooknative/src/CenterView/index.js
@@ -4,18 +4,20 @@ import * as React from "react";
 import { View } from "react-native";
 
 import styled from "../styled";
+import { AlignItems } from "../Types";
 
-const Main = styled(View, {
+const Main = styled(View, props => ({
   flexGrow: 1,
   justifyContent: "center",
-  alignItems: "center",
+  alignItems: props.align || "center",
   backgroundColor: "#f5fcff",
-});
+}));
 
 type Props = {|
-  children: React.ChildrenArray<React.Node> | null,
+  +children: React.ChildrenArray<React.Node> | null,
+  +align: AlignItems,
 |};
 
 export default function CenterView(props: Props) {
-  return <Main>{props.children}</Main>;
+  return <Main {...props}>{props.children}</Main>;
 }
--- a/storybooknative/src/FormLabel/index.js
+++ b/storybooknative/src/FormLabel/index.js
@@ -1,0 +1,51 @@
+// @flow
+import React from "react";
+import { Text } from "react-native";
+
+import styled from "../styled";
+import defaultTokens from "../defaultTokens";
+
+import type { Props, AsteriskProps } from "./index";
+
+const StyledAsteriskWrapper = styled(Text, props => {
+  const { theme, filled } = props;
+  return {
+    fontWeight: theme.orbit.fontWeightBold,
+    color: filled ? theme.orbit.colorFormLabelFilled : theme.orbit.colorTextError,
+    fontSize: theme.orbit.fontSizeFormLabel,
+  };
+});
+
+StyledAsteriskWrapper.defaultProps = {
+  theme: defaultTokens,
+};
+
+export const StyledAsterisk = ({ filled }: AsteriskProps) => (
+  <StyledAsteriskWrapper filled={filled}>* </StyledAsteriskWrapper>
+);
+
+const StyledLabel = styled(Text, props => {
+  const { theme, disabled, filled } = props;
+  return {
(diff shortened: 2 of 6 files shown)


storybooknative/src/CenterView/index.js
```diff
@@ -5,17 +5,18 @@ import { View } from "react-native";
...
   justifyContent: "center",
-  alignItems: "center",
+  alignItems: props.align || "center",
   backgroundColor: "#f5fcff",
-});
+}));
 
 type Props = {|
   children: React.ChildrenArray<React.Node> | null,
+  align: "flex-start" | "center" | "flex-end" | "stretch",
```
tbergq: Props should be marked readonly `+align: "flex-start" | "center" | "flex-end" | "stretch"`

Also, maybe we should extract this align-enum somewhere and export it. It could be used by other components in the future also?

storybooknative/src/FormLabel/index.js
```diff
@@ -0,0 +1,47 @@
...
+StyledAsterisk.defaultProps = {
+  theme: defaultTokens,
+};
+
+const StyledLabel = styled(Text, props => {
+  const { theme, disabled, filled } = props;
+  return {
+    fontSize: theme.orbit.fontSizeFormLabel,
+    color: !filled || disabled ? theme.orbit.colorFormLabel : theme.orbit.colorFormLabelFilled,
+    // lineHeight: theme.orbit.lineHeightText, #TODO remove comment before merge
```
tbergq: why is this commented?

storybooknative/src/FormLabel/index.js.flow
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+export type Props = {
+  children: React$Element,
+  filled?: boolean,
+  disabled?: boolean,
+  required?: boolean,
```
tbergq: read-only

storybooknative/src/InputField/index.js
```diff
@@ -70,8 +91,8 @@ const StyledTextInput = styled(TextInput, props => {
     color: disabled ? theme.orbit.colorTextInputDisabled : theme.orbit.colorTextInput,
     fontSize: getToken(TOKENS.fontSizeInput, props),
     width: "100%",
-    height: "100%",
-    lineHeight: getToken(TOKENS.heightInput, props),
+    height: getToken(TOKENS.heightInput, props),
+    // lineHeight: getToken(TOKENS.heightInput, props), #TODO remove comment before merge
     zIndex: 2,
```
tbergq: Don't we have some zIndex values in orbit-tokens?

storybooknative/src/InputField/index.js
```diff
@@ -94,14 +115,23 @@ const InputField = (props: Props) => {
...
+  const isFilled = !!props.value;
   return (
     <React.Fragment>
-      {label !== undefined && label}
+      {label && !inlineLabel && <FormLabel label={label} isFilled={isFilled} required={required} />}
       <StyledInputContainer size={size} disabled={disabled} error={error}>
+        {label &&
+          inlineLabel && (
+            <StyledInlineLabel size={size} filled={isFilled} disabled={disabled}>
+              {required && <StyledAsterisk filled={isFilled}>* </StyledAsterisk>}
```
tbergq: Maybe <StyledAsterisk could be exported with the `*` so we don't have to fill it every time?

```
Const StyledAsteriskWrapper = styled(Text, {...});
export StyledAsterisk = props => <StyledAsteriskWrapper {...props}>* </StyledAsteriskWrapper>
```

storybooknative/src/InputField/index.js
denbalogh: We do but not for this case. These values are from the web version.

tbergq: And why does it matter that they are from the web version?

denbalogh: I thought that web and RN should be as consistent as possible.

storybooknative/src/FormLabel/index.js.flow
```diff
@@ -1,8 +1,12 @@
...
-  filled?: boolean,
-  disabled?: boolean,
-  required?: boolean,
+  +children: React$Element,
+  +filled?: boolean,
+  +disabled?: boolean,
+  +required?: boolean,
+};
+
+export type asteriskProps = {
```
tbergq: `AsteriskProps`

storybooknative/src/InputField/index.js
tbergq: Ok, let me rephrase, we can't we use zIndex from orbit-tokens?

I am not saying this is wrong here, I just don't understand completely why 🙂

PR #521, opened 2018-10-24
Error and help messages.

denbalogh:

### DIFF
--- a/storybooknative/src/InputField/index.js.flow
+++ b/storybooknative/src/InputField/index.js.flow
@@ -1,9 +1,5 @@
 // @flow
 
-import * as React from "react";
-
-import Text from "../Text";
-
 export type Props = {|
   +size?: "small" | "normal",
   +type?: "text" | "number" | "email" | "password",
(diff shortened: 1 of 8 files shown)


storybooknative/src/InputField/index.js.flow
```diff
@@ -1,24 +1,20 @@
...
-
 export type Props = {|
   +size?: "small" | "normal",
   +type?: "text" | "number" | "email" | "password",
   +defaultValue?: string,
   +autoFocus?: boolean,
   +keyboardType?: "email-address" | "numeric" | "number-pad",
   +secureTextEntry?: boolean,
-  +label?: React.Element<typeof Text>,
+  +label?: string,
```
tbergq: These are props that the consumer will pass in, this will not work with our setup, since we use translation component. It should also work for components. 

Same goes for help and error I guess.

storybooknative/src/InputField/index.js.flow
```diff
@@ -15,10 +14,10 @@ export type Props = {|
   +inlineLabel?: boolean,
   +value?: string,
   +placeholder?: string,
```
tbergq: placeholder also needs to be a component, see 
https://github.com/kiwicom/mobile/blob/master/packages/shared/src/forms/TextInput.js

storybooknative/src/InputField/index.js.flow
```diff
@@ -15,10 +14,10 @@ export type Props = {|
...
-  +prefix?: React$Element,
-  +suffix?: React$Element,
-  +help?: React$Element,
-  +error?: React$Element,
+  +prefix?: React$Node,
+  +suffix?: React$Node,
+  +help?: React.Element<typeof Text> | string,
```
tbergq: This is also too specific, we cannot pass our Translation component inn here

storybooknative/src/InputField/index.js.flow
denbalogh: I will do a separate PR for placeholder.

PR #539, opened 2018-10-25
UPDATE: React.Node for label in Radio and Checkbox

wEprous:
The `label` prop opened for `React.Node`, for a case when TextLink is needed.

<br/><br/><br/><url>LiveURL: https://orbit-components-update-checkbox-radio-label.surge.sh</url>


tomashapl:
Please check my MR #444 we could use this after merge

tomashapl:
We couldn't open this prop for React.Node, just open for 

```
React$Element<typeof Text>
```

wEprous:
Actually, we have to. We discussed this with @darkwindcz  and by design, this will be probably the only case where we open `string` prop for `React$Node`. Devs need to insert tooltips etc. and kind of ASAP.

wEprous:
Also the `Checkbox` will be used in #571

PR #540, opened 2018-10-25
InputField - Placeholder

denbalogh:

### DIFF
--- a/storybooknative/src/InputField/index.js
+++ b/storybooknative/src/InputField/index.js
@@ -70,6 +70,20 @@ StyledFakeInput.defaultProps = {
   theme: defaultTokens,
 };
 
+const StyledPlaceholder = styled(Text, props => {
+  const { theme } = props;
+  return {
+    color: theme.orbit.colorPlaceholderInput,
+    fontSize: getToken(TOKENS.fontSizeInput, props),
+    padding: getToken(TOKENS.paddingInput, props),
+    position: "absolute",
+  };
+});
+
+StyledPlaceholder.defaultProps = {
+  theme: defaultTokens,
+};
+
 const StyledInlineLabel = styled(Text, props => {
   const { theme, filled, disabled } = props;
   return {
@@ -89,6 +103,12 @@ const StyledInlineLabel = styled(Text, props => {
 StyledInlineLabel.defaultProps = {
   theme: defaultTokens,
 };
+
+const StyledInputWrapper = styled(View, () => ({
+  width: "100%",
+  zIndex: 2,
+  justifyContent: "center",
+}));
 
 const StyledTextInput = styled(TextInput, props => {
   const { theme, disabled } = props;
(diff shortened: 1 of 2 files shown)


storybooknative/src/InputField/index.js
```diff
@@ -90,6 +104,12 @@ StyledInlineLabel.defaultProps = {
   theme: defaultTokens,
 };
 
+const StyledInputWrapper = styled(View, () => ({
+  width: "100%",
+  zIndex: 2,
```
RobinCsl: I believe there are some specific values for zIndex in orbit-design-tokens.

storybooknative/src/InputField/index.js
```diff
@@ -165,17 +203,24 @@ class InputField extends React.Component<Props, State> {
...
-            placeholderTextColor={defaultTokens.orbit.colorPlaceholderInput}
-            editable={!disabled}
-            size={size}
-            type={type}
-            {...rest}
-            onFocus={this.onFocus}
-            onBlur={this.onBlur}
-          />
+          <StyledInputWrapper>
+            {!!this.props.placeholder &&
```
RobinCsl:
```suggestion
            {this.props.placeholder != null &&
```

storybooknative/src/InputField/index.js
```diff
@@ -165,17 +203,24 @@ class InputField extends React.Component<Props, State> {
...
+              this.state.displayPlaceholder && (
+                <StyledPlaceholder size={size}>{placeholder}</StyledPlaceholder>
+              )}
+            <StyledTextInput
+              underlineColorAndroid="transparent"
```
RobinCsl: What do you think about extracting platform specific code to other files?

storybooknative/src/InputField/index.js
RobinCsl: @denbalogh Could you please have a look at this?

storybooknative/src/InputField/index.js
RobinCsl: @tbergq what do you think about this?

tbergq: underlineColorAndroid is a prop on the input from react-native. So I would leave it here. 

Generally, we should be very careful with having platform specific files, since it often leads to duplication of code. I find it useful if we have two files when using different native dependencies, like DatePicker component from mobile repo. In this case, they cannot be in the same file. 

at some point we might want to add the PlatformStyleSheet in our styled components wrapper 🤔

RobinCsl: Fair enough, I'll accept it as it is then. 

Regarding `PlatformStylesheet`, @FilipMessa is tackling it in his PR https://github.com/kiwicom/orbit-components/pull/550

storybooknative/src/InputField/index.js
RobinCsl: @wEprous Should there be more z-index tokens for this use case or we stick with the custom value 2 that you use in the web version?

wEprous: Stick to the same `z-index` as we use in React. We still didn't decide how to approach to z-indexes and the values of `1` or `2` are good, it's just relative to a parent 😉

PR #550, opened 2018-10-28
RN: New simple card component

### LINKED ISSUE #536: Extract the SimpleCard component
https://github.com/kiwicom/mobile/blob/master/packages/shared/src/SimpleCard.js

FilipMessa:

**This branch is base on the `rn-button-component branch`, so the MR number #507 should be merged as first.**

closes #536 

### Summary
- Created `createPlatformStyles` helper. 
- Added the SimpleCard component.

I created  `createPlatformStyles` helper, now you can be able to add styles for a specific platform.
The API is same as we already used in the [mobile project.](https://github.com/kiwicom/mobile/blob/master/packages/shared/src/PlatformStyleSheet.js)

 For improvement to the future, I see space for converting specific react-native styles (paddingHorizontal, paddingVertical, ...etc) to the web styles.

<br/><br/><br/><url>LiveURL: https://orbit-components-rn-simple-card-component.surge.sh</url>


storybooknative/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@kiwicom/orbit-design-tokens": "^0.2.1",
+    "lodash.omit": "^4.5.0",
```
RobinCsl: With the existing solution in https://github.com/kiwicom/mobile/blob/master/packages/shared/src/PlatformStyleSheet.js, it probably won't be necessary to add this library.

FilipMessa: I took inspiration from the existing solution but, I decided to a different approach. In the existing solution, we iterate through every style's property.

I mine solution, I iterate through platform types and check if styles for particular platform exists. I suppose that is more effective in many cases then go through all property in style object.

Also, the difference between what we use on **mobile** and what we need for the **RNW Orbit** is in requirements.
Because we expected in the **RNW Orbit** that we can send `Object` and `Function` as the parameter to the `styled` function.

If we don't want to use omit from the `lodash`, I can write own omit function.  But I would like to know where is the problem with this approach?

possible custom omit function implementation.
```
const omit = (object, prop) => {
  const objectWithoutProp = Object.assign({}, object);
  delete objectWithoutProp[prop];
  return objectWithoutProp;
};
```

storybooknative/src/styled/index.js
```diff
@@ -1,10 +1,25 @@
...
+import { Platform } from "react-native";
 import styled from "styled-components";
+import omit from "lodash.omit";
 
 import type { Style as _Style } from "../types/Style";
 
+type Styles = _Style | ((props: Object) => _Style);
+
+const reduceStyles = styles =>
+  ["ios", "android", "web"].reduce((acc, type) => {
```
RobinCsl: Could you explain why deleting an object property is a better approach in this case?
It is possible to use the same approach as before with destructuring the styles object.

```
const reduceStyles = (styles: _Style): _Style => {
  let { ios, android, web, ...style } = { ...styles };
  if (ios && Platform.OS === "ios") {
    style = {...style, ...ios};
  }
  if (android && Platform.OS === 'android') {
     style = { ...style, ...android };
   }
  if (web && Platform.OS === 'web') {
     style = { ...style, ...android };
   }
   return style;
}; 
```
and in that case we do not need any loop, any library, etc.

Let me know your thoughts 😃

storybooknative/package.json
tbergq: I definitely don't think we should make our own implementation of what already exists. The best code is the one you don't have to maintain 😄 

I do agree here that could be more efficient not to copy all properties always yes. 

Then there is the never ending discussion lodash vs ramdba 🤷‍♂️

storybooknative/src/styled/index.js
FilipMessa: Yes, this should work. It was my first implementation, but because I love `reduce` and this is much shorter, I prefer my solution. 
But I am OK with your suggestion, so I will rewrite it to your way.

storybooknative/package.json
FilipMessa: I like the ramda also. 
But on the end, I am going it rewrite it to the how @RobinCsl suggest. 😉
https://github.com/kiwicom/orbit-components/pull/550#discussion_r229665315

storybooknative/src/SimpleCard/index.js
```diff
@@ -0,0 +1,53 @@
+// @flow strict
+// TODO create PlatformStyleSheet
```
RobinCsl: This comment can probably be removed now.
```suggestion
```

storybooknative/src/SimpleCard/SimpleCard.stories.js
```diff
@@ -0,0 +1,20 @@
...
+import SimpleCard from "./index";
+
+storiesOf("SimpleCard", module)
+  .addDecorator(withKnobs)
+  .addDecorator(getStory => <CenterView>{getStory()}</CenterView>)
+  .add("Default", () => (
+    <SimpleCard onPress={action("clicked")}>
+      <Text>Simple Card</Text>
+    </SimpleCard>
+  ));
```
RobinCsl: Maybe add a story with a card with no `onPress` props.

storybooknative/src/styled/__tests__/index.test.js
```diff
@@ -0,0 +1,109 @@
+// @flow
+/* eslint-disable no-return-assign */
+import { Platform } from "react-native";
+
+import { createPlatformStyles } from "../index";
+
+let originalPlatform;
+// $FlowFixMe
```
RobinCsl:
```suggestion
beforeEach(() => {
  originalPlatform = Platform.OS;
});
afterEach(() => {
  Platform.OS = originalPlatform;
});
```
and that means top `eslint-disable` is not necessary anymore.

storybooknative/src/styled/__tests__/index.test.js
```diff
@@ -0,0 +1,109 @@
+// @flow
+/* eslint-disable no-return-assign */
```
RobinCsl:
```suggestion
```

storybooknative/src/styled/__tests__/index.test.js
```diff
@@ -0,0 +1,109 @@
...
+    backgroundColor: "blue",
+  },
+  android: {
+    backgroundColor: "green",
+  },
+  web: {
+    backgroundColor: "orange",
+  },
+};
+const fontSize = 10;
```
RobinCsl: I am not sure about the `fontSize`'s being numbers. I think for it to work with react-native-web, we need them to be in `px`, like `"10px"`.

FilipMessa: good point, thanks

storybooknative/src/styled/__tests__/index.test.js
FilipMessa: done

storybooknative/src/styled/__tests__/index.test.js
FilipMessa: done

storybooknative/src/SimpleCard/SimpleCard.stories.js
FilipMessa: thanks for a good idea 👍

storybooknative/src/SimpleCard/index.js
FilipMessa: done

storybooknative/src/SimpleCard/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+import styled from "../styled";
+import defaultTokens from "../defaultTokens";
+import Touchable from "../Touchable";
+
+import { Props } from "./index";
+
+const addStylesHelper = props => ({
+  borderStyle: "solid",
+  paddingLeft: `10px`,
```
tbergq: we should use paddingEnd and paddingStart instead

FilipMessa: Sure but with our current `styled` implementation this won't work for the web.
Because it will be converted to `padding-end` and `padding-start, and this styles property doesn't exist for the web. 

I realize that before and suggest creating some parser for react-native style to the web.
But yesterday I found that this conversion works with StyleSheet.create from the `react-native`.

So I am thinking about two solutions now:
Avoid `styled-components` and use instead of `StyleSheet.create()`.
Or try implemented this conversion on our side, maby there will be way how we can reuse `createDOMstyles` from the 'react-native-web' (https://goo.gl/U8r7or)

But first, the important question is do we really need `styled-components`?
I know it is used on the Orbit Web components, but why we need it for Orbit RNW?

FilipMessa: Here is an example where you can see that with `StyleSheet.create()` it's out of the box.
https://codesandbox.io/s/jn85q7k49
The `marginVertical` and `paddingEnd` are converted to web style automatically, if you use `StyleSheet` for creating styles.

tbergq: I really thought styled-components was smarter, but I guess it is not. 

I guess the reason we wanted to use styled-components was to be able to reuse some styles from web. But also adding react-native-web means we could potentially just replace existing implementation 🤔 

So the the case for re-use really is eliminated. 

I am not opposed to replacing styled-components, as I would personally not chose it for any RN project in the first place.

FilipMessa: BTW: What do you suggest for this MR for now?
Should I write padding for all specific platforms separately?

FilipMessa: `paddingEnd` for **Android** and **IOS** and `paddingRight` for **Web**

RobinCsl: but since it is the same padding everywhere, `padding: "10px",` would suffice?

RobinCsl: Otherwise, for the future, since we can have platform specific styles, we could leverage that and not care too much (for now) of duplication of styles for lack of a better solution that does not require heavy work. Thoughts?

FilipMessa: 💯 @RobinCsl  definitely `padding: "10px"` should work. 
I feel stupid, that I didn't realize that 👽

tbergq: > I feel stupid

It's friday, so it is ok 🍺

RobinCsl: @FilipMessa No problem at all! That's what code reviews are for 😃 and it's Friday too 😄

RobinCsl:
Maybe just one last thing @FilipMessa, would you mind squashing your commits?

FilipMessa:
sure I will, just a minute

storybooknative/src/types/Style.js
```diff
@@ -10,11 +10,23 @@
...
   ...ShapedStyles,
+  +ios?: ExactStyles,
+  +android?: ExactStyles,
+  +web?: ExactStyles,
+|};
+
+export type OutputStyle = {|
+  ...ExactStyles,
   +borderWidth?: number | AnimatedNode | string,
-  +lineHeight?: number | string,
```
RobinCsl: I am wondering whether we should have the same:
```
  +fontSize?: string,
  +borderWidth?: number | AnimatedNode | string,
  +lineHeight?: number | string,
```

in `OutputStyle` too?

FilipMessa: They are inside because it is part of `ExactStyles` now. 
But there is this confusing and not necessary type `ShapedStyles ` so I will remove it.

storybooknative/src/styled/index.js
```diff
@@ -15,10 +40,24 @@ export { default as lineHeight } from "./lineheight";
...
  * const ColoredView = styled(View, (props) => ({ backgroundColor: props.color}))
  *
- * and we get flow validating our styles
+ * and we get flow validating our styles.
+ *
+ * Also you can define platform differences easily:
+ *
+ *  const Component = styled(View, {
+ *     fontSize: "23px",
+ *     iso: {
```
tbergq: ios?

FilipMessa: :see_no_evil:

storybooknative/src/types/Style.js
RobinCsl: So maybe we can even remove `+borderWidth?: number | AnimatedNode | string,` too?

FilipMessa: Good point. Thank you

wEprous:
Hope I'm not late. But any specific reason that we can not call this component `Card` as we do in `React`? :)

RobinCsl:
I do not think there is any specific reason *not* to call it `Card`. @FilipMessa, would you mind renaming it, please?

storybooknative/src/types/Style.js
```diff
@@ -10,11 +10,21 @@
...
+
+export type Style = {|
+  ...ExactStyles,
+  +ios?: ExactStyles,
+  +android?: ExactStyles,
+  +web?: ExactStyles,
+|};
+
+export type OutputStyle = {|
+  ...ExactStyles,
```
RobinCsl: Maybe this is not necessary to spread `ExactStyles` here and just export it?

FilipMessa: I tried it, but then there was some a flow error. And with this it works OK

PR #559, opened 2018-10-30
InputField tests

denbalogh:
**Placeholder** needs to be merged first.


storybooknative/src/InputField/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,3901 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
RobinCsl: I do not like these snapshot files being 4k lines when really they are just including the `theme` props everywhere... @wEprous Is there anything we can do about it?

wEprous: I don't like it either, but there is nothing we can do  🙁

tbergq: What is the value of a 4K snapshot test, since some has to manually check when something changes

RobinCsl: Especially if there is a modification to the theme object, there will be tons of changes that are just noise and we might overlook the necessary parts that needed checking.

wEprous: Well, modification of the theme object should be done in own PR, when new tokens are released. Otherwise, something is wrong.

Personally, I use snapshots for props checking and for the theme and testing of visuals, we introduced `loki`.

RobinCsl: I guess you're right, changing the theme object should only be done in its own PR. 

I need to have a look at `loki` when I have time and see how this could work with RN.

storybooknative/src/InputField/index.js
```diff
@@ -90,6 +104,12 @@ StyledInlineLabel.defaultProps = {
   theme: defaultTokens,
 };
 
+const StyledInputWrapper = styled(View, () => ({
+  width: "100%",
+  zIndex: 2,
```
RobinCsl: Same issue as here: https://github.com/kiwicom/orbit-components/pull/540#discussion_r229334754

PR #564, opened 2018-10-30
Extracted checkbox code from kiwi-mobile(link in description)

denbalogh:
https://github.com/kiwicom/mobile/blob/master/packages/shared/src/forms/Checkbox.js

### DIFF
--- a/storybooknative/src/Checkbox/index.js
+++ b/storybooknative/src/Checkbox/index.js
@@ -1,0 +1,32 @@
+// @flow
+
+import * as React from "react";
+import { View } from "react-native";
+
+import styled from "../styled";
+import Touchable from "../Touchable";
+
+import type { Props } from "./index";
+
+const StyledIconPlaceholder = styled(View, () => ({}));
+
+const StyledWrapper = styled(View, () => ({}));
+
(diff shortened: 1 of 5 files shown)


RobinCsl:
Please, rename the issue to be more specific with what code you extracted. Thank you!

storybooknative/src/Checkbox/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+import styled from "../styled";
+import Touchable from "../Touchable";
+
+import Props from "./index";
+
+export default function Checkbox(props: Props) {
+  const StyledIconPlaceholder = styled(View, () => ({}));
+
+  const StyledWrapper = styled(View, () => ({}));
```
tbergq: Why are this styled wrapper without any styles used instead of just a `View`?

tbergq: Also if the intent is to add styles here later, the definition should still be outside of the component. Now it is being recreated on each render.

denbalogh: Fixed the definition scope and yes, it is just a set up for next iteration where I am adding the styles.

PR #572, opened 2018-11-01
UPDATE: Breakpoints queries

wEprous:


src/Hide/index.js
```diff
@@ -0,0 +1,37 @@
...
+import styled, { css } from "styled-components";
+
+import mediaQueries from "../utils/mediaQuery";
+import { DEVICES } from "../utils/mediaQuery/consts";
+
+import type { Props } from "./index";
+
+const sortedBreakpoints = Object.keys(DEVICES).sort((a, b) => DEVICES[a] - DEVICES[b]);
+
+const StyledHide = styled.div`
```
PetrSnobelt: Isn't it better use span for inline rather then restyle div?

src/Hide/index.js
```diff
@@ -0,0 +1,37 @@
...
+  ${({ on, inline }) =>
+    sortedBreakpoints.map(
+      viewport =>
+        viewport in mediaQueries
+          ? css`
+              ${mediaQueries[viewport](css`
+                display: ${on.indexOf(viewport) !== -1
```
mainframev: checking if is element in array can be done by includes. Something like: on.includes(viewport).

wEprous: Array.includes is not supported in IEs 😞

wEprous:
I gonna refactor it a bit. I need also to extend Stack functionality to support all specific breakpoints.

src/Hide/index.js
mainframev: @wEprous another options is: ```if (~[1, 2, 3].indexOf(1))```. In this case ```indexOf(viewport) !== -1``` is the best decision since includes is not supported by IE11 without babel-polyfill, just share practice :)

wEprous:
@PetrSnobelt @mainframev 
**Refactored**, can you look once more plz? 🙏

PetrSnobelt:
Looks good

PR #574, opened 2018-11-01
RN: Checkbox - Orbit Styles

denbalogh:

### DIFF
--- a/storybooknative/src/Checkbox/Checkbox.stories.js
+++ b/storybooknative/src/Checkbox/Checkbox.stories.js
@@ -1,0 +1,33 @@
+// @flow
+
+/* eslint-disable jsx-a11y/accessible-emoji, import/no-unresolved, import/extensions */
+
+import React from "react";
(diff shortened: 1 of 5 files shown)


storybooknative/src/Checkbox/Checkbox.stories.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+/* eslint-disable jsx-a11y/accessible-emoji, import/no-unresolved, import/extensions */
```
RobinCsl:
```suggestion
/* eslint-disable import/no-unresolved, import/extensions */
```
I believe the eslint rule regarding emoji's is not necessary anymore.

RobinCsl:
This is the existing Checkbox on the web:
screen shot 2018-11-02 at 15 11 41

And this is the result of this implementation:
screen shot 2018-11-02 at 15 11 47

It seems to me the alignment is different. Could you fix this?

PR #581, opened 2018-11-01
DEV: Upgrade to babel 7.0.0

wEprous:
TODO:
After the update, the `--copy-files` option copies all ignored files `*.stories.js`, `*.test.js` and `*.storyshot`. But we want to copy only `*.js.flow` and `README.md` files.

Same for the `es` build. Also, it would be nice to define babel config for `build:modules` as was suggested once.
<br/><br/><br/><url>LiveURL: https://orbit-components-dev-update-babel.surge.sh</url>


wEprous:
TODO:
After the update, the `--copy-files` option copies all ignored files `*.stories.js`, `*.test.js` and `*.storyshot`. But we want to copy only `*.js.flow` and `README.md` files.

Same for the `es` build. Also, it would be nice to define babel config for `build:modules` as was suggested once.

PR #601, opened 2018-11-06
UPDATE: styled-components 4.0.3

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-styled-compoonents-upgrade.surge.sh</url>


src/ClickOutside/index.js
```diff
@@ -37,7 +37,7 @@ class ClickOutside extends React.PureComponent<Props> {
 
     return (
       <Inner
-        innerRef={node => {
+        ref={node => {
```
wEprous: Maybe we can use `createRef()`

tomashapl: 👍

PR #609, opened 2018-11-07
Add script to copy files to lib folder 

tbergq:
<br/><br/><br/><url>LiveURL: https://orbit-components-npmignore.surge.sh</url>

### DIFF
--- a/storybooknative/.gitignore
+++ b/storybooknative/.gitignore
@@ -58,3 +58,6 @@ storybook-static
 
 # Bundle artifact
 *.jsbundle
+
+# Build
+lib/
--- a/storybooknative/index.js
+++ b/storybooknative/index.js
@@ -1,6 +1,0 @@
-/** @format @flow */
-
-export { default as Button } from "./src/Button";
-export { default as Text } from "./src/Text";
-export { default as Touchable } from "./src/Touchable";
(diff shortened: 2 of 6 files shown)


tomashapl:
@tbergq in main repo we are using whitelisting instead of blacklisting, please check `files` section in `package.json` we had some issues with `.npmignore`

tbergq:
Ok, let me see if I can make it work with whitelisting as well

tbergq:
I don't think we can make this work without `.npmignore`.

What was the issue you had with it @tomashapl? 

We want to provide the src to the consumer, but he does not need the stories.

tomashapl:
> I don't think we can make this work without `.npmignore`.
> 
> What was the issue you had with it @tomashapl?
> 
> We want to provide the src to the consumer, but he does not need the stories.

Sometimes `npm publish` completely ignore `.npmignore` ¯\_(ツ)_/¯

tbergq:
Yeah, after adding whitelisting, blacklisting seems to fail. 😞

tbergq:
Ok, so I changed the approach a little bit. I made a script that copies all files expect test folders and `.stories.js` files to lib folder, then whitelist the lib folder.

PR #630, opened 2018-11-13
FIX: Box sizing

### LINKED ISSUE #627: <TripSegment> has defective layout
`<TripSegment>` has these issues used in project:

- It has no spacing below each segment when used without` <TripLayover>`
- Carrier logo is deformed and positioned behind toggle arrows on the right side
- When component is toggled to display `<List> `eg. flight number, max-height is counted wrongly.

## Expected Behavior
- When `<TripSegment>` is used without `<TripLayover>` it should have some spacing below
- Carrier logo is not deformed and positioned properly
- On toggle the max-height of `<List>` wrapper should be counted correctly to display whole information.

## Steps to Reproduce

Use only several `<TripSegment>` without `<TripSector>` or `<TripLayover>` 

screen shot 2018-11-12 at 14 38 29

### LINKED ISSUE #631: <TripSector>'s 100% height breaks layout
`<TripSector>` breaks layout as it has rule of 100% height.

## Expected Behavior
This component shouldn't break layout

screen shot 2018-11-13 at 13 44 19

wEprous:
Add this css style into `preview-head.html`
```html
<style>
  *, *:before, *:after {
    box-sizing: border-box;
  }
</style>
```

Start storybook and run `yarn loki test`. There should not be any visual difference.

Closes #627 
Closes #631


wEprous:
We won't support projects that use this selector:
```css
*, *:before, *:after {
    box-sizing: border-box;
  }
```

PR #647, opened 2018-11-21
UPDATE: support tabIndex

### LINKED ISSUE #646: Support tabIndex prop in Button

Keyboard navigation (on landing page of Search) breaks when user tries to tab through all inputs to search button. That one is skipped and focus goes back to the top of the page, because Orbit button cannot have tabindex assigned.

Please add tabIndex as optional prop to buttons, so focus order of elements can be overridden when necessary.

I have considered wrapping Orbit button into another normal button and move logic there as an ugly hack. Button is focusable in correct order with tab and triggerable with space/enter, but React complains that button tag shouldn't be nested within another button.

Solution nr. 2 was to use different element type (just plain div) as wrapper, but that would force me to attach custom listeners to watch for enter/space keypresses, which would be even more work for bad solution.

tomashapl:
Added support tabIndexes to

```
Button
ButtonLink
Checkbox
InputField
InputFile
InputStepper
Radio
Select
Textarea
TextLink
```

Closes #646<br/><br/><br/><url>LiveURL: https://orbit-components-add-tabindex.surge.sh</url>

### DIFF
--- a/src/Button/__tests__/__snapshots__/index.test.js.snap
+++ b/src/Button/__tests__/__snapshots__/index.test.js.snap
@@ -24,6 +24,7 @@ exports[`Button with icon should match snapshot 1`] = `
   onlyIcon={false}
   size="normal"
   sizeIcon="medium"
+  tabIndex="-1"
   theme={
     Object {
       "orbit": Object {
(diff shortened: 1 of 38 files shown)


wEprous:
@tomashapl can you add some sort of tests to be sure that this property is rendered into the DOM?

tomashapl:
> @tomashapl can you add some sort of tests to be sure that this property is rendered into the DOM?

Adjusted

PR #654, opened 2018-11-22
Rtl

mainframev:


src/utils/rtl/index.js.flow
```diff
@@ -9,6 +9,8 @@ export type BorderRadius = (value: string) => (theme: ThemeProps) => string;
 
 export type TextAlign = (value: "left" | "right") => (theme: ThemeProps) => string | LeftToRight;
 
+export type Translate3d = (x: string, y: string, z: string) => (themeProps: ThemeProps) => string;
```
wEprous: I would declare one parameter - `string`, so all utils for rtl are used in the same pattern. See [this](https://github.com/kiwicom/orbit-components/tree/master/src/utils/rtl#borderradius)

So usage would be:
```jsx
transform: ${translate3d("10px 20px 30px")};
```

PR #685, opened 2018-12-04
FIX: rendering children in loading

MartinChroboczek:


tomashapl:
Closing, due to many issues connected with it

PR #691, opened 2018-12-10
FIX: Text wrong weight value

### LINKED ISSUE #651: Wrong prop value in Text component in Storybook
`Text` component in storybook in prop `weight` has as a knob values of `normal` and `bold` whereas proper enum values for this prop are `regular` it's confusing and pushes user to look at the docs or code for proper values.

wEprous:
Small possibility of breaking change - if someone explicitly used `weight=regular`.

Closes #651 <br/><br/><br/><url>LiveURL: https://orbit-components-fix-text-wrong-weight.surge.sh</url>


tomashapl:
weight: `normal` replicates `font-weight: normal` i am not sure but it was more understandable

PR #704, opened 2018-12-13
UPDATE: Modal usage without ModalSection

wEprous:
Modal is now possible to use without `ModalSection`. It means in combination with `ModalHeader` and `ModalFooter`.<br/><br/><br/><url>LiveURL: https://orbit-components-fix-modal-without-section.surge.sh</url>

### DIFF
--- a/src/Modal/index.js
+++ b/src/Modal/index.js
@@ -8,8 +8,8 @@ import Close from "../icons/Close";
 import { SIZES, CLOSE_BUTTON_DATA_TEST } from "./consts";
 import media from "../utils/media";
 import { StyledModalFooter } from "./ModalFooter";
-import { MobileHeader } from "./ModalHeader";
-import { StyledModalSection } from "./ModalSection";
+import { MobileHeader, StyledModalHeader } from "./ModalHeader";
+import ModalSection, { StyledModalSection } from "./ModalSection";
 import { StyledHeading } from "../Heading";
 import { right } from "../utils/rtl";
 import transition from "../utils/transition";
@@ -177,6 +177,10 @@ const ModalWrapperContent = styled.div`
     visibility ${theme.orbit.durationFast} ease-in-out ${theme.orbit.durationFast}`};
   }
 
+  ${StyledModalHeader} {
+    margin-bottom: ${({ hasModalSection, theme }) => !hasModalSection && theme.orbit.spaceXLarge};
+  }
+
   ${media.desktop`
     position: relative;
     bottom: initial;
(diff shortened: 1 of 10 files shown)


tomashapl:
@wEprous Could you extend it of componentWillUnmount to set `this.state.hasModalSection`

i would prefer something like this:

```
setModalSection => hasModalSection = {
  this.setState({
    hasModalSection
  })
}
```

tomashapl:
@wEprous But i am not sure about more ModalSections, it will rerender more times, isn't better solution to check if one of children is ModalSection ?

wEprous:
I think that this approach is more understandable and easier, then e.g. this:
https://github.com/kiwicom/orbit-components/blob/62fff668f9cbd719d6051efbd4679448284e25ae/src/InputGroup/index.js#L168-L173

It will also be called only one time due to the condition.

tomashapl:
> I think that this approach is more understandable and easier, then e.g. this:
> [orbit-components/src/InputGroup/index.js](https://github.com/kiwicom/orbit-components/blob/62fff668f9cbd719d6051efbd4679448284e25ae/src/InputGroup/index.js#L168-L173)
> 
> Lines 168 to 173 in [62fff66](/kiwicom/orbit-components/commit/62fff668f9cbd719d6051efbd4679448284e25ae)
> 
>  this.setState({ 
>    filled: !React.Children.map( 
>      this.props.children, 
>      child => child.props.value !== undefined && child.props.value !== "", 
>    ).includes(false), 
>  }); 
> It will also be called only one time due to the condition.

So my question is:

1. Is possible to have more `<ModalSections />` ?
2. if not please implement componentWillUnmount fci to reset state of hasModalSection
3. if yes we have to go through children and check if one of children is ModalSection

src/Modal/index.js
```diff
@@ -186,12 +190,16 @@ const ModalWrapperContent = styled.div`
     overflow: visible;
     ${StyledModalSection}:last-of-type {
       padding-bottom: ${({ theme }) => theme.orbit.spaceXXLarge};
-      margin-bottom: ${({ fixedFooter, footerHeight }) =>
-        fixedFooter ? `${footerHeight}px` : "0"};
+      margin-bottom: ${({ hasModalSection, fixedFooter, footerHeight }) =>
+        hasModalSection && fixedFooter ? `${footerHeight}px` : "0"};
```
tomashapl: `hasModalSection` is unnecessary condition in this case because you are styling `${StyledModalSection}:last-of-type` so the evaluation will be always true

src/Modal/index.js
```diff
@@ -318,14 +328,30 @@ class Modal extends React.PureComponent<Props, State> {
     }
   };
 
+  decideModalSection = () => {
+    const children = React.Children.toArray(this.props.children);
+    const hasModalSection =
+      children.map(child => child.type.displayName).indexOf(ModalSection.displayName) !== -1;
```
tomashapl: I am not sure if `.displayName` is right, in the other files we are using just `.name`

PR #707, opened 2018-12-13
DEV: Update flow to 0.87.0

wEprous:
Update flow to the same version as nitro has. Should be probably merged after update on FE.

Merge #701 first.


tomashapl:
Firstly we should update FE

PR #773, opened 2019-01-21
FIX: BreadcrumbItem hover underline

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-breadcrumb-item-hover-underline.surge.sh</url>


wEprous:
Should be the hover/focus color `paletteInkLightHover`?

PR #634, opened 2018-11-16
UPDATE: Add reverseOnRtl to Icon

wEprous:
Closes #764
<br/><br/><br/><url>LiveURL: https://orbit-components-update-reverseonrtl-icon.surge.sh</url>

PR #776, opened 2019-01-22
FIX: Textlink arrow in RTL story

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-textlink-rtl-story.surge.sh</url>


wEprous:
It should be fixed via theming context, #634

PR #781, opened 2019-01-29
DEV: Add icons.json list

wEprous:
Create build script for generating the `icons.json` file that will be used on orbit.kiwi for displaying icons on page https://orbit.kiwi/visual-style/icons/.

Also `Sort` and `Uber` icon was added - it was completely missing. @darkwindcz just forgot to add them 🙂 
<br/><br/><br/><url>LiveURL: https://orbit-components-add-icon-list-json.surge.sh</url>


tomashapl:
could you please a little bit rewrite this function ?
please use `map` and if you want to convert array to object use spread operator `...`

tomashapl:
also would be useful async map
check example and `fs.readFile`

```
Promise.all(ids.map(function(id) { 
    return getItem(id).then(function(result) { 
        return result.name;
    });
})).then(function(results) {
    // results is an array of names
})
```

PR #787, opened 2019-01-30
DOCS: Better typing for call backs

wEprous:
Just better typing for docs for callbacks.
<br/><br/><br/><url>LiveURL: https://orbit-components-docs-better-callbacks.surge.sh</url>


tomashapl:
Nice 👏 I would add also Promise, because you could pass this:

```
onChange={() => new Promise((resolve, reject) => {...})}
```

wEprous:
So, should I update to this format: `() => void | Promise`?

tomashapl:
@wEprous Yes please 🙏

wEprous:
Updated. Somewhere typing for Promises is missing, we should fix it in another PR, I gonna create an issue for that.

PR #801, opened 2019-02-11
UPDATE: Relaxed country props in CountryFlag

michalsanger:
Accept nullable string as code prop. Fallback to "anywhere" on unknown country code.

<br/><br/><br/><url>LiveURL: https://orbit-components-relaxed-country-flag.surge.sh</url>


darkwindcz:
cc @wEprous @tomashapl 

From my point of view, why not. But it would be great if we logged somewhere if it fallback to anywhere so we have some indication about missing flag.

michalsanger:
Added warning into console on unsupported country code. Please review

michalsanger:
So what do you think @wEprous @tomashapl ?

tomashapl:
@michalsanger Great 👏  Thanks

michalsanger:
@tomashapl how do you it in this project? Could you point me to some code?

wEprous:
LGTM.
I'm not sure if a giving a console warn to a developer only is good enough in terms of maintainability. For me, it would be the best solution to have a log in Sentry and also give a developer warning.

mrtnzlml:
@michalsanger This warning should not be called in production by design. It uses `__DEV__` behind the scenes.

tomashapl:
> LGTM.
> I'm not sure if a giving a console warn to a developer only is good enough in terms of maintainability. For me, it would be the best solution to have a log in Sentry and also give a developer warning.

We will have also bad requests in sentry, but we don't want to spam client's console

tomashapl:
> @michalsanger This warning should not be called in production by design. It uses `__DEV__` behind the scenes.

So this shou

> @michalsanger This warning should not be called in production by design. It uses `__DEV__` behind the scenes.

So this should be installed as only `dev` dependencies ?

src/CountryFlag/index.js
```diff
@@ -18,8 +19,17 @@ StyledCountryFlag.defaultProps = {
   theme: defaultTokens,
 };
 
+export function getCountryProps(code: ?string, name: ?string): { code: string, name: string } {
+  const codeNormalized = !code ? "ANYWHERE" : code.toUpperCase();
+  warning(CODES[codeNormalized], `Country code not supported: ${JSON.stringify(code)}`);
```
mrtnzlml: Alternatively thanks to the underlying `sprintf`:

```diff
-  warning(CODES[codeNormalized], `Country code not supported: ${JSON.stringify(code)}`);
+  warning(CODES[codeNormalized], 'Country code not supported: %s', code); // is the JSON.stringify needed? 🤔 
```

🤓

mrtnzlml:
> So could be `@kiwicom/js` installed just only in `dev` dependencies?

@tomashapl Probably not in your case because it's a function that's gonna be called in production. In our case (monorepo) we can have it as a dev dependency because it's being stripped out completely (WIP).

You'd have to use https://www.npmjs.com/package/@kiwicom/babel-preset-kiwicom but the invariant/warning removal is not implemented there yet. Awaiting here: https://gitlab.skypicker.com/incubator/universe/issues/432

TL;DR: keep to as a regular dependency for now.

tomashapl:
> > So could be `@kiwicom/js` installed just only in `dev` dependencies?
> 
> @tomashapl Probably not in your case because it's a function that's gonna be called in production. In our case (monorepo) we can have it as a dev dependency because it's being stripped out completely (WIP).
> 
> You'd have to use https://www.npmjs.com/package/@kiwicom/babel-preset-kiwicom but the invariant/warning removal is not implemented there yet. Awaiting here: https://gitlab.skypicker.com/incubator/universe/issues/432
> 
> TL;DR: keep to as a regular dependency for now.

Thanks for explanation 👏 We add `implement @kiwicom/babel-preset-kiwicom` to our backlog.

michalsanger:
Ups, I forget to change also the `src/CountryFlag/README.md`

src/CountryFlag/index.js.flow
```diff
@@ -1,526 +1,18 @@
 // @flow
```
tomashapl: What is the reason for getting rid off whole `codes` ?

michalsanger: Where would you use that type?

tomashapl: if you provide to `CountryFlag` wrong code flow error appears. We want to be strict what component could receive.

michalsanger: @tomashapl but this PR is about relax on this strictness. It should be the `CountryFlag` responsibility to handle various input. It's the "be generous on input strict on output" principle.

michalsanger:
@tomashapl @wEprous could you check this PR please? It has been here for more then one month.

PR #836, opened 2019-02-20
ICONS: Add svg content to icons.json

wEprous:
Remover `url` from the `icons.json` and added the SVG content.<br/><br/><br/><url>LiveURL: https://orbit-components-icons-svg-content-list.surge.sh</url>


tomashapl:
Is it really needed ? it could slow down build and also `icons.json` will have very large size

wEprous:
The build is processed around ~ 4 seconds so I don't think that would be an issue. I'm waiting to @darkwindcz to try it out on orbit.kiwi, as he couldn't use just URL to fetch the SVGs content.

PR #878, opened 2019-03-04
UPDATE: Adding block prop to Hide component

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-hide-display-prop.surge.sh</url>

### DIFF
--- a/src/Hide/Hide.stories.js
+++ b/src/Hide/Hide.stories.js
@@ -20,7 +20,7 @@ storiesOf("Hide", module)
     }),
   )
   .addWithChapters("With Separator", () => {
-    const block = boolean("display", false);
+    const block = boolean("display", true);
 
     return {
       info:
@@ -30,7 +30,7 @@ storiesOf("Hide", module)
           sections: [
             {
               sectionFn: () => (
-                <Hide on={[]} block={block}>
+                <Hide on={["largeMobile"]} block={block}>
                   <Separator />
                 </Hide>
               ),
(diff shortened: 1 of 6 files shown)


src/Hide/Hide.stories.js
```diff
@@ -18,13 +19,35 @@ storiesOf("Hide", module)
       padding: "20px",
     }),
   )
+  .addWithChapters("With Separator", () => {
+    const block = boolean("display", false);
```
wEprous: Default true 🙏 so the Separator is displayed.

src/Hide/Hide.stories.js
```diff
@@ -18,13 +19,35 @@ storiesOf("Hide", module)
...
+
+    return {
+      info:
+        "Configuration with Separator, for separator to work correctly block property has to be set.",
+      chapters: [
+        {
+          sections: [
+            {
+              sectionFn: () => (
+                <Hide on={[]} block={block}>
```
wEprous: Please use some breakpoint, e.g. `largeMobile`

src/Hide/_tests_/index.test.js
```diff
@@ -35,6 +37,12 @@ describe("Hide", () => {
     );
     expect(component).toHaveStyleRule("display", "none");
   });
+  it("should be displayed block", () => {
+    component.setProps({ block: true, on: breakpoints[on[1]] });
```
wEprous: If you want to do it this way, should be
```
component.setProps({ block: true, on: on[1] });
```

Otherwise, you are setting a value of `(min-width: something)` not e.g. `smallMobile`.

You can also create a new describe if you want to.

PR #880, opened 2019-03-04
FIX: TripSegment translatation props

### LINKED ISSUE #879: <TripSegment> to accept React.Node in props
We use `nitro`'s `Translate` and custom`DateTime` components and would like to use them as values in `TripSegment`.
## Expected Behavior
`departureTime`, `arrivalTime` and `duration` props should allow values of `React.Node` type 
## Current Behavior
only `string` type is supported

## Possible Solution
fix similar to https://github.com/kiwicom/orbit-components/pull/444

mainframev:
Closes #879 
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-tripsegment.surge.sh</url>

### DIFF
--- a/src/TripSegment/README.md
+++ b/src/TripSegment/README.md
@@ -25,14 +25,14 @@ After adding import into your project you can use it simply like:
 
 Table below contains all types of the props available in TripSegment component.
 
-| Name              | Type                    | Default     | Description
-| :---------------- | :---------------------- | :---------- | :---------------------------------------------------------------------------------------- |
-| **arrival**       | `string`                |             | The arrival city name and short code.
-| **arrivalTime**   | `string`                |             | The arrival time.
-| **carrier**       | `Carrier`               |             | It's used to render proper icon in the Milestone, by Carrier type. [`See Carrier type`](https://github.com/kiwicom/orbit-components/tree/master/src/CarrierLogo#user-content-carrier)
-| **children**      | `React.Node`            |             | The content of the component.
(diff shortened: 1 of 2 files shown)


src/TripSegment/index.js.flow
```diff
@@ -18,11 +18,11 @@ export type Props = {|
   ...Globals,
   +children: React$Node,
   +carrier: Carrier,
-  +departure: string,
-  +departureTime: string,
-  +arrival: string,
-  +arrivalTime: string,
-  +duration: string | number,
+  +departure: React.Node,
```
wEprous: I think that using `Translate` type should be enough. See `common.js.flow`.

src/TripSegment/README.md
```diff
@@ -25,14 +25,14 @@ After adding import into your project you can use it simply like:
...
+| **arrival**       | `Translation`           |         | The arrival city name and short code.                                                                                                                                                 |
+| **arrivalTime**   | `Translation`           |         | The arrival time.                                                                                                                                                                     |
+| **carrier**       | `Carrier`               |         | It's used to render proper icon in the Milestone, by Carrier type. [`See Carrier type`](https://github.com/kiwicom/orbit-components/tree/master/src/CarrierLogo#user-content-carrier) |
+| **children**      | `Translation`           |         | The content of the component.                                                                                                                                                         |
```
wEprous: Revert plz 🙏 children is still node

mainframev: it's still node, just mistake by copy-paste in documentation.

PR #881, opened 2019-03-05
FIX: Flow type for exact children

### LINKED ISSUE #670: ListItem cannot be undefined or null
Example from our code:
```
<List>
{document && <ListItem>{document}</ListItem>}
<ListItem>{insurance}</ListItem>
</List>
```

If there is no document, we don't want to display ListItem, so we used a condition, but flow finds an error: 
Flow: Cannot create `List` element because inexact null or undefined [1] is incompatible with exact `React.Element` [2] in array element of property `children`.

vepor:
What someone uses conditional rendering with boolean, object or array in `List` or `Modal`, the flow cries that inexact is compatible with React.Element.

This can be fixed by the proposed type:
```
type StrictNode<T> = boolean | React$Element<T | Fragment>;

export type StrictChildren<T> = ?StrictNode<T> | Array<?StrictNode<T>>;
```

So the used children can be: `all falsey values`, `the exact type` or `React.Fragment`.

For some cases we gonna stick to using `React$Node`.
Closes #670 <br/><br/><br/><url>LiveURL: https://orbit-components-fix-type-exact-children.surge.sh</url>

### DIFF
--- a/src/Breadcrumbs/index.js.flow
+++ b/src/Breadcrumbs/index.js.flow
@@ -2,14 +2,12 @@
 /*
   DOCUMENTATION: https://orbit.kiwi/components/breadcrumbs/
 */
-import type { Globals } from "../common/common.js.flow";
+import type { Globals, StrictChildren } from "../common/common.js.flow";
 import typeof BreadcrumbsItemType from "./BreadcrumbsItem";
-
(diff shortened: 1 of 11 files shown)


vepor:
Superseded by #925

PR #911, opened 2019-03-12
UPDATE: Add setScrollPosition for Modal

### LINKED ISSUE #845: <Modal>: allow scroll to top of the modal window
If I re-render the modal contents dynamically depending on the user input (like in a multistep wizard form), it may happen that I need to be able to scroll to top of the modal window (such is the case when the content is longer than the fold)

Allow for passing a ref on which then scrollTop could be set.

Allow for a prop which would accept scrollTop

**Additional context**
This need has come up from nesting a multistep content-heavy wizard form inside a modal

wEprous:
Closes #845 
<br/><br/><br/><url>LiveURL: https://orbit-components-update-add-setscroll-modal.surge.sh</url>

### DIFF
--- a/src/Modal/README.md
+++ b/src/Modal/README.md
@@ -24,7 +24,6 @@ Table below contains all types of the props available in the Modal component.
 | children      | `React.Node`                |                 | The content of the Modal. [See Subcomponents](#sub-components)
 | dataTest      | `string`                    |                 | Optional prop for testing purposes.
 | fixedFooter   | `boolean`                   | `false`         | If `true` the ModalFooter will be fixed to the bottom of window.
-| setScroll     | `number`                    |                 | Property that can set up scrollTop position for the Modal component. It setups the scroll whenever changed.
 | size          | [`enum`](#modal-enum)       | `"normal"`      | The maximum width of the Modal on desktop viewport.
 | onClose       | `event => void \| Promise`  |                 | Function for handling onClose event. If you don't pass any function the Close button will not be displayed and it will not be possible to close the Modal. [See Functional specs](#functional-specs)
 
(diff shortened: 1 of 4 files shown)


tomashapl:
Please remove `setScroll` prop. You could do in another way:

```
const modalRef = React.createRef();

return (<Modal ref={modalRef}>....)
```

and then you can easily call modalRef?.current?.setScrollPosition(0) 😉

src/Modal/README.md
```diff
@@ -38,7 +38,30 @@ Table below contains all types of the props available in the Modal component.
 ### Functional specs
 * Also, if you want to select the Close Button element for testing purposes, use [data-test="ModalCloseButton"] selector.
 
-
+* If you need to set scrollTop position of the Modal component, you can use instance of the component and method `setScrollPoint` like this:
```
tomashapl: ... you can use instance of the component and method `setScrollPosition` please 😉

src/Modal/index.js
```diff
@@ -308,6 +309,17 @@ class Modal extends React.PureComponent<Props, State> {
     }
   }
 
+  setScrollPosition = (value: number) => {
+    const { modalContent, modalBody } = this;
+    if (window?.innerWidth >= DEVICES_WIDTH.largeMobile) {
+      if (modalBody?.current) {
+        modalBody.current.scrollTop = value;
```
tomashapl: you can simplify this without conditions

```
if (window?.innerWidth >= DEVICES_WIDTH.largeMobile) {
   modalBody?.current?.scrollTop = value;
```

src/Modal/index.js
```diff
@@ -308,6 +309,17 @@ class Modal extends React.PureComponent<Props, State> {
...
+      }
+    } else if (modalContent?.current) {
```
tomashapl:
```
else {
     modalContent?.current?.scrollTop = value;
}
```

src/Modal/index.js
tomashapl: ok, so in this case please add full condition

```
if (modalBody?.current?.scrollTop) {
 modalBody.current.scrollTop = value;
}
```

PR #918, opened 2019-03-13
Remove overflow hidden and nowrap from Select help text

### LINKED ISSUE #485: <Select />: help text is overflow hidden and no wrap
## Expected Behavior
The whole text to show

## Current Behavior
Text being partially hidden

image

## Possible Solution
Remove overflow hidden and nowrap

## Context (Environment)
using orbit 0.15.0

Taym95:

Try to fix #485


vepor:
Sorry for late respond @Taym95.

This proposed solution can't work now as we don't want jumping components. If someone will change content of `error` property, there is chance that next component after e.g. `InputField` is going to move a bit.

We need to fix this differently, probably via some other wrapper components.

PR #927, opened 2019-03-19
UPDATE: Card accessibility

filipdanisko:
Updating card for accessibility, card is now possible navigate using keyboard and focus state is present.
<br/><br/><br/><url>LiveURL: https://orbit-components-update-card-a11y.surge.sh</url>


src/Card/CardSection/index.js
```diff
@@ -90,12 +91,34 @@ class CardSection extends React.Component<any, Props> {
     }
   };
 
+  handleKeyPress = (ev: SyntheticKeyboardEvent<HTMLInputElement>) => {
+    if (ev.key === "Enter" || ev.key === " ") {
+      ev.preventDefault();
+      this.injectCallbackAndToggleSection();
+
+      const { onClose, onExpand, expanded } = this.props;
```
tomashapl: This code below does do the same as `injectCallbackAndToggleSection` please remove 🙏

src/Card/CardSection/index.js
```diff
@@ -90,12 +91,34 @@ class CardSection extends React.Component<any, Props> {
     }
   };
 
+  handleKeyPress = (ev: SyntheticKeyboardEvent<HTMLInputElement>) => {
+    if (ev.key === "Enter" || ev.key === " ") {
```
tomashapl: In this case would be better to use `ev.code` it returns `Enter` or `Space`

filipdanisko: I tried `.code` before, not sure why but I'm getting `undefined` not sure what am I missing.

src/Card/CardSection/index.js
```diff
@@ -91,20 +91,11 @@ class CardSection extends React.Component<any, Props> {
     }
   };
 
-  handleKeyPress = (ev: SyntheticKeyboardEvent<HTMLInputElement>) => {
-    if (ev.key === "Enter" || ev.key === " ") {
+  handleDownPress = (ev: SyntheticKeyboardEvent<HTMLDivElement>) => {
```
tomashapl: handleKeyDownPress 🙏

PR #513, opened 2018-10-22
FIX: ButtonLink and Button hover/active states

mainframev:
This Pull Request meets the following criteria:

* [ ] Tests have been added/adjusted for my new feature
* [ ] New Components are registered in index.js of my project

PR #928, opened 2019-03-20
UPDATE: Button & ButtonLink accessibility

filipdanisko:
- Adding AriaControls and AriaExpanded props to both Button and ButtonLink
- Updating Focus state to be more visible and consistent
- Updating Button disabled visual state to match ButtonLink behaviour

<br/><br/><br/><url>LiveURL: https://orbit-components-update-button-a11y.surge.sh</url>

### DIFF
--- a/src/Button/README.md
+++ b/src/Button/README.md
@@ -31,6 +31,8 @@ Table below contains all types of the props available in Button component.
 | submit        | `boolean`                         | `false`         | If `true`, the Button will have `type="submit"` attribute, otherwise `type="button"`.
 | **type**      | [`enum`](#enum)                   | `"primary"`     | The type of Button.
 | width         | `number`                          | `0`             | The width of the Button. Number is defined in `px`.
+| ariaControls  | `string`                          |                 | Id of the element the button controls.
+| ariaExpanded  | `boolean`                         |                 | Tells screen reader the controlled element from `ariaControls` is expanded
 
 ### enum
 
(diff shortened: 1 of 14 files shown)


vepor:
@filipdanisko can you remove the `enabled` selector as in #513? 🙏

src/ButtonLink/README.md
```diff
@@ -59,3 +62,13 @@ const YourComponent = props => <div {...props}>YourComponent</div>
 <ButtonLink component={YourComponent}>Title</ButtonLink>
 ```
 This example will render ButtonLink in div with children **YourComponent** and not **Title**.
+
+
+## Accessibility
+if you use `ButtonLink` without `href` so it's rendered as a `<Button>` element then you can use props below. In case you are passing `href` and element is rendered as `<a>` then it should be treated as a link and `ariaControls`, `ariaExpanded` shouldn't be useful.
```
vepor: I would use `<button>` so it's clear that you mean HTML element and not a component

vepor: Also, use a capital `I` at the start of the sentence

src/Button/README.md
```diff
@@ -62,3 +64,12 @@ Table below contains all types of the props available in Button component.
   If you specify the children of **YourComponent** component, it will override the children prop of Button component, e.g.:
   ```jsx
   const YourComponent = props => <div {...props}>YourComponent</div>
+
+## Accessibility
+A button is mainly used for indicating an action, submitting a data, opening a modal etc... If you want to use Button for navigation consider using a `<TextLink>` for that.
```
vepor: `etc.`

PR #931, opened 2019-03-20
UPDATE: Checkbox accessibility improvement

filipdanisko:
Making input element visible for readers, making focus work using tab

https://orbit-components-update-checkbox-a11y.surge.sh<br/><br/><br/><url>LiveURL: https://orbit-components-update-checkbox-a11y.surge.sh</url>


vepor:
There was a problem in `Cypress` that input in Checkbox or Radio could not be selected as it was visually hidden (display and visibility). I think that this probably fixes this problem, can you check - discuss it with Lukas?

filipdanisko:
Should be good according to @vepor and Lukas

PR #934, opened 2019-03-20
UPDATE: RadioButton focus state

filipdanisko:
Updating focus state on RadioButton, to be visible when using a keyboard.<br/><br/><br/><url>LiveURL: https://orbit-components-update-radioButton-a11y.surge.sh</url>


vepor:
Some comment as https://github.com/kiwicom/orbit-components/pull/931#issuecomment-475625592

filipdanisko:
Same answer, should be good

PR #936, opened 2019-03-21
UPDATE: InputStepper accessibility

filipdanisko:
- Updating focus on a buttons to be visible
- Moving to prevent default in the key handler to an inside of a condition to prevent focus trap.<br/><br/><br/><url>LiveURL: https://orbit-components-update-inputStepper.surge.sh</url>


src/InputStepper/index.js
```diff
@@ -118,24 +151,28 @@ class InputStepper extends React.Component<Props & ForwardedRef, State> {
...
           prefix={
-            <ButtonLink
-              disabled={disabled || value <= +minValue}
-              iconLeft={<MinusCircle color="secondary" />}
-              size={size}
-              onClick={this.decrementCounter}
-              transparent
-              component={PrefixSuffix}
-            />
+            <StyledButtonWrapper role="button" tabIndex="0" onKeyDown={this.handleKeyDecrement}>
```
vepor: Shouldn't be this handled by `ButtonLink` itself? I mean all properties (can be added, make sense for me) and also the focus/active state.

filipdanisko: There are no types for that, so this is a workaround. Should I add the types?

src/InputStepper/index.js
```diff
@@ -78,6 +99,18 @@ class InputStepper extends React.Component<Props & ForwardedRef, State> {
     }
   };
 
+  handleKeyDecrement = (ev: SyntheticKeyboardEvent<HTMLInputElement>) => {
+    if (ev.keyCode === 32 || ev.keyCode === 13) {
```
vepor: Please add the key codes into `keyMaps.js` 🙏

src/InputStepper/index.js
vepor: Well, we can use your solution for now. But it make sense to have these props available in Button/ButtonLink component as other devs can use them inside their projects.

filipdanisko: @vepor I agree but I would rather do that in a separate PR.

PR #940, opened 2019-03-21
UPDATE: Tooltip accessibility

filipdanisko:
- Tooltip is now able to be open with focus
- Addiing aria tags
- Adding helper function to generate randomid<br/><br/><br/><url>LiveURL: https://orbit-components-update-tooltip-a11y.surge.sh</url>


src/Tooltip/index.js
```diff
@@ -437,13 +442,16 @@ class Tooltip extends React.PureComponent<Props, State> {
           onFocus={this.handleIn}
           onBlur={this.handleOut}
           ref={this.container}
+          aria-describedby={this.tooltipId}
+          tabIndex="0"
         >
           {children}
         </StyledTooltipChildren>
         <Portal element="tooltips">
           <StyledTooltip role="tooltip" data-test={dataTest}>
```
vepor: Tooltip has now `role` twice. Is in intentional?

src/utils/randomId/index.flow.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+export type Random = (value: string) => string;
+
+declare export default function randomId(name: string): string;
```
vepor:
```
export type RandomID = (value: string) => string;

declare export default RandomID;
```

src/utils/randomId/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import type { Random } from "./index.flow";
+
+const randomId: Random = (name: string) => {
+  return name + Math.floor(Math.random() * 1000000000) + Math.floor(Math.random() * 1000000000);
```
vepor: 😅 won't be `thousand` enough?

vepor: Also, plz `randomID` 🙏

filipdanisko: 1000 is a length of 3, that's not sufficient I had it originally on that, but after running a benchmark, it returned match multiple times on 1000 items. So I cracked that up, but maybe we don't need an addition of extra random number in that case?

Also, randomID is not proper camel-case

src/Tooltip/index.js
filipdanisko: Good catch!

src/Tooltip/index.js
```diff
@@ -32,14 +32,14 @@ import {
 import { isAlignCenter, isAlignEnd, isAlignStart } from "./helpers/isAlign";
 import { DEVICES_WIDTH } from "../utils/mediaQuery/consts";
 import tooltipPadding from "./helpers/tooltipPadding";
+import generateRandomId from "../utils/randomId";
```
vepor: Always use the same name as the file has 🙏

src/Tooltip/index.js
```diff
@@ -437,13 +442,16 @@ class Tooltip extends React.PureComponent<Props, State> {
           onFocus={this.handleIn}
           onBlur={this.handleOut}
           ref={this.container}
+          aria-describedby={this.tooltipId}
+          tabIndex="0"
```
vepor: https://github.com/kiwicom/orbit-components/pull/945#discussion_r270847115

PR #943, opened 2019-03-22
UPDATE: DestinationCard accessibility

filipdanisko:
- Card is now focusable
- Card now calls onClick when pressing `enter`
- Adding aria tags
- rendering Heading to H2 element, maybe even better to do smaller then that what do you think @vepor 

Also @vepor is there better way to extend Heading to contain `id`? I wrapped content of header in `span` which is not ideal to me. Any help appreciated.

<br/><br/><br/><url>LiveURL: https://orbit-components-update-destinationCard.surge.sh</url>


src/DestinationCard/index.js
```diff
@@ -189,7 +194,14 @@ class DestinationCard extends React.PureComponent<Props, State> {
     }
   };
 
+  handleKeyDown = (ev: SyntheticKeyboardEvent<HTMLElement>) => {
+    if (ev.keyCode === 13 && this.props.onClick !== undefined) {
```
vepor: Can you save the keyCode into `keyMaps.js`? 🙏

src/DestinationCard/index.js
```diff
@@ -212,7 +224,11 @@ class DestinationCard extends React.PureComponent<Props, State> {
       <StyledDestinationCard
         data-test={dataTest}
         onClick={onClick}
+        onKeyDown={this.handleKeyDown}
         height={height >= SMALLEST_HEIGHT ? height : SMALLEST_HEIGHT}
+        tabIndex="0"
```
vepor: Same comment as https://github.com/kiwicom/orbit-components/pull/945#discussion_r270847115

PR #945, opened 2019-03-22
UPDATE: TripSegment accessibility

filipdanisko:
- Tripsegment is now focusable and can be opened by pressing `enter`<br/><br/><br/><url>LiveURL: https://orbit-components-update-tripSegmet-a11y.surge.sh</url>


src/TripSegment/index.js
```diff
@@ -284,7 +315,13 @@ class TripSegment extends React.PureComponent<Props, State> {
     const { expanded, initialExpanded, contentHeight } = this.state;
 
     return (
-      <StyledTripSegment dataTest={dataTest}>
+      <StyledTripSegment
+        dataTest={dataTest}
+        tabIndex="0"
```
vepor: When someone is gonna use this component on e.g. search, won't be the `zero` a problem? I think that we should provide a property for specifying the tabIndex - based on the document hierarchy, right?

filipdanisko: tabindex="0" is respecting document DOM. Can you be more specific where it could be an issue?

vepor: I meant a case when someone wants to use custom `tabIndex` and don't use the document hierarchy. Like the `0` use default, but can be specified explicitly.

filipdanisko: Oh I see, yeah that would make sense, with defaulting to 0. Will Add!

PR #950, opened 2019-03-25
FIX: Stepper and InputStepper onChange callback

### LINKED ISSUE #938: <InputStepper> triggers onChange handler on every rerender
InputStepper rerenders with every parent render and triggers onChange handler after it. Causes infinite rerender if props are changed as result of onChange function.

## Expected Behavior

onChange should only be triggered when stepper state changes

## Current Behavior

onChange is triggered every time parent rerenders

## Possible Solution

Change InputStepper from Component to PureComponent. Stepper is PureComponent already and doesnt seem to suffer from this problem.

## Steps to Reproduce

Do something in onChange that changes props of parent component(like changing redux store state), triggering another render.

## Context (Environment)

Causes infinite rerender if props are changed as result of onChange function.

vepor:
Closes #938 

https://orbit-components-fix-stepper-inputstepper-onchange.surge.sh<br/><br/><br/><url>LiveURL: https://orbit-components-fix-stepper-inputstepper-onchange.surge.sh</url>


src/InputStepper/index.js
```diff
@@ -32,31 +32,32 @@ class InputStepper extends React.Component<Props & ForwardedRef, State> {
...
     const { value } = this.state;
     const { minValue, step = 1 } = this.props;
     const newValue = value - step;
-
-    this.setState({
-      value: newValue <= +minValue ? minValue : newValue,
-    });
+    const stateValue = newValue <= +minValue ? minValue : newValue;
+    if (stateValue && stateValue !== value) {
+      this.onChangeCallback(stateValue);
```
tomashapl: `this.onChangeCallback` -> `setValueAndInjectCallback` 🙏

PR #958, opened 2019-03-26
UPDATE: Modal accessibility

filipdanisko:
- Modal now contains role, and aria labels to be visible for screenreaders.
- Opened modal now places focus on a first focusable element
- Opened modal now traps focus inside.

There is one more thing about focus returning after modal is closed. I think that should be handled by end user with onClose event, should I just add that to readme? 

https://orbit-components-update-modal-a11y.surge.sh<br/><br/><br/><url>LiveURL: https://orbit-components-update-modal-a11y.surge.sh</url>


src/Modal/consts.js
```diff
@@ -6,3 +6,10 @@ export const SIZES = {
 };
 
 export const CLOSE_BUTTON_DATA_TEST = "ModalCloseButton";
+
+export const FOCUSABLE_ELEMENT_SELECTORS =
+  'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, [tabindex="0"], [contenteditable]';
+
+export const KEY_CODE_MAP = {
```
tomashapl: Could be `KEY_CODE_MAP` in some common file ? we should use it in also in another components

filipdanisko:
update randomID to HASH-12345-id-12345

src/Modal/ModalHeader/index.js
```diff
@@ -157,13 +157,21 @@ class ModalHeader extends React.PureComponent<Props> {
...
+      dataTest,
+      modalID,
+    } = this.props;
     const hasHeader = title || description;
     return (
       <StyledModalHeader illustration={!!illustration} suppressed={suppressed} data-test={dataTest}>
         {illustration}
         {hasHeader && (
-          <ModalTitle illustration={!!illustration}>
+          <ModalTitle illustration={!!illustration} id={modalID}>
```
tomashapl: I think `id` should be in something that is always rendered like `ModalWrapper` not only in `ModalHeader`

PR #967, opened 2019-04-01
UPDATE: Tooltip render

filipdanisko:
Updating Tooltip to render to DOM only when `open` is triggered.<br/><br/><br/><url>LiveURL: https://orbit-components-update-tooltip-render.surge.sh</url>

### DIFF
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -89,7 +89,6 @@ const StyledTooltipWrapper = styled.div`
   bottom: ${({ shownMobile, tooltipWidth }) => (shownMobile ? "0" : `-${tooltipWidth}px`)};
   left: 0;
   right: 0;
-
   img {
     max-width: 100%;
   }
--- a/src/Tooltip/index.js
+++ b/src/Tooltip/index.js
@@ -61,15 +61,6 @@ const StyledTooltip = styled.div`
   width: 100%;
 `;
 
-const fadeIn = keyframes`
-  0% {
-    opacity: 0;
-  }
-  100% {
-    opacity: 1;
-  }
-`;
-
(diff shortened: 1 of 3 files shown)


src/Tooltip/index.js
```diff
@@ -205,6 +215,17 @@ class Tooltip extends React.PureComponent<Props, State> {
     align: ALIGNS.CENTER,
     shown: false,
     shownMobile: false,
+    positions: {
```
vepor: It was really necessary to save positions in the state? the method `getDimensions` is called every time it needs to calculate it.

filipdanisko: I'm open to discussion, I was thinking about just a boolean state variable, which would force reRender in similar way, but that seem pointless because positions are changed in the static anyway and adding state that doesn't really do anything seemed strange.But it might be more performant. What do you think, do you have any other solutions in mind?

src/Tooltip/index.js
```diff
@@ -459,40 +471,42 @@ class Tooltip extends React.PureComponent<Props, State> {
...
-              <StyledTooltipContent ref={this.content}>{content}</StyledTooltipContent>
-              <StyledTooltipClose>
-                <Button type="secondary" block onClick={this.handleClose} ref={this.closeButton}>
-                  {closeText}
-                </Button>
-              </StyledTooltipClose>
-            </StyledTooltipWrapper>
-          </StyledTooltip>
-        </Portal>
+        {(shown || shownMobile) && (
```
vepor: I would add another state that handles rendering into DOM and the `shown` and `showMobile` would be used only for visually displaying it after all calculations etc. is ready.

PR #968, opened 2019-04-01
UPDATE: Breadcrumbs onBack button

filipdanisko:
Adding a optional back button to breadcrumb navigation<br/><br/><br/><url>LiveURL: https://orbit-components-update-breadcrumbs-backButton.surge.sh</url>


src/Breadcrumbs/README.md
```diff
@@ -15,7 +15,9 @@ Table below contains all types of the props available in Breadcrumbs component.
 | Name          | Type                                    | Default         | Description                      |
 | :------------ | :-------------------------------------- | :-------------- | :------------------------------- |
 | dataTest      | `string`                                |                 | Optional prop for testing purposes.
-| **children**  | `React.Node`                            |                 | The content of the Breadcrumbs, normally [`BreadcrumbsItem`](#breadcrumbsitem).
+| **children**  | `React.Node`                            |                 | The content of the Breadcrumbs, normally [`BreadcrumbsItem`]
+(#breadcrumbsitem).
```
vepor: The link needs to be connected 😅

src/Breadcrumbs/index.js
```diff
@@ -23,15 +25,32 @@ const StyledBreadcrumbsList = styled.ol`
...
     <StyledBreadcrumbsList vocab="http://schema.org/" typeof="BreadcrumbList">
-      {React.Children.map(children, (item, key) =>
-        React.cloneElement(item, {
+      {back && (
+        <StyledBackButton
+          iconLeft={<ChevronLeft />}
+          circled
+          type="secondary"
+          size="small"
+          onClick={back}
```
vepor: `back` is not self-descriptive, please use `onBack` 🙏

src/Breadcrumbs/index.js
```diff
@@ -23,15 +25,32 @@ const StyledBreadcrumbsList = styled.ol`
   padding: 0;
 `;
 
-const Breadcrumbs = ({ children, dataTest }: Props) => (
+const StyledBackButton = styled(Button)`
```
vepor: Even that this is possible because we use rest props inside `Button` component, I would not extend it as we don't provide this option officially.

filipdanisko: @vepor what would you recommend as a best approach? Wrapping that inside new inline styled component, or using Stack?

src/Breadcrumbs/README.md
```diff
@@ -16,6 +16,7 @@ Table below contains all types of the props available in Breadcrumbs component.
 | :------------ | :-------------------------------------- | :-------------- | :------------------------------- |
 | dataTest      | `string`                                |                 | Optional prop for testing purposes.
 | **children**  | `React.Node`                            |                 | The content of the Breadcrumbs, normally [`BreadcrumbsItem`](#breadcrumbsitem).
+| back          | `event => void \| Promise`              |                 | Callback for handling back button action. If present the back button is visible
```
vepor: `onBack`

tomashapl:
Sorry I am not OK with `onBack` prop, for me is better use ¨onNavigateBack`

vepor:
@tomashapl We use `onGoBack` in the `DestinationHeader` component so I would use it rather than and `onNavigateBack`. Think that it suits the case properly 🙂

tomashapl:
Agree with ´onGoBack´ but in props is only ´onBack´ which is not clear.

vepor:
@tomashapl OK 👍

PR #976, opened 2019-04-03
FIX: Flow React import in Sticky

### LINKED ISSUE #975: Sticky flow definitions does not import React
`src/Sticky/index.js.flow` file has `Props` definition including `children: React.Node` but it does not import React. How does it even pass your CI?

## Expected Behavior
Runing flow should pass without errors.

## Current Behavior
Got Error `Cannot resolve name React.` in abovementioned file when try to run flow in my project.

## Possible Solution
import React in abovementioned file.

tomashapl:
React was not imported in flow declaration `¯\_(ツ)_/¯`

Fixes #975 <br/><br/><br/><url>LiveURL: https://orbit-components-flow-fix-sticky-react-import.surge.sh</url>

### DIFF
--- a/src/Sticky/index.js.flow
+++ b/src/Sticky/index.js.flow
@@ -1,9 +1,8 @@
 // @flow
-import * as React from "react";
 
 export type Props = {|
   +offset?: number,
-  +children: React.Node,
+  +children: React$Node,
 |};
 
 export type State = {|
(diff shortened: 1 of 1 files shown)


vepor:
Why just not to use `React$Node`? 🙂

tomashapl:
> Why just not to use `React$Node`? 🙂

There was no reason, just fixed by proposal solution from issue

PR #979, opened 2019-04-04
FIX: Stack default direction in breakpoints

filipdanisko:
Screenshot 2019-04-04 at 15 04 23

Live URL: https://orbit-components-update-stack-spacing.surge.sh


vepor:
I just gonna test in with loki

PR #985, opened 2019-04-05
FIX: CountryFlag warning

vepor:
`@kiwicom/js` is not compiled properly and it's breaking usage of `orbit-components` in IE10+.

Removed for now.<br/><br/><br/><url>LiveURL: https://orbit-components-fix-countryflag-warning.surge.sh</url>

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,6 @@
     "styled-components": "^4.2.0"
   },
   "dependencies": {
-    "@kiwicom/js": "^0.6.0",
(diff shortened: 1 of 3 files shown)


michalsanger:
Could you create an issue please so we can fix it in `@kiwicom/js`? Btw there is version `0.7.0` already.

vepor:
@michalsanger done! 🙂 (problem is present even in `0.7.0`)

PR #986, opened 2019-04-05
UPDATE: Use divs for Text in TripSegment

mainframev:
Changed because of react warning, which says, that div can't appear as descendant of `<p>` element.
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-tripsegment.surge.sh</url>


mainframev:
Changed because of react warning, which says, that div can't be descendent of ```<p>``` element.

PR #998, opened 2019-04-15
FIX: overflowing TripSegment in card

tomashapl:
Little bit expand TripSector to prevent:

image
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-overflowing-tripsegment-in-card.surge.sh</url>


vepor:
Why `yarn.lock` is changed? 🤔 @tomashapl

PR #1009, opened 2019-04-23
NEW: NotificationBadge

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-notification-badge.surge.sh</url>


filipdanisko:
@honzatmn @tomashapl Please guys, how this is going to be used? 

I think it would be good idea to support aria tags to reflect notification behaviour. Minimum I would say would be to support aria-label, but more advanced solution for inspiration is [here.](https://davidetriso.github.io/aria-badge/) Containing `aria-live`, `aria-atomic` and `role`

src/NotificationBadge/index.js
```diff
@@ -0,0 +1,34 @@
...
+StyledNotificationBadge.defaultProps = {
+  theme: defaultTheme,
+};
+
+const NotificationBadge = (props: Props) => {
+  const { type = TYPE_OPTIONS.NEUTRAL, children, icon, ariaLabel, dataTest } = props;
+
+  return (
+    <StyledNotificationBadge>
+      <Badge type={type} dataTest={dataTest} icon={icon} ariaLabel={ariaLabel}>
```
vepor: Should we support `icon` for NotificationBadge? @honzatmn

src/NotificationBadge/README.md
```diff
@@ -0,0 +1,36 @@
+# Badge
```
vepor: NotificationBadge in the entire file 🙏

tomashapl: nice catch 👏 updated

PR #1015, opened 2019-04-24
UPDATE: InputGroup accessibility

filipdanisko:
- Adding `role="group"` and `aria-labelledby` to better support screen readers 
- Extending `<FormLabel>` to contain `id` attribute.<br/><br/><br/><url>LiveURL: https://orbit-components-update-inputgroup-a11y-role.surge.sh</url>


vepor:
Snapshotting the mounted component that uses some randomizing won't work. Can please update the test @filipdanisko? Probably switching the snapshot to shallow will be enough.

Issue #460, opened 2018-10-09
<Button /> and <ButtonLink />: make aria-label prop required if button text omitted


Radek-Palisa:
## Expected Behavior
Screenreader to announce what the button does.

## Current Behavior
If you use e.g.
`<Button icon={<Check />} />` there will be nothing for the screenreader to announce.

## Possible Solution
make including aria-label compulsory if the children of the button are empty

## Context (Environment)
using orbit 0.14.0

Issue #395, opened 2018-09-27
Support accessibility


vepor:
We need to support ARIA in all components, so all components are accessible.

 
*Example:*
```html
<button type="button" aria-label="Close">
    <span aria-hidden="true">&times;</span>
  </button>
```

TODO:
- [x] Modal #435 
- [ ] Button, ButtonLink #460 

Add more related reported issues.

PR #1019, opened 2019-04-26
UPDATE: Adding Button title prop for accessibility

filipdanisko:
Updating Button and ButtonLink:
- Adding `title` prop for `aria-label`.
- Pushing warninig to dev when title should be used.
Closes: #460
References: #395<br/><br/><br/><url>LiveURL: https://orbit-components-update-button-a11y.surge.sh</url>


src/Button/index.js
```diff
@@ -471,12 +474,18 @@ const Button = React.forwardRef((props: Props, ref: Ref) => {
...
+    title,
   } = props;
   const iconLeft = props.iconLeft || icon;
   const sizeIcon = size === ICON_SIZES.SMALL ? ICON_SIZES.SMALL : ICON_SIZES.MEDIUM;
   const onlyIcon = iconLeft && !children;
   const isDisabled = loading || disabled;
 
+  warning(
+    !(!children && !title),
+    "Warning: title is missing on Button. Use title prop to add aria-label to be accessible for screen readers",
```
vepor: I would use:
```
Warning: children or title property is missing on Button. Use title property to add aria-label to be accessible for screen readers.
```
We can also append a link for documentation or README to the warning, if you want 🙂

src/ButtonLink/ButtonLink.stories.js
```diff
@@ -58,7 +59,7 @@ storiesOf("ButtonLink", module)
   .add(
     "Playground",
     () => {
-      const title = text("Title", "ButtonLink");
+      const label = text("Label", "ButtonLink");
```
vepor: Need to say that I'm quite confused that children has knob `label` and the `title` is used for `aria-label` actually. Wouldn't be better to have just `children` knob so it is more obvious?

PR #1038, opened 2019-05-07
UPDATE: Add none spaceAfter option

vepor:
Adding `none` option to `spaceAfter` prop - function, so developers can you it like this:
```jsx
<Text spaceAfter={someProperty ? "largest" : "none"} />
```

We can't use maybe type for that, as we need to support `spaceAfter` in media queries in the Stack component and it can be rendered with `null` or `undefined`.<br/><br/><br/><url>LiveURL: https://orbit-components-update-none-spaceafter.surge.sh</url>

### DIFF
--- a/src/common/getSpacingToken/README.md
+++ b/src/common/getSpacingToken/README.md
@@ -23,16 +23,17 @@ const App = () => <MyComponent spaceAfter="small" />
 ## Parameter
 This function receives one parameter - object. With usage in `styled-components` it can receive it *automatically*.
 
-| Name         | Type                     | Description                      |
-| :-------     | :----------------------- | :------------------------------- |
-| spaceAfter   | [`enum`](#enum)          | The value to be applied.
-| theme        | `typeof defaultTheme`    | Theme object with tokens and rtl.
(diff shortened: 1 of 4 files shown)


src/common/getSpacingToken/README.md
```diff
@@ -23,16 +23,17 @@ const App = () => <MyComponent spaceAfter="small" />
...
+| :-------     | :----------------------- | :------------ | :------------------------------- |
+| spaceAfter   | [`enum`](#enum)          | `"none"`      | The value to be applied.
+| theme        | `typeof defaultTheme`    |               | Theme object with tokens and rtl.
 
 > Please note that the *token value* in the documentation may not be up to date. Check [orbit.kiwi](https://orbit.kiwi/design-tokens/).
 
 ### enum
 | name              | used token      | token value |
 | :---------------- | :-------------- | :---------- |
+| `"none"`          | `spaceXXSmall`  | `4px`       |
```
filipdanisko: None shouldn’t have a positive value, right?

vepor: Yep, thanks!

PR #1054, opened 2019-05-22
NEW: Slider component

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-new-slider-component.surge.sh</url>


src/Slider/_tests_/index.test.js
```diff
@@ -0,0 +1,488 @@
...
+    expect(instance.findClosestKey(20, [1, 12, 24])).toEqual(2);
+  });
+  test("moveValueByStep should return proper value", () => {
+    expect(instance.moveValueByStep(1)).toEqual(2);
+    expect(instance.moveValueByStep(-1)).toEqual(1);
+    expect(instance.moveValueByStep(5)).toEqual(6);
+    expect(instance.moveValueByStep(-5)).toEqual(1);
+    expect(instance.moveValueByStep(-10)).toEqual(1);
+    instance.setState({ value: 100 });
+    expect(instance.moveValueByStep(1)).toEqual(100);
```
sampi: does this not go to `101`, because the maximum is set to `100` somewhere?

src/Slider/_tests_/index.test.js
```diff
@@ -0,0 +1,488 @@
...
+  });
+  test("injectCallbackAndSetState should call", () => {
+    instance.injectCallbackAndSetState(onChange, 95);
+    expect(onChange).toHaveBeenCalled();
+    instance.injectCallbackAndSetState(onChangeBefore, 90);
+    expect(onChangeBefore).toHaveBeenCalled();
+    instance.injectCallbackAndSetState(onChangeAfter, 90, true);
+    expect(onChangeAfter).toHaveBeenCalled();
+  });
+  it("handleBarMouseDown execute onChange", () => {
```
sampi: you're using `test()` everywhere, except here

src/Slider/_tests_/index.test.js
```diff
@@ -0,0 +1,488 @@
...
+    expect(instance.findClosestKey(1, [1, 12])).toEqual(0);
+    expect(instance.findClosestKey(2, [1, 12])).toEqual(0);
+    expect(instance.findClosestKey(6, [1, 12])).toEqual(0);
+    expect(instance.findClosestKey(7, [1, 12])).toEqual(1);
+    expect(instance.findClosestKey(11, [1, 12])).toEqual(1);
+    expect(instance.findClosestKey(11, [1, 12, 24])).toEqual(1);
+    expect(instance.findClosestKey(20, [1, 12, 24])).toEqual(2);
+  });
+  test("moveValueByStep should return proper value", () => {
+    expect(instance.moveValueByStep(1)).toEqual(2);
```
sampi: if for some reason this test runs after `"alignValueToStep should return proper value"`, then it will fail because the `step` size would not allow to get to `2`.

Maybe it would be a good idea to reset the `instance` before each test.

src/Slider/_tests_/index.test.js
```diff
@@ -0,0 +1,488 @@
...
+  test("defaultValue changes the value state", () => {
+    component.setProps({ defaultValue: [1, 100] });
+    expect(instance.state.value).toEqual([1, 100]);
+  });
+  it("should execute keyDown event with arrow up", () => {
+    instance.setState({ handleIndex: 0, value: [50, 60] });
+    instance.handleKeyDown({
+      stopPropagation: () => {},
+      preventDefault: () => {},
+      keyCode: 38,
```
sampi: maybe you could import the keys from `keyMaps.js` for better readability of the test

sampi: it also looks like this part is repeating quite a lot, so it could be more terse to break it out into a helper function 

```js
// somewhere outside of `describe` and `test`
function testKeyTriggeringEvent(instance, expect, initialState, keyCode, stateValue, onHandler) {
    instance.setState(initialState);
    instance.handleKeyDown({
      stopPropagation: () => {},
      preventDefault: () => {},
      keyCode,
    });
    expect(instance.state.value).toEqual(stateValue);
    expect(onHandler).toHaveBeenCalled();
}

// later in the test code

it("should execute keyDown event with arrow up", () =>
    testKeyTriggeringEvent(instance, expect, { handleIndex: 0, value: [50, 60] }, keyMaps.ARROW_UP, [51, 60], onChange)
);
```

src/Slider/_tests_/index.test.js
```diff
@@ -0,0 +1,488 @@
...
+  test("should have Bar component", () => {
+    const children = slider.find("Slider__StyledSliderInput").find("Bar");
+    children.forEach(node => {
+      expect(node.prop("max")).toBe(max);
+      expect(node.prop("min")).toBe(min);
+      expect(node.prop("value")).toEqual([1, 24]);
+    });
+  });
+  it("should match snapshot", () => {
+    expect(component).toMatchSnapshot();
```
sampi: snapshot tests should be used sparingly, and I don't see a reason to use it here.
it's just something that will break each time you modify the Slider, but it won't give you any real information.

src/Slider/components/Bar/index.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+  };
+})`
+  position: absolute;
+  height: 4px;
+  top: 10px;
+  border-radius: 4px;
+  background-color: ${({ theme, active }) =>
+    active ? theme.orbit.paletteProductNormal : theme.orbit.paletteInkLighter};
+  z-index: ${({ active }) => (active ? 30 : 20)};
```
sampi: I thought we started using new `z-index` values from tokens, is that just a draft still?

src/Slider/components/Handle/_tests_/index.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const renderedHandle = handle.render();
+    expect(renderedHandle.prop("role")).toBe("slider");
+    expect(renderedHandle.prop("aria-valuemax")).toBe(max.toString());
+    expect(renderedHandle.prop("aria-valuemin")).toBe(min.toString());
+    expect(renderedHandle.prop("aria-valuenow")).toBe(value[index].toString());
+    expect(renderedHandle.prop("aria-label")).toBe(ariaLabel[index]);
+    expect(renderedHandle.prop("aria-valuetext")).toBe(ariaValueText);
+  });
+  it("should match snapshot", () => {
+    expect(component).toMatchSnapshot();
```
sampi: again, is this snapshot test useful for later development or testing?

src/Slider/components/Handle/index.js
```diff
@@ -0,0 +1,136 @@
...
+)`
+  display: flex;
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  flex: 0 0 24px;
+  align-items: center;
+  justify-content: center;
+  box-sizing: border-box;
+  width: 24px;
```
sampi: is it normal to hardcode pixel values in the css?
shouldn't these all be some sort of tokens for units or similar?

src/Slider/components/Histogram/_tests_/index.test.js
```diff
@@ -0,0 +1,44 @@
...
+    const dataActive = [false, false, true, true];
+    const children = component
+      .find("Histogram__StyledHistogram")
+      .find("Histogram__StyledHistogramColumn");
+    children.forEach((node, index) => {
+      expect(node.prop("active")).toBe(dataActive[index]);
+    });
+  });
+  it("should match snapshot", () => {
+    expect(component).toMatchSnapshot();
```
sampi: another snapshot test, that is not very useful.

src/Slider/utils/calculateCountOf.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import type { CalculateCountOf } from "./calculateCountOf";
+
+const countValues = (data, [from, to], min) => {
```
sampi: this one could be simplified a little bit, by removing `return`:


```js
const countValues = (data, [from, to], min) =>
  data.reduce(
    ([selected, totalCount], curr, index) => [
      index + min >= from && index + min <= to ? selected + curr : selected,
      totalCount + curr
    ],
    [0, 0]
  );
```

src/Slider/_tests_/index.test.js
vepor: Yep, the default value of `max` is `100`

src/Slider/_tests_/index.test.js
vepor: For now, the standard is to have it for each component, so we should keep it that way. We will decide what to do with snapshots lates 🙂

src/Slider/components/Bar/index.js
vepor: It's still draft and we don't have tokens for it, yet. But the values are actually are the proposed levels.

src/Slider/components/Handle/index.js
vepor: It's normal for new components. We create component and then tokenize it.

src/Slider/README.md
```diff
@@ -0,0 +1,86 @@
...
+<Slider defaultValue={5} onChange={(value) => doSomething(value)} />
+```
+
+## Props
+Table below contains all types of the props available in the Slider component.
+
+| Name                  | Type                                  | Default         | Description                     |
+| :-------------------- | :------------------------------------ | :-------------- | :------------------------------ |
+| ariaLabel             | `Translation or Translation[]`        |                 | `aria-label` attribute or attributes for handles. See [functional specs](#functional-specs).
+| ariaValueText         | `Translation`                         |                 | Readable text alternative of current value (aria-valueNow).
```
filipdanisko: This in ideal case won't be used at all, it's just for edge cases and it should be changed when value is changed. This would probably deserves better documentation.

PR #1062, opened 2019-05-27
UPDATE: Added missing translations in components

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-added-missing-translations-in-components.surge.sh</url>


filipdanisko:
Just found another missing translation inside InputFields with ButtonLink suffix.

PR #1069, opened 2019-05-29
DEV: fix incorrect optional chain usages

mrtnzlml:

<br/><br/><br/><url>LiveURL: https://orbit-components-flow-unnecessary-optional-chain.surge.sh</url>


src/Card/CardSection/CardSectionContent/index.js
```diff
@@ -54,7 +54,7 @@ class CardSectionContent extends React.Component<Props, State> {
     contentHeight: 0,
   };
 
-  node: { current: any | HTMLDivElement } = React.createRef();
```
mrtnzlml: This is wrong everywhere you use `{ current: any | Something }`, see: https://github.com/facebook/flow/blob/master/lib/react.js#L246-L247. I left it everywhere there it uncovered additional bugs. Maybe worth checking it.

vepor:
Thanks @mrtnzlml. Just out of curiosity, is it incorrect usage, or just unnecessary? I believe that we needed to use it in previous versions of `flow`.

mrtnzlml:
@vepor Could be considered unnecessary, yes. Even the lint is called like this. I am more strict at myself so this is _incorrect_ for me... 😊

I think this combo of optional chaining and unnecessary chain lint works like this from the beginning.

mrtnzlml:
How about this? :)

PR #1090, opened 2019-06-06
UPDATE: Alert inlineActions prop

filipdanisko:
Adding `inlinActions` prop for cases where there is only a heading and displaying a button under heading wouldn't look ideal.<br/><br/><br/><url>LiveURL: https://orbit-components-update-alert-inline.surge.sh</url>

### DIFF
--- a/src/Alert/README.md
+++ b/src/Alert/README.md
@@ -16,6 +16,7 @@ Table below contains all types of the props available in Alert component.
 | closable      | `boolean`                       | `false`         | If `true`, the Close icon will be displayed. [See Functional specs](#functional-specs)
 | dataTest      | `string`                        |                 | Optional prop for testing purposes.
 | icon          | `React.Element<any> \| boolean` |                 | The displayed icon. [See Functional specs](#functional-specs)
+| inlineActions | `React.Node`                    |                 | Renders action to a right side of a Alert. [See Functional specs](#functional-specs)
 | onClose       | `() => void \| Promise`         |                 | Function for handling Alert onClose.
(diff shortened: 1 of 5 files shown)


src/Alert/index.js
```diff
@@ -125,6 +125,8 @@ const IconContainer = styled(StyledDiv)`
   flex-shrink: 0;
   margin: ${({ theme }) => rtlSpacing(`0 ${theme.orbit.spaceSmall} 0 0`)};
   color: ${getTypeToken(TOKENS.colorIconAlert)};
+  display: ${({ inlineActions }) => (inlineActions ? "flex" : "")};
```
vepor: `inlineActions && "flex"` 🙏 
Same for the condition on line 129 and 141.

src/Alert/README.md
```diff
@@ -34,3 +35,5 @@ Table below contains all types of the props available in Alert component.
 * By passing the `closable` prop into Alert, you will be able to handle `onClose` function and Close icon will be displayed. Also, if you want to select the Close Button element for testing purposes, use [data-test="AlertCloseButton"] selector.
 
 * If you set up `icon` props as `true`, Alert will have it's own icon, based on selected `type`. If you want to use different icon, just pass it to the `icon` prop as `React.Element`. Alert without `icon` prop doesn't have any icon.
+
+* Passing a `inlineActions` will cause `children` to be ignored.
```
vepor: Can you just mention something about the usage? E.g., the property should be used only for buttons, nothing else. 🙏

PR #1112, opened 2019-06-19
UPDATE: Tooltip enabled prop

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-add-enabled-prop-to-tooltip.surge.sh</url>


vepor:
We need to also switch the focusability of the wrapped children.
Snímek obrazovky 2019-06-19 v 15 24 13

PR #1119, opened 2019-06-25
UPDATE: InputGroup onChange behaviour

infi-pc:
This is potentially breaking change, update should be checked.

When e.g. `onChange` is passed to the children, e.g. `InputField` it will be used instead of the `handleChange` declared inside the `InputGroup`.

<br/><br/><br/><url>LiveURL: https://orbit-components-inputgroup-onchange.surge.sh</url>


src/InputGroup/index.js
```diff
@@ -250,7 +250,7 @@ class InputGroup extends React.PureComponent<Props, State> {
                     label={undefined}
                     help={undefined}
                     error={undefined}
-                    onChange={this.handleChange}
+                    onChange={this.props.onChange ? this.handleChange : item.props.onChange}
```
tomashapl: Could we add both behaviour ? e.g. If onChange is in InputField and also in InputGroup, it could be useful to call both of it

infi-pc: this seemed to me a little bit cleaner. I think it could lead to weird behavior when somebody uses both. Do you have some use case?

tomashapl: Not sure if real use case. But I have on my mind one situation. for example birthday input group, there could be some normalizers on inputs, but in the end you have to check `onChange` event on InputGroup due to handling error, but normalizers will check onChange event on each input

PR #1129, opened 2019-07-15
UPDATE: Better compression for Illustrations

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-illustration-compression.surge.sh</url>


src/Illustration/index.js
```diff
@@ -18,12 +18,11 @@ const getHeightToken = ({ theme, size }) => {
 
 export const StyledImage = styled.img.attrs(({ theme, size, illustrationName }) => {
   const height = parseInt(getHeightToken({ theme, size }), 10);
-
+  const illustrationPath = `${illustrationName}-q85.png`;
```
tomashapl: Could we add prop `quality` ? By default it should be 85. And how it will work on lambda side ? it will generate  files with quality postfix to cache or ?

vepor: We can, but I would not do it. Quality of 85 percent should be enough for every use case. If we gonna introduce this property, someone can change and we would load big images for nothing 😞 
And yes, it will generate cached images for every Illustration + compression percentage (doesn't work now on production lambda).

PR #1159, opened 2019-08-05
FIX: Tooltip content scrolling

vepor:
https://take.ms/87h5Og

<br/><br/><br/><url>LiveURL: https://orbit-components-fix-tooltip-scrolling.surge.sh</url>

### DIFF
--- a/src/Tooltip/Tooltip.stories.js
+++ b/src/Tooltip/Tooltip.stories.js
@@ -131,6 +131,24 @@ storiesOf("Tooltip", module)
     },
   )
   .add(
+    "With long content",
+    () => {
+      const content = text(
+        "content",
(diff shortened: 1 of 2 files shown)


tomashapl:
Could you please write some story for this case ? @vepor

vepor:
I would not expose it, as it should not be used in this way - it's just legacy edge case. Also it won't be captured visually, as Tooltip is hidden by default. @tomashapl

tomashapl:
> I would not expose it, as it should not be used in this way - it's just legacy edge case. Also it won't be captured visually, as Tooltip is hidden by default. @tomashapl

Understand, but it should be nice to have story with bigger content of tooltip to see if your fix works

vepor:
Added @tomashapl

src/Tooltip/Tooltip.stories.js
```diff
@@ -130,6 +130,24 @@ storiesOf("Tooltip", module)
         "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
     },
   )
+  .add(
```
tomashapl: Please use there `lokiSkip` instead of `add` to prevent capturing

PR #1163, opened 2019-08-07
DEV: Uploading icons font to npm

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-uploading-icons-font-to-npm.surge.sh</url>


package.json
```diff
@@ -50,6 +50,7 @@
     "es",
     "lib",
     "umd",
+    "orbit-icons-font",
```
vepor: I would also publish the generated zip file in root folder. It would be nice to have it linked on orbit.kiwi, the folder itself can be useful for React Native etc.

PR #1165, opened 2019-08-07
FIX: Tooltip bounce

vepor:
If you set preferredPosition to right and some bigger text into content the component will shake little bit when you hover onto the tooltip itself.<br/><br/><br/><url>LiveURL: https://orbit-components-fix-tooltip-bounce.surge.sh</url>


vepor:
Closed in behalf of #1181

PR #1182, opened 2019-08-20
DEV: Monorepo preparation

tomashapl:
<br/><br/><br/><url>LiveURL: https://orbit-components-monorepo-prepare.surge.sh</url>


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+# Default ignored files
+/workspace.xml
```
mrtnzlml: 🤔 Folder `.idea` should probably not be committed.

PR #1188, opened 2019-08-23
FIX: Tooltip calculation when scrolled

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-tooltip-calculation-scrolled.surge.sh</url>

### DIFF
--- a/src/Tooltip/Tooltip.stories.js
+++ b/src/Tooltip/Tooltip.stories.js
@@ -164,25 +164,16 @@ storiesOf("Tooltip", module)
       );
 
       return (
-        <div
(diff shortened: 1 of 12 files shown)


tomashapl:
`resolveContainerAlign,js` is still using `containerTop` why this is needed ? THX

vepor:
@tomashapl `containerTop` is used because you need to apply absolute position of the container (it's top position from `getBoundingClientRect` and `scrollY` of the window). On other hand, `containerTopLeft` is relative top position of the container to window and it changes upon every scroll or resize event.

PR #1206, opened 2019-09-04
UPDATE: Render property basis in Stack

vepor:
Currently, when developers want to use proper sizing or spacing inside `basis` property of `Stack`, they hardcode every value, e.g. like this:
```jsx
<Stack direction="row" spacing="comfy" shrink desktop={{ basis: "calc(75% - 12px)" }}>
  <Button>Primary</Button>
  <Button type="secondary">Primary</Button>
</Stack>
```

This can be a big maintainability and consistency issue in the future, when the spacing or sizing system will change.

Proposed solution is to have `basis` property with type: `string | ((theme: Theme) => string)` that would be used like this:
```jsx
<Stack direction="row" spacing="comfy" shrink desktop={{ basis: theme => `calc(75% - ${theme.orbit.spaceSmall})` }}>
  <Button>Primary</Button>
  <Button type="secondary">Primary</Button>
</Stack>
```

The same approach can be used for `Grid` components where guys have to use sizings or spacings for e.g. `gap`.
<br/><br/><br/><url>LiveURL: https://orbit-components-update-stack-basis-render-prop.surge.sh</url>


honzatmn:
Nice idea to make it connected to tokens 👍 

Only thing I think about is writing a construction like this:
```
basis: theme => `calc(75% - ${theme.orbit.spaceSmall})`
```

... as it seems that `theme => ` part will be repeated all over again everywhere. Couldn't we add it somehow automatically?

vepor:
Also @honzatmn to elaborate, guys can also use it as now as using `function` won't be mandatory:
```jsx
<Stack direction="row" spacing="comfy" shrink desktop={{ basis: "50%" }}>
  <Button>Primary</Button>
  <Button type="secondary">Primary</Button>
</Stack>
```

PR #1214, opened 2019-09-11
right to iconRight

straxico:


vepor:
Hey @straxico, first of all, thanks a lot for this PR!
But need to say that I'm not sure if this fixes anything. The `right` attribute of `getIconSpacing` is passed to `IconContainer` here: https://github.com/kiwicom/orbit-components/blob/master/src/Button/index.js#L295-L301

It means, that the function gets it correctly, or is there some other issue or reason, why is this change necessary?

vepor:
Do you have any particular reasons besides TypeScript to rewrite all components? I'm not sure how much is Orbit flexible for you use case, but we aim to be used also in other projects, not only in projects related to Kiwi.com. Also, hopefully, we will support TypeScript definitions until the end of year and fully switch to it, later.

Yes, we just wrote some components before it was possible to use hooks, so let's say that it's just legacy code. But still, there are some advantages (in some use cases) when comes to Class components.

To be honest, to have five DOM elements in Button is kind of necessary: if you want to align everything inside button into the center and also Safari in older versions doesn't support usage of `flex` on `button` primitive.

But anyway, looking forward to any other PR you might make in the future!

PR #1285, opened 2019-09-30
UPDATE: Add aria-labelledby prop to button

tbergq:

<br/><br/><br/><url>LiveURL: https://orbit-components-button-aria-labelledby.surge.sh</url>

### DIFF
--- a/src/Button/__tests__/__snapshots__/index.test.js.snap
+++ b/src/Button/__tests__/__snapshots__/index.test.js.snap
@@ -4,6 +4,7 @@ exports[`Button with icon should match snapshot 1`] = `
 <Button__StyledButton
   ariaControls="element"
   ariaExpanded={true}
+  ariaLabelledBy="my-aria-span"
   buttonRef={null}
   component="button"
   dataTest="test"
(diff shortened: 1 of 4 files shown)


vepor:
Hello @tbergq, thanks for the PR!
Can you just please specify the **exact** use case, why is this enhancement necessary? 🙏

tbergq:
Yes, we have some abstraction around translations, so that we want to use a translation component. Since the `title` prop needs a string, we can not use it with our translation component. 

Hence, I want to do something like 

```
<Button ariaLabelledBy="my-id">
 <SettingsIcon />
<span id="my-id" style={{opacity: 0, position: 'absolute}}'><Translation id="my.translation.key" />< /span>
</Button>
```

vepor:
Thanks for the info @tbergq.
Why not then to have translate function? I think that with this property, we're fixing the problem with bad approach.
https://github.com/kiwicom/orbit-components/blob/master/src/Translate/index.js#L10-L22

tbergq:
> Why not then to have translate function?

We do, but it is just exposed to those components handling translations, like our Translation component. 

I don't fancy too much `<div>{translate('my-key'}</div>`, because we loose track of our translations. I prefer solving problems with components, rather than passing around functions.

tbergq:
Is there anything blocking this PR from being merged?

tbergq:
Closing since this feature is not wanted.

PR #1292, opened 2019-10-02
FIX: Scrolling in long content popovers

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-popover-mobile-overflow.surge.sh</url>


src/Popover/components/ContentWrapper.js
```diff
@@ -54,6 +54,24 @@ StyledPopoverParent.defaultProps = {
   theme: defaultTheme,
 };
 
+const StyledChildrenWrapper = styled.div`
+  height: 400px;
```
vepor: Why static values @filipdanisko?

PR #1333, opened 2019-10-17
DEV: Refactor Table component to React hooks

mainframev:

<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-hookies.surge.sh</url>


src/Table/index.js
```diff
@@ -84,68 +84,57 @@ StyledTable.defaultProps = {
...
-  }
-
-  componentWillUnmount() {
-    window.removeEventListener("resize", this.handleResize);
-  }
-
-  handleResize = () => {
-    const { table, outer } = this;
+  const handleResize = () => {
     if (table && outer) {
```
vepor: Wrap the `handleResize` into `useCallback` and use it as dependency of `useEffect`. Or declare it inside the `useEffect` and use empty dependency array `[]`.

package.json
```diff
@@ -78,8 +78,8 @@
   "author": "kiwi.com",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^16.8.3",
-    "react-dom": "^16.8.3",
+    "react": "^16.9.0",
```
vepor: Please don't change or add dependencies when it's not necessary. We don't even know if it can be raised due to main repo, where `16.8.3` is used also - update needs to be done there first.

src/Table/index.js
```diff
@@ -84,68 +84,57 @@ StyledTable.defaultProps = {
...
+
+  const outer: { current: any | HTMLElement } = React.useRef(null);
+  const inner: { current: any | HTMLElement } = React.useRef(null);
+  const table: { current: any | HTMLElement } = React.useRef(null);
+
+  const handleScroll = () => {
+    if (shadows && inner && table && outer) {
+      setLeft(inner.current?.scrollLeft >= 5);
+      setRight(
+        inner.current?.scrollLeft + outer.current?.clientWidth <= table.current?.clientWidth,
```
vepor: Also I've noticed that when you fully scroll to the right, the right shadow won't disappear. Can you check? 🙏

package.json
mainframev: that's not a major version, it will not break something for sure.

PR #1347, opened 2019-10-23
DEV: InputStepper to hooks

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-inputStepperHookie.surge.sh</url>


src/InputStepper/index.js
```diff
@@ -7,140 +7,143 @@ import InputStepperStateless from "./InputStepperStateless";
...
+      onDecrement={handleDecrementCounter}
+      onIncrement={handleIncrementCounter}
+      titleIncrement={titleIncrement}
+      titleDecrement={titleDecrement}
+    />
+  );
+};
 
 // $FlowExpected
 const ForwardedInputStepper = React.forwardRef((props, ref) => (
```
vepor: You can forward ref immediately on L17, check `Button` for instance. This legacy code was due to usage of `Class` component.

src/InputStepper/index.js
```diff
@@ -7,140 +7,143 @@ import InputStepperStateless from "./InputStepperStateless";
...
 
-  setValueAndInjectCallback = (value: number) => {
-    const { onChange } = this.props;
-    if (onChange) {
-      onChange(value);
-    }
-    this.setState({ value });
-  };
+const InputStepper = ({ defaultValue, onChange, ...props }: AllProps) => {
+  const [value, setValue] = React.useState<number>(defaultValue || 0);
```
vepor: You can use `useStateWithCallback` hook 🙂

src/InputStepper/index.js
```diff
@@ -1,153 +1,141 @@
...
+const InputStepper = React.forwardRef<Props, HTMLInputElement>((
+  {
+    onChange,
+    minValue = Number.NEGATIVE_INFINITY,
+    defaultValue = 0,
+    maxValue = Number.POSITIVE_INFINITY,
+    step = 1,
+    ...props
+  }: Props,
+  // $FlowExpected
```
vepor: What's the problem with the `Ref` type? Can you fix that properly instead of suppressing it with comment? 🙏

readme.md
```diff
@@ -45,7 +45,7 @@ Or via CSS:
 
 ```jsx
 import Alert from "@kiwicom/orbit-components/lib/Alert"
-    
+
```
vepor: Please revert changes that are not in the scope of this PR 🙏

src/InputStepper/index.js
mainframev: nope. Could not find solution. Just try to do same with other components where is suppressed comment - it will be same. No idea how to fix.

PR #1348, opened 2019-10-23
DEV: Stepper to hooks

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-stepper.surge.sh</url>

### DIFF
--- a/src/Stepper/_tests_/index.test.js
+++ b/src/Stepper/_tests_/index.test.js
@@ -2,12 +2,10 @@
 import * as React from "react";
 import { shallow } from "enzyme";
 
-import Stepper from "../index";
 import StepperStateless from "../StepperStateless";
 
 describe("Stepper", () => {
   const defaultValue = 1;
-  const step = 2;
   const name = "name";
   const disabled = false;
   const maxValue = 100;
@@ -16,21 +14,6 @@ describe("Stepper", () => {
   const onChange = jest.fn();
   const onFocus = jest.fn();
   const onBlur = jest.fn();
-
-  const component = shallow(
-    <Stepper
-      defaultValue={defaultValue}
-      step={step}
-      name={name}
-      maxValue={maxValue}
-      minValue={minValue}
-      disabled={disabled}
-      dataTest={dataTest}
-      onChange={onChange}
-      onFocus={onFocus}
-      onBlur={onBlur}
-    />,
-  );
 
   const statelessCompoennt = shallow(
     <StepperStateless
@@ -63,11 +46,7 @@ describe("Stepper", () => {
         .prop("data-test"),
     ).toBe(dataTest);
   });
-  it("should execute onChange method", () => {
-    const instance = component.instance();
-    instance.incrementCounter();
-    expect(onChange).toHaveBeenCalled();
-  });
+
   it("should execute onFocus method", () => {
     input.simulate("focus");
     expect(onFocus).toHaveBeenCalled();
--- a/src/Stepper/index.js
+++ b/src/Stepper/index.js
@@ -4,79 +4,66 @@ import * as React from "react";
 import StepperStateless from "./StepperStateless";
 import validateIncrement from "../utils/validateIncrement";
 import validateDecrement from "../utils/validateDecrement";
+import useStateWithCallback from "../hooks/useStateWithCallback";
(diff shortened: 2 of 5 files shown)


src/Stepper/index.js
```diff
@@ -4,79 +4,71 @@ import * as React from "react";
...
 
-class Stepper extends React.PureComponent<Props, State> {
-  state = {
-    value: this.props.defaultValue || 0,
-  };
-
-  setValueAndInjectCallback = (value: number) => {
-    const { onChange } = this.props;
+const Stepper = ({ onChange, defaultValue, ...props }: Props) => {
+  const [value, setValue] = useStateWithCallback<number>(defaultValue || 0, val => {
```
vepor: I noticed that we can enhance the `useStateWithCallback` a bit. It would great to use it only like this:
```js
const [value, setValue] = useStateWithCallback<number>(defaultValue, onChange);
```
And add the callback eval into the `useStateWithCallback` directly:
```js
if (value !== prevValue) {
  if(callback) callback(value);
}
```
Can you? 🙏 
https://github.com/kiwicom/orbit-components/blob/master/src/hooks/useStateWithCallback.js#-L11L14


Another note: because you passing function bind, the callback function inside the hook would be re-created upon every render of this component. Don't do this for memoized functions/components:
```js
val => {
  if(onChange) {
    onChange(val);
  }
}
```
Rather than, use `useCallback` separately. It's not necessary if you gonna extend the `useStateWithCallback` as I wrote above.

src/Stepper/index.js
```diff
@@ -4,79 +4,71 @@ import * as React from "react";
...
+import type { Props } from "./index";
 
-class Stepper extends React.PureComponent<Props, State> {
-  state = {
-    value: this.props.defaultValue || 0,
-  };
-
-  setValueAndInjectCallback = (value: number) => {
-    const { onChange } = this.props;
+const Stepper = ({ onChange, defaultValue, ...props }: Props) => {
```
vepor: I would unify the approach when comes to default values. e.g. `defaultValue = 0` and don't use it on line 12.

src/Stepper/index.js
```diff
@@ -4,79 +4,71 @@ import * as React from "react";
...
-    this.setValueAndInjectCallback(validateIncrement({ value, maxValue, step }));
-  };
+  const incrementCounter = React.useCallback(() => {
+    const { maxValue = Number.POSITIVE_INFINITY, step = 1 } = props;
 
-  decrementCounter = () => {
-    const { value } = this.state;
-    const { minValue = Number.NEGATIVE_INFINITY, step = 1 } = this.props;
+    setValue(validateIncrement({ value, maxValue, step }));
+  }, [props, setValue, value]);
```
vepor: I understand the reason why you are desctructurizing the props in the callback directly, but using `props` inside dependency array would cause to re-creation of the callback upon every change of the `props` object (even when e.g. `dataTest` will change). Can you look into it and find another way how to do it?

src/Stepper/_tests_/index.test.js
```diff
@@ -63,11 +46,7 @@ describe("Stepper", () => {
         .prop("data-test"),
     ).toBe(dataTest);
   });
-  it("should execute onChange method", () => {
```
vepor: Don't you wanna try the `@testing-library/react` now? This refactor doesn't have to be merged asap and it would be great to try it 🙂

mainframev: @vepor not sure how to test it with that library now

src/Stepper/index.js
```diff
@@ -4,79 +4,59 @@ import * as React from "react";
...
-  }
-}
+  const { onBlur, onFocus, disabled, name, dataTest, titleIncrement, titleDecrement } = props;
+  return (
+    <StepperStateless
+      disabled={disabled}
+      dataTest={dataTest}
+      value={value}
+      name={name}
+      minValue={minValue}
```
vepor: Looks good 💪 , only problem is that `minValue` and `maxValue` renders with `Infinity` or `-Infinity` now 😕 

Can you look and fix @mainframev? 🙏

mainframev: fixed

PR #1349, opened 2019-10-23
FIX: Scrolling of Popover inside another scrollable element

filipdanisko:


src/Popover/helpers/getScrollableParent.js.flow
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export type GetScrollableParent = (node: React.node) => React.node;
```
vepor: `HTMLElement` should be probably used. `React.Node` is completely different type.

src/Popover/helpers/getScrollableParent.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import type { GetScrollableParent } from "./getScrollableParent";
+
+const regex = /(auto|scroll)/;
+
+const style = (node, prop) => getComputedStyle(node).getPropertyValue(prop);
+
+const scroll = node =>
+  regex.test(style(node, "overflow") + style(node, "overflow-y") + style(node, "overflow-x"));
```
vepor: Is it checking just for `overflow` attributes enough? What if `overflow` will be present, but the actual `scrollingHeight` won't be bigger than `clientHeight`?

filipdanisko: Good point I will investigate that

filipdanisko: In that case, listener will be attached but won't callback anything, due to element not being scrollable, we can talk about if we want to attach listener to element which is not scrollable but in my mind is ok due to resize or dynamic content change

vepor: IMHO, it's needed to add tests only for `getScrollableParent`. The other functions are not exported and they doesn't have to be - there is no other use case yet.
I would rather add comments to all functions that are declared there 🙂

vepor: If we need also to test other cases - if `style` works alright, we can use these cases for `getScrollableParent`.

src/Popover/hooks/useDimensions.js
```diff
@@ -50,14 +51,20 @@ const useDimensions: UseDimensions = ({ containerRef, popover, content, fixed })
         });
       }
     };
+    const scrollableParrent = getScrollableParent(containerRef.current);
```
vepor: One small thing @filipdanisko. Because using this recursion function can be quite expensive, I would run it only the very first render with `useMemo` and empty dependencies `[]`. IMHO, it's impossible that the parent element will change without affection it's children (not triggering unmount).

vepor: And also, there is slight typo `scrollableParrent` 😄

filipdanisko: I can't really call useMemo here, that would go against Hook usage rules 🤔

PR #1350, opened 2019-10-23
FIX: Popover overflow

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-popover-overflow.surge.sh</url>

### DIFF
--- a/src/Popover/components/ContentWrapper.js
+++ b/src/Popover/components/ContentWrapper.js
@@ -31,7 +31,7 @@ const StyledPopoverParent = styled.div`
   padding: ${({ theme, noPadding }) => (noPadding ? 0 : theme.orbit.spaceMedium)};
   box-shadow: ${({ theme }) => theme.orbit.boxShadowRaisedReverse};
   overflow: auto;
-  z-index: ${({ isInsideModal }) => (isInsideModal ? "1000" : "600")};
+  z-index: 1000;
   transition: ${transition(["opacity", "transform"], "fast", "ease-in-out")};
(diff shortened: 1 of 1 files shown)


src/Popover/components/ContentWrapper.js
```diff
@@ -30,8 +30,8 @@ const StyledPopoverParent = styled.div`
   background-color: ${({ theme }) => theme.orbit.backgroundModal}; // TODO: Add token
   padding: ${({ theme, noPadding }) => (noPadding ? 0 : theme.orbit.spaceMedium)};
   box-shadow: ${({ theme }) => theme.orbit.boxShadowRaisedReverse};
-  overflow: hidden;
-  z-index: 1000;
+  overflow: auto;
+  z-index: ${({ isInsideModal }) => (isInsideModal ? "1000" : "600")};
```
vepor: I would not change z-indexes now, it can be unsafe and unnecessarily BC.

src/Popover/components/ContentWrapper.js
```diff
@@ -78,7 +78,6 @@ const StyledOverlay = styled.div`
 
   ${media.largeMobile(css`
     display: none;
-    z-index: ${({ isInsideModal }) => (isInsideModal ? "999" : "599")};
```
vepor: This should be reverted, right @filipdanisko?

vepor: @filipdanisko 🙏

filipdanisko: No need, due to display none present :)

PR #1351, opened 2019-10-23
FIX: Tile's title wrapping in IE

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-tile-title-wrapping.surge.sh</url>


jaroslav-kubicek:
it works, thanks! 🚀

PR #1356, opened 2019-11-01
UPDATE: Popover getScrollableParent execution optimization

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-popover-parentrecursive.surge.sh</url>

### DIFF
--- a/src/Popover/components/ContentWrapper.js
+++ b/src/Popover/components/ContentWrapper.js
@@ -1,5 +1,5 @@
 // @flow
-import React, { useRef, useEffect, useContext } from "react";
+import React, { useRef, useEffect, useContext, useMemo } from "react";
 import styled, { css } from "styled-components";
 import convertHexToRgba from "@kiwicom/orbit-design-tokens/lib/convertHexToRgba";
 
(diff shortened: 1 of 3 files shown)


src/Popover/components/ContentWrapper.js
```diff
@@ -117,7 +118,8 @@ const PopoverContentWrapper = ({
   const content: { current: React$ElementRef<*> } = useRef(null);
   const overlay: { current: React$ElementRef<*> } = useRef(null);
   const position = calculatePopoverPosition(preferredPosition, preferredAlign);
-  const dimensions = useDimensions({ containerRef, popover, content, fixed });
+  const scrollableParent = useMemo(() => getScrollableParent(containerRef.current), [containerRef]);
```
vepor: I'm not sure if the `deps` are okay. Does `containerRef` changes at least once? I mean in this use case, not theoretically. Also, it's confusing when the function is dependent on `containerRef.current`, but just `containerRef` is in `deps`. Another thing is that by default, `eslint-plugin-react-hooks` doesn't add content of `useRef` to dependencies at all - this is bypassed here because it's prop.

filipdanisko: Theoretically, we don't need any deps at all, right? containerRef won't change and if child of popover will change, this will be recalculated anyway.

PR #1360, opened 2019-11-05
UPDATE: Popover fixed actions

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-popover-fixedactions.surge.sh</url>


src/Popover/index.js
```diff
@@ -23,6 +23,11 @@ StyledActions.defaultProps = {
   theme: defaultTheme,
 };
 
+const StyledContentWrapper = styled.div`
+  overflow: auto;
+  max-height: calc(100vh - 120px);
```
vepor: What's `120px`? 🤔

filipdanisko: Good point, it's a size of top margin and size of actions, but I will make it more readable and also I forgot condition here if there are no actions 😅

PR #1373, opened 2019-11-13
FIX: Forwarding of refs

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-frwrdref.surge.sh</url>

### DIFF
--- a/src/Button/index.js.flow
+++ b/src/Button/index.js.flow
@@ -45,6 +45,8 @@ export type Props = {|
   +ariaExpanded?: boolean,
   +ariaControls?: string,
   +role?: string,
+  // Don't use, it's deprecated and will be removed in the major version
+  +className?: string,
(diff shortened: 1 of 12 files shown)


src/Button/index.js.flow
```diff
@@ -45,6 +45,7 @@ export type Props = {|
   +ariaExpanded?: boolean,
   +ariaControls?: string,
   +role?: string,
+  +className?: string,
```
vepor: @mainframev can you please just explicitly write comment for the className property: "Don't use, it's deprecated and will be removed in the near future"? 🙏 The same for ButtonLink.

Because of wrong typing before, we didn't have to have this in flow file, and we don't want to have it in docs as it's deprecated property - will be removed and usage of e.g. `styled(Button)` won't work in the future.

PR #1377, opened 2019-11-20
FIX: Missing aria-labels in our codebase

### LINKED ISSUE #1375: Collapse component console warning
Missing `aria-label` on ButtonLink inside [Collapse component](https://github.com/kiwicom/orbit-components/blob/2b6196a215bd16f20d9cee1e57bd1081225e96d7/src/Collapse/index.js#L136) is throwing [console error](https://monosnap.com/file/SqsuU8IgMR1GtEpkSmSZr3a26LLtIj).

## Possible Solution
Define `aria-label` prop.

filipdanisko:
Closes #1375<br/><br/><br/><url>LiveURL: https://orbit-components-fix-collapse-aria.surge.sh</url>


src/InputStepper/__tests__/index.test.js
```diff
@@ -42,6 +42,8 @@ describe(`InputStepper with help, prefix and suffix`, () => {
       onFocus={onFocus}
       onBlur={onBlur}
       spaceAfter={spaceAfter}
+      titleIncrement="Add"
```
vepor: Can we put this into translations? I would use this properties for overrides. `Add` and `Remove` is totally okay for 90 percent of usages.

filipdanisko: @vepor we are not providing a default, so it is really needed?

vepor: That's the thing, I think that we should - that accessibility is part of all components and when someone needs some different labels etc. there is a way

PR #1391, opened 2019-11-28
FIX: Tile description scaling on landscape

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-tile-description-scaling.surge.sh</url>


vepor:
Will be part of #1381

PR #1401, opened 2019-12-05
FIX: TranslateString type for HTML attributes

mainframev:
[JIRA, Orbit-802](https://jira.kiwi.com/browse/ORBIT-802)

Updated that Translation type, because it caused a problem with [Object object] when people were trying to insert component there. 
<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-plainhtml.surge.sh</url>


src/Select/index.js.flow
```diff
@@ -23,7 +23,7 @@ export type Props = {|
   +required?: boolean,
   +size?: Size,
   +label?: Translation,
-  +placeholder?: Translation,
+  +placeholder?: string,
```
vepor: Why not `TranslationString` also here?

mainframev: because it's used not only in html prop

PR #1515, opened 2020-02-07
UPDATE: Compass icon

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-updat-compasicon.surge.sh</url>


filipdanisko:
> Path definition is broken, doesn't work for 24x24 viewBox.

Should be fixed

PR #1522, opened 2020-02-10
BREAKING: Introduce ButtonPrimitive, implementing planned breaking changes

mainframev:
Refactored Button and ButtonLink to ButtonPrimitive to share common DOM structure, functionality and properties.

There are few breaking changes:
- removing unsafe className (deprecated, not public API)
- removing icon property (deprecated)
- changing type of width to string (planned)

<br/><br/><br/><url>LiveURL: https://orbit-components-mainframe-yolobuttonprimitive.surge.sh</url>


mainframev:
@vepor, @filipdanisko 

can somebody CR? I'll fix it, plus there are warning in the console, not sure how to fix that. Probably that caused by that function which spreads props.

src/Button/index.js
```diff
@@ -1,302 +1,68 @@
...
+      bordered = false,
+      disabled = false,
+      iconLeft,
+      ...props
+    },
+    ref,
+  ) => {
+    const theme = useTheme();
+    const onlyIcon = Boolean(iconLeft && !children);
+    const sizeIcon = getIconSize(size);
```
filipdanisko, line 31: These are IMO a good candidate for moving them in to ButtonPrimitive,

PR #1535, opened 2020-02-17
FIX: Modal close when you make drop outside of modalWrapper

### LINKED ISSUE #964: Modal closes when you make drop outside of modal body
## Current Behavior
modals are closing when you make some selection and drop outside of the modal on chrome (here it is compared to mozilla: https://take.ms/1Yzfe) this is particularly annoying when you try to select something you typed. the issue seems to have started when chrome was updated to v 73.0.3683.86

## Expected Behavior
Should behave like in non chrome browsers, when user moves outside of a modal body with previous click inside body of modal. Modal shouldn't be closed.

## Possible Solution
Create own mouseClick handler and check if there was previous click inside body of modal before calling close function

## Steps to Reproduce
Use nay orbit modal on kiwi.com inside latest Chrome  73.0.3683.86

filipdanisko:
Closes #964<br/><br/><br/><url>LiveURL: https://orbit-components-fix-clickoutside.surge.sh</url>

### DIFF
--- a/src/Modal/index.js
+++ b/src/Modal/index.js
@@ -364,6 +364,8 @@ export class PureModal extends React.PureComponent<Props & ThemeProps, State> {
 
   timeout: TimeoutID;
 
+  clickedModalBody: boolean = false;
+
   static defaultProps = {
     theme: defaultTheme,
   };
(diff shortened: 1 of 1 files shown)


src/Modal/index.js
```diff
@@ -532,6 +535,15 @@ export class PureModal extends React.PureComponent<Props & ThemeProps, State> {
       // If is clicked outside of modal
       onClose(ev);
     }
+    this.clickedModalBody = false;
+  };
+
+  handleMouseDown = () => {
+    /*
+      This is due to issue where it's was possible to close Modal,
+      even though click started (onKeyDown) in ModalWrapper.
```
vepor: `even though click started (onMouseDown) in ModalWrapper.`

PR #1544, opened 2020-02-18
feat(PictureCard): Add src attribute

aexvir:
I also adjusted `LazyImage` to be more flexible, so technically, if `PictureCard` is adjusted accordingly, it should be possible to use the `srcset` feature with arbitrary images, but I considered that to be outisde of the scope for now.

<br/><br/><br/><url>LiveURL: https://orbit-components-aexvir-add-picturecard-src.surge.sh</url>


CollierCZ:
Closes #1834

PR #1551, opened 2020-02-19
DEV: fix broken component names in storybook deployment

jaroslav-kubicek:
Use `@babel/env` with default settings for storybook because `storybook-build` is not compatible with `@adeira/babel-preset-adeira` - modules would need to be set to false and `@babel/plugin-transform-runtime` removed: https://github.com/adeira/universe/blob/master/src/babel-preset-adeira/src/index.js#L98 => otherwise `No Display Name` is shown instead of component name<br/><br/><br/><url>LiveURL: https://orbit-components-fix-storybook-deploy.surge.sh</url>


vepor:
Seems to work fine, thanks a lot @jaroslav-kubicek 💪 
Only thing we need to check is if we had normal names before update to `@adeira` on IE11. But it's not merge blocker as source code won't be copied from IE, at least I think that this kind of suicidal person doesn't exist 😅

vepor:
FYI, it was wrong even before 🙂

PR #1584, opened 2020-03-02
chore(ci): add auto-generated release notes

fallion:
Current release notes are hand-generated. This action will automate this for all commits. Works best if commits follow the conventional commit standard.

There will also be support for highlighting breaking changes added this week. Again, requires conventional commit standard to identify breaking changes.


aexvir:
Agree, but this is up to @vepor 
I'd really like to add both `commitsar` and `release-notary` and enforce the commit styling

vepor:
Hey @fallion, thanks for the PR! It will be very useful 🙂 Just one question, can we use regex for the tag name, e.g. `/\d+\.\d+\.\d+/` so we don't have to use e.g. `v0.70.0`?

fallion:
Will adjust it. It doesn't require the `v` part.

fallion:
Fixed.

PR #1657, opened 2020-03-11
fix(popover): Opening with onKeyDown event

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-popover-keydown-open.surge.sh</url>


filipdanisko:
Looks good, but could we remove the story from storybook? As I don't think it's appropriate ussage of popover

PR #1666, opened 2020-03-13
fix(Breadcrumbs): add correct html5 structured microdata for breadcrumbs

aalises:
**- Problem**

Right now, the Breadcrumbs on Orbit contain structured microdata that helps google identify the different parts of the breadcrumbs for SEO purposes (https://developers.google.com/search/docs/data-types/breadcrumb).

When parsing the HTML, Google takes the `@id` property from the `href` property of a given anchor, to identify the URL of the breadcrumb. However, when the `href` is not present i.e when you are only using the `onClick` handler, this `@id` property is currently not set.

This is caused because the `id` property on the Breadcrumbs do not add the correct property on the breadcrumbs for google to identify them. The correct property to add is `itemid`.

**- Solution**

This MR adds valid microdata for the Breadcrumbs by setting the `itemid` when the `id` prop is present and defaulting to `href` if no `id` is set.

By that, https://search.google.com/structured-data/testing-tool/u/0/ effectively passes and the SEO of the Breadcrumbs is improved when no `href` is set.

**- To test**

Create a Breadcrumb without `href`s (Added the option on `Storybook playground`), and paste the HTML code in the structured-data console. Check that everything passes nicely e.g

Screenshot 2020-03-13 at 09 18 53


aalises:
@vepor Seems like I need a valid surge account for deploying the Storybook? Any account would do it?

PR #1691, opened 2020-03-19
feat(Tile): Adding title attribute

filipdanisko:
Adding title attribute to enable native tooltip on hover to appear, useful for adding extra piece of information.

@vepor What do you think about the naming? I was not able to think of anything better that that, due to already having a Title a Description taken 🤔 <br/><br/><br/><url>LiveURL: https://orbit-components-update-tile-title.surge.sh</url>


sampi:
LGTM!

PR #1727, opened 2020-03-27
docs(grid): Add children count limitation

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-docs-grid-children-limitation.surge.sh</url>

### DIFF
--- a/src/utils/Grid/README.md
+++ b/src/utils/Grid/README.md
@@ -71,10 +71,36 @@ By default, this Grid component solves:
 
 ## Templates known limitations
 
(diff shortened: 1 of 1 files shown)


src/utils/Grid/README.md
```diff
@@ -71,10 +71,36 @@ By default, this Grid component solves:
...
+
+```jsx
+import Grid from "@kiwicom/orbit-components/lib/utils/Grid";
+
+import Cell from "./Cell";
+
+const Wrapper = ({ children }) => {
+  // some magic
+  return children;
+};
```
tbergq: Could you also possibly specify that `React.Fragment` won't work either?

PR #1730, opened 2020-03-27
feat: Adding commitlint and husky

filipdanisko:
Adding commitlint, spliting #1661 to separate MR as we are not sure yet about the changelog stuff
<br/><br/><br/><url>LiveURL: https://orbit-components-update-commit-lint.surge.sh</url>

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "yarn prettier:fix"
+      "pre-push": "yarn prettier:fix"
     }
   },
   "main": "lib/index.js",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,12 @@
       }
     }
   },
+  "husky": {
+    "hooks": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -89,7 +95,10 @@
   },
   "dependencies": {
     "@adeira/js": "^1.2.2",
-    "@kiwicom/orbit-design-tokens": "~0.9.0"
+    "@commitlint/cli": "^8.3.5",
```
vepor: everything should be dev deps 🙏 @filipdanisko

package.json
```diff
@@ -78,6 +78,12 @@
       }
     }
   },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
+      "pre-push": "yarn prettier:fix"
```
vepor: do we really want prettier:fix runned on pre-push? I would have it also for commit

vepor:
Also, can you remove prettier:check from CI?

package.json
filipdanisko: Well `commit-msg` is validated after `pre-commit` making it little painfull to wait for prettier to run before commit is validated, it still takes couple of seconds which is totally ok for push, not so sure about commit though

PR #1732, opened 2020-03-27
feat(Tile): make tile padding large on desktop viewport 

aalises:
This way it's aligned with the behaviour of `Card`


aalises:
@vepor @williamkolmacka Instead of downsizing the `Card` to 16px, could we instead adapt the `Tile` to be also `24px` on desktop viewports?

src/Tile/components/TileContent/index.js
```diff
@@ -2,20 +2,31 @@
...
+        padding: ${theme.orbit.spaceLarge};
+      `)}
     `;
   }
+
   return css`
     padding: ${theme.orbit.spaceMedium} 0;
     margin: 0 ${theme.orbit.spaceMedium};
+
+    ${mq.largeMobile(css`
```
vepor: It needs to be like this:
```
    ${mq.largeMobile(css`
      padding: ${theme.orbit.spaceLarge} 0;
      margin: 0 ${theme.orbit.spaceLarge};
    `)}
```

Otherwise the expandable Tile will be broken 🙂

aalises: Done @vepor 🎉

PR #1882, opened 2020-05-22
feat(Storybook): integration of Playroom

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-feat-playroom.surge.sh</url>


mainframev:
Outdated. We will create new PR with new playground.

vepor:
@mainframev IMO not needed it didn't work as expected and I would not spend time on it right now, since we're researching our own solution

PR #1918, opened 2020-06-03
feat(core): tailored styled API

vepor:


src/core/core.stories.js
```diff
@@ -0,0 +1,72 @@
...
+          ${largeMobile`
+            color: ${theme.orbit.paletteGreenNormal};
+          `};
+        `,
+        ({ width }) => ({
+          as: "span",
+          attrs: { style: { width } },
+          blockList: ["width"],
+        }),
+      );
```
silvenon, line 61: Regarding the API itself, how about mimicking styled-components entirely?

```js
const Box = orbited.attrs(({ width }) => ({
  style: { width },
}))`
  ${({ css, theme, largeMobile }) => css`
    color: ${theme.orbit.paletteRedNormal};
    ${largeMobile`
      color: ${theme.orbit.paletteGreenNormal};
    `};
  `}
`
```

We could filter out props like `width` internally like styled-components, instead of `blockList` and `allowOnly`, but I feel like I'm missing some context here.

vepor:
Archiving for now, we're investigating other options.

PR #1931, opened 2020-06-09
fix(Tag): onRemove now shows close button

filipdanisko:


CollierCZ:
The docs still say it's possible to pass only `selected` and have the close button appear, but is that true?

filipdanisko:
> The docs still say it's possible to pass only `selected` and have the close button appear, but is that true, @filipdanisko ?

Should be all good now

PR #2024, opened 2020-07-15
fix(InputStepper): onChange is triggered when disabled

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-stepper-disabled.surge.sh</url>

### DIFF
--- a/src/InputStepper/__tests__/index.test.js
+++ b/src/InputStepper/__tests__/index.test.js
@@ -5,6 +5,7 @@ import { mount } from "enzyme";
 import InputStepper from "../index";
 import { SIZE_OPTIONS } from "../../InputField/consts";
 import SPACINGS_AFTER from "../../common/getSpacingToken/consts";
+import ButtonLink from "../../ButtonLink";
 
 describe(`InputStepper with help, prefix and suffix`, () => {
   const size = SIZE_OPTIONS.NORMAL;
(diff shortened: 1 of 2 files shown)


silvenon:
Works correctly in the Storybook, but a test wouldn't hurt 🙂 Something like this:

```jsx
it("should not work if disabled", () => {
  const onChangeMock = jest.fn();
  component.setProps({ disabled: true, onChange: onChangeMock });
  const btnDecrement = component.find(ButtonLink).first();
  btnDecrement.simulate("click");
  expect(onChangeMock).not.toHaveBeenCalled();
  // cleanup
  component.setProps({ disabled, onChange });
});
```

but this doesn't work for some reason, setting `disabled` to `false` makes it pass as well. Maybe you can finish it? I'm a little rusty with Enzyme.

silvenon:
Oh, I realized that it passed either way because of `minValue`. You can e.g. select increment button instead:

```jsx
it("should not work if disabled", () => {
  const onChangeMock = jest.fn();
  component.setProps({ disabled: true, onChange: onChangeMock });
  const btnIncrement = component.find(ButtonLink).last();
  btnIncrement.simulate("click");
  expect(onChangeMock).not.toHaveBeenCalled();
  // cleanup
  component.setProps({ disabled, onChange });
});
```

PR #2025, opened 2020-07-15
fix(Layout): inner usage of Card – edge to edge behavior

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-edge-card-layout.surge.sh</url>


src/Layout/LayoutColumn/index.js
```diff
@@ -1,14 +1,27 @@
...
+import { StyledCard } from "../../Card";
+import mq from "../../utils/mediaQuery";
 
 import type { Props } from "./index";
 
 const StyledColumn = styled.div`
   ${({ hideOn }) => !!hideOn && getViewportHideStyles(hideOn)};
+
+  ${StyledCard} {
+    margin: 0 -${({ theme }) => theme.orbit.spaceMedium};
```
silvenon, line 16: My recommendation is to set strictly what's necessary, with this shorthand you can accidentally override a possible `margin-top` or `margin-bottom` down the road.

```suggestion
    margin-left: -${({ theme }) => theme.orbit.spaceMedium};
    margin-right: -${({ theme }) => theme.orbit.spaceMedium};
```

src/Layout/LayoutColumn/index.js
```diff
@@ -1,14 +1,27 @@
...
+    width: auto;
+  }
+  ${mq.tablet(css`
+    ${StyledCard} {
+      margin: auto;
```
silvenon, line 21: Is there a reason for setting to `auto` instead of `0`? `0` is the initial value of `margin`.

```suggestion
      margin-left: 0;
      margin-right: 0;
```

src/Layout/LayoutColumn/index.js
vepor: You're right! Especially if more Cards would be stacked with `Stack` component.
We can also don't have to solve RTL with functions that we have. It is set up for both horizontal sides, so it would be redundant.

src/Layout/LayoutColumn/index.js
vepor: As I'm thinking about it, we should not use either `auto` or `0`, as it can be problematic for some cases. Instead of that, we should implement the other way of media query – `max-width` so we don't have to set up the default value once again, we don't even know what the value will be.

silvenon: I like this solution as well ✅

PR #2028, opened 2020-07-16
feat(Buttons): adding rel attribute

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-update-link-nofollow.surge.sh</url>


src/primitives/ButtonPrimitive/index.js
```diff
@@ -55,7 +66,7 @@ export const StyledButtonPrimitive = styled(
         disabled={disabled}
         href={!disabled ? href : null}
         target={!disabled && href && external ? "_blank" : undefined}
-        rel={!disabled && href && external ? "noopener noreferrer" : undefined}
+        rel={relValues.join(" ")}
```
silvenon, line 69: This is most probably not a problem, but there are attributes where an empty string is not equivalent to nothing, like `alt` and `src`, so I would err on the side of using `undefined` instead.

```suggestion
        rel={relValues.length > 0 ? relValues.join(" ") : undefined}
```

filipdanisko: It should be fine in case of rel, but you are right, just to be safe :)

PR #2035, opened 2020-07-20
chore: build on prepublishOnly instead of prepare

silvenon:
The script `prepare` runs every time we install dependencies, which is unnecessary and painful in development because the `build` script is slow. `prepublishOnly` script is usually a much better fit, so this extracts the necessary stuff to `postinstall`.

https://docs.npmjs.com/misc/scripts<br/><br/><br/><url>LiveURL: https://orbit-components-chore-prepare-to-prepublishonly.surge.sh</url>


silvenon:
Does this change affect GitHub actions as well? I don't know what they do, so I didn't touch them.

vepor:
@silvenon We have currently two actions – one is for creation of changelog for each tag release on GitHub. The second one runs on every PR – triggers build for master and PR branch and diff its size – based on `src/index.js`. Maybe we should trigger  `yarn build:icons` for it also.

silvenon:
@vepor `main.yml` uses `preactjs/compressed-size-action`, which triggers the `build` script, and `build:icons` is a part of that script.

The reason why I was using the `build:icons` script separately in other places is because that's the script that Storybook needs, but it's also included in the `build` script.

I don't think there's any changes necessary, or am I missing something?

silvenon:
I stopped being a smartass and added back the `prepare` script, but now it only runs `build:icons` and `build:typeFiles`, which are necessary for getting started.

vepor (approved):

silvenon:
@vepor I rebased this and made a minor change: renamed `prepare` to `postinstall`, because `prepare` runs before publish, which is unnecessary as we already have `prepublishOnly` for that.

silvenon:
Hm, I guess there are some changes to compressed size 🤔

silvenon:
Waiting for #2183.

vepor (approved):

PR #2037, opened 2020-07-20
docs: add examples for TextLink

CollierCZ:
<br/><br/><br/><url>LiveURL: https://orbit-components-examples-add-textlink.surge.sh</url>


src/TextLink/__examples__/ACCESSIBILITY.js
```diff
@@ -0,0 +1,63 @@
...
+
+export default {
+  Example: () => (
+    <Stack direction="column">
+      <Stack shrink direction="column" spacing="condensed">
+        <Heading type="title3" as="h3">
+          Tab index
+        </Heading>
+        <Text>
+          The <code>tabIndex</code> prop makes the given element focusable in keyboard navigation.
```
filipdanisko, line 18: @CollierCZ This is not exactly true. It's true with positive numbers but it can also be used to hide an element from the accessibility tree. This prop in general should be used only for edge-cases for example when you are rendering TextLink in to non actionable elements like `div` or you want to hide that element from accessibility tree. Could you mention this there as now it can be seen we promote that as preferred solution to make something focusable? :)

filipdanisko (changes requested):

src/TextLink/__examples__/ACCESSIBILITY.js
CollierCZ: Better now?

filipdanisko (approved):

PR #2045, opened 2020-07-23
docs: add examples for InputStepper and Stepper

CollierCZ:
<br/><br/><br/><url>LiveURL: https://orbit-components-examples-add-steppers.surge.sh</url>


filipdanisko:
@CollierCZ is there any specific reason for 10rem max-width? It might be nicer for stories as for example inputstepperstep where heading is broken in to 2 lines

CollierCZ:
> @CollierCZ is there any specific reason for 10rem max-width? It might be nicer for stories as for example inputstepperstep where heading is broken in to 2 lines

Not a specific reason. It just looks ugly when it's the full width of a wide screen. Could make it 11 to keep the label from being broken into 2 lines.

filipdanisko (approved):

PR #2052, opened 2020-07-27
Simplify running ESLint with overrides and automatically fix on pre-commit

silvenon:
By using `overrides` we can resolve the ESLint configuration for JS and TS within `.eslintrc.js` itself, so we don't have to specify the configuration file when running ESLint. This also fixes the issue where text editor extension didn't work for TS files because it was using the default configuration, which is for JS files.

Also, by running ESLint on `pre-commit` with `--fix`, we can catch ESLint problems sooner, so I added that as well.<br/><br/><br/><url>LiveURL: https://orbit-components-eslint-overrides.surge.sh</url>


silvenon:
Btw, `*.ts` covers `*.d.ts` as well.

PR #2058, opened 2020-07-29
chore: lint .js.flow files as well

silvenon:
<br/><br/><br/><url>LiveURL: https://orbit-components-lint-js-flow-files.surge.sh</url>


silvenon:
Fixed, it was linting autogenerated files, so I ignored those.

PR #2063, opened 2020-07-30
chore: bump ESLint dependencies

silvenon:
Mainly, bump ESLint to v7, which figures out extensions from `overrides`, so we no longer need to pass `--ext`. I disabled some errors while updating other ESLint dependencies.<br/><br/><br/><url>LiveURL: https://orbit-components-bump-eslint-v7.surge.sh</url>


silvenon:
ESLint v7 lints the configuration file now as well, that's why there are so many changes there.

PR #2077, opened 2020-08-03
fix(Rtl): flow definition of destructured assigment

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-rtl-def.surge.sh</url>


vepor:
@filipdanisko are TS definitions okay? 🙂

PR #2083, opened 2020-08-04
fix(Typescript): updating definitions

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-ts-errors.surge.sh</url>


src/utils/mediaQuery/index.d.ts
```diff
@@ -16,4 +17,6 @@ export type Devices =
   | "smallMobile";
 
 declare const MediaQuery: { [key in Devices]: QueryFunction };
-export { MediaQuery, MediaQuery as default };
+declare const GetBreakpointWidth: (name: string, theme: ThemeShape, pure?: boolean) => string;
```
vepor, line 20: Should not be it called `getBreakpointWidth` so it corresponds with the actual name? 🤔

PR #2091, opened 2020-08-05
feat(Hooks): useIntersect

mainframev:
I've added a new hook, which might be useful for further development or for anybody who uses orbit. Usage of Observer has some advantages over standard approaches with event listeners.<br/><br/><br/><url>LiveURL: https://orbit-mainframe-useIntersect.surge.sh</url>


vepor:
@mainframev could you also please `README.md` for this hook? 🙏  It would also mention there that the Observer doesn't work in IE 10/11 🙂

mainframev:
> @mainframev could you also please `README.md` for this hook? 🙏 It would also mention there that the Observer doesn't work in IE 10/11 🙂
done ;)

silvenon:
I couldn't figure out why something still felt off in the implementation, so I did some googling and found [this](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) in React Hooks FAQ. This is the right way to do it, so I reimplemented `useIntersect` using `useCallback`.

mainframev:
> I couldn't figure out why something still felt off in the implementation, so I did some googling and found [this](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) in React Hooks FAQ. This is the right way to do it, so I reimplemented `useIntersect` with `useCallback`, and added tests.

I thought it was ok already. Check this [article](https://medium.com/the-non-traditional-developer/how-to-use-an-intersectionobserver-in-a-react-hook-9fb061ac6cb5). In the end guy explains the issue with useRef inside, correct me if i am wrong :) Is there another issue you faced?

silvenon:
Will do! I should've asked how you came up with the implementation 🤦‍♂️

mainframev:
> Will do! I should've asked how you came up with the implementation 🤦‍♂️

I needed Observer API in my own app, so i found this hook :) It's just nice to have, because there are a lot of usecases for it. 

I'll look later to the changes you made :)

silvenon:
Ok, I reverted my `useCallback` implementation because yours passes the basic tests I've written. There's one unlikely edge case where it will fail, but we can fix it later.

silvenon:
So in total I made 3 changes:

1. removed `if (observer.current) observer.current.disconnect();` on top of `useEffect` because I think the cleanup already solves this (the author is wrong that cleanup only happens on unmount, it happens every time the hook updates)
2. added the missing `alt` attribute in the destructuring statement in the `LazyLoaded` example
3. added the test

vepor (approved):

PR #2117, opened 2020-08-12
fix(Browsers): added kiwicom supported browsers to the project

dinodsaurus:
Added a company-wide browser list to the project so we can run tests on it and see if everything behaves as previously 🤞 
Will test it also on few other projects just to ensure that everything behaves the same 😉 
<br/><br/><br/><url>LiveURL: https://orbit-components-fix-supported-browsers.surge.sh</url>

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+extends @kiwicom/browserslist-config
--- a/.github/contribution/testing-conventions.md
+++ b/.github/contribution/testing-conventions.md
@@ -59,8 +59,23 @@ describe("Component", () => {
 
 This part of testing is often abandoned by people, but we are aware of how important this is for users. It’s recommended that you test your new components across different platforms and browsers. New components should work on these browsers:
 
-- **Chrome**
-- **Firefox**
-- **Safari**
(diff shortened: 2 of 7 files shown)


silvenon:
Also, I think this should be treated as a breaking change, what do you think?

dinodsaurus:
Wow, thanks a lot for pointing this out! ❤️ 
Yea this could be treated as a breaking change 🤔 
This package is still in the testing phase..
Could you pls take over this MR and make it work on orbit and see if there are some big problems occurring 🙏 Also it would be worth to check bundle size

silvenon:
There appears to be no changes in bundle size.

filipdanisko:
Can you also please update this section of readme? As it is in relation to this issue :)  https://github.com/kiwicom/orbit-components/blob/master/.github/contribution/testing-conventions.md#browser-compatibility

dinodsaurus:
Love the last commit! I copied it to our supported browsers package 😍  🦖

silvenon:
@filipdanisko with the last commit we shouldn't have to worry about updating the list anymore. Let me know what you think.

PR #2136, opened 2020-08-18
docs: add examples for Loading

CollierCZ:
<br/><br/><br/><url>LiveURL: https://orbit-examples-add-loading.surge.sh</url>


packages/orbit-components/src/Loading/__examples__/LOADING_CHILDREN.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default {
+  Example: () => {
+    const [loadingNow, setLoadingNow] = React.useState(true);
+    return (
+      <Stack direction="column">
+        <Loading loading={loadingNow} type="inlineLoader" text="Saving your request">
+          <Text>Your request has been saved!</Text>
+        </Loading>
+        <Loading loading={loadingNow} type="inlineLoader" text="Saving your request" />
```
vepor, line 17: Isn't this redundant, or did you want to really have it there @CollierCZ? 🙂

packages/orbit-components/src/Loading/__examples__/TYPES.js
```diff
@@ -0,0 +1,69 @@
...
+      </Stack>
+      <Stack spacing="extraTight">
+        <Heading as="h3" type="title3">
+          Button loader
+        </Heading>
+        <Text>
+          Useful when a user has interacted with a button to show that something is happening.
+        </Text>
+        <Button loading>Save</Button>
+        <Button iconLeft={<Loading type="buttonLoader" />} disabled>
```
vepor, line 45: I'm not sure @CollierCZ if this is a right use case. It would be if we would not have `loading` boolean property on the Button component.

vepor (changes requested):

packages/orbit-components/src/Loading/__examples__/LOADING_CHILDREN.js
CollierCZ: The point is that the Loading component without children is always shown, even when you toggle the state. So it was intentional.

packages/orbit-components/src/Loading/__examples__/TYPES.js
CollierCZ: OK. So there's no case to use `buttonLoader` directly then? Thought there might be.

packages/orbit-components/src/Loading/__examples__/LOADING_CHILDREN.js
vepor: Okay, fair point! 🙂

vepor (approved):

PR #2138, opened 2020-08-19
ci: perform Prettier check on Markdown files

silvenon:
Prettier mistakes for Markdown files were still slipping past us because Prettier wasn't being enforced in CI. This is now fixed, along with fixing Prettier mistakes where necessary. Also, I deleted the `test-ci` script because we weren't using it anywhere, I don't think it's necessary to bundle these scripts together, if CI check fails we will only run the script which fails, not all of them.


filipdanisko:
I think `test-ci` is mainly for local development as I use it for that sometimes rather then running all scripts separately, so I'm ok with keeping that one.

silvenon:
Ok, but I changed CI config to run `test-ci`, so now it's truly "ci".

filipdanisko (dismissed):

PR #2143, opened 2020-08-20
Monorepo: Phase One

silvenon:
This PR implements Lerna, adding `orbit-components` as the package. There are many things going on here, I'll write the overview here, and explain the lower-level stuff in review comments.

## How to review this

IMO GitHub UI is not equipped for reviewing PRs of this size, so I recommend reviewing this locally because most of the changes are just moved files, so you can save yourself a lot of scrolling. This will show you the entire diff of the last (and the only) commit, but without moved files.

```
git show --diff-filter=r
```

This should open the diff in less, so press:

- <kbd>J</kbd> to go down
- <kbd>K</kbd> to go up
- <kbd>F</kbd> to jump down by the size of the screen
- <kbd>B</kbd> to jump up by the size of the screen

When you find something to comment on, find the change in GitHub and leave the comment.

Moving on…

## Self-containing

I attempted to move everything related to `orbit-components` in the package directory. Alternatively I could have left stuff like Storybook in the project root, but I decided to go down the route of moving things to the root only if it's related to all packages. I'm open to second opinions, of course.

## Versioning and publishing

This is an excellent opportunity to start utilizing our diligent use of Conventional Commits. Lerna's `version` command supports the `--conventional-commits` flag which automatically knows how to bump the version, and also generates an entry in `CHANGELOG.md`. Combining it with `--create-release github` creates a GitHub release as well.

Because versioning will be independent, `lerna version` will create git tags like `orbit-components@0.90.0`, `orbit-design-tokens@0.11.0` etc.

If we want to continue with the current release strategy, locally we should run `lerna version ...` (the entire command is saved as `yarn release`), then CI will run `lerna publish from-git`. I wasn't able to test the latter locally because I didn't want to publish anything, so if that won't work as expected we'll come up with a different strategy, for example running `lerna publish` ourselves without `from-git`, which will run `lerna version` under the hood.

## Exact cross-dependency versions?

One decision we can make while we're doing this is whether we're going to switch to exact versioning of cross-dependencies. For example, `orbit-components` depends on `orbit-design-tokens`, so if a new patch version of `orbit-design-tokens` is released, how will consumers of `orbit-components` be able to update?

Exact versioning means that instead of `"orbit-design-tokens": "~0.10.0"` we can have `"orbit-design-tokens": "0.10.0"`, and Lerna will know what to do. If `orbit-design-tokens` needs to be released, Lerna will release a new version of `orbit-components` as well with updated version of `orbit-design-tokens` in its `package.json`.

I personally think this is better, but might be an overkill at this point if nobody had problems with this yet.

<br/><br/><br/><url>LiveURL: https://orbit-components-monorepo-phase-1.surge.sh</url>


.gitignore
```diff
@@ -63,24 +63,3 @@ typings/
...
-src/FeatureIcon/index.js.flow
-src/Illustration/index.d.ts
-src/AirportIllustration/index.d.ts
-src/ServiceLogo/index.d.ts
-src/FeatureIcon/index.d.ts
-orbit-icons-font
-orbit-email-icons
-orbit-icons-font.zip
-orbit-svgs.zip
-
```
silvenon, line 86: This is moved to `packages/orbit-components/.gitignore`.

babel.config.js
```diff
@@ -25,18 +9,12 @@ const environments = {
...
+module.exports = {
+  presets: [["@adeira/babel-preset-adeira", { target, environments }]],
+  plugins: ["babel-plugin-styled-components"],
+  env: {
+    test: {
+      plugins: ["require-context-hook"],
     },
-  };
+  },
 };
```
silvenon, line 20: I simplified this, now this serves as the base preset which packages extend as needed, e.g. `orbit-components` modifies it if `BABEL_TARGET` is `storybook`.

package.json
```diff
@@ -1,80 +1,18 @@
...
-        "target": "chrome.docker",
-        "width": 1366,
-        "height": 768
-      },
-      "chrome.iphone7": {
-        "target": "chrome.docker",
-        "preset": "iPhone 7"
-      }
-    }
+    "release": "lerna version --conventional-commits"
```
silvenon, line 15: This is the script we should run for creating releases.

packages/orbit-components/babel.config.js
```diff
@@ -0,0 +1,7 @@
+// @noflow
+const baseConfig = require("../../babel.config");
+
+module.exports = {
+  ...baseConfig,
+  presets: process.env.BABEL_TARGET === "storybook" ? [] : baseConfig.presets,
```
silvenon, line 6: Because Storybook is only used for `orbit-components` (for now), I moved that part of Babel config logic here.

packages/orbit-components/package.json
```diff
@@ -0,0 +1,143 @@
+{
```
silvenon, line 1: Nothing new in this `package.json` IIRC, I just extracted the parts of the root `package.json` that were related to `orbit-components`.

readme.md
```diff
@@ -7,7 +7,11 @@
 <br />
 
 <div align="center">
-  <strong>Orbit-components is a React component library which provides developers with the easiest possible way of building Kiwi.com’s products.</strong>
+
+[![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg)](https://lerna.js.org/)
```
silvenon, line 11: Just added the badge ✨

.github/workflows/main.yml
```diff
@@ -10,5 +10,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: preactjs/compressed-size-action@v2
         with:
+          build-script: yarn orbit-components build
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           pattern: "lib/**/*.js"
```
silvenon, line 15: Update this!

filipdanisko (approved):

PR #2180, opened 2020-08-26
chore: enforcing yarn's version

filipdanisko:
The monorepo does not work on older Yarn version. I'm leveraging the `yarn policies set-version` here which adds a copy of specified yarn version to repository so every yarn command is run from this copy ensuring same yarn behaviour across the project. more on this topic https://classic.yarnpkg.com/en/docs/cli/policies/<br/><br/><br/><url>LiveURL: https://orbit-yarn-policies.surge.sh</url>


silvenon (changes requested):
Could you also add `engines` to `package.json`?

```json
{
  "engines": {
    "yarn": ">=1.22.4"
  }
}
```

or just `1.22.4`, however you want.

This will cause Yarn to throw an error on install if it's the wrong version.

silvenon (approved):

PR #2188, opened 2020-08-27
feat(Packages): import orbit-design-tokens

vepor:
Didn’t change any implementation, except the fix for TS definitions. Other than that, few notes: 

- I needed to temporarily suppress some ESLint warnings/error, because it doesn’t make sense to refactor it right now – will be deleted or refactored anyway.
- Also I’ve created separate `build` script, so it can be used locally and also on the CI, via `build-ci`  and `test-ci`. It was also necessary to build the tokens first.

Also, please just merge and don't squash.<br/><br/><br/><url>LiveURL: https://orbit-feat-import-tokens.surge.sh</url>


silvenon:
You need to add `orbit-design-tokens`' `build` script in `build-ci`.

packages/orbit-design-tokens/src/index.d.ts
```diff
@@ -0,0 +1,623 @@
...
+type CustomBase = Partial<Base>;
+
+type CustomFoundation = Partial<{
+  palette: CustomPalette;
+  base: CustomBase;
+}>;
+
+export declare const ConvertHexToRgba: (color: string, opacity: number) => string;
+export declare const FromPlainObject: (themePaletteColors: ThemePaletteColors) => Tokens;
+export declare const GetTokens: (customFoundation?: CustomFoundation) => Tokens;
```
filipdanisko, line 622: Just a quick fix of exported types, did not wanted to do that in the orbit-design-tokens repository
```suggestion
export declare const fromPlainObject: (themePaletteColors: ThemePaletteColors) => Tokens;
export declare const getTokens: (customFoundation?: CustomFoundation) => Tokens;
```

.github/workflows/test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock', '.yarnclean') }}
 
       - name: Install dependencies
-        run: yarn install && yarn bootstrap
+        run: yarn install && yarn bootstrap && yarn orbit-design-tokens build
```
silvenon, line 44: Better add another step below named `Build design tokens` for now, just to be explicit about it, later we will work on adding more sense to our scripts.

packages/orbit-design-tokens/.gitignore
```diff
@@ -0,0 +1,5 @@
+/node_modules
+.idea
+es
+lib
+.vscode
```
silvenon, line 5: You can remove `.node_modules`, `.vscode` and `.idea`, they are already covered by our root `.gitignore`.

packages/orbit-design-tokens/.npmrc
```diff
@@ -0,0 +1,2 @@
+scope=@kiwicom
+access=public
```
silvenon, line 2: Deleting this file should be fine, but just in case add `"private": false` to `package.json`.

silvenon (changes requested):

silvenon (approved):

PR #2198, opened 2020-08-28
chore/add prettier to orbit components

silvenon:
- chore: add missing prettier dependency
- ci: fix ESLint check script


silvenon:
Accidentally opened #2199 twice, so this is a duplicate.

PR #2202, opened 2020-08-28
fix:  updating typescript definitions to match flow

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-ts-fixes.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/SocialButton/index.d.ts
+++ b/packages/orbit-components/src/SocialButton/index.d.ts
@@ -10,14 +10,7 @@ declare module "@kiwicom/orbit-components/lib/Button";
 
 type Type = "apple" | "facebook" | "google" | "twitter";
 
-type OmittedButtonCommonProps = Exclude<
-  ButtonCommonProps,
-  {
-    readonly iconLeft?: React.ReactNode;
-    readonly iconRight?: React.ReactNode;
-    readonly circled?: boolean;
-  }
->;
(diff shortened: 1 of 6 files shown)


packages/orbit-components/src/SocialButton/index.d.ts
```diff
@@ -10,14 +10,7 @@ declare module "@kiwicom/orbit-components/lib/Button";
 
 type Type = "apple" | "facebook" | "google" | "twitter";
```
silvenon, line 11: According to the [Flow counterpart](https://github.com/kiwicom/orbit/blob/809b0c34f888a411752896c87ffe759284c249ed/packages/orbit-components/src/SocialButton/index.js.flow#L9) this needs to be exported as well, no?

packages/orbit-components/src/TextLink/index.d.ts
```diff
@@ -8,9 +8,9 @@ import * as Common from "../common/common";
 
 declare module "@kiwicom/orbit-components/lib/TextLink";
 
-type Type = "primary" | "secondary";
+export type Type = "primary" | "secondary";
```
silvenon, line 11: I don't see this being exported from the Flow counterpart, am I missing something? It probably should be, though, feel free to add it.

silvenon (changes requested):

silvenon (approved):

PR #2230, opened 2020-09-09
feat(Tooltip): refactoring Tooltip into 2 separate components

filipdanisko:
## What
Current Tooltip implementation contained 2 component in one, I have separated the functionality of Tooltip on desktop in to `TooltipPrimitive` and the Mobile version of a Tooltip in to `MobileDialogPrimitive`. The `Tooltip` component now just renders conditionally those 2 components. (names of the components are open to discussion, as they are not truly a primitive components). 

## Why
There are 2 reasons why we want to do this:
1. Similar components to MobileDialogPrimitive are used in other components (Popover, Modal, Dialog). However current implementation of those components is heavily tight in to them. Separation in this PR is a first step to make them reusable so we can later refactor Popover and other components to use MobileDialogPrimitive too.

2. To have a TooltipPrimitive, meaning a tooltip that not reacts to media query is a key step to be able to use them inside new error form implementation.
<br/><br/><br/><url>LiveURL: https://orbit-feat-tooltip-refactor.surge.sh</url>


packages/orbit-components/src/Tooltip/__tests__/index.test.js
```diff
@@ -1,13 +1,21 @@
...
 import Airplane from "../../icons/Airplane";
 
+jest.mock("../../hooks/useMediaQuery", () => {
+  return () => {
+    return {
+      isLargeMobile: true,
+    };
+  };
+});
+
```
silvenon, line 15: Nice one 👍

packages/orbit-components/src/primitives/MobileDialogPrimitive/components/DialogContent.flow.js
```diff
@@ -0,0 +1,13 @@
...
+export type Props = {|
+  shown: boolean,
+  dialogId: ?string,
+  children: React$Node,
+  parent: ?React$Node,
+  onClose: () => void,
+  ...Globals,
+|};
+
+declare export default React.ComponentType<Props>;
```
silvenon, line 13: Should there be a TypeScript definition for this? Probably not if it's a private component, right?

packages/orbit-components/src/Tooltip/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -1,18 +1,2145 @@
...
+                </StyledComponent>
+              </Icon__StyledIcon>
+            </OrbitIcon>
+          </Airplane>
+        </span>
+      </StyledComponent>
+    </TooltipPrimitive__StyledTooltipChildren>
+  </TooltipPrimitive>
+</Tooltip>
 `;
```
silvenon, line 2145: OMFG, we have to get rid of snapshot tests like this 😆

silvenon (commented):
Otherwise LGTM!

packages/orbit-components/src/primitives/MobileDialogPrimitive/components/DialogContent.flow.js
filipdanisko: Exactly as it is not intended to be exported then flow should be sufficient.

packages/orbit-components/src/Tooltip/__tests__/index.test.js
filipdanisko: Do you have any idea how I could re-mock that for a different test? Eg I want to test for `isLargeMobile: true` and also for `false`. I was not able to do that simply 🤔

packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
```diff
@@ -0,0 +1,79 @@
...
+        tabIndex={enabled ? tabIndex : undefined}
+        enabled={enabled}
+        removeUnderlinedText={removeUnderlinedText}
+      >
+        {children}
+      </StyledTooltipChildren>
+      {enabled && render && (
+        <Portal renderInto="dialogs">
+          <DialogContent
+            parent={children}
```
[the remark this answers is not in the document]
filipdanisko: I missed that when cleaning up, good catch!

packages/orbit-components/src/Tooltip/__tests__/index.test.js
silvenon: [`jest.doMock`](https://jestjs.io/docs/en/jest-object#jestdomockmodulename-factory-options) is for mocking per-test. Is that what you meant? But I'm not sure if `require` will work here, if it doesn't try appending `.default` to the result of `require` call.

vepor (changes requested):

packages/orbit-components/src/Tooltip/index.js
```diff
@@ -47,94 +20,33 @@ const Tooltip = ({
...
-    setRenderWithTimeout(false);
-  }, [setRenderWithTimeout, setShownMobile]);
-
-  useEffect(() => {
-    setTooltipId(RandomID("tooltip"));
-  }, []);
-  if (!enabled) return children;
+  const { isLargeMobile } = useMediaQuery();
   return (
     <>
```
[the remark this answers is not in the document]
silvenon: I didn't say anything because it makes the diff nicer. 😄

packages/orbit-components/src/Tooltip/__tests__/index.test.js
```diff
@@ -1,13 +1,21 @@
...
 describe("Tooltip", () => {
   const content = "Write some message to the user";
-  const component = shallow(
+  const component = mount(
```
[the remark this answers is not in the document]
silvenon: He did remove the snapshot. https://github.com/kiwicom/orbit/pull/2230/files#diff-2672540b2fc19e5fba04fbf75a3bf727R24

mainframev (dismissed):

packages/orbit-components/src/primitives/MobileDialogPrimitive/components/DialogContent.flow.js
```diff
@@ -0,0 +1,12 @@
...
+
+export type Props = {|
+  shown: boolean,
+  dialogId: ?string,
+  children: React$Node,
+  onClose: () => void,
+  ...Globals,
+|};
+
+declare export default React.ComponentType<Props>;
```
vepor, line 12: We're using `React.ComponentType` but it's not imported. Also we should prohibit usage of `React$Node`.

vepor (changes requested):

mainframev:
@silvenon @vepor can me werge it?

vepor (approved):
Yes, LGTM! 💪

mainframev (approved):

PR #2235, opened 2020-09-11
fix(InputGroup): filled should be when values are set to all children

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-fix-inputgroup-filled.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/InputGroup/index.js
+++ b/packages/orbit-components/src/InputGroup/index.js
@@ -159,7 +159,7 @@ const findPropInChild = (propToFind, children) => {
       if (el.props && typeof el.props[propToFind] !== "undefined") return el.props[propToFind];
       return null;
     })
-    .filter(el => el !== null);
+    .filter(el => el !== null && el !== "");
 };
 
 const InputGroup = ({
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/InputGroup/index.js
```diff
@@ -156,10 +156,10 @@ StyledInputGroup.defaultProps = {
 const findPropInChild = (propToFind, children) => {
   return React.Children.toArray(children)
     .map(el => {
-      if (el.props && el.props[propToFind]) return el.props[propToFind];
+      if (el.props && typeof el.props[propToFind] !== "undefined") return el.props[propToFind];
       return null;
     })
-    .filter(Boolean);
+    .filter(el => el !== null);
```
silvenon, line 162: So empty strings are fine?

vepor: 👍

vepor, line 162:
```suggestion
    .filter(el => el !== null && el !== "");
```

filipdanisko (approved):

PR #2242, opened 2020-09-15
feat: introduce Box component

filipdanisko:
<br/><br/><br/><url>LiveURL: https://orbit-new-box.surge.sh</url>


mainframev:
Probably later we will have to unify logic between `Inline` and `Box`, because it has almost same logic implement twice. 
I suggest to review both of them, merge them and later to refactor. Also it has some similarities with `Stack`

silvenon (changes requested):

silvenon (dismissed):

packages/orbit-components/src/Box/Box.stories.js
```diff
@@ -0,0 +1,272 @@
...
+      </Box>
+    );
+  })
+  .add("TextAlign", () => {
+    const textAlign = select("textAlign", TEXT_ALIGN, TEXT_ALIGN.CENTER);
+
+    return (
+      <Box
+        background={DEFAULT_COLOR}
+        textAlign={textAlign}
```
vepor, line 68: The `align` is not visible since `Text` component has its own `text-align`. I would not use the `Text` inside and that's it.

packages/orbit-components/src/Inline/index.d.ts
```diff
@@ -0,0 +1,35 @@
+// Type definitions for @kiwicom/orbit-components
```
vepor, line 1: This file is for `Inline` component so it should not exist in this PR 😨

packages/orbit-components/src/Box/index.js.flow
```diff
@@ -0,0 +1,147 @@
...
+  +top?: string,
+  +right?: string,
+  +bottom?: string,
+  +align?: "start" | "end" | "center" | "stretch",
+  +justify?: "center" | "start" | "end" | "between" | "around",
+  +textAlign?: "left" | "right" | "center",
+  +elevation?: Elevation,
+  +color?: ColorTokens,
+  +background?: ColorTokens,
+  +borderRadius?: "small" | "normal" | "large" | "radiusCircle",
```
vepor, line 135: The use of `radiusCircle` won't work, because the final normalized token name will be `borderRadiusRadiusCircle`.

It should use the `borderRadiusCircle` token, hence the short name should be `circle`.

Don't forget to update it in all places.

packages/orbit-components/src/Box/Box.stories.js
```diff
@@ -0,0 +1,272 @@
...
+    const display = select("display", ["flex", "inline-flex"], "flex");
+    const align = select("align", ALIGN, ALIGN[0]);
+    const justify = select("justify", JUSTIFY, "start");
+    const wrap = select("wrap", WRAP, WRAP[0]);
+
+    return (
+      <Box
+        display={display}
+        justify={justify}
+        background="blueNormal"
```
vepor, line 189: Would delete the background in this case and would change the background of the second example rectangle, otherwise is too confusing.

packages/orbit-components/src/Box/Box.stories.js
```diff
@@ -0,0 +1,272 @@
...
+        </Box>
+      );
+    },
+    {
+      info:
+        "This is the default configuration of this component. Visit Orbit.Kiwi for more detailed guidelines.",
+    },
+  )
+  .add("Overflow", () => {
+    const overflow = select("overflow", OVERFLOW, OVERFLOW.AUTO);
```
vepor, line 48: Since we have changed the definition of consts from arrays to objects, we should also change it in the stories.
e.g. `const overflow = select("overflow", Object.keys(OVERFLOW), OVERFLOW.AUTO);`

I noticed that in some cases the `OVERFLOW[0]` is used and that's also incorrect.

packages/orbit-components/src/Box/consts.js
```diff
@@ -0,0 +1,156 @@
+// @flow
+import theme from "../defaultTheme";
+
```
vepor, line 3: Most of the defined constants are being used only in the stories, but a fragment of it is being imported into `normalize.js` helper function. Therefore it will increase the bundle size with unused stuff.

Would split it – leave `WIDTH_AND_HEIGHT` and `TOKENS` here, the rest can be moved to the stories file for playground purposes.

packages/orbit-components/src/Box/normalize.js
```diff
@@ -0,0 +1,98 @@
...
+const isJsxProp = (key): boolean => Object.keys(jsxPropsMap).includes(key);
+
+const formatCSS = (key, value): string => `${key}: ${value};`;
+
+const firstToUpper = string => string.charAt(0).toUpperCase() + string.slice(1);
+
+const normalizeSpacing = (el, property): string[] => {
+  if (typeof el === "object") {
+    return Object.entries(el).map(([key, val]: any) => {
+      return formatCSS(`${property}-${key}`, TOKENS[val]);
```
vepor, line 28: Normalizing spacing like this will block themeability, because the `theme` is directly used and imported in the `consts.js`. It should use the proper value from context provider.

Recommending to put another helper function to this file that would use `theme` property passed to `normalize` function.

packages/orbit-components/src/Box/Box.stories.js
```diff
@@ -0,0 +1,272 @@
...
+      const maxHeight = text("max-height", "100px");
+      const elevation = select("elevation", ELEVATION, ELEVATION.ACTION);
+      const borderRadius = select("border-radius", BORDER_RADIUS, BORDER_RADIUS.NORMAL);
+      const position = select("position", POSITION, POSITION.RELATIVE);
+      const top = text("top", "10px");
+      const right = text("right", "10px");
+      const bottom = text("left", "10px");
+      const left = text("bottom", "10px");
+      const margin = select("margin", SPACINGS, SPACINGS[0]);
+      const padding = object("padding", { top: "small", left: "x-small", right: "xx-small" }, "");
```
vepor, line 225: Usage of the old spacing values.

packages/orbit-components/src/Box/Box.stories.js
```diff
@@ -0,0 +1,272 @@
...
+    },
+    {
+      info: "Some description about this type of component. ",
+    },
+  )
+  .add(
+    "RTL",
+    () => (
+      <RenderInRtl>
+        <Box>Box in RTL</Box>
```
vepor, line 266: Having screenshotted this won't be efficient enough.

We should add properties that need to be RTL aware, e.g. `left`, `right`, horizontal `margin` and `padding` and `textAlign`.

packages/orbit-components/src/Box/index.js
```diff
@@ -0,0 +1,56 @@
...
+import { DEVICES as DEVICES_CONSTS } from "../utils/mediaQuery/consts";
+import normalize from "./normalize";
+
+import type { Props } from "./index";
+
+const StyledBox = styled(({ className, as: Element, children, dataTest }) => (
+  <Element className={className} data-test={dataTest}>
+    {children}
+  </Element>
+)).attrs(({ dataTest }) => ({ "data-test": dataTest }))`
```
vepor, line 16: Redundant usage of `attrs`. `data-test` is being rendered directly in the return of `styled` factory.

vepor (changes requested):

vepor:
Also, what I miss is one important file – `README.md` 😅

packages/orbit-components/src/Box/consts.js
```diff
@@ -0,0 +1,156 @@
...
+  "large",
+  "Xlarge",
+  "XXlarge",
+  "XXXlarge",
+];
+
+export const TOKENS = {
+  none: null,
+  XXXSmall: theme.orbit.spaceXXXSmall,
+  XXsmall: theme.orbit.spaceXXSmall,
```
vepor, line 148: All the spacing names should be in `camelCase`. Because we use also `T-Shirt` sizes it should look like `XXSmall`. Check #2457.

packages/orbit-components/src/Box/index.js
mainframev: @vepor  it's for the test, it will not work without it

packages/orbit-components/src/Box/consts.js
mainframev: it was just misspelled

packages/orbit-components/src/Box/Box.stories.js
silvenon: I wanted `Text` to have the correct `font-family`. But considering that this component can set `text-align`, maybe it makes sense that it sets `font-family` just like `Text`?

vepor: That sounds great 🙂  Why not – it won't be any side effect that we would not like.

packages/orbit-components/src/Box/index.js.flow
```diff
@@ -0,0 +1,147 @@
...
+  | "blueNormal"
+  | "blueNormalHover"
+  | "blueNormalActive"
+  | "blueDark"
+  | "blueDarkHover"
+  | "blueDarkActive"
+  | "blueDarker"
+  | "socialFacebook"
+  | "socialFacebookHover"
+  | "socialFacebookActive";
```
silvenon, line 84: I wish there was a way to not have to repeat all of this twice manually 😅 I'll think about automating this one day. (Just thinking out loud, feel free to resolve this comment.)

silvenon (changes requested):

packages/orbit-components/src/Box/Box.stories.js
silvenon: There are still `Text` components the stories and `Box` should have `font-family` set so we don't have to use `Text` to avoid Times New Roman.

packages/orbit-components/src/Box/normalize.js
```diff
@@ -0,0 +1,102 @@
...
+
+  if (el !== "none") {
+    return [formatCSS(property, TOKENS(theme)[el])];
+  }
+
+  return [];
+};
+
+const normalizeToken = (token, property, prefix, theme: Theme): string => {
+  if (token === "radiusCircle") return formatCSS(property, theme.orbit.borderRadiusCircle);
```
vepor, line 40: As suggested before, I still would use `circle` name for this option of `borderRadius`. Because we sticked to `radiusCircle` special condition is needed and I don't see any advantage of it 🤔

packages/orbit-components/src/Box/__tests__/index.test.js
```diff
@@ -0,0 +1,166 @@
...
+
+  it.each(Object.entries(DIRECTIONS))("should have directions", (key, direction) => {
+    render(
+      <Box dataTest={`${dataTest}-${key}`} display="flex" direction={DIRECTIONS[key]}>
+        kek
+      </Box>,
+    );
+
+    expect(screen.getByTestId(`${dataTest}-${key}`)).toHaveStyle({ "flex-direction": direction });
+  });
```
silvenon, line 89: - the test name needs to vary based on the value, otherwise the terminal output is confusing because the same test name is being printed several times
- you don't need keys here, only values, so use `Object.values`
- you don't need to construct `dataTest`, just use the the same one because these are separate tests, so `cleanup` runs before each one

The same goes for all other `it.each` tests.

```suggestion
  it.each(Object.values(DIRECTIONS))("should have directions: %s", (direction) => {
    render(
      <Box dataTest={dataTest} display="flex" direction={direction}>
        kek
      </Box>,
    );

    expect(screen.getByTestId(dataTest)).toHaveStyle({ "flex-direction": direction });
  });
```

packages/orbit-components/src/Box/README.md
```diff
@@ -0,0 +1,161 @@
...
+| `"action"`  |
+| `"fixed"`   |
+| `"overlay"` |
+| `"raised"`  |
+
+### spacingObject
+
+| key        | value                          |
+| :--------- | :----------------------------- |
+| `"top"`    | [`spacingToken](#spacingtoken) |
```
vepor, line 143: Missing backtick on all the lines. Also it was not clear to me, how I should define the `spacingObject`, because it refers to it very oddly.

Would attach simple example or note below the `spacingObject` title, e.g.:
```
If you want to define `spacingObject`, you can define specific directions like this:
<Box margin={{ top: "XXSmall", bottom: "XXSmall" }} />
```

vepor (changes requested):
Few things before merge 🚀

silvenon (changes requested):

silvenon (dismissed):

vepor (approved):

PR #2254, opened 2020-09-19
ci: fix invalid tag pattern

silvenon:
I incorrectly fixed it in #2240. See the [filter pattern cheat sheet](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet).<br/><br/><br/><url>LiveURL: https://orbit-ci-fix-tag-pattern.surge.sh</url>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - "@kiwicom/[a-z-]+@[0-9]+.[0-9]+.[0-9]+"
+      - "@kiwicom/*@[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   release-notary:


.github/workflows/release.yml
```diff
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - "@kiwicom/[a-z-]+@[0-9]+.[0-9]+.[0-9]+"
+      - "@kiwicom/*@[0-9]+.[0-9]+.[0-9]+"
```
mainframev, line 5: this should work better `@kiwicom/*@[0-9]*.[0-9]*.[0-9]*`

silvenon: The thing is that in workflows `*` doesn't refer to preceding characters, like `+` does. The pattern you suggested would match `@kiwicom/orbit-components@0ab.92.0a1c_`.

silvenon: Why do you propose `*` over `+`?

mainframev: well, i tried it [here](https://globster.xyz/) and it did not work, but you are right, + should match preceding chars, so never mind :)

mainframev (approved):

.github/workflows/release.yml
silvenon: You'll tell me tomorrow what exactly didn't work because globster is for matching files, and we're matching the tag name pattern.

PR #2255, opened 2020-09-21
feat: introduce Inline component

filipdanisko:
Adding an Inline component 
<br/><br/><br/><url>LiveURL: https://orbit-new-inline-component.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Inline/Inline.stories.js
+++ b/packages/orbit-components/src/Inline/Inline.stories.js
@@ -8,8 +8,8 @@ import { ALIGN, JUSTIFY, SPACING } from "./consts";
 
 import Inline from "./index";
 
-const PlaceHolder = () => {
-  return <div style={{ width: "50px", height: "50px", background: "grey" }} />;
+const PlaceHolder = ({ height = 50 }) => {
+  return <div style={{ width: "50px", height: `${height}px`, background: "grey" }} />;
 };
 
 const dataTest = "test";
@@ -18,11 +18,11 @@ const TestChildren = () => {
   return (
     <>
       <PlaceHolder />
-      <PlaceHolder />
-      <PlaceHolder />
-      <PlaceHolder />
-      <PlaceHolder />
-      <PlaceHolder />
+      <PlaceHolder height={60} />
+      <PlaceHolder height={70} />
+      <PlaceHolder height={80} />
+      <PlaceHolder height={90} />
+      <PlaceHolder height={100} />
     </>
   );
 };
@@ -47,8 +47,8 @@ storiesOf("Inline", module)
     "Playground",
     () => {
       const as = text("As", "div");
-      const align = select("align", ALIGN, ALIGN[0]);
-      const justify = select("justify", JUSTIFY, JUSTIFY[0]);
+      const align = select("align", ALIGN, ALIGN.START);
+      const justify = select("justify", JUSTIFY, JUSTIFY.START);
       const spacing = select("spacing", SPACING, SPACING[1]);
 
       return (
@@ -58,7 +58,8 @@ storiesOf("Inline", module)
       );
     },
     {
-      info: "Some description about this type of component. ",
+      info:
+        "info: You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
     },
   )
   .add(
--- a/packages/orbit-components/src/Inline/Inline.stories.js
+++ b/packages/orbit-components/src/Inline/Inline.stories.js
@@ -1,0 +1,77 @@
+// @flow
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import { text, select } from "@storybook/addon-knobs";
+
+import RenderInRtl from "../utils/rtl/RenderInRtl";
+import { ALIGN, JUSTIFY, SPACING } from "./consts";
+
+import Inline from "./index";
+
+const PlaceHolder = () => {
+  return <div style={{ width: "50px", height: "50px", background: "grey" }} />;
+};
+
+const TestChildrens = () => {
+  return (
+    <>
+      <PlaceHolder />
+      <PlaceHolder />
+      <PlaceHolder />
+      <PlaceHolder />
+      <PlaceHolder />
+      <PlaceHolder />
+    </>
+  );
+};
+
+storiesOf("Inline", module)
+  .add(
+    "Default",
+    () => {
+      const spacing = select("spacing", SPACING, SPACING[1]);
+      return (
+        <Inline spacing={spacing}>
+          <TestChildrens />
+        </Inline>
+      );
+    },
+    {
+      info:
+        "This is the default configuration of this component. Visit Orbit.Kiwi for more detailed guidelines.",
+    },
+  )
+  .add(
+    "Playground",
+    () => {
+      const as = text("As", "div");
+      const align = select("align", ALIGN, ALIGN[0]);
+      const justify = select("justify", JUSTIFY, JUSTIFY[0]);
+      const spacing = select("spacing", SPACING, SPACING[1]);
+
+      return (
+        <Inline as={as} align={align} justify={justify} spacing={spacing}>
+          <TestChildrens />
+        </Inline>
+      );
+    },
+    {
+      info: "Some description about this type of component. ",
+    },
+  )
+  .add(
+    "RTL",
+    () => {
+      const spacing = select("spacing", SPACING, SPACING[1]);
+      return (
+        <RenderInRtl>
+          <Inline spacing={spacing}>
+            <TestChildrens />
+          </Inline>
+        </RenderInRtl>
+      );
+    },
+    {
+      info: "This is a preview of this component in RTL setup.",
+    },
+  );
--- a/packages/orbit-components/src/Inline/Inline.stories.js
+++ b/packages/orbit-components/src/Inline/Inline.stories.js
@@ -12,7 +12,9 @@ const PlaceHolder = () => {
   return <div style={{ width: "50px", height: "50px", background: "grey" }} />;
 };
 
-const TestChildrens = () => {
+const dataTest = "test";
+
+const TestChildren = () => {
   return (
     <>
       <PlaceHolder />
@@ -31,8 +33,8 @@ storiesOf("Inline", module)
     () => {
       const spacing = select("spacing", SPACING, SPACING[1]);
       return (
-        <Inline spacing={spacing}>
-          <TestChildrens />
+        <Inline spacing={spacing} dataTest={dataTest}>
+          <TestChildren />
         </Inline>
       );
     },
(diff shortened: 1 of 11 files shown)


packages/orbit-components/src/Inline/Inline.stories.js
```diff
@@ -0,0 +1,79 @@
...
+      const spacing = select("spacing", SPACING, SPACING[1]);
+
+      return (
+        <Inline as={as} align={align} justify={justify} dataTest={dataTest} spacing={spacing}>
+          <TestChildren />
+        </Inline>
+      );
+    },
+    {
+      info: "Some description about this type of component. ",
```
silvenon, line 61: This looks like a placeholder, you can copy the description used by some components:
```suggestion
      info: "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
```

packages/orbit-components/src/Inline/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+  const tokens = {
+    none: "",
+    "xx-small": theme.orbit.spaceXXSmall,
+    "x-small": theme.orbit.spaceXSmall,
+    small: theme.orbit.spaceSmall,
+    medium: theme.orbit.spaceMedium,
+    large: theme.orbit.spaceLarge,
+    "x-large": theme.orbit.spaceXLarge,
+    "xx-large": theme.orbit.spaceXLarge,
+    "xxx-large": theme.orbit.spaceXLarge,
```
silvenon, line 28: Also, just to point out that these three sizes are mapped to the same token.

silvenon (approved):

packages/orbit-components/src/Inline/consts.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+export const ALIGN = ["start", "end", "center"];
```
vepor, line 3: Can we convert all arrays do objects? We use it everywhere else and it's better in the usage in my opinion.
Instead `ALIGN[0]` to use `ALIGN.START`, since I don't remember what's the first in the array.

packages/orbit-components/src/Inline/index.js
```diff
@@ -0,0 +1,58 @@
...
+  <Element>{children}</Element>
+)).attrs(({ dataTest }) => ({ "data-test": dataTest }))`
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+
+  // map to mediaQueries
+  ${({ viewportSizes }) => {
+    return Object.entries(viewportSizes).map(([query, value]: any) => {
+      // DEVICES_CONSTS[0] === "smallMobile"
```
vepor, line 22: Comments should not have look like the same as the condition below them 😅 Can we rephrase it, even though that's kinda easy condition? 🙏

vepor (changes requested):

packages/orbit-components/src/Inline/index.js.flow
```diff
@@ -0,0 +1,40 @@
...
+*/
+import * as React from "react";
+
+import type { Globals } from "../common/common.js.flow";
+
+export type Position = "start" | "end" | "center";
+// spacing tokens can then be imported from Box component
+export type SpacingToken =
+  | "none"
+  | "xx-small"
```
silvenon, line 13: Is `xxx-small` missing from both type definitions on purpose?

silvenon (changes requested):

mainframev:
@vepor @silvenon is ready for merge

packages/orbit-components/src/Inline/__tests__/index.test.js
```diff
@@ -0,0 +1,99 @@
...
+          kek
+        </div>
+        <div>bur</div>
+      </Inline>
+    ));
+
+    expect(getComputedStyle(render(spacings[0]).getByTestId("none"))).toHaveProperty(
+      "margin-left",
+      "",
+    );
```
silvenon, line 57: I recommend deleting this `it` block entirely, you're just asserting that `spacing` maps to correct tokens, and a mistake happening there is just as likely as making a mistake in the test file itself. 😉

silvenon (commented):
Some more thoughts while we're waiting to unify naming for spacing props.

packages/orbit-components/src/Inline/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    `;
+  }
+
+  return "";
+};
+
+type PropObject = { [key: Prop]: Position };
+
+// TODO: kinda weird, but it's well known problem in flow with Object.values
```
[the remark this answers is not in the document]
mainframev: nope, it's not related to `any` below, i used it blow, because i did not find any other option how to solve it. That's basically issue in Flow, TS does not have such problems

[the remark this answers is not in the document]
mainframev: Screenshot 2020-10-30 at 00 11 50

the problem is with `mixed` value in `Flow`. I tried to type it, with generics and etc, nothing helped, you can try to play with it and maybe you'll find solution, there are a lot of similar issues when i searched it

packages/orbit-components/src/Inline/index.js
```diff
@@ -0,0 +1,56 @@
...
+import defaultTheme from "../defaultTheme";
+import { DEVICES as DEVICES_CONSTS } from "../utils/mediaQuery/consts";
+import media from "../utils/mediaQuery";
+import { normalize } from "./helpers";
+
+import type { Props } from "./index";
+
+const StyledInline = styled(({ as: Element, children }) => (
+  <Element>{children}</Element>
+)).attrs(({ dataTest }) => ({ "data-test": dataTest }))`
```
vepor, line 14: Would define the `data-test` on the return inside `styled` factory. It seems redundant to call another function.

packages/orbit-components/src/Inline/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+export const normalize = (object: PropObject) => ({ theme }: ThemeProps) => {
+  if (!object) return null;
+
+  return Object.entries(object).reduce((acc, [key, val]: [string, any]) => {
+    if (key === "justify") {
+      return [...acc, formatCSS("justify-content", getJustify(val))];
+    }
+
+    if (key === "align") {
+      return [...acc, formatCSS("align-items", val)];
```
vepor, line 59: When used `start` and `end` it won't be correct value. It needs to be prefixed with `flex-`.

packages/orbit-components/src/Inline/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+  if (justify === "start") return "flex-start";
+  if (justify === "end") return "flex-end";
+
+  return "center";
+};
+
+export const normalizeSpacing = (el: SpacingToken, theme: Theme): string => {
+  const tokens = {
+    none: "",
+    XXXsmall: theme.orbit.spaceXXXSmall,
```
vepor, line 21: All the spacing names should be in `camelCase`. Because we use also `T-Shirt` sizes it should look like `XXXSmall`. Check #2457.

vepor (changes requested):
Also we need to add `README.md` for this component.

packages/orbit-components/src/Inline/consts.js
```diff
@@ -0,0 +1,18 @@
...
+  "none",
+  "XXXsmall",
+  "XXsmall",
+  "Xsmall",
+  "small",
+  "medium",
+  "large",
+  "Xlarge",
+  "XXlarge",
+  "XXXlarge",
```
silvenon, line 17:
```suggestion
  "XXXSmall",
  "XXSmall",
  "XSmall",
  "small",
  "medium",
  "large",
  "XLarge",
  "XXLarge",
  "XXXLarge",
```

packages/orbit-components/src/Inline/__tests__/index.test.js
```diff
@@ -0,0 +1,89 @@
...
+  it("should have props", () => {
+    render(
+      <Inline align="center" justify="start" dataTest={dataTest}>
+        <Elements />
+      </Inline>,
+    );
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+    const styles = screen.getByTestId(dataTest);
+    expect(styles).toHaveStyle({ alignItems: "center" });
+    expect(styles).toHaveStyle({ justifyContent: "flex-start" });
```
silvenon, line 42: The value of `styles` is a DOM element, so that name is a bit confusing. I recommend something like `container`:
```suggestion
    const container = screen.getByTestId(dataTest);
    expect(container).toHaveStyle({ alignItems: "center" });
    expect(container).toHaveStyle({ justifyContent: "flex-start" });
```

packages/orbit-components/src/Inline/__tests__/index.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it("should have spacings", () => {
+    const spacings = Object.keys(tokens).map(token => (
+      <Inline key={token} align="start" justify="center" spacing={token} dataTest={dataTest}>
+        <div data-test={token} style={{ marginLeft: tokens[token] }}>
+          kek
+        </div>
+        <div>bur</div>
+      </Inline>
+    ));
```
silvenon, line 53: I testing this with [`it.each`](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout) instead, for example:

```js
it.each(Object.entries(tokens))('should have expected spacing for token "%s"', (token, spacing) => {
  render(
    <Inline align="start" justify="center" spacing={token} dataTest={dataTest}>
      <div data-test={token} style={{ marginLeft: tokens[token] }}>
        kek
        </div>
      <div>bur</div>
    </Inline>
  );
  expect(screen.getByTestId(dataTest)).toHaveStyle({ marginLeft: spacing });
});
```

😎 (I didn't check whether this works.)

silvenon (changes requested):

packages/orbit-components/src/Inline/__tests__/index.test.js
mainframev: works, thx :)

packages/orbit-components/src/Inline/__tests__/index.test.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(container).toHaveStyle({ alignItems: "center" });
+    expect(container).toHaveStyle({ justifyContent: "flex-start" });
+  });
+
+  it.each(Object.entries(tokens))(
+    'should have expected spacing for token "%s"',
+    (token, spacing) => {
+      render(
+        <Inline align="start" justify="center" spacing={token} dataTest={dataTest}>
+          <div data-test={token} style={{ marginLeft: tokens[token] }}>
```
silvenon, line 50: In my previous review I missed two things here:

1. Why is left margin being set here? I assume the intention is to test the computed left margin given by `Inline`, no?
2. You can remove the `dataTest` prop from `Inline` considering that you're not using it.

silvenon (approved):

vepor (approved):

PR #2263, opened 2020-09-28
chore: update storybook to 6v

mainframev:
- Updated `storybook` to 6 version<br/><br/><br/><url>LiveURL: https://orbit-mainframe-storybook.surge.sh</url>


packages/orbit-components/.storybook/orbitDecorator.js
```diff
@@ -1,29 +1,29 @@
...
+      <div style={{ marginTop: 20 }}>
         <React.Fragment>
           <Code code={jsxToString(children, options)} language="jsx" format={false} />
         </React.Fragment>
       </div>
     </div>
-  )
-}
+  );
+};
```
silvenon, line 27: I was wondering how this got past ESLint. It's because ESLint by default ignores all paths starting with `.`, which includes `.storybook`, so please also add `!.storybook` to `.eslintignore`.

packages/orbit-components/package.json
```diff
@@ -91,11 +91,11 @@
     "@commitlint/config-conventional": "^9.1.1",
     "@emotion/core": "^10.0.28",
     "@octokit/rest": "^15.13.1",
-    "@storybook/addon-actions": "^5.3.19",
-    "@storybook/addon-info": "5.3.19",
-    "@storybook/addon-knobs": "^5.3.19",
-    "@storybook/addons": "5.3.19",
-    "@storybook/react": "^5.3.19",
+    "@storybook/addon-actions": "^6.0.21",
+    "@storybook/addon-info": "^6.0.21",
```
silvenon, line 95: The latest version of `@storybook/addon-info` is `6.0.0-alpha.2`.

silvenon (changes requested):

packages/orbit-components/.storybook/orbitDecorator.js
mainframev: thanks, did not notice that

packages/orbit-components/.storybook/manager.js
```diff
@@ -1,10 +1,11 @@
-import { create } from "@storybook/theming"
+// @flow
+/* eslint-disable import/no-extraneous-dependencies */
```
silvenon, line 2: Instead of adding disable directives in `.storybook` you can include `**/.storybook` in [this list](https://github.com/kiwicom/orbit/blob/7bd4ab3cc74b7c42a6d3817d6712865709a7e1bd/config/eslintJS.js#L22-L30). Dealing with ESLint can sometimes be so tedious… 😓

silvenon (dismissed):
Other than the suggestion below, this update appears to work, thanks! ❤️

packages/orbit-components/.storybook/manager.js
mainframev: yes, i know, it was first thing which came to my mind, but later i noticed that there are already eslint disables in these files, so i picked the easy way out :D

silvenon (approved):
Looks good, feel free to merge if it's ready.

PR #2265, opened 2020-09-28
feat: add Wizard component

silvenon:
<br/><br/><br/><url>LiveURL: https://orbit-feat-wizard.surge.sh</url>


packages/orbit-components/src/Wizard/WizardStep/index.js
```diff
@@ -0,0 +1,246 @@
...
+          ${active &&
+          css`
+            box-shadow: 0 0 0 4px ${activeGlowColor};
+          `}
+        }
+        svg {
+          position: relative;
+          display: block;
+        }
+      `}
```
mainframev, line 77: Just curious. Is there any reason why you don't use SC and pass styles with `css` template literals via css?

packages/orbit-components/src/Wizard/WizardStep/index.js
```diff
@@ -0,0 +1,246 @@
+// @flow
+import * as React from "react";
+import { css } from "styled-components";
+import { hex2rgb } from "@csstools/convert-colors";
```
mainframev, line 4: i suppose we can write that function, no need to add another dependency, which was updated last time 2 years ago :)

packages/orbit-components/src/Wizard/WizardStep/index.js
```diff
@@ -0,0 +1,246 @@
...
+    .map(n => 255 * (n / 100))
+    .concat(0.2)
+    .join(", ")})`;
+
+  /**
+   * We're explicitly rounding values in layout styles because browsers seem to
+   * be rounding them inconsistently horizontally vs vertically.
+   */
+
+  const icon = (
```
mainframev, line 44: Consider to make it as separate component inside WizardStep :) 

I think it's better to split it here, less lines of code, better reading and later code maintain etc.

packages/orbit-components/src/Wizard/WizardStep/index.js
silvenon: I was wondering what to do as well, but reinventing the function seemed unnecessary; `@csstools/convert-colors` is tree-shakable, so I don't think there will be any bundle size repercussions, and as far as the time they were last updated, I assume it's because these calculations are pretty straightforward, and possibly not widely used, but I've been collaborating with Jonathan Neal a bit and have been following him for a long time, so I trust his code.

packages/orbit-components/src/Wizard/WizardStep/index.js
silvenon: I'm aware that so far we probably haven't used `css` anywhere, but there are several reasons why I like using it:

- avoiding having to come up with a lot of nonsense component names 😅
- I thinking creating a component for every set of styles is disrupting to the reading flow of the code, I try to keep it "top to bottom" as much as possible
- all that this component does is contain styles, so I see no need to put the styles outside of it because there is no logic to flesh out

I think this is exactly what `css` was designed for, and `styled` might be more suitable for reusable sets of styles, which can be extended into styled new components as needed, etc.

packages/orbit-components/src/Wizard/WizardStep/index.js
silvenon: Good idea! This is a bit awkward.

packages/orbit-components/src/Wizard/WizardStep/index.js
vepor: We've got `convertHexToRgba` in `orbit-design-tokens`, so feel free to use it 🙂 
https://github.com/kiwicom/orbit-design-tokens/blob/master/src/convertHexToRgba.js

packages/orbit-components/src/Wizard/index.js
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <nav>
+      <ul
+        css={css`
+          list-style-type: none;
+          padding: 0;
+          margin: 0;
+          display: flex;
+          > * {
+            flex: 1 1 0%;
```
mainframev, line 143: why? you know, that all childs are `li`, select them by type, not with `>` selector

packages/orbit-components/src/Wizard/index.js
```diff
@@ -0,0 +1,154 @@
...
+  const stepStatuses = childrenArray.map((step, index) => {
+    let status;
+    if (index < completedSteps) {
+      status = "completed";
+    } else if (index === completedSteps) {
+      status = "available";
+    } else {
+      status = "disabled";
+    }
+    return status;
```
mainframev, line 35: probably this will look better:

```
if (index < completedSteps) return "completed"
if (index === completedSteps) return "available";
return "disabled";

File: packages/orbit-components/src/Wizard/index.js (hunk 4 above)
Comment (silvenon) [reply packages/orbit-components/src/Wizard/index.js:143]: I'm fine with either, there is a slight difference in specificity, but it's not important for this use case.

Comment (mainframev) [reply packages/orbit-components/src/Wizard/index.js:143]: sure, it will not have any significant value on perfomance, just good tone to use more performant css selectors

File: packages/orbit-components/src/Wizard/index.js (hunk 5 above)
Comment (silvenon) [reply packages/orbit-components/src/Wizard/index.js:35]: Yep 👍

File: packages/orbit-components/src/Wizard/Wizard.stories.js [hunk 6]
```diff
@@ -0,0 +1,46 @@
+// @flow
+
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { number, array } from "@storybook/addon-knobs";
+import { action } from "@storybook/addon-actions";
+
+import Wizard, { WizardStep } from ".";
+
+storiesOf("Wizard", module).add(
```
vepor, line 10: @silvenon could you also add a story with RTL version? 🙏

vepor (changes requested):
Generally I like the code you've implemented! 💪 

But I would be careful and definitely think twice about the code style for the styles – usage of `css`.

At the first look, I didn't know on what exact component I'm looking on – I mean I need to compare every used component/element with the rendered DOM structure to see what is what. There's a lot of LOC that I need to scroll or see with my eyes – I can't focus only on JSX (functional perspective of the component) or styles – it's together, mixed. It's simply harder to read for me.

As far we were using `styled-components` for past two and half years I think that we had agreed on unwritten coding style conventions that are very good. Even though that in some cases could be improved.

With choosing `css` over `styled.div` I think that we are loosing the main benefits that `styled-components` offer and that is, from my perspective:

Simple and quick reusability of defined styles (components) as we can just jump to specific file, make it more abstract and reuse in different components/files. With `css` I just can't see it the same way. If we decide to do it, it will be more work to do so. For no benefit at the first place. 

IMHO, it lacks the `styles are coupled with component` properly. Its used on element that doesn't have any name, e.g. `StyledDiv`. It also does have a lot of selectors `svg { }` that are a bit confusing for me at the first look.

I would consider to actual use the same way we do in every single other component. I currently don't see any benefits with this approach, even though that can be some cases in the future, but only for some small cases 🙂 

Also one question came to my mind and that is:
Will be `css` usages reevaluated every single time, with every performed re-render? In the past, the general guideline was to **not** declare any styled components in the scope of component. Example:
```jsx
const Example = () => {
  const StyledDiv = styled.div`
    display: block;
  `
  return <StyledDiv />;
}

**Comment by silvenon** [comment]
Note to self: this component also suffers from a usability problem in modal view. Our `Modal` component currently isn't flexible enough to accommodate removing close button. I'll need to either extend `Modal` to support this or add some dirty fix.

File: packages/orbit-components/src/Wizard/Wizard.stories.js (hunk 6 above)
Comment (silvenon) [reply packages/orbit-components/src/Wizard/Wizard.stories.js]: Added a RTL story and fixed RTL issues, _however_ the text in the toggle button in mobile doesn't look right to me:

Screen Shot 2020-10-15 at 12 45 48

How come it says "of 5 4"…?

Comment (vepor) [reply packages/orbit-components/src/Wizard/Wizard.stories.js]: This is okay, because `direction: rtl` can't work with normal latin text 🙂 
The browser is solving this automatically, we just need to worry about CSS, nothing more 😉

File: packages/orbit-components/src/Wizard/index.js [hunk 7]
```diff
@@ -0,0 +1,161 @@
...
+          </ButtonLink>
+        </div>
+        <Portal>
+          <div
+            id={id}
+            css={css`
+              ${ModalCloseContainer} {
+                display: none;
+              }
+            `}
```
silvenon, line 97: I removed Modal's close button because it's not in the design, and the `Modal` component doesn't support not rendering the close button yet.

packages/orbit-components/src/primitives/ButtonPrimitive/components/ButtonPrimitiveContentChildren.js
```diff
@@ -2,13 +2,14 @@
...
 
+import { left } from "../../../utils/rtl";
 import defaultTheme from "../../../defaultTheme";
 import type { Props } from "./ButtonPrimitiveContentChildren";
 
 const StyledButtonPrimitiveContentChildren = styled.div`
   display: inline-block;
   width: ${({ contentWidth }) => contentWidth};
-  text-align: ${({ hasIcon }) => hasIcon && "left"};
+  text-align: ${({ hasIcon }) => hasIcon && left};
```
silvenon, line 12: I'm pretty sure this is the right thing to do. The text inside button components with `fullWidth` is currently always aligned to the left, but I assume it should be aligned to the right? This will have an impact on all button components, so it's an important decision, but the Wizard component didn't look right without it.

Before:

Screen Shot 2020-10-15 at 12 53 18

After:

Screen Shot 2020-10-15 at 12 52 27

silvenon (commented):
Ok, I think everything is done now! Including the documentation.

silvenon:
Added TypeScript definitions. I hope that's all I forgot. 😞

silvenon:
Added `Wizard` and `WizardStep` to `src/index.js.flow` and `src/index.d.ts`.

packages/orbit-components/src/Wizard/index.js
```diff
@@ -0,0 +1,155 @@
...
+                {activeStepTitle}
+              </span>
+            </Stack>
+          </ButtonLink>
+        </div>
+        <Portal>
+          <div id={id}>
+            {open && (
+              <Modal
+                hasCloseButton={false}
```
silvenon, line 94: Ok, I solved the close button problem by adding a new prop to `Modal` called `hasCloseButton`. I tried `display: none` first, but the tests were failing because they were still able to find the button, which is weird because Testing Library seems to detect when a wrapper of an element has `display: none` and hides those, like I mentioned on Orbit Show 🤷‍♂️ but whatever, this is cleaner anyway.

packages/orbit-components/src/Wizard/Wizard.stories.js
```diff
@@ -0,0 +1,73 @@
...
+import { action } from "@storybook/addon-actions";
+
+import RenderInRtl from "../utils/rtl/RenderInRtl";
+
+import Wizard, { WizardStep } from ".";
+
+storiesOf("Wizard", module)
+  .add("Default", () => {
+    return (
+      <Wizard id="wizard" completedSteps={3} activeStep={3} onChangeStep={() => {}}>
```
mainframev, line 15: add `action("onChangeStep")` to onChangeStep prop :)

packages/orbit-components/src/Wizard/index.js.flow
```diff
@@ -0,0 +1,19 @@
+// @flow
+import * as React from "react";
+
+import type { Globals } from "../common/common.js.flow";
+import typeof WizardStepComponent from "./WizardStep";
+
+export type Status = "available" | "completed" | "disabled";
```
mainframev, line 7: Consider to move to Context, because you use it only there and in icon component :)

packages/orbit-components/src/Wizard/index.js
```diff
@@ -0,0 +1,155 @@
...
+                onClose={() => {
+                  setOpen(false);
+                }}
+              >
+                <nav
+                  css={css`
+                    /* matching this to ModalBody's border-radius */
+                    padding-top: 9px;
```
mainframev, line 102: It should be exactly nine pixels? because if it can be `8px` you can use a special token for it, or to calculate it with token and + 1px

packages/orbit-components/src/Wizard/__tests__/compact.test.js
```diff
@@ -0,0 +1,104 @@
...
+import React from "react";
+import { render, screen, within } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import Wizard, { WizardStep } from "..";
+
+jest.mock("../../hooks/useMediaQuery", () => () => ({ isTablet: false }));
+
+describe("Wizard", () => {
+  describe("compact", () => {
```
mainframev, line 11: Seems like redundant describe within describe, you can just write `describe("#Wizard - compact")`, it might be useful if you have several tests inside.

packages/orbit-components/src/Wizard/WizardStepIcon.js
```diff
@@ -0,0 +1,130 @@
...
+  ${({ status }) =>
+    status === "disabled" &&
+    css`
+      background: ${({ theme }) => theme.orbit.paletteCloudNormalHover};
+    `}
+  ${({ active, hasGlow }) =>
+    active &&
+    hasGlow &&
+    css`
+      box-shadow: 0 0 0 4px ${({ theme }) => convertHexToRgba(theme.orbit.paletteProductNormal, 20)};
```
mainframev, line 41: You can destructure it everything on the top of the component :) it will be easier to read and add

packages/orbit-components/src/Wizard/WizardStep.js
```diff
@@ -0,0 +1,197 @@
...
+};
+
+const StyledLink = styled.a`
+  display: block;
+  text-decoration: none;
+  ${({ theme, active }) =>
+    active
+      ? css`
+          span {
+            font-weight: 500;
```
mainframev, line 88: Use token here :) `fontWeightMedium`

packages/orbit-components/src/primitives/ButtonPrimitive/components/ButtonPrimitiveContentChildren.js
mainframev: Not sure :) should icon be visible in that case? 

Screenshot 2020-10-15 at 20 46 52

mainframev (approved):

PR #2268, opened 2020-09-29
docs: fix example generation

CollierCZ:
Move primitives to separate folders based on component name
to avoid conflicts.<br/><br/><br/><url>LiveURL: https://orbit-fix-example-generation.surge.sh</url>


packages/orbit-components/config/buildExample.js
```diff
@@ -160,11 +160,16 @@ async function readFileAndCreateJSON(pathToFile) {
   const ast = parse(code, {
     sourceType: "module",
   });
+  const firstLetter = nameSplit[1].charAt(0);
+  const componentName =
+    firstLetter === firstLetter.toLowerCase() && firstLetter !== firstLetter.toUpperCase()
```
mainframev, line 165: @CollierCZ here `firstLetter === firstLetter.toLowerCase()` is enough for check :)

CollierCZ: I thought so too, but someone recommended both. I'll put it the simpler way. 🙂

mainframev (approved):

PR #2269, opened 2020-09-29
test: improve failure message for the Babel plugin

silvenon:
When I got the error in the former screenshot (it was for another module), it took me a while to figure out what it meant even though I was the one who reviewed the PR that added this test, so I wanted to improve the error message.

Before:

Screen Shot 2020-09-29 at 12 47 11

After:

Screen Shot 2020-09-29 at 12 40 22
<br/><br/><br/><url>LiveURL: https://orbit-test-babel-plugin-dx.surge.sh</url>


tbergq (approved):
Thanks, this will improve DX

PR #2270, opened 2020-09-29
fix: force eol to lf

mainframev:
// every contributor now pushes git changes from unix OS as far as i know,
// but just in case if someone would like to contribute from windows,
// it will be useful to enforce eol to lf
<br/><br/><br/><url>LiveURL: https://orbit-fix-line-endings.surge.sh</url>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,10 @@
+* text=auto eol=lf
+*.gif binary
+*.jpg binary
+*.eot binary
+*.woff binary
+*.svg binary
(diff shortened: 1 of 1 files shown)


silvenon (approved):

silvenon:
How come your commit message body is formatted as JS comments?

mainframev:
@silvenon 
when you write a short message in git after normal commit message - it will be added to description automatically, i just get used to use `//` there, weird, i know

PR #2277, opened 2020-09-30
fix(Hide): span to div

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-fix-hide.surge.sh</url>


silvenon (approved):
Let's hope that this doesn't affect existing UI 🤞

PR #2280, opened 2020-10-01
feat(Accordion): added Accordion component

CroTweet:
Created Accordion component as agreed in this [doc](https://docs.google.com/document/d/1QF1MV3wS1pgaGYMXUGqjsLznh8435BUSpAmH3JWb0-0/edit#).

Closing ticket: [ `AN-1079`](https://jira.kiwi.com/browse/AN-1079)


packages/orbit-components/src/Accordion/AccordionSection/components/SectionContent.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import * as React from "react";
+import styled from "styled-components";
+
+import defaultTheme from "../../../defaultTheme";
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  padding: ${({ theme }) => `0 ${theme.orbit.spaceLarge} ${theme.orbit.spaceLarge}`};
```
mainframev, line 10: You can use `Stack` for same functionality

packages/orbit-components/src/Accordion/AccordionSection/components/SectionContent.js
```diff
@@ -0,0 +1,31 @@
...
+
+Wrapper.defaultProps = {
+  theme: defaultTheme,
+};
+
+type Props = {
+  children: React.Node,
+  expanded: boolean,
+  hasFooter: boolean,
+};
```
mainframev, line 21: `type Props {| ... |}` :)

packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.js
```diff
@@ -0,0 +1,67 @@
...
+import styled from "styled-components";
+
+import Button from "../../../Button";
+import defaultTheme from "../../../defaultTheme";
+
+const Wrapper = styled.div`
+  display: flex;
+  padding: ${({ theme, noPadding }) =>
+    !noPadding && `${theme.orbit.spaceXLarge} ${theme.orbit.spaceLarge}`};
+  background-color: white;
```
mainframev, line 12: tokens should be used for colors: `background-color: ${({theme}) => theme.orbit.paletteWhite}`

packages/orbit-components/src/Accordion/AccordionSection/index.js
```diff
@@ -0,0 +1,63 @@
...
+import AccordionWrapper from "../components/AccordionWrapper";
+import SectionHeader from "./components/SectionHeader";
+import SectionFooter from "./components/SectionFooter";
+import SectionContent from "./components/SectionContent";
+
+import type { Props } from "./index";
+
+const ExpandableContent = styled.div`
+  dispaly: flex;
+  flex-direction: column;
```
mainframev, line 18: `Stack` also here, you can turn off spacing, if you dont need it, but no need to create additional component here

packages/orbit-components/src/Accordion/index.js
```diff
@@ -0,0 +1,91 @@
...
+            // This is used for the case when user wants to map sections and change their order
+            // related issue: #1005
+            const index = Number(item.key) || key;
+
+            // Either use provided id or item index
+            const sectionId = (typeof id !== "undefined" && id) || index;
+            // Determine if section is expanded
+            const isExpanded = expandedSection === sectionId;
+            const handleDefaultExpand = () => {
+              if (onExpand) {
```
mainframev, line 64: it's  better to have it inline `if (onExpand) onExpand()` :)

packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.js
```diff
@@ -0,0 +1,67 @@
...
+const HeaderContent = styled.div`
+  display: flex;
+  flex-grow: 1;
+  align-items: center;
+`;
+
+const HeaderActions = styled.div`
+  display: flex;
+  margin-left: ${({ theme }) => theme.orbit.spaceLarge};
+`;
```
mainframev, line 25: Stack also can be applied here, in both cases

packages/orbit-components/src/Accordion/index.js
```diff
@@ -0,0 +1,91 @@
...
+            const { id, onExpand } = item.props;
+
+            // This is used for the case when user wants to map sections and change their order
+            // related issue: #1005
+            const index = Number(item.key) || key;
+
+            // Either use provided id or item index
+            const sectionId = (typeof id !== "undefined" && id) || index;
+            // Determine if section is expanded
+            const isExpanded = expandedSection === sectionId;
```
mainframev, line 62: Probably here is a mistake:

You are comparing `expandedSection` which is `number` to sectionId which can be a `string`
1 === "1" will be false

mainframev:
I will take a look to this PR later, i just noticed few things here, but did not make proper CR for it. Also it took me a while to understand, that this is from your fork, could not find branch in git :D

silvenon:
> Also it took me a while to understand, that this is from your fork, could not find branch in git :D

`gh pr checkout 2280` 😎

packages/orbit-components/src/Accordion/index.js
CroTweet: Oh, okay 😄 I thought this would be more readable, but I'll change it 👍

packages/orbit-components/src/Accordion/AccordionSection/index.js
CroTweet: I've tried using Stack here but looks like Stack is blocking `expanded` prop required for Slide component to work. I can use plain div in this case, no styles needed here.

packages/orbit-components/src/Accordion/index.js
CroTweet: The idea is to have ability to pass string ids (e.g. flightId, segmentCode, ...). That's why we have `id` prop on AccordionSection component.

packages/orbit-components/src/Accordion/AccordionSection/index.js
```diff
@@ -0,0 +1,53 @@
...
+          actions={actions}
+          expanded={Boolean(expanded)}
+          onExpand={onExpand}
+          expandable={expandable}
+        >
+          {header}
+        </SectionHeader>
+      )}
+
+      <Slide maxHeight={height} expanded={expanded} id={slideID} ariaLabelledBy={labelID}>
```
silvenon, line 43: The value of [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) should be same as the `id` attribute of an element that contains its label. Or something like that, I'm not familiar with this part of a11y, but the internet is.

packages/orbit-components/src/Accordion/AccordionSection/index.js
```diff
@@ -0,0 +1,53 @@
...
+        <div expanded={expanded} ref={ref}>
```
silvenon, line 44: I assume passing `expanded` to this `div` is accidental?

packages/orbit-components/src/Accordion/AccordionSection/index.js
```diff
@@ -0,0 +1,53 @@
...
+  const labelID = React.useMemo(() => randomID("labelID"), []);
+
+  const [{ height }, ref] = useBoundingRect({ height: expanded ? null : 0 });
+
+  return (
+    <AccordionWrapper dataTest={dataTest}>
+      {header && (
+        <SectionHeader
+          actions={actions}
+          expanded={Boolean(expanded)}
```
silvenon, line 35: Is typecasting here necessary considering that you're enforcing the type with a static type, and that `expanded` is required?
```suggestion
          expanded={expanded}
```

silvenon (changes requested):
The transition doesn't look smooth; when the button disappears the content height suddenly shrinks to whatever the height of the text is, that's why you see a visual "jump". It seems like actions need to be positioned absolutely or something, they shouldn't be affecting the content flow when they're removed or added. 🤔

Also, some boring stuff that for now needs to be done manually, until we figure out how to automate/enforce it:

- export `AccordionSection` from `packages/orbit-components/src/Accordion/index.js`
- export `Accordion` and `AccordionSection` components from `packages/orbit-components/src/index.js` (also update `index.js.flow` and `index.d.ts`)
- add `AccordionSection` to `packages/babel-plugin-orbit-components/index.js` (`AccordionSection: "Accordion/AccordionSection"`)

packages/orbit-components/src/Accordion/Accordion.stories.js
```diff
@@ -0,0 +1,156 @@
...
+          This is a content <br />
+          <br />
+          This is a content <br />
+          This is a content <br />
+          This is a content <br />
+          This is a content <br />
+          <br />
+          This is a content <br />
+          This is a content <br />
+          This is a content <br />
```
silvenon, line 128: Concerning stories, I recommend replacing placeholders with some real-world examples, then it's immediately clear what this component is good for. You can use the examples in your Figma design as inspiration.

packages/orbit-components/src/Accordion/AccordionSection/index.js
CroTweet: `<Slide />` requires child component with `expanded` prop. And I can't use `<Stack />` because it will not take `expanded` as prop. I can't even use the `React.Fragment`.

silvenon:
Maybe the smooth transition problem can be solved by simply setting `min-width` to the content before expanding the Accordion (and before the buttons are removed)? 🤷 Just brainstorming, I'm sure you'll come up with something.

packages/orbit-components/src/Accordion/README.md
```diff
@@ -0,0 +1,66 @@
+# Accordion
+
+To implement Accordion component into your project you'll need to the import at least the Accordion and the [AccordionSection](#Accordionsection):
```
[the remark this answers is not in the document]
CroTweet: Yes, you need at least `Accordion` and `AccordionSection` in order for it to work as designed.
Looks like I've forgot to add:

`export { default as AccordionSection } from "./AccordionSection";`

inside Accordion/index.js

Fixed with last commit.

packages/orbit-components/src/Accordion/AccordionSection/index.js
silvenon: I suggested removing the typecasting because (the way I see it) the value of `expanded` is already boolean, do you disagree then?

packages/orbit-components/src/Accordion/AccordionSection/components/SectionContent.js
silvenon: What's the status of this?

silvenon (changes requested):
The transition still doesn't look smooth because the one-line text jumps from these two heights (because the button is higher than one line of text). See the height of the content box (the blue box).

Screen Shot 2020-10-12 at 18 09 18

Screen Shot 2020-10-12 at 18 09 04

silvenon (commented):
Also, we will definitely need tests, so that this doesn't get broken in the future. Have you ever done this before? And do you know Testing Library?

packages/orbit-components/src/Accordion/README.md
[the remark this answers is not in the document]
CollierCZ: I agree with you, @silvenon.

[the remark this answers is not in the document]
CroTweet: @silvenon What I wanted to say is that you can use "naked"  `Accordion` along with `AccordionSection` without controlled state and other props. Maybe I should phrase it better 🤔

packages/orbit-components/src/Accordion/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+const Accordion = ({ children, dataTest, spaceAfter, expanded, loading }: Props) => {
+  // Make first section expanded by default (index 0)
+  const [expandedSection, setExpandedSection] = React.useState(0);
+
+  React.useEffect(() => {
+    if (typeof expanded !== "undefined") {
+      setExpandedSection(expanded);
+    }
+  }, [expanded]);
```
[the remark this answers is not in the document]
CroTweet: The idea is to provide user with option to use Accordion in a controlled and uncontrolled way.
With this we can both manage state by default and enable users to control it if they have need for it.

packages/orbit-components/src/Accordion/AccordionSection/components/SectionContent.js
CroTweet: @silvenon @mainframev  I've removed `display-flex` and `flex-direction` but I'm not sure how to add padding on Stack component 🤔

mainframev: you had padding on right side and bottom, Stack will add only on bottom automatically, when they align as column. Probably you don't need that right padding inside SectionContent, consider to add it inside `Accordion` 🤔

packages/orbit-components/src/Accordion/index.js
CroTweet: We can use provided `AccordionSection` IDs to store in our state as expanded segment IDs. There shouldn't be any problem with that, right?

packages/orbit-components/src/Accordion/AccordionSection/components/SectionContent.js
CroTweet: I can use `Stack` for bottom spacing 👍, but I need to keep left-right padding inside this component because of sticky footer component which needs 100% width.

packages/orbit-components/src/Accordion/Accordion.stories.js
silvenon: We'll solve this some other time.

silvenon (dismissed):
Looks good!

packages/orbit-components/src/Accordion/Accordion.stories.js
```diff
@@ -21,13 +22,37 @@ storiesOf("Accordion", module)
...
+        <AccordionSection
+          id="0X0"
+          header={
+            <Stack spacing="compact">
+              <Text type="primdary">This is a header label</Text>
+              <Text size="small">This is a header label</Text>
+            </Stack>
+          }
+        >
           This is a section content
```
silvenon, line 34: Wrap these inside `Text`.

silvenon (changes requested):

silvenon (approved):

PR #2287, opened 2020-10-02
feat(Timeline): init component

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-feat-timeline.surge.sh</url>


packages/orbit-components/src/Timeline/TimelineContext.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const TimelineStatusProvider = ({ children }: { children: React.Node }) => {
+  const [state, dispatch] = React.useReducer(reducer, initialState);
+
+  return (
+    <TimelineStatusContext.Provider value={{ state, dispatch }}>
+      {children}
+    </TimelineStatusContext.Provider>
+  );
+};
```
silvenon, line 40: `useState` seems to be enough here, this way you can offload most of the logic to the `TimelineStep` component so it's less split:

```suggestion
export const TimelineStatusProvider = ({ children }: { children: React.Node }) => {
  const [statuses, setStatuses] = React.useState([]);

  return (
    <TimelineStatusContext.Provider value={{ statuses, setStatuses }}>
      {children}
    </TimelineStatusContext.Provider>
  );
};
```

packages/orbit-components/src/Timeline/components/TimelineStep/index.js
```diff
@@ -0,0 +1,158 @@
...
+
+const RelativeStyled = styled.div`
+  position: relative;
+`;
+
+const TimelineStep = ({ children, step, time, status }: Props) => {
+  const { state, dispatch } = useStatuses();
+  const { index, last } = useStep();
+
+  const getStatus = state.statuses[index + 1];
```
silvenon, line 96:
```suggestion
  const nextStatus = state.statuses[index + 1];
```

packages/orbit-components/src/Timeline/components/TimelineStep/index.js
```diff
@@ -0,0 +1,158 @@
...
+                <Text type={status ? "primary" : "secondary"}>{children}</Text>
+              </IndentStyled>
+            </Stack>
+          </Stack>
+        </RelativeStyled>
+      </Hide>
+      <Hide on={["smallMobile", "mediumMobile", "largeMobile", "tablet"]}>
+        <RelativeStyled>
+          <Stack inline direction="column" align="center" spacing="compact" spaceAfter="large">
+            <TimeWrapperStyled visible={!!status}>
```
silvenon, line 128:
```suggestion
            <TimeWrapperStyled visible={Boolean(status)}>
```

packages/orbit-components/src/Timeline/components/TimelineStep/index.js
```diff
@@ -0,0 +1,158 @@
...
+              <Text size="small">{time}</Text>
+            </TimeWrapperStyled>
+            <ProgressLineStyled data-test="progressLine" desktop isLeft status={status} />
+            <IconWrapperStyled>
+              <StatusIcon status={status} />
+            </IconWrapperStyled>
+            <ProgressLineStyled
+              data-testid="progressLine"
```
silvenon, line 136: We configured Testing Library to look for `data-test` instead of the default `data-testid` because that's the convention we were already using in Orbit.
```suggestion
              data-test="progressLine"
```

packages/orbit-components/src/Timeline/index.js
```diff
@@ -0,0 +1,48 @@
...
+import useMediaQuery from "../hooks/useMediaQuery";
+import { TimelineStatusProvider, TimelineStepContext } from "./TimelineContext";
+import getSpacingToken from "../common/getSpacingToken";
+import themeDefault from "../defaultTheme";
+
+import type { Props } from "./index";
+
+const WrapperStyled = styled.div`
+  position: relative;
+  overflow: hidden;
```
silvenon, line 15: This component is cutting text off on small resolutions.

silvenon (changes requested):

packages/orbit-components/src/Timeline/TimelineContext.js
mainframev: good point, thx, my approach was a bit over-engineered :D

packages/orbit-components/src/Timeline/components/TimelineStep/index.js
mainframev: in the end it's the same, but thanks :)

packages/orbit-components/src/Timeline/Timeline.stories.js
```diff
@@ -0,0 +1,110 @@
...
+        </Timeline>
+      );
+    },
+    { info: "This is example of Timeline with warning status" },
+  )
+  .add(
+    "with critical",
+    () => {
+      const step = text("step", "Non refundable");
+      const status = text("status", "warning");
```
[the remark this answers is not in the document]
mainframev: did not notice that, thx :)

packages/orbit-components/src/Timeline/components/TimelineStep/index.js
mainframev: You could notice, that I had `data-test` set already, it's just one forgotten. Thanks, that you notice it :)

packages/orbit-components/src/Timeline/TimelineContext.js
```diff
@@ -0,0 +1,43 @@
...
+export const TimelineStepContext: React.Context<StepContext> = React.createContext({
+  index: 0,
+  last: false,
+});
+
+const reducer = (state, action) => {
+  switch (action.type) {
+    case "ADD_STATUS":
+      return {
+        statuses: [...state.statuses, action.payload],
```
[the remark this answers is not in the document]
mainframev: good point, but i'll just use object instead then :)

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+  return theme.orbit.paletteCloudNormalHover;
+};
+
+const ProgressLineStyled = styled.span`
+  ${({ desktop, theme, status, isLeft }) => css`
+    ${!desktop && `position: absolute`};
+    width: ${desktop ? "50%" : "2px"};
+    ${isLeft ? `left: 0` : `right: -20px`};
+    z-index: -1;
```
vepor, line 54: I would not prefer to use negative z-index here. It will cause a trouble on IE 10/11 and because of it, you're forced to use `useModalContext`, especially `isInsideModal` boolean. In the end it will add unnecessary complexity that will make future refactoring of modal related components harder and we will have to consider this also in z-index framework.

What I suggest is to get rid off negative z-index, the usage of `isInsideModal` and rather to use `relative` position on the `icon` or its wrapper. We can also tweak the height, since we're using `calc` already, so it will be fine.

There will be probably a small problem with the usage of `Circle` icon, that has different inner size, but it can be tweak with other interpolations – render functions.

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+import CheckCircle from "../../icons/CheckCircle";
+import Circle from "../../icons/CircleEmpty";
+import AlertCircle from "../../icons/AlertCircle";
+import CloseCircle from "../../icons/CloseCircle";
+import { useStatuses, useStep } from "../TimelineContext";
+import useMediaQuery from "../../hooks/useMediaQuery";
+
+import type { Props } from "./index";
+
+const IconWrapperStyled = styled.div`
```
vepor, line 18: As far most of our components are using unwritten naming taxonomy `StyledName` I prefer that you would stick to it and wouldn't have it reversed, so it code base is consistent 🙂

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+    background: ${renderStatus(status, theme)};
+    height: ${desktop ? `2px` : `calc(100% + 20px)`};
+  `}
+`;
+
+ProgressLineStyled.defaultProps = {
+  theme: defaultTheme,
+};
+
+const TimeWrapperStyled = styled.div`
```
vepor, line 65: Please don't use specific – explicit name for any of the text properties. There can be more usages rather than just `date/time` in the timeline step. Moreover, usage of this text property should be optional, therefore the condition is unnecessary. It will be required by design, it should show up and not be hidden just visually.

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+import type { Props } from "./index";
+
+const IconWrapperStyled = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  min-width: ${({ theme }) => theme.orbit.spaceLarge};
+  min-height: ${({ theme }) => theme.orbit.spaceLarge};
+  svg {
```
vepor, line 24: Should not be necessary to use when the progress indicator will be rendered from the bottom edge of icon to the second one.

packages/orbit-components/src/Timeline/Timeline.stories.js
```diff
@@ -0,0 +1,137 @@
...
+          </TimelineStep>
+          <TimelineStep step={step} time={time} status={status}>
+            {children}
+          </TimelineStep>
+        </Timeline>
+      );
+    },
+    { info: "This is example of Timeline with critical status" },
+  )
+  .add("inside Modal", () => {
```
vepor, line 112: I'm missing story for RTL version, could you add it, so we know that it will work and we can have it screenshotted 🙂

packages/orbit-components/src/Timeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### Props
+
+Table below contains all types of the props in **TimelineStep** component.
+
+| Name         | Type            | Default | Description                                                               |
+| :----------- | :-------------- | :------ | :------------------------------------------------------------------------ |
+| **children** | `React.Node`    |         | The content of the component                                              |
+| step         | `string`        |         | Text for [`Badge`](https://orbit.kiwi/components/badge/) component inside |
+| status       | [`enum`](#enum) |         | Status of current process step                                            |
```
vepor, line 45: Even though that this component is based upon status colors we have in our foundation I would definitely use `type` name of this property that you're calling `status` currently. It will ensure consistent experience.

packages/orbit-components/src/Timeline/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+#### Props
+
+Table below contains all types of the props in **TimelineStep** component.
+
+| Name         | Type            | Default | Description                                                               |
+| :----------- | :-------------- | :------ | :------------------------------------------------------------------------ |
+| **children** | `React.Node`    |         | The content of the component                                              |
+| step         | `string`        |         | Text for [`Badge`](https://orbit.kiwi/components/badge/) component inside |
```
vepor, line 44: Last but not least, I would use `label` name here, as `step` is basically the whole thing and not just the content of the `Badge`.

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+  const { index, last } = useStep();
+  const { isDesktop } = useMediaQuery();
+
+  const nextStatus = statuses[index + 1];
+
+  React.useEffect(() => {
+    setStatuses(prev => ({ ...prev, [index]: status || null }));
+  }, [setStatuses, status, index]);
+
+  return isDesktop ? (
```
vepor, line 108: The last suggestion I have is to actually split the `TimelineStep` to two components – one for each version mobile & desktop. It will be more readable when updates or fixes will be needed 🙂

packages/orbit-components/src/Timeline/TimelineStep/index.js
```diff
@@ -0,0 +1,160 @@
...
+          <ProgressLineStyled
+            data-test="progressLine"
+            desktop
+            status={nextStatus || (last && status)}
+          />
+        </RelativeStyled>
+        <Stack flex align="center" direction="column">
+          <Stack inline spacing="condensed" align="center">
+            <Badge type={status}>
+              <Text type={status || "secondary"}>{step}</Text>
```
vepor, line 128: We can use `Text` component here (also in the mobile version). It has totally different color and line height, moreover the used size and weight is wrong.

I would use only the `Badge` component in these cases. To implement `secondary` color for the text we can:
- either use `BadgePrimitive` to recreate `neutral` type of the `Badge` and just tweak the color,
- or use `neutral` badge, and overwrite its color to `secondary`.

With both approaches it will look like in the design specification.

vepor (changes requested):
Good job @mainframev 💪

I just have some things to have polished before merge 🙂

packages/orbit-components/src/Timeline/TimelineStep/index.js
mainframev: should  i make changes to neutral badge or it's gonna be discussed with Will first?

vepor: No, as I mentioned, please either use `BadgePrimitive` to **recreate** the type `neutral` of the `Badge` component and just tweak the font color to match the design specification.

Or, alternatively, you can use `<Badge type="secondary" />` and just overwrite the font color locally by selector in your styled component.

mainframev: aha, got it, thanks :)

packages/orbit-components/src/Timeline/index.js
```diff
@@ -0,0 +1,51 @@
...
+  margin-bottom: ${getSpacingToken};
+  z-index: ${({ isInsideModal }) => isInsideModal && `1000`};
+`;
+
+WrapperStyled.defaultProps = {
+  theme: themeDefault,
+};
+
+const Timeline = ({ children, spaceAfter, dataTest }: Props) => {
+  const childs = React.Children.toArray(children);
```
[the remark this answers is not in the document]
mainframev: @vepor  [english](https://lh3.googleusercontent.com/proxy/HkX1mWSMWFmQPyEaICBqz7nSf2EBcPScdw2GYUq257fwF6bnsQ4vr6FySfrdlORxZsIHkExliFvQY-xnhPqscdUP-SUT7ZkmmWHPMZf4URyum0MwodgT-Z5BnhPUSQuyexBI0uIai_sa8fZvtmNW8QI469SqVA3rGYI) :D

packages/orbit-components/src/Timeline/TimelineStep/index.js
mainframev: It should be hidden visually, there is case, when it can be provided, but status will be something like waiting

mainframev: It should be hidden, in case, when the status is not success (aka label) and not current step, user will have to provide time, but it should be hidden and it's easier to handle height in that case

vepor: It has to be, from design point of view that was used in the specification.
After discussion with @williamkolmacka we want to have it opened so it will be up to the developer or designer to fit their case the most.

silvenon (changes requested):
1. The text in labels looks like it should be centered:

Screen Shot 2020-10-15 at 15 02 32

2. The text is still being cut off on mobile:

Screen Shot 2020-10-15 at 15 12 52

3. The line is missing in RTL on mobile:

Screen Shot 2020-10-15 at 15 11 11

silvenon (approved):
Looks great!

Issue #2305, opened 2020-10-07
Add "improve" type to commitlint


vepor:
This type is very suitable for platform changes. This would be purely a semantic addition, but working on a very useful change only to tag it with "chore" is a bit demotivating.

PR #2330, opened 2020-10-13
improve(Commitlint): add the improve type to commitlint

aalises:
Adds the `improve` type to the commitlint configuration, so now you can do commits with the `improve` type. This is useful for huge tooling and platform changes, like the commit of this MR of course 😂

e.g. of usage (also in the commit)

```
improve(Something): adds a lot of awesome changes
```

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -3,5 +3,23 @@ module.exports = {
   extends: ["@commitlint/config-conventional"],
   rules: {
     "scope-case": [2, "always", "pascal-case"],
+    "type-enum": [
+      2,
+      "always",
+      [
+        "build",
+        "chore",
+        "ci",
+        "docs",
+        "feat",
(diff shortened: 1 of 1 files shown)


mainframev:
I would stick with standard conventional commits type, i think it's better to specify custom scope for it, not commit type

silvenon:
@mainframev I suggested this in #2330.

vepor:
As suggested in the #2305 I don't think that we have stick to standard conventional commits type.

aalises:
@mainframev I am fine with whatever! I think in that case it might be okay cc @vepor

silvenon (approved):
Ok, Victor and I synced on this. 😃 Thank you so much, I can't wait to use `improve`! ❤️❤️❤️

vepor (approved):

PR #2331, opened 2020-10-13
feat(Eslint): add rules for enforcing readOnly types on TS/Flow

aalises:
Adds rules for enforcing `readonly` and `exact` types on Typescript and Flow

`prefer-readonly-parameter-types` throws a `Maximum Call Stack Exceeded` error on `MediaQuery.d.ts` ... this has been reported previously on https://github.com/typescript-eslint/typescript-eslint/issues/1665 but was fixed.. something wrong happening with the version

We are not using that many parameters anyways... Also, they were some weird interactions with some third party library types (styled-components, react...) and some of our types, so had to disable some flow rule in a couple of files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,7 @@ module.exports = {
       files: "*.ts",
       extends: ["./config/eslintTS"],
     },
+    // some ESLint rules fail in certain cases, so we're disabling them
     {
       files: ["packages/orbit-components/src/utils/**/*"],
       rules: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,5 +15,11 @@ module.exports = {
       files: "*.ts",
       extends: ["./config/eslintTS"],
     },
+    {
+      files: ["packages/orbit-components/src/utils/**/*"],
+      rules: {
+        "@typescript-eslint/prefer-readonly-parameter-types": "OFF",
+      },
+    },
   ],
 };
--- a/config/eslintJS.js
+++ b/config/eslintJS.js
@@ -16,6 +16,7 @@ module.exports = {
   rules: {
     "no-console": ["error", { allow: ["warn", "error"] }],
     "prettier/prettier": "error",
+    "flowtype/require-exact-type": "error",
     "import/no-extraneous-dependencies": [
       "error",
       {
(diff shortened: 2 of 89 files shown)


config/eslintJS.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
   rules: {
     "no-console": ["error", { allow: ["warn", "error"] }],
     "prettier/prettier": "error",
+    "flowtype/require-exact-type": "error",
+    "flowtype/require-readonly-react-props": "error",
```
silvenon, line 20: It turns out that `flowtype/require-readonly-react-props` unfortunately doesn't detect props types in definition files, so it isn't useful for us, you can remove it.

config/eslintTS.js
```diff
@@ -6,7 +6,9 @@ module.exports = {
     "prettier/@typescript-eslint",
     "plugin:prettier/recommended",
   ],
+  plugins: ["import", "react"],
   rules: {
+    "@typescript-eslint/prefer-readonly": "error",
```
silvenon, line 11: We don't need this rule as it's not affecting our code. Can you disable `prefer-readonly-parameter-types` for the problematic file using an ESLint disable directive or `overrides` field?

silvenon (changes requested):
I was slightly nervous about these exact types breaking consumer code, but I tried it in `frontend/frontend` and everything worked fine. 🙂

package.json
```diff
@@ -57,7 +57,7 @@
     "caniuse-db": "^1.0.30001142",
     "eslint": "^7.7.0",
     "eslint-config-airbnb": "^18.2.0",
-    "eslint-config-prettier": "^6.11.0",
+    "eslint-config-prettier": "^6.12.0",
```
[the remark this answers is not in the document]
aalises: Not quite!! was part of some tests to see what was wrong with the `Maximum Call Exceeded`

vepor (dismissed):

.eslintrc.js
```diff
@@ -15,5 +15,11 @@ module.exports = {
       files: "*.ts",
       extends: ["./config/eslintTS"],
     },
+    {
```
silvenon, line 18:
```suggestion
    // some ESLint rules fail in certain cases, so we're disabling them
    {
```

silvenon (approved):
Amazing, thank you so much!!

silvenon (approved):

PR #2332, opened 2020-10-13
Add script for checking icons and add it on Pre-Commit

aalises:
Added a `checkIcons` script which checks that icons comply with given properties

- Have unique HTML comments
- Do not have `fill` attributes
- Have an standard size (24px x 24px)


silvenon (commented):
Awesome! 💯 We only really need to check the icons when they are added, so do you think you can tweak the implementation to take an argument which SVG file to check, and if no params are provided then all icons are checked?

That way you can:

1. run `check:icons` in `lint-staged` for staged `icons/svg/*.svg` files
2. run `check:icons` for all icons in CI (so without params)

We don't need to run it in `postinstall`, I believe. Let me know what you think! I'll check the code tomorrow.

silvenon (approved):
Wonderful, thanks!

PR #2339, opened 2020-10-14
fix(Slide): initialize max-height to work SSR

silvenon:
We were setting the `maxHeight` state in `componentDidMount`, and
considering that this method only runs in the browser, `max-height`
ended up being `0` in server-side rendering. There is no reason not to
initialize the state with the `maxHeight` prop right away.
<br/><br/><br/><url>LiveURL: https://orbit-fix-slide-expanded-ssr.surge.sh</url>


silvenon:
Forgot to complete tests, hold on…

packages/orbit-components/src/utils/Slide/index.js
```diff
@@ -28,7 +28,7 @@ StyledSlide.defaultProps = {
 
 class Slide extends React.Component<Props, State> {
   state = {
-    maxHeight: 0,
+    maxHeight: typeof this.props.maxHeight !== "undefined" ? this.props.maxHeight : null,
```
mainframev, line 31: should it be null? or if false set to `0` maybe?

silvenon: Yep!

silvenon:
@mainframev fixed, ready for review 👍

mainframev (approved):

PR #2349, opened 2020-10-15
chore: eslint jest rules

### LINKED ISSUE #2348: Prohibit use of .only() in Jest tests via ESLint
`eslint-plugin-jest` has the rule to prohibit the use of `.only()`, we should use it. It would probably be best to do this after #2347, but it's fine either way.

mainframev:
- forbid usage of [focused-tests](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-focused-tests.md)
- consistent `it` [consistent-test-it](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md)
- removed identical titles [no-identicial-title](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-identical-title.md)

Fixes #2348.

and etc. <br/><br/><br/><url>LiveURL: https://orbit-chore-eslint-jest-only.surge.sh</url>

### DIFF
--- a/config/eslintJS.js
+++ b/config/eslintJS.js
@@ -62,5 +62,10 @@ module.exports = {
     "no-await-in-loop": "off",
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
+    "jest/no-identical-title": "error",
+    "jest/no-standalone-expect": "error",
+    "jest/no-if": "error",
+    "jest/consistent-test-it": "warn",
+    "jest/no-focused-tests": "error",
   },
 };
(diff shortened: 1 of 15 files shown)


silvenon (commented):
I agree with most of these rules, but I would like to discuss `jest/consistent-test-it`. I like the freedom of sometimes using `it` and sometimes `test` because `it` tests can be read like "it uses media queries", where `test` tests can be read like "test accessibility" and "test styles". I don't think enforcing either one brings value, I like combining them.

mainframev:
@silvenon i think it's better to have consistency and left only `it`, description of `it` can be written as you want :)

vepor:
I don't have any strong opinion for `consistent-test-it`, I don't see any major problem with having both. It doesn't mean that it will be harder to read tests if we use both – the syntax is the same.

mainframev:
`it` should fit better with `@testing-library` since `it` focused on output, like: `it("should render ...")`

they are the same, just naming differences

silvenon (approved):
Ok, I'm fine with either 👍 please resolve conflicts.

PR #2355, opened 2020-10-16
feat(ListChoice): add disabled boolean property

vepor:
Added `disabled` boolean property, so it can be used in the cases when it's pre-filled with `true` for `selected` property and it user should not be able to change the value. <br/><br/><br/><url>LiveURL: https://orbit-feat-listchoice-disabled.surge.sh</url>


packages/orbit-components/src/ListChoice/index.js
```diff
@@ -41,19 +41,23 @@ const StyledListChoice = styled.div`
...
-      color: ${({ theme }) => theme.orbit.colorIconPrimary};
-    }
     outline: none;
+    ${({ disabled }) =>
+      !disabled &&
+      css`
+        background-color: ${({ theme }) => theme.orbit.paletteCloudLight};
+        ${StyledListChoiceIcon} svg {
+          color: ${({ theme }) => theme.orbit.colorIconPrimary};
+        }
```
mainframev, line 59: destructure theme together with disabled :)

packages/orbit-components/src/ListChoice/index.js
```diff
@@ -78,17 +82,19 @@ StyledListChoiceContent.defaultProps = {
...
     ...(selectable ? { "aria-checked": selected } : null),
   };
 
   return (
     <StyledListChoice
-      onClick={onClick}
+      onClick={!disabled ? onClick : null}
       data-test={dataTest}
-      onKeyDown={handleKeyDown(onClick)}
+      onKeyDown={!disabled ? handleKeyDown(onClick) : null}
```
mainframev, line 94: i suppose here should be undefined, not null :)

packages/orbit-components/src/ListChoice/index.js.flow
```diff
@@ -10,6 +10,7 @@ export type Props = {|
   +description?: Translation,
   +selectable?: boolean,
   +selected?: boolean,
+  +disabled?: boolean,
   +icon?: React$Node,
```
mainframev, line 14: i know, that you add only `disabled` prop, but here is `React$Node` below, you could replace it to `React.Node` too :)

packages/orbit-components/src/ListChoice/index.js
vepor: It's the same, it won't do anything 🤷

packages/orbit-components/src/ListChoice/index.js
vepor: Thanks, good catch! 🙂

packages/orbit-components/src/ListChoice/index.js.flow
vepor: ✅

packages/orbit-components/src/ListChoice/index.js
mainframev: i know :) it's SC, in normal components we usually pass undefined in such cases, so i would follow same logic here, makes no difference in the end, but for consistency. Also i suppose it's better from perspective `null` and `undefined` differences, like here is more about something that hasn't been initialized

mainframev (approved):

packages/orbit-components/src/ListChoice/index.js
vepor: That would mean that we have to go through the whole code base, we're not consistent when comes to usage of `null` or `undefined`.

PR #2358, opened 2020-10-17
feat(Types): expose enum types from ts/flow definitions of components

aalises:
Expose `ts/flow` enum types from components:

- `Text`: Weight, Type, As, Align
- `Illustration`: Name
- `Alert`: Type
- `Button`: Type
- `ButtonLink`: Type
- `Popover`: AlignsCore (used for preferredAlign) and Position
- `Badge`: Type
- `TextLink`: Type

### DIFF
--- a/packages/orbit-components/src/Alert/README.md
+++ b/packages/orbit-components/src/Alert/README.md
@@ -37,6 +37,15 @@ The table below contains all types of the props available in Alert component.
 | `"warning"`  |
 | `"critical"` |
 
+### Types
+
+You can import the `Type` `enum` as a type from the component definition
+
+```js
+import type { Type } from "@kiwicom/orbit-components/lib/Alert"; // Flow
+import { Type } from "@kiwicom/orbit-components/lib/Alert"; // TS
+```
+
 ## Functional specs
 
 - By passing the `closable` prop into Alert, you will be able to handle `onClose` function and Close icon will be displayed. Also, if you want to select the Close Button element for testing purposes, use [data-test="AlertCloseButton"] selector.
--- a/packages/orbit-components/src/Alert/index.d.ts
+++ b/packages/orbit-components/src/Alert/index.d.ts
@@ -7,9 +7,9 @@ import * as React from "react";
 import * as Common from "../common/common";
 import AlertButton from "./AlertButton";
 
-type Type = "info" | "success" | "warning" | "critical";
-
 declare module "@kiwicom/orbit-components/lib/Alert";
+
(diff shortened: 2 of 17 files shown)


RobinCsl:
I would prefer to extract the necessary information from the component's props instead, e.g. see paragraph 7 of https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680

I believe a similar way exists with Flow.

aalises:
@RobinCsl that would actually work without any changes, as we are already exposing `Props` right? so just doing `Props["type"]` should work, or some utility magic in Flow...

RobinCsl:
> @RobinCsl that would actually work without any changes, as we are already exposing `Props` right? so just doing `Props["type"]` should work, or some utility magic in Flow...

Yes, in Flow it could be done with
```
import React from 'react';

type Props = {
     firstName: "John" | "Uhtred" | "Ragnar",
     lastName: string,
}

const Greet = ({firstName, lastName}: Props) => {
     const fullName = `${firstName} ${lastName}`;
     return (
          <h1>Hello {fullName}</h1>
     )
}

type firstName = $PropertyType<React$ElementProps<typeof Greet>, 'firstName'>;

// so mainstream: no error
const john: firstName = 'John';

// Such a rebel: error
const aragorn: firstName = 'Aragorn';
```
Try Flow Playground: 
https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocmATzCTgAUcwBnOAXjgG9y4+-MwKCxgA5FCCQAuOACIAUhAAWAO1lwAPnICqSmEQAm6rbIQoA5ipRRZAGl784AGxQjxkmSKjAV5+wF9KNAgVETgAcSIkeE4ACi5BYTEJJFtnV2TJfxkmCFYASg4APm4HfmDQrABXJyd3ek4AAwASBKE3FP84VpcOrMaKRzgiGCqoFThYsqG+AB4lAEYigAkkWohuTBq6ztmAekWi6bh88kCqWnpEvobu3LpYagAVS9nkdBhmgFEnJEkVGD3FizGh0CCYCJRGBFNL4a6ZJD4IoUch7PZwFgbEAoHxeVAgcgVMIAK2UKhk8PqHAIilUZEoaLgAGUqmglHAUMMkAAjNaEkJhawWaDkuCUlLU-AAQSgwvG9KAA

silvenon:
We decided that it's better to extract types from props instead of exporting needed types individually. It doesn't require any changes from us and consumers can still get any type they need.

PR #2369, opened 2020-10-19
test(TextLink): use testing library

vepor:
<br/><br/><br/><url>LiveURL: https://orbit-test-textlink-library.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/TextLink/__tests__/index.test.js
+++ b/packages/orbit-components/src/TextLink/__tests__/index.test.js
@@ -1,73 +1,68 @@
 // @flow
 import * as React from "react";
-import { shallow } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import TextLink from "../index";
 import ChevronRight from "../../icons/ChevronRight";
 
-describe("TextLink", () => {
-  const title = "My text link";
-  const href = "https://kiwi.com";
-  const onClick = jest.fn();
-  const type = "primary";
-  const tabIndex = "-1";
-  const dataTest = "test";
-  const rel = "nofollow";
+const title = "My text link";
+const dataTest = "test";
 
-  const component = shallow(
-    <TextLink
-      onClick={onClick}
-      href={href}
-      type={type}
-      rel={rel}
-      external
-      icon={<ChevronRight />}
-      tabIndex={tabIndex}
-      dataTest={dataTest}
-    >
-      {title}
-    </TextLink>,
-  );
+describe("#TextLink", () => {
+  it("should be focusable and have button role", () => {
+    render(<TextLink>{title}</TextLink>);
+    userEvent.tab();
+    expect(screen.getByText(title)).toHaveFocus();
+    expect(screen.getByRole("button")).toBeInTheDocument();
+  });
 
-  const componentWithoutHref = shallow(<TextLink>{title}</TextLink>);
-
-  it("should contain a children", () => {
-    expect(component.children().exists()).toBe(true);
-  });
-  it("should contain an href", () => {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/TextLink/__tests__/index.test.js
```diff
@@ -1,73 +1,68 @@
...
-    expect(component.render().prop("tabindex")).toBe(tabIndex);
-  });
-  it("should contain an external href", () => {
-    expect(component.render().prop("target")).toBe("_blank");
-  });
-  it("should contain SVG", () => {
-    expect(component.find("ChevronRight").exists()).toBe(true);
+describe("#TextLink", () => {
+  it("should be focusable and have button role", () => {
+    render(<TextLink>{title}</TextLink>);
```
silvenon, line 14: Considering that you're testing the role, it's best to set it explicitly, otherwise changing the default `asComponent` will needlessly break the test.

```suggestion
    render(<TextLink asComponent="button">{title}</TextLink>);
```

packages/orbit-components/src/TextLink/__tests__/index.test.js
```diff
@@ -1,73 +1,68 @@
...
+        href={href}
+        icon={<ChevronRight dataTest={dataTestIcon} />}
+        tabIndex={tabIndex}
+        dataTest={dataTest}
+      >
+        {title}
+      </TextLink>,
+    );
+    expect(screen.getByText(title)).toBeInTheDocument();
+    expect(screen.getByText(title).closest("a")).toHaveAttribute("href", href);
```
silvenon, line 42: Seems like this should be enough:
```suggestion
    expect(screen.getByRole("link")).toHaveAttribute("href", href);
```

packages/orbit-components/src/TextLink/__tests__/index.test.js
```diff
@@ -1,73 +1,68 @@
...
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+    expect(screen.getByTestId(dataTestIcon)).toBeInTheDocument();
+    expect(screen.getByTestId(dataTest)).not.toHaveAttribute("role");
```
silvenon, line 45: This is an example of testing an implementation detail; whether the role is being set explicitly or not is not important, you just want to ensure that it is what you expect. In this case it is _implicit_ and equals to `link` (see [this reference of implicit roles](https://www.w3.org/TR/html-aria/)). You can test that like this:
```suggestion
    expect(screen.getByRole("link")).toBeInTheDocument();
```

packages/orbit-components/src/TextLink/__tests__/index.test.js
```diff
@@ -1,73 +1,68 @@
...
+    );
+    expect(screen.getByText(title).closest("a")).toHaveAttribute(
+      "rel",
+      expect.stringContaining("noreferrer"),
+    );
+    expect(screen.getByText(title).closest("a")).toHaveAttribute(
+      "rel",
+      expect.stringContaining(rel),
+    );
+    expect(screen.getByText(title).closest("a")).toHaveAttribute("target", "_blank");
```
silvenon, line 66: It would be easier on the eyes to save the link element to a variable.

silvenon (changes requested):
Excellent! Other than some minor stuff. Also, Testing Library recommends to lean towards using `ByRole` queries, unless it's not possible or for some other reason it makes more sense to use another query.

packages/orbit-components/src/TextLink/__tests__/index.test.js
vepor: Make total sense! 🙂

silvenon (approved):

PR #2370, opened 2020-10-19
fix(Timeline): IE fixes

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-fix-timeline.surge.sh</url>


packages/orbit-components/src/Timeline/TimelineStep/primitives/StyledIconWrapper.js
```diff
@@ -4,12 +4,11 @@ import styled from "styled-components";
...
 
 const StyledIconWrapper = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: center;
   min-width: ${({ theme, mobile }) => mobile && theme.orbit.spaceLarge};
   min-height: ${({ theme }) => theme.orbit.spaceLarge};
   z-index: 1;
+  text-align: center;
+  line-height: 1.3;
```
vepor, line 11: Could you just explain why value `1.3` exactly? 🙂

vepor (changes requested):
There is last bug: the text overflows the parent containers 🙁 
Timeline bug

packages/orbit-components/src/Timeline/TimelineStep/primitives/StyledIconWrapper.js
mainframev: @vepor 

well, it looked like optimal line-height in IE and Chrome, we have tokens for 1.2 and 1.4 🤔

vepor: Not sure if we can use `flexbox` as we do for now. It would make sense to use height token then. If not I'm fine with explicit `line-height` also. Usage of some specific token would be the best, if that won't work, please just attach some comment 😉

mainframev:
@vepor, @silvenon can we merge it?

vepor:
Going to check @mainframev!

vepor (approved):
Everything works properly! 🎉

PR #2371, opened 2020-10-19
chore: publish

silvenon:
- @kiwicom/babel-plugin-orbit-components@1.7.1
 - @kiwicom/orbit-components@0.94.0


vepor (approved):

silvenon:
The major version got bumped because of Grid feature.

PR #2373, opened 2020-10-19
feat(TextLink)!: add standAlone, noUnderline & iconRight, rename icon to iconLeft

vepor:
- Added `noUnderline` property to TextLink
- Added `standAlone` property to TextLink
- Renamed `icon` to `iconRight` on TextLink
- Added `iconLeft` to TextLink

<br/><br/><br/><url>LiveURL: https://orbit-improve-textlink-primitive.surge.sh</url>


silvenon (commented):
As much as I like abstractions, I think this could impose some restrictions later on if features of `ButtonPrimitive` start to diverge from `TextLink`. Repetition is probably a better strategy here. Also, semantically it's a little confusing that something called "link" is based on something called "button", as those are two different concepts in UI. Although I'm aware that links and buttons are pretty interchangeable in Orbit.

packages/orbit-components/src/primitives/ButtonPrimitive/common/createRel.js
```diff
@@ -0,0 +1,18 @@
...
+    if (!relValues.includes("noopener")) {
+      relValues.push("noopener");
+    }
+    if (!relValues.includes("noreferrer")) {
+      relValues.push("noreferrer");
+    }
+  }
+  return relValues.length > 0 ? relValues.join(" ") : undefined;
+};
+
```
mainframev, line 17: consider to cover it with test :)

mainframev (approved):

packages/orbit-components/src/primitives/ButtonPrimitive/common/createRel.js
vepor: I've just noticed that the coverage of tests of Button, ButtonLink and ButtonPrimitive is very low, so we should introduce more tests regardless this change 🙁

mainframev: then it can be done by separate MR i think :)

vepor: I'll put it into Jira – and connect it with move to `testing-library/react`

williamkolmacka (approved):
Looks good, approved from a design perspective. 🎉

PR #2377, opened 2020-10-20
feat(ButtonLink)!: remove background for compact, delete inline type

vepor:
- Remove type `inline` from ButtonLink – the usage can be replaced by TextLink now
- Removed interactive states of background when `compact` is used<br/><br/><br/><url>LiveURL: https://orbit-feat-buttonlink-remove-inline.surge.sh</url>


packages/orbit-components/src/Breadcrumbs/index.js.flow
```diff
@@ -9,7 +9,7 @@ import type { spaceAfter } from "../common/getSpacingToken/index";
 export type Props = {|
   ...Globals,
   +children: React$Node,
-  +onGoBack?: (ev: SyntheticEvent<HTMLButtonElement>) => void | Promise<any>,
+  +onGoBack?: (ev: SyntheticEvent<HTMLElement>) => void | Promise<any>,
```
silvenon, line 12: Why is this not `HTMLAnchorElement` as well?

williamkolmacka (dismissed):
Everything is okay except the hover state of secondary type ButtonLink in compact mode. It's almost invisible, so I would suggest to solve it similarly as we have it on secondary TextLink. Try to color it with product normal on hover state, it should work better. :)

silvenon (approved):
Looks good!

silvenon:
Also, I would recommend to rebase and check for ESLint errors, just in case.

packages/orbit-components/src/Breadcrumbs/index.js.flow
vepor: Because `onGoBack` is used both on `TextLink` and `Button` there's a conflict between `HTMLButtonElement` and `HTMLAnchorElement`. Therefore I used `HTMLElement`.

silvenon: Ok!

mainframev (dismissed):

silvenon (approved):
From the code perspective looks good!

mainframev (approved):

PR #2381, opened 2020-10-21
chore(Flow): replace `React$*` with `React.*`

### LINKED ISSUE #2345: Prohibit usage of global React Flow types (e.g. React$Node)
Global React Flow types are never mentioned in the Flow documentation, which strongly suggests that they don't like people using them. Also, using them in Orbit apparently caused some problems for others, so we should definitely stop using them and prohibit their usage via ESLint. However, I'm not aware of any ESLint rule that can do this, maybe there is one, I haven't thoroughly checked, but in case there isn't it should this is a great opportunity to learn how to build one because as far as complexity goes, this should be one of the simpler ones. ESLint has great documentation for [working with rules](https://eslint.org/docs/developer-guide/working-with-rules) and [testing them](https://eslint.org/docs/developer-guide/unit-tests), and [AST explorer](https://astexplorer.net/) is your friend here.

sampi:
closes #2345

> ESLint is now going to throw an error on usage of `React$*`.

-----------

the `error` prop of `FormFeedback` was made optional, components calling FormFeedback have the field already set to be optional, so this field has been optional implicitly already.<br/><br/><br/><url>LiveURL: https://orbit-sampi-fix-2345.surge.sh</url>


mainframev (approved):

mainframev:
@sampi looks good, thx :)

mainframev (approved):

PR #2383, opened 2020-10-21
chore(BabelPlugin): move devDependency to peer

### LINKED ISSUE #2368: Avoid unnecessary version bumps of @kiwicom/babel-plugin-orbit-components
`@kiwicom/babel-plugin-orbit-components` always gets a version bump because it has `@kiwicom/orbit-components` in `devDependencies`. It will probably be enough to remove it, otherwise we should consider removing all `devDependencies` from both packages.

Success can be checked by running `yarn release`, then seeing whether the version is unjustifiably bumped.

lerna/lerna#1737

sampi:
closes #2368

--------
remove internal dependency (it will use it anyway, but won't affect version bumps)

<br/><br/><br/><url>LiveURL: https://orbit-sampi-fix-2368.surge.sh</url>

### DIFF
--- a/packages/babel-plugin-orbit-components/package.json
+++ b/packages/babel-plugin-orbit-components/package.json
@@ -22,7 +22,6 @@
   "devDependencies": {
     "@babel/cli": "^7.12.0",
     "@babel/core": "^7.11.6",
-    "@kiwicom/orbit-components": "^0.94.0",
     "jest": "^26.5.2",
     "prettier": "^2.1.2"
   },


silvenon (changes requested):
It's not appropriate to move it to `peerDependencies` because `@kiwicom/orbit-components` is only needed for testing. Even though I can't imagine the situation where you would use the Babel plugin without it, technically speaking it's not needed.

We don't expect adding tests for this, only that you check locally whether it works. With `yarn release` you can bump versions and resetting locally, so you can test that way. I tested it myself, if you simply remove `@kiwicom/orbit-components` from `devDependencies`, tests for the Babel plugin will still work and its version will no longer be bumped unnecessarily. So doing this is good enough for us!

silvenon (approved):

PR #2385, opened 2020-10-21
chore: move devDependencies to root

### LINKED ISSUE #2384: Move all devDependencies to root package.json
Bumping `devDependencies` also bumps the version, as we've found out, so it's most likely best to move them all to the root `package.json`. The added benefit is that way we avoid having to bump the same development dependency in multiple places. I think it will be easier to manage the monorepo this way.

sampi:
closes #2384

* ran `lerna link convert`
* manually checked that all dependencies are still available
* modified `eslintJS.js` to look for `devDependencies` in the root `package.json` as well

--------------------

<br/><br/><br/><url>LiveURL: https://orbit-sampi-fix-2384.surge.sh</url>


silvenon (approved):
Thanks so much! ❤️

PR #2386, opened 2020-10-21
docs: Create guide on how to use component types

### LINKED ISSUE #2379: Create guide how to extract types from props
As discussed in #2358, we don't have to explicitly export and document `types`.

Instead of it, we can take leverage of utility types that Typescript and Flow offers. See https://github.com/kiwicom/orbit/pull/2358#issuecomment-712224827

That means we can just create official guide how to use it and put it on orbit.kiwi.

aalises:
Small guide (for developers!) on how to use component's exported types (using built-in utility types in `Flow` and `Typescript`).

As discussed with @CollierCZ , I placed them under `docs/guides`. It should not matter much for now as right now the guides are generated in `Wordpress` but definitely, in the future, more content should be moved under that folder!

The subtitle, would be something like: `Guidelines on how to use and make the most out of the component exported types in Typescript and Flow`

Closes #2379


vepor (dismissed):
Looks great! Thanks a lot @aalises 🙂

CollierCZ (approved):

PR #2388, opened 2020-10-21
chore: switch to native Dependabot

silvenon:
Seems like it would be good to switch to v2 because we'll be benefiting from better documentation and more features. See [Keeping your dependencies updated automatically](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically).<br/><br/><br/><url>LiveURL: https://orbit-chore-dependabot-v2.surge.sh</url>


silvenon:
If you agree, I the old Dependabot needs to be disabled in the web app.

silvenon:
@vepor removed `yarn autoclean`, it's a good thing you noticed!

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
mainframev, line 6: do we need it daily? maybe it's worth to consider to do it weekly? 🤔

mainframev (approved):

PR #2389, opened 2020-10-21
feat(Loki): skip code blocks

### LINKED ISSUE #2346: Remove code blocks from visual regression testing
We're using [Loki](https://www.npmjs.com/package/loki) for visual regression testing. Currently our screenshots include code blocks, which is a bit problematic because it increases the number of false positives. When we change the code a bit but the output of the component hasn't changed, that's still being treated as a visual regression, which slows down publishing a new release. It would be great if we can somehow remove code blocks while performing visual regression testing.

sampi:
closes #2346

The idea is that we add a wrapping `<div>` with a specific ID to the decorator that adds the source code at the bottom. This way we can tell Loki to only take a screenshot of that `<div>` and not the whole page, therefore skipping story-code changes.

The area of the screenshot to be taken by Loki:

Screenshot 2020-10-21 at 19 38 42

### TODO

Before merging:
* Help me run the visual regression tests (I don't have docker locally, the documentation is "coming soon", and it's getting late) cc @silvenon ,
* Make sure that absolutely positioned elements are captured correctly (that is the only caveat of this wrapper `<div>` solution.

---------------------------------

<br/><br/><br/><url>LiveURL: https://orbit-sampi-fix-2346.surge.sh</url>

### DIFF
--- a/packages/orbit-components/.storybook/orbitDecorator.js
+++ b/packages/orbit-components/.storybook/orbitDecorator.js
@@ -18,7 +18,7 @@ const orbitDecorator = (storyFn, context) => {
     <div style={{ padding: "20px" }}>
       <Heading spaceAfter="medium">{context.kind}</Heading>
       <Text spaceAfter="largest">{context.parameters?.info}</Text>
-      {children}
+      <div id="component_preview">{children}</div>
       <div style={{ marginTop: 20 }}>
         <>
           <Code code={jsxToString(children, options)} language="jsx" format={false} />
--- a/packages/orbit-components/package.json
+++ b/packages/orbit-components/package.json
@@ -58,6 +58,7 @@
     "yarn.lock"
   ],
   "loki": {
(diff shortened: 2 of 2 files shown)


mainframev:
great! thanks a lot :)

mainframev (approved):

silvenon (changes requested):
Could you instead create a npm script in the root `package.json` called `dev:test` (I plan to add regular `dev` later) which contains something like this:

```
cross-env NODE_ENV=test yarn workspace @kiwicom/orbit-components storybook
```

then in `orbitDecorator.js` find this part:

```jsx
<div style={{ marginTop: 20 }}>
  <>
    <Code code={jsxToString(children, options)} language="jsx" format={false} />
  </>
</div>
```

and add it only if `process.env.NODE_ENV !== "test"`. This approach might be more stable.

PR #2391, opened 2020-10-21
feat(AlertButton): set small as default size for alert button

### LINKED ISSUE #2307: Remove all sizes from AlertButton besides small
We currently offer a possibility to have different sizes on `AlertButton` component, but from design perspective only `small` is feasible. Therefore we can do a breaking change and delete `normal` and `large` size from this component and make `small` as the default and only one.

ggolubic:
Closes #2307 

- Set small as default size for AlertButton
- Updated ts and flow types
- Updated storybooks and docs


vepor (approved):
Amazing! Thanks a lot @ggolubic! 🙂

PR #2403, opened 2020-10-26
Tag component changes

mainframev:
BREAKING CHANGE
- Icon prop removed
- Different colors
- Remove focusing if onClick/onRemove not provided
- rewrote to testing-library

3 commits in the end

I hope i got it right about colors, plz check if it's correct tokens @vepor 
<br/><br/><br/><url>LiveURL: https://orbit-feat-tag.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Tag/__tests__/index.test.js
+++ b/packages/orbit-components/src/Tag/__tests__/index.test.js
@@ -26,4 +26,3 @@ describe("Tag", () => {
     expect(onRemove).toHaveBeenCalled();
     expect(screen.getByText(content)).toBeInTheDocument();
   });
-});
--- a/packages/orbit-components/src/Tag/__tests__/index.test.js
+++ b/packages/orbit-components/src/Tag/__tests__/index.test.js
@@ -26,3 +26,4 @@ describe("Tag", () => {
     expect(onRemove).toHaveBeenCalled();
     expect(screen.getByText(content)).toBeInTheDocument();
   });
+});
--- a/packages/orbit-components/src/Tag/index.js
+++ b/packages/orbit-components/src/Tag/index.js
@@ -3,7 +3,7 @@ import * as React from "react";
 import styled, { css } from "styled-components";
 
 import defaultTheme from "../defaultTheme";
-import { rtlSpacing, left, right } from "../utils/rtl";
+import { rtlSpacing, left } from "../utils/rtl";
 import CloseCircle from "../icons/CloseCircle";
 import { SIZES, STATES } from "./consts";
 import KEY_CODE_MAP from "../common/keyMaps";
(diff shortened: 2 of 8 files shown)


silvenon (changes requested):
Otherwise looks good, and stories work fine. 👍

packages/orbit-components/src/Tag/index.js
```diff
@@ -19,103 +20,106 @@ const getFontSize = ({ theme, size }) => {
...
+    padding: ${rtlSpacing(theme.orbit.paddingTag)};
+    transition: color ${theme.orbit.durationFast} ease-in-out,
+      box-shadow ${theme.orbit.durationFast} ease-in-out,
+      background ${theme.orbit.durationFast} ease-in-out;
+
+    &:focus {
+      outline: 0;
+    }
+
+    ${({ actionable }) =>
```
vepor, line 92: Could be moved to the first function in the top - if we want to go with one function 🙂

packages/orbit-components/src/Tag/index.js
```diff
@@ -153,35 +157,34 @@ const Tag = (props: Props) => {
...
-      icon={!!icon}
-      tabIndex="0"
+      tabIndex={(onClick || onRemove) && "0"}
       role="button"
-      onKeyDown={ev => buttonClickEmulation(ev, onClick)}
+      onKeyDown={buttonClickEmulation(onClick)}
     >
+      {children}
       {!!onRemove && (
         <CloseContainer
```
vepor, line 166: The `CloseContainer` component doesn't have any transition 😕

vepor (changes requested):
One last thing to resolve - the states of colors. Put it into the original thread on Slack.

packages/orbit-components/src/Tag/__tests__/index.test.js
```diff
@@ -1,42 +1,29 @@
...
-
-  it("should have passed props", () => {
-    expect(component.render().prop("data-test")).toBe(dataTest);
-  });
-  it("should contain a content", () => {
-    expect(component.render().text()).toBe(content);
-  });
-  it("should contain a icon", () => {
-    expect(component.find("Bus").exists()).toBe(true);
+    userEvent.click(screen.getByLabelText("close"));
```
silvenon, line 25:
```suggestion
    userEvent.click(screen.getByRole("button", { name: "close" }));
```

silvenon (changes requested):
Other than this minor suggestion everything else looks good. Regarding colors I'm not in the loop (not sure where that Slack thread is), so I'll leave it up to @vepor to confirm.

silvenon (approved):

silvenon (approved):
Looks good, and the close button has transitions now 👍

PR #2409, opened 2020-10-26
feat(Itinerary): init

mainframev:
Storybook: https://orbit-silvenon-feat-Itinerary.surge.sh


packages/orbit-components/src/Itinerary/index.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const Itinerary = ({ children, dataTest, spaceAfter }: Props) => {
+  const content = React.Children.toArray(children);
+
+  return (
+    <StyledItineraryWrapper data-test={dataTest} spaceAfter={spaceAfter}>
+      {content &&
+        content.length > 0 &&
+        React.Children.map(content, el => {
```
vepor, line 26: Is this mapping really necessary?

packages/orbit-components/src/Itinerary/index.js
```diff
@@ -0,0 +1,35 @@
...
+          if (!React.isValidElement(el)) return null;
+          return el;
+        })}
+    </StyledItineraryWrapper>
+  );
+};
+
+export { ItineraryPart };
```
vepor, line 34: As with `ItineraryPart` I would export from the `Itinerary/index.js` also other related Itinerary components, e.g. `ItineraryPartPlace` etc.

packages/orbit-components/src/Itinerary/ItineraryPart/index.js
```diff
@@ -0,0 +1,73 @@
...
+            ? // eslint-disable-next-line no-return-assign
+              React.cloneElement(el, { ref: r => (refs.current[i] = r) })
+            : el}
+        </ItineraryPartProvider>
+      );
+    });
+
+  return (
+    <StyledWrapper spaceAfter={spaceAfter} data-test={dataTest}>
+      {status ? (
```
vepor, line 58: I would extract it to render function that would look like this, so we don't have to have duplicate of `<Stack direction="column">{parts}</Stack>`:

```jsx
const renderWithStatus = children => {
  if(status) {
    return (
      <ItineraryPartStatus type={status} label={label} offset={calculatedWidth}>
        {children}
      </ItineraryPartStatus>
  }
  return children;
}
return (
  <StyledWrapper spaceAfter={spaceAfter} data-test={dataTest}>
    {renderWithStatus(<Stack direction="column">{parts}</Stack>)}
  </StyledWrapper>
)
```

packages/orbit-components/src/Itinerary/ItineraryPart/index.js
```diff
@@ -0,0 +1,73 @@
...
+        <ItineraryPartProvider
+          index={i}
+          last={i === content.length - 1}
+          isNextHidden={content[i + 1] && content[i + 1].props.hidden}
+          count={content.length}
+          isHidden={el.props.hidden}
+          hasStatus={!!status}
+          calculatedWidth={calculatedWidth}
+        >
+          {el.type.displayName === "ItineraryPartPlace"
```
vepor, line 48: I remember that we had problems in the past with usage of `displayName`.
https://github.com/kiwicom/orbit/pull/733/files#diff-79769f43c5aef4029a31e3e93d7dac9e5f2bae8e87782dc40952611f49994ef6L331

It won't work in some cases, or browsers, so the best would be to have different approach to it.

vepor: I see that the ref is being used for re-calculation if there's some part has bigger width than the others. Since we want to cover this edge case I think that we can simply have a callback in the context that will be called after hydratation. The ref can be local for the `ItineraryPartPlace` or any other component and it will be trigger the state update in `useLayoutEffect`.

packages/orbit-components/src/Itinerary/ItineraryPart/components/ItineraryPartStatus/index.js
```diff
@@ -0,0 +1,104 @@
+// @flow
```
vepor, line 1: Since all components in `*/ItineraryPart/components/*` are public API, I would suggest to have it as in folder structure like:`Itinerary/ItineraryPart/ItineraryPartStatus`. So it's aligned with the rest of our code base and API.

Components in `*/components` folder are being used only as internal one – for instance check `Card` component.

vepor (dismissed):

packages/orbit-components/src/Itinerary/ItineraryPart/index.js
mainframev: There is a better way then, just to wrap parts in `Stack` and no need to create function then :)

vepor: If there is, let's go with it. But I would generally avoid code duplication that can potentially result in mistakes when updating the component 🙂

packages/orbit-components/src/Itinerary/ItineraryPart/index.js
mainframev: good point about `useLayoutEffect` did not notice that flickering, so it's exactly the case for it, thanks :)

mainframev:
released as RC version [0.110.1rc](https://www.npmjs.com/package/@kiwicom/orbit-components/v/0.110.1-rc.0) for temporary usage.

mainframev:
> released as RC version [0.110.1rc](https://www.npmjs.com/package/@kiwicom/orbit-components/v/0.110.1-rc.0) for temporary usage.

`0.110.1rc` has been deprecated. Use [@kiwicom/orbit-components@0.117.1-rc.0](https://www.npmjs.com/package/@kiwicom/orbit-components/v/0.117.1-rc.0)

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.js
```diff
@@ -0,0 +1,165 @@
...
+                    <>
+                      <StyledHeadingOffset>
+                        <Text weight="bold" spaceAfter="medium">
+                          {heading}
+                        </Text>
+                      </StyledHeadingOffset>
+                      <Stack direction="column" spacing="XSmall" spaceAfter="medium">
+                        {items.map(({ icon, text, additional }) => {
+                          return (
+                            <Stack flex grow={false} align="center">
```
silvenon, line 144: Needs unique `key`.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.js
```diff
@@ -0,0 +1,165 @@
...
+            {opened ? <ChevronUp /> : <ChevronDown />}
+          </Stack>
+        </StyledInnerWrapper>
+        <Slide maxHeight={height} expanded={opened} id={slideID} ariaLabelledBy={labelID}>
+          <StyledExpandable ref={ref} onClick={setOpened}>
+            <StyledExpandableContent offset={calculatedWidth}>
+              {content &&
+                content.map(({ heading, items }) => {
+                  return (
+                    <>
```
silvenon, line 135: Needs unique `key`, use `React.Fragment` instead.

silvenon (changes requested):
Here's what I have for now. After you implement last visual changes I'll carefully go over the CSS.

Btw, when seeing all this new Flow code I can't help but to think how nice it would be to get #3047 merged and rewrite this to TypeScript. You would probably run into some bugs with the setup because it would be the first Orbit component in TypeScript, so I'm sure it has some blind spots, but it would serve as a nice thorough test of my setup. 😄

silvenon:
You also need to run `yarn bootstrap`, because this change is missing from `yarn.lock`.
```diff
diff --git a/yarn.lock b/yarn.lock
index 533aee114..e75c18b71 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -2450,6 +2450,15 @@
   resolved "https://registry.yarnpkg.com/@kiwicom/browserslist-config/-/browserslist-config-1.0.0.tgz#4abe67d42fd037f87e5abc20eb9e392a14639a30"
   integrity sha512-PoLLgols9+D58y58m5dbFCSxe6SqT7v+80zPqvDY/qfb/ZYbZ8dcCDrceMgdu2z0oP0D7ZtGK+7lpPXsIqRDYw==
 
+"@kiwicom/orbit-components@^0.117.0":
+  version "0.117.0"
+  resolved "https://registry.yarnpkg.com/@kiwicom/orbit-components/-/orbit-components-0.117.0.tgz#90bac6e807cc42446d1671f7b8ccea3a7983aae1"
+  integrity sha512-NRcYvSzI4fYdGLfxpkzy9Taq3lKXzmmbGoEkTsDgVtKiX6Z2/hC5Ktfzp4tK1epeJWQfk4uLcm//cX6pfMRmdA==
+  dependencies:
+    "@adeira/js" "^1.3.0"
+    "@kiwicom/orbit-design-tokens" "^0.13.9"
+    body-scroll-lock "^4.0.0-beta.0"
+
 "@lerna/add@4.0.0":
   version "4.0.0"
   resolved "https://registry.yarnpkg.com/@lerna/add/-/add-4.0.0.tgz#c36f57d132502a57b9e7058d1548b7a565ef183f"
```

packages/orbit-components/src/Itinerary/ItineraryPart/ItineraryPartDetail/index.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+const StyledExpandable = styled.div``;
+
+const ItineraryPartDetail = ({ duration, summary, children }: Props) => {
+  const { expanded, setExpanded, hasStatus } = usePart();
+  const { calculatedWidth } = useWidth();
+  const [{ height }, ref] = useBoundingRect({ height: 0 });
+
+  const slideID = React.useMemo(() => randomID("slideID"), []);
```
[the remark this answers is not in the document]
mainframev: is for the `Slide` component, we use it everywhere, i guess your question should be `why ariaLabelledBy prop is missing?` :D

packages/orbit-components/src/Itinerary/ItinerarySegment/index.js
```diff
@@ -0,0 +1,81 @@
...
+    if (onClick) onClick(ev);
+    setOpened(!opened);
+  };
+
+  return (
+    <StyledWrapper
+      spaceAfter={spaceAfter}
+      data-test={dataTest}
+      tabIndex={0}
+      onKeyDown={ev => handleKeyDown(ev, setOpened(!opened))}
```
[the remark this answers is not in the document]
mainframev: For such a case, I'd like to left for now both keys. We can discuss it later

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.js
```diff
@@ -0,0 +1,165 @@
...
+              <StyledSummary>{summary}</StyledSummary>
+            </Stack>
+            {opened ? <ChevronUp /> : <ChevronDown />}
+          </Stack>
+        </StyledInnerWrapper>
+        <Slide maxHeight={height} expanded={opened} id={slideID} ariaLabelledBy={labelID}>
+          <StyledExpandable ref={ref} onClick={setOpened}>
+            <StyledExpandableContent offset={calculatedWidth}>
+              {content &&
+                content.map(({ heading, items }) => {
```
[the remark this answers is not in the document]
mainframev: I thought the same about it but could not find a better solution. At the same time, it does not seem bad in that particular case, especially that guys on FE will map the response from the backend and that prop can easily allow them to pass it to the component. 🤔

packages/orbit-components/src/Itinerary/ItinerarySegment/context.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import * as React from "react";
+
+import type { Context, ProviderProps } from "./context";
+
+export const ItinerarySegmentContext: React.Context<Context> = React.createContext({
```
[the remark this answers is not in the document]
mainframev: i did not do that because of the flow:
`Cannot build a typed interface for this module. You should annotate the exports of this module with types. Cannot determine the type of this call expression. Please provide an annotation, e.g., by adding a type cast around this expression.Flow(signature-verification-failure)
`

silvenon (commented):
Regarding toggling when pressing a key, I marked that one as unresolved because _any_ key toggles it. You mentioned "both keys", which ones did you mean? Btw, if you're going to pick spacebar at the end, make sure to prevent default, I think that prevents space from scrolling down. 🤔

mainframev:
> Regarding toggling when pressing a key, I marked that one as unresolved because _any_ key toggles it. You mentioned "both keys", which ones did you mean? Btw, if you're going to pick spacebar at the end, make sure to prevent default, I think that prevents space from scrolling down. 🤔

I meant `space` and `enter`. That's why our `util` `onKeyDown` does 🤔

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.js
```diff
@@ -126,22 +125,24 @@ const ItinerarySegmentDetail = ({ duration, summary, content }: Props): React.No
...
+        <Slide maxHeight={height} expanded={opened} id={slideID} ariaLabelledBy={slideID}>
+          <StyledExpandable ref={ref} onClick={toggleOpened}>
             <StyledExpandableContent offset={calculatedWidth}>
               {content &&
-                content.map(({ heading, items }) => {
+                content.map(({ heading, items }, idx) => {
                   return (
-                    <>
+                    // eslint-disable-next-line react/no-array-index-key
+                    <React.Fragment key={idx}>
```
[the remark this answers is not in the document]
mainframev: let's wait then for #3055, we can use here the function from render prop component `UID` ;)

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  /** The time of ItineraryPartPlace */
+  readonly time: string;
+  /** The city of ItineraryPartPlace */
+  readonly city: string;
+  /** The place of ItineraryPartPlace */
+  readonly station: string;
+  /** ItineraryPartPlace which status is HiddenCity */
+  readonly hidden?: boolean;
+  /** Changes the icon of current ItineraryPartPlace, if there is some important information for user */
+  readonly warning?: boolean;
```
silvenon, line 20: Orange doesn't seem to be the only possibility, I see the blue one as well:

Screen Shot 2021-11-11 at 22 10 13

Instead of adding a flag for whether it's warning or not, I recommend adding a `type` prop instead to allow for multiple types.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  hidden,
+  minWidth = 60,
+  warning,
+}: Props): React.Node => {
+  const { calculatedWidth, setWidths } = useWidth();
+  const ref = React.useRef<HTMLDivElement | null>(null);
+
+  React.useEffect(() => {
+    if (ref.current) setWidths(prev => (ref.current ? [...prev, ref.current.clientWidth] : prev));
+  }, [ref, setWidths]);
```
silvenon, line 49: ESLint shouldn't be complaining about leaving out `ref`, it knows that it's a ref object because it's being defined in the same file.
```suggestion
  }, [setWidths]);
```

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  station,
+  hidden,
+  minWidth = 60,
+  warning,
+}: Props): React.Node => {
+  const { calculatedWidth, setWidths } = useWidth();
+  const ref = React.useRef<HTMLDivElement | null>(null);
+
+  React.useEffect(() => {
+    if (ref.current) setWidths(prev => (ref.current ? [...prev, ref.current.clientWidth] : prev));
```
silvenon, line 48: Not necessary considering that you're checking inside.
```suggestion
    setWidths(prev => (ref.current ? [...prev, ref.current.clientWidth] : prev));
```

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  icon,
+  time,
+  city,
+  station,
+  hidden,
+  minWidth = 60,
+  warning,
+}: Props): React.Node => {
+  const { calculatedWidth, setWidths } = useWidth();
+  const ref = React.useRef<HTMLDivElement | null>(null);
```
silvenon, line 45: Use distinctive names.
```suggestion
  const dateRef = React.useRef<HTMLDivElement | null>(null);
```

packages/orbit-components/src/Itinerary/ItinerarySegment/context.jsx
```diff
@@ -0,0 +1,33 @@
+// @flow
+import * as React from "react";
+
+import type { Context } from "./context";
+import typeof { ItinerarySegmentProvider as ItinerarySegmentProviderType } from "./context";
+
+export const ItinerarySegmentContext: React.Context<Context> = React.createContext({
```
silvenon, line 7:
```suggestion
export const ItinerarySegmentContext = React.createContext<Context>({
```

packages/orbit-components/src/Itinerary/ItinerarySegment/context.jsx
```diff
@@ -0,0 +1,33 @@
...
+  toggleOpened: () => {},
+  index: 0,
+  count: 0,
+  last: false,
+});
+
+export const ItinerarySegmentProvider: ItinerarySegmentProviderType = ({
+  children,
+  ...ctx
+}): React.Node => {
```
silvenon, line 21: No need for return type because it's already a part of `ItinerarySegmentProviderType`.
```suggestion
}) => {
```

packages/orbit-components/src/Itinerary/ItinerarySegment/context.jsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <ItinerarySegmentContext.Provider
+      value={{
+        ...ctx,
+      }}
+    >
+      {children}
+    </ItinerarySegmentContext.Provider>
```
silvenon, line 29: What is the benefit of `ItinerarySegmentProvider` over using `ItinerarySegmentContext.Provider` directly? Syntax of passing props rather than an object?

packages/orbit-components/src/Itinerary/README.MD
```diff
@@ -0,0 +1,111 @@
...
+To implement the Itinerary component into your project you'll need to add the import:
+
+```jsx
+import Itinerary, {
+  ItinerarySegment,
+  ItineraryStatus,
+  ItineraryBadgeList,
+  ItinerarySegmentStop,
+  ItinerarySegmentDetail,
+} from "@kiwicom/orbit-components";
```
silvenon, line 12:
```suggestion
} from "@kiwicom/orbit-components/lib/Itinerary";
```

packages/orbit-components/src/Itinerary/context.jsx
```diff
@@ -0,0 +1,33 @@
...
+  setWidths: () => {},
+  setCalculatedWidth: () => {},
+  calculatedWidth: 0,
+});
+
+export const ItineraryProvider: ItineraryProviderType = ({ children }): React.Node => {
+  const [widths, setWidths] = React.useState([]);
+  const [calculatedWidth, setCalculatedWidth] = React.useState(0);
+
+  React.useLayoutEffect(() => {
```
silvenon, line 17: Unless you're reading layout from the DOM you shouldn't need `useLayoutEffect`.
```suggestion
  React.useEffect(() => {
```

packages/orbit-components/src/Itinerary/context.jsx
```diff
@@ -0,0 +1,33 @@
...
+    setCalculatedWidth(Math.max(...widths));
+  }, [widths, setCalculatedWidth]);
```
silvenon, line 19:
```suggestion
  }, [widths]);
```

silvenon (changes requested):
This is what I found so far. In the next review I'm going to spend more time comparing the component to the design to make sure I didn't miss anything.

packages/orbit-components/src/Itinerary/ItineraryStatus/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+const ItineraryStatus = ({
+  type,
+  label,
+  spaceAfter = "medium",
+  children,
+  offset = 0,
+}: Props): React.Node => {
+  return (
+    <StyledWrapper $type={type} spaceAfter={spaceAfter} tabIndex={0}>
+      <StyledStatusHeader $type={type}>
```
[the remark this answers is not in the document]
mainframev: are you sure about that? did you check the correct tab in figma? In the `Latest` it looks different

packages/orbit-components/src/Itinerary/ItinerarySegment/ItineraryIcon.jsx
```diff
@@ -0,0 +1,83 @@
...
+  theme: defaultTheme,
+};
+
+const Icon = ({ warning, isDetails, icon }) => {
+  if (icon) return icon;
+  if (warning) return <AlertCircle size="small" color="warning" />;
+  if (isDetails) return <AirplaneDown size="small" />;
+
+  return <Circle size="small" color="secondary" />;
+};
```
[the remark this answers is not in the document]
mainframev: I don't like that way with `lets`, in some side projects I have even `fp/no-let` rule, personal preference. Not sure why you think it's overkill, even a small thing can be a component, and I [prefer that way], to split into more small things, easier to read, reuse, etc (
![https://i.ytimg.com/vi/LaiN63o_BxA/maxresdefault.jpg]).

packages/orbit-components/src/Itinerary/ItineraryStatus/index.jsx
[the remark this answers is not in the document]
silvenon: You're right, I was looking at the wrong design, the one linked in the Jira ticket. I updated the Jira description to link to the latest design.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItineraryIcon.jsx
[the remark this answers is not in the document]
silvenon: Ok 👌 There's no objective upside to it, more of a preference.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItineraryIcon.jsx
```diff
@@ -0,0 +1,83 @@
...
+  justify-content: center;
+  z-index: 1;
+
+  ${({ index, last, count, isNextHidden, theme }) => css`
+    ${index > 0 &&
+    count > 0 &&
+    css`
+      &:before {
+        top: -9px;
+        border: 1px solid ${theme.orbit.paletteCloudNormalHover};
```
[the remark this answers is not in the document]
silvenon: I was looking at the wrong design, the latest one doesn't have the border.

packages/orbit-components/src/primitives/BadgePrimitive/index.jsx
```diff
@@ -73,8 +94,11 @@ const BadgePrimitive = (props: Props): React.Node => {
       ariaLabel={ariaLabel}
       borderColor={borderColor}
     >
+      {carriers && <CarrierLogo carriers={carriers} rounded size="medium" />}
```
silvenon, line 97: The spacing of the CarrierLogo doesn't look quite right. I used a red background so that you can see the problem.

Screen Shot 2021-11-15 at 18 19 12

silvenon (changes requested):
I found several more.

packages/orbit-components/src/Itinerary/index.js.flow
```diff
@@ -14,6 +14,7 @@ import typeof ItinerarySegmentDetailComponent from "./ItinerarySegment/Itinerary
 import type { Globals } from "../common/common.js.flow";
 import type { spaceAfter } from "../common/getSpacingToken";
 
+export type Status = "warning" | "critical" | "info" | "success";
 export type Props = {|
   ...Globals,
   ...spaceAfter,
```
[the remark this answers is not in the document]
mainframev: I don't think we should, i found that prop sometimes really useful.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+  React.useEffect(() => {
+    setWidths(prev => (dateRef.current ? [...prev, dateRef.current.clientWidth] : prev));
+  }, [setWidths]);
+
+  return (
+    <StyledWrapper $hidden={hidden}>
+      <Stack inline align="center" spacing="small">
+        <StyledDate minWidth={calculatedWidth || minWidth} ref={dateRef}>
+          <Stack flex direction="column" spacing="XXXSmall" align="end">
+            <Text weight="bold">{time}</Text>
```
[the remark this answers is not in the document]
mainframev: Let's wait until design review, ok? Because even in the design colors do not match, for instance, we have different `inkLight` color than in figma, because of [this](https://jira.kiwi.com/browse/ORBIT-1820) and i think it should be updated in the design.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.jsx
```diff
@@ -161,7 +175,9 @@ const ItinerarySegmentDetail = ({ duration, summary, content, icon }: Props): Re
                         {items.map(({ icon: itemIcon, name, value }, id) => {
                           return (
                             <Stack flex grow={false} align="center" key={randomId(String(id))}>
-                              <StyledIcon>{itemIcon}</StyledIcon>
+                              <StyledIcon isFirst={id === 0} isLast={items.length - 1 === id}>
```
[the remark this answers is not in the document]
mainframev: :D

silvenon (approved):
LGTM! Let's see what @majlan says.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.jsx
```diff
@@ -0,0 +1,209 @@
...
+        expanded={opened}
+        id={randomId("slide")}
+        ariaLabelledBy={randomId("slide")}
+      >
+        <StyledExpandable ref={slideRef} onClick={toggleOpened}>
+          <StyledExpandableContent $offset={calculatedWidth}>
+            {content &&
+              content.map(({ title, items }, idx) => {
+                return (
+                  <React.Fragment key={randomId(String(idx))}>
```
silvenon, line 178: Scratch what I said about avoiding index. In this case it would be really hard to improve render performance, so we should think about this separately as we reevaluate our components as a separate task. But you can use `idx` directly, you don't need to use `randomId`, `key` only needs to be unique for siblings.

silvenon (changes requested):

silvenon (approved):

PR #2411, opened 2020-10-27
refactor(Modal): use hooks

vborszeki:

- [x] New Components has both `.flow` and `d.ts` files and are exprorted in `index.js.flow` and `index.d.ts`<br/><br/><br/>
<url>LiveURL: https://orbit-refactor-modal.surge.sh</url>


packages/orbit-components/src/Modal/index.js
```diff
@@ -323,360 +322,310 @@ ModalWrapperContent.defaultProps = {
...
+
+  const modalContent: {| current: any | HTMLElement |} = React.useRef();
+  const modalBody: {| current: any | HTMLElement |} = React.useRef();
+  const modalID: string = randomID("modalID");
+
+  const prevChildren = usePrevious(children);
+
+  /* eslint-disable no-use-before-define */
+  React.useEffect(() => {
+    const timer: TimeoutID = setTimeout(() => {
```
mainframev, line 357: we have a hook `useStateWithTimeout` :) might be useful in this case, but it looks good already, so it's up to you

mainframev (approved):
Also, the whole component is jumping on `Removable section` tab in storybook

vborszeki:
hey @mainframev, 
blinking removable section is fixed 🛠

packages/orbit-components/src/hooks/usePrevious/index.js
```diff
@@ -0,0 +1,14 @@
...
+import * as React from "react";
+
+const usePrevious = (value: any) => {
+  const ref = React.useRef();
+
+  React.useEffect(() => {
+    ref.current = value;
+  });
+
+  return ref.current;
```
mainframev, line 11: Seems like redundant hook, it used only once and probably it will be better just to use inside component without moving that logic to custom hook :)

silvenon (changes requested):

packages/orbit-components/src/hooks/usePrevious/index.js
vborszeki: That's true, it's not used elsewhere yet. On the other hand, as the Modal component is already pretty huge, I tried to extract as much irrelevant logic as possible, to make the file a bit more readable and easier to skim through.

Does it make sense?

If you prefer it colocated with the component, I can put it there of course.

packages/orbit-components/src/Modal/__tests__/index.test.js
```diff
@@ -79,49 +82,46 @@ describe("Large Modal", () => {
...
   });
 
-  it("close button should have dataTest", () => {
-    expect(closeButton.prop("dataTest")).toBe(CLOSE_BUTTON_DATA_TEST);
-    closeButton.simulate("click");
-    expect(onClose).toHaveBeenCalled();
+  it("should call callback function when clicking on close button", () => {
+    render(LargeModal);
+
+    fireEvent.click(screen.getByTestId(CLOSE_BUTTON_DATA_TEST));
```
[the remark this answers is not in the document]
vborszeki: fixed it, I've just noticed it has been also added to the [testing conventions](https://github.com/kiwicom/orbit/blob/master/.github/contribution/testing-conventions.md) recently 🙃

packages/orbit-components/src/Modal/__tests__/index.test.js
```diff
@@ -79,49 +82,46 @@ describe("Large Modal", () => {
...
-      <ModalHeader title={title} />
-      <ModalFooter>
-        <Button fullWidth>Continue to Payment</Button>
-      </ModalFooter>
-    </Modal>,
-  );
-  it("should match snapshot", () => {
-    const instance = component.instance();
-    expect(instance.state.hasModalSection).toBe(false);
+    expect(container.querySelector("section")).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
vborszeki: Querying by role doesn't work as there is no `section` [WAI-ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles).

> But we won't know whether it does unless we try the opposite in the previous where the section should exist

Totally agree. I just found it more idiomatic in the first test to get the element by test id, but later forgot to rewrite. 😿

packages/orbit-components/src/Modal/__tests__/index.test.js
```diff
@@ -55,15 +56,17 @@ describe("Large Modal", () => {
...
-  const closeButton = component.find("Modal__CloseContainer").find("ModalCloseButton");
 
   it("should have passed props", () => {
+    const component = shallow(LargeModal);
+
+    const modalWrapper = component.find("Modal__ModalWrapper");
+    const modalHeader = component.find("ModalHeader");
+    const modalSection = component.find("ModalSection");
+    const modalFooter = component.find("ModalFooter");
+
```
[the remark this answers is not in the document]
vborszeki: Good idea! I was also thinking about that.

I have restructured the tests a bit, having a separate `it` block for checking the presence of passed props that are not affecting the visual appearance and another one to check if passed props are rendered correctly.

packages/orbit-components/src/Modal/__tests__/index.test.js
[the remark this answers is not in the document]
silvenon: My bad, the default role is `region`, read [this reference](https://www.w3.org/TR/html-aria/).

Screen Shot 2020-11-05 at 09 40 27

packages/orbit-components/src/Modal/__tests__/index.test.js
```diff
@@ -14,114 +15,101 @@ import Wallet from "../../icons/Wallet";
 import Button from "../../Button";
 import ChevronLeft from "../../icons/ChevronLeft";
 
-// for testing purposes only
 const Div = (props: any) => <div>{props.children}</div>;
+const onClose = jest.fn();
```
silvenon, line 19: `onClose` needs to be defined in every `it` block that's using it, otherwise calls are adding up between tests because `onClose` is always the same mock.

But with the way that you're creating a React element outside of tests this isn't possible, so I suggest the following: instead of rendering one modal that contains everything necessary to test anything, render a separate modal in every test that contains only what's necessary for that particular test, see [InputField test](https://github.com/kiwicom/orbit/blob/4325df7e71ede36fdfe76ab8082fb533728a2732/packages/orbit-components/src/InputField/__tests__/index.test.js) as an example.

silvenon (changes requested):
One more thing and it's great!

packages/orbit-components/src/Modal/__tests__/index.test.js
sampi: or to keep things clean, a `beforeEach()` could be defined with `onClose.mockClear()`

silvenon (approved):

PR #2419, opened 2020-10-27
docs: update testing conventions

silvenon:
Considering that we're rewriting tests, I think it would be useful to establish some conventions, not only for others but for ourselves. I wrote down some basic stuff, do you have some thoughts and ideas for more?<br/><br/><br/><url>LiveURL: https://orbit-docs-testing-conventions.surge.sh</url>


silvenon:
I'm not too crazy about the "should" convention because it doesn't always seem to be applicable, maybe it's unnecessary.

mainframev (approved):
nice! good idea :)

PR #2426, opened 2020-10-29
feat(Button): div behave like button, a11y

mainframev:
[jira ticket](https://jira.kiwi.com/browse/ORBIT-681)

If button rendered as div, it should react onEnter and onSpace clicks.

Hope i did not miss something, @vepor can you check? <br/><br/><br/><url>LiveURL: https://orbit-fix-button-a11y.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/primitives/ButtonPrimitive/__tests__/index.test.js
+++ b/packages/orbit-components/src/primitives/ButtonPrimitive/__tests__/index.test.js
@@ -1,6 +1,6 @@
 // @flow
 import * as React from "react";
-import { render, screen } from "@testing-library/react";
+import { render, screen, fireEvent } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import ButtonPrimitive from "../index";
@@ -42,11 +42,13 @@ describe("ButtonPrimitive", () => {
     userEvent.click(button);
     expect(onClick).toHaveBeenCalled();
   });
+
   describe("disabled", () => {
     it("default", () => {
       render(<ButtonPrimitive disabled>Lorem ipsum</ButtonPrimitive>);
       expect(screen.getByRole("button")).toBeDisabled();
     });
+
     it("with href", () => {
       const onClick = jest.fn();
       render(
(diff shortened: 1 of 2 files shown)


silvenon (changes requested):
Could you add a test for this? This is what I would (try to) do:

1. focus the button with [`userEvent.tab`](https://github.com/testing-library/user-event#tabshift-focustrap)
2. simulate pressing <kbd>⮐</kbd> using [`fireEvent.keyDown`](https://testing-library.com/docs/dom-testing-library/api-events#fireeventeventname)
3. assert that `onClick` has been called

Regarding `fireEvent`, note that `@testing-library/react` exports everything from `@testing-library/dom`, including `fireEvent`, so you don't have to install `@testing-library/dom`.

mainframev:
@silvenon i'll add, thx :)

packages/orbit-components/src/primitives/ButtonPrimitive/index.js
```diff
@@ -50,6 +51,12 @@ export const StyledButtonPrimitive = styled(
       const Component = isButtonWithHref ? "a" : asComponent;
       const buttonType = submit ? "submit" : "button";
 
+      const handleKeyDown = () => {
```
vepor, line 54: Having the `handleKeyDown` won't work – it can't be a function in this particular case. It doesn't have any parameter, even though it should have – to pass down the event.
```js
const handleKeyDown = ev => {
  if (asComponent === "div") {
    return onKeyDown(onClick)(ev);
  }
  return null;
};
```

Better would be to have have it as:
```js
const handleKeyDown = asComponent === "div" ? onKeyDown(onClick) : null;
```

vepor (changes requested):
And as @silvenon wrote, please add the tests because as you can see, without them it can be hard to say if it really works or not 😅

packages/orbit-components/src/primitives/ButtonPrimitive/__tests__/index.test.js
```diff
@@ -75,4 +79,21 @@ describe("ButtonPrimitive", () => {
...
+    render(
+      <ButtonPrimitive dataTest={dataTest} tabIndex={0} asComponent="div">
+        {children}
+      </ButtonPrimitive>,
+    );
+
+    const button = screen.getByTestId(dataTest);
+
+    userEvent.tab();
+    fireEvent.keyDown(button);
```
silvenon, line 96: To check whether `onClick` has been called when <kbd>⮐</kbd> is pressed you need to pass the event object as the second argument (and the `onClick` mock to `ButtonPrimitive`).

silvenon (changes requested):
Not sure if this was ready for another review.

packages/orbit-components/src/primitives/ButtonPrimitive/__tests__/index.test.js
mainframev: yes, you are right, thx, fixed

silvenon (approved):

PR #2438, opened 2020-10-30
fix(Tooltip): do not propagate the onClick event on mobile when stopPropagate is set to true

mattapet:

- [ ] New Components has both `.flow` and `d.ts` files and are exprorted in `index.js.flow` and `index.d.ts`

### DIFF
--- a/packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
+++ b/packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
@@ -49,7 +49,7 @@ const MobileDialog = ({
       setshown(false);
       setRenderWithTimeout(false);
     },
-    [setRenderWithTimeout, setshown],
+    [setRenderWithTimeout, setshown, stopPropagation],
   );
 
   if (!enabled) return children;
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
```diff
@@ -41,10 +41,16 @@ const MobileDialog = ({
...
-  }, [setRenderWithTimeout, setshown]);
+  const handleOutMobile = useCallback(
+    ev => {
+      if (stopPropagation) {
+        ev.stopPropagation();
+      }
+      setshown(false);
+      setRenderWithTimeout(false);
+    },
+    [setRenderWithTimeout, setshown],
```
silvenon, line 52: ESLint correctly warns about this:
```suggestion
    [setRenderWithTimeout, setshown, stopPropagation],
```

silvenon (changes requested):
Thanks so much!

packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
mattapet: Thanks for catching this! ESLint was not cooperating with me well today so this slipped 😅

silvenon (approved):

PR #2451, opened 2020-11-02
improve: add codemod for spacing of Stack

vepor:
Adding codemod to be able to shift usages of Stack to updated spacings that are coming in #2457.

```bash
yarn global add jscodeshift
jscodeshift --transform https://raw.githubusercontent.com/kiwicom/orbit/master/packages/orbit-components/transforms/Stacj-spacing.js [path] --parser=babel|babylon|flow|ts|tsx
```


silvenon (approved):
Nice!

mainframev (approved):

PR #2459, opened 2020-11-03
fix(ButtonPrimitive): Button hover state when clicked

filipdanisko:
When Button is clicked, chrome keeps it in focus state thus hiding any future hover states.


vepor:
Thanks @filipdanisko!

Does it work also for different browsers? Just to know it that was tester or not 🙂

filipdanisko:
@vepor 
Other browsers act differently, FF and Safari (these I tested) does not keep the focus after click. Unsure about IE 🤔

mainframev:
solved in #3242

PR #2460, opened 2020-11-04
fix(InputFile): use button tag for select button

silvenon:
This improves accessibility. For some reason this was being set as `div` from the beginning (#231), but I don't see a reason for it (anymore).
<br/><br/><br/><url>LiveURL: https://orbit-fix-input-file-select-button-a11y.surge.sh</url>


silvenon:
I realized why: `<button>` doesn't trigger the file input when clicked. I'll rewrite this PR to add that comment instead.

PR #2461, opened 2020-11-04
fix(InputFile): forward onFocus event handler

silvenon:
Focusing the input was previously triggering the `onChange` event handler, now `onFocus` is properly being forwarded. This should not be a problem, but is technically breaking for anyone that was relying on this behavior.

This behavior was there from the beginning, I assume it's a mistake because the component takes `onFocus` which it doesn't use.
<br/><br/><br/><url>LiveURL: https://orbit-fix-input-file-onFocus.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/InputFile/index.js
+++ b/packages/orbit-components/src/InputFile/index.js
@@ -112,6 +112,7 @@ const InputFile = React.forwardRef<Props, HTMLInputElement>(
       error,
       help,
       onChange,
+      onFocus,
       onBlur,
       allowedFileTypes,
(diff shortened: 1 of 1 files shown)


vepor (approved):
It was a mistake, probably. However I don't believe that someone is relying on this behavior – it doesn't make sense and in 99.9 % of cases you will handle just the `onChange`.

silvenon:
Ok, I edited the commit message to remove the breaking change status, because it's a bug fix.

vepor (approved):

PR #2469, opened 2020-11-05
fix(Inline): add inner wrapper for negative margin

silvenon:
This way if someone applies margin to `Inline`, e.g. if it's nested
inside a component like `Stack`, its negative margin won't be
overridden.
<br/><br/><br/><url>LiveURL: https://orbit-fix-inline-inner.surge.sh</url>


silvenon:
See https://skypicker.slack.com/archives/GSGN9BN6Q/p1604584396005800.

mainframev (approved):

PR #2471, opened 2020-11-06
docs: update Inline props table

CollierCZ:
To match RFC and type definitions.
<br/><br/><br/><url>LiveURL: https://orbit-docs-inline.surge.sh</url>


mainframev:
that change is already in my PR, closing

PR #2475, opened 2020-11-06
fix: Unify asComponents prop name

mainframev:


packages/orbit-components/src/Heading/Heading.stories.js
```diff
@@ -167,7 +167,7 @@ storiesOf("Heading", module)
       return (
         <Heading
           id={id}
-          as={as}
+          asComponentСomponent={asComponent}
```
silvenon, line 170:
```suggestion
          asСomponent={asComponent}
```

silvenon (changes requested):
Thanks, but we should have first agreed on which one to pick, I actually prefer `as` because it's shorter to type and it matches what `styled-components` uses. Also, considering that this is a breaking change, it's important to edit your commit message with `git commit --amend` and add `BREAKING CHANGE: <description>` to the body, for example:

```
BREAKING CHANGE: in some components we were using polymorphic prop `as`, in other ones `asComponent`, this change ensures that we're always using the same, so that people don't have to look it up.
```

We should definitely hold off this change because it's a breaking change without spectacular benefit, I would actually release it as part of `1.0.0`. Also, it would be good to write a codemod for this as well (I can do that part).

silvenon:
It's perfectly fine to keep this open, we'll mark it as a draft and pick it up again when we're ready.

silvenon:
Now that we know that there is a reason for this, we should close this because it would be an unnecessary breaking change.

PR #2477, opened 2020-11-06
chore: add new Loki references

silvenon:
If `yarn test:loki` creates new references (untracked files), the workflow should be like this:

1. first review and commit untracked files, those are new references
2. review differences and run `yarn loki:approve`

If new references aren't commmitted first, `yarn loki:approve` will delete them and only update the existing ones.
<br/><br/><br/><url>LiveURL: https://orbit-chore-new-loki-references.surge.sh</url>


vepor:
@silvenon could we put this info somewhere? e.g. contribution guidelines? I think that the right order can be forgotten easily 🙂

silvenon:
Done #2485

vepor (approved):

PR #2483, opened 2020-11-09
fix(Accordion): Accordion sticky footer story

mainframev:
Not the problem of Accordion component, but how story example with Stack is written

[ORBIT-1461](https://jira.kiwi.com/browse/ORBIT-1461)

<br/><br/><br/><url>LiveURL: https://orbit-fix-accordion-footer.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Accordion/AccordionSection/components/SectionFooter.js
+++ b/packages/orbit-components/src/Accordion/AccordionSection/components/SectionFooter.js
@@ -11,6 +11,7 @@ const Wrapper = styled.div`
   background-color: ${({ theme }) => theme.orbit.paletteWhite};
   position: sticky;
   bottom: 0;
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/Accordion/AccordionSection/components/SectionFooter.js
```diff
@@ -11,6 +11,7 @@ const Wrapper = styled.div`
   background-color: ${({ theme }) => theme.orbit.paletteWhite};
   position: sticky;
   bottom: 0;
+  box-sizing: border-box;
```
silvenon, line 14: How does this help? I tried turning it off and on and it doesn't seem to make a difference.

silvenon (approved):
It works, just one question.

PR #2486, opened 2020-11-09
docs: add examples for Wizard and Timeline

CollierCZ:
<br/><br/><br/><url>LiveURL: https://orbit-examples-add-progress.surge.sh</url>


packages/orbit-components/src/Wizard/__examples__/DEFAULT.js
```diff
@@ -0,0 +1,82 @@
...
+            text: "Select seating.",
+          });
+          break;
+        default:
+          setHeadingAndText({
+            heading: "Overview and payment",
+            text: "Enter payment details.",
+          });
+      }
+    };
```
mainframev, line 52: i would limit number of clicks, it should not go to infinity :) 

now if im clicking on `next step` it goes beyond number of steps, it should return to the first step i think

CollierCZ: 😱 ♾️

CollierCZ: Infinity is bad in this case. Because it's supposed to show default behavior, I'd go for hiding the button on the last step rather than cycling back (so it doesn't seem like that's the default way to get to the start).

PR #2491, opened 2020-11-10
improve: fix Stack spacing codemod for TypeScript

silvenon:


silvenon:
I love how easy it was to define test for Flow vs TypeScript!

PR #2492, opened 2020-11-10
ListChoice tests

mainframev:
- Should not be squashed (as we agreed, i suppose it's time me to start use that our agreement) 
- 2 commits (switch to testing-library and focus test)<br/><br/><br/><url>LiveURL: https://orbit-feat-testing-library-listchoice.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/ListChoice/__tests__/index.test.js
+++ b/packages/orbit-components/src/ListChoice/__tests__/index.test.js
@@ -1,119 +1,77 @@
 // @flow
 import * as React from "react";
-import { shallow } from "enzyme";
-import { render } from "react-dom";
-import { nullthrows } from "@adeira/js";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import ListChoice from "../index";
 import Accommodation from "../../icons/Accommodation";
 
-let screen;
-const body = nullthrows(document.body, "Expected document to have a body");
-
-beforeEach(() => {
-  screen = document.createElement("div");
-  body.appendChild(screen);
-});
-
-afterEach(() => {
-  body.removeChild(screen);
-});
-
 describe("ListChoice", () => {
   const title = "Choice Title";
-  const description = "Further description";
   const dataTest = "test";
-  const selected = true;
-  const selectable = true;
   const onClick = jest.fn();
 
-  const component = shallow(
-    <ListChoice
-      title={title}
-      description={description}
-      selectable={selectable}
-      selected={selected}
-      disabled
-      icon={<Accommodation />}
-      onClick={onClick}
-      dataTest={dataTest}
-    />,
-  );
-  const content = component.find("ListChoice__StyledListChoiceContent");
+  render(<ListChoice title={title} dataTest={dataTest} />);
+  it("should have data-test", () => {
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+  });
+
   it("should render icon", () => {
-    expect(component.find("Accommodation").exists()).toBe(true);
-  });
-  it("should render data-test", () => {
-    expect(component.render().prop("data-test")).toBe(dataTest);
-  });
-  it("should render title and description", () => {
-    expect(content.find("Heading").children().text()).toBe(title);
-    expect(content.find("Text").children().text()).toBe(description);
-  });
-  it("should render checkbox with checked false", () => {
-    expect(component.find("Checkbox").prop("checked")).toBe(true);
-  });
-  it("should NOT execute onClick method", () => {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/ListChoice/__tests__/index.test.js
```diff
@@ -1,119 +1,78 @@
...
-      onClick={onClick}
-      dataTest={dataTest}
-    />,
-  );
-  const content = component.find("ListChoice__StyledListChoiceContent");
-  it("should render icon", () => {
-    expect(component.find("Accommodation").exists()).toBe(true);
+  render(<ListChoice title={title} dataTest={dataTest} />);
+  it("should have data-test", () => {
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
```
silvenon, line 16:
```suggestion
  it("should have data-test", () => {
    render(<ListChoice title={title} dataTest={dataTest} />);
    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
```

packages/orbit-components/src/ListChoice/__tests__/index.test.js
```diff
@@ -1,119 +1,78 @@
...
-        dataTest={dataTest}
-      />,
-      screen,
-    );
-
-    const listChoice = screen.querySelector(`[data-test="${dataTest}"]`);
-    expect(listChoice?.getAttribute("role")).toBe("checkbox");
-    expect(listChoice?.getAttribute("aria-checked")).toBe("false");
+    render(<ListChoice title={title} selectable dataTest={dataTest} />);
+    expect(screen.getByTestId(dataTest)).toHaveAttribute("role", "checkbox");
```
silvenon, line 63:
```suggestion
    expect(screen.getByRole("checkbox")).toBeInTheDocument();
```

packages/orbit-components/src/ListChoice/__tests__/index.test.js
```diff
@@ -1,119 +1,78 @@
...
-        icon={<Accommodation />}
-        onClick={onClick}
-        dataTest={dataTest}
-      />,
-      screen,
-    );
+  it("should have focus", () => {
+    render(<ListChoice title={title} selectable dataTest={dataTest} />);
+    userEvent.tab();
+    expect(screen.getByTestId(dataTest)).toHaveAttribute("tabindex", "0");
```
silvenon, line 69: The `tabindex` check is unnecessary because it's an implementation detail which might change, you're only interested in whether the element has focus, like the test says.

packages/orbit-components/src/ListChoice/__tests__/index.test.js
```diff
@@ -1,119 +1,78 @@
...
+    expect(screen.getByTestId(dataTest)).toHaveFocus();
+  });
 
-    const listChoice = screen.querySelector(`[data-test="${dataTest}"]`);
-    expect(listChoice?.getAttribute("role")).toBe("checkbox");
-    expect(listChoice?.getAttribute("aria-checked")).toBe("true");
+  it("should not have focus", () => {
+    render(<ListChoice title={title} disabled selectable dataTest={dataTest} />);
+    userEvent.tab();
+    expect(screen.getByTestId(dataTest)).toHaveAttribute("tabindex", "-1");
```
silvenon, line 76:
```suggestion
    expect(screen.getByTestId(dataTest)).not.toHaveFocus();
```

silvenon (changes requested):

silvenon:
Regarding squashing, I think it's fine to do it in this case, I'm changing tests as well as part of rewriting. And besides, these commits won't end up in the changelog, so we don't need to put too much effort into separating them.

packages/orbit-components/src/ListChoice/__tests__/index.test.js
mainframev: @silvenon  i did it on purpose, for some reason it takes test multiple times, not sure why, if you can help - i'll be happy)

silvenon: This is because both the container and the nested `Checkbox` have the role `checkbox`. I don't think it's necessary for container to imitate a checkbox with `role="checkbox"` and `aria-checked` considering that there is already a `Checkbox` which has these a11y properties naturally, but I'm not sure what's the right thing to do here, maybe add `aria-hidden="true"` to `Checkbox` but it doesn't support it currently, so leave it for later and think about it.

silvenon (approved):

silvenon (changes requested):
Commits should be `test`, not `feat`, they are only related to tests and shouldn't bump the version.

PR #2493, opened 2020-11-10
feat(Select): added readOnly prop

mainframev:
- readOnly prop for `Select`
- test that ensures that prop is present<br/><br/><br/><url>LiveURL: https://orbit-feat-select-readonly.surge.sh</url>


silvenon (approved):
It would be nice to also make it configurable in the "Playground" story because this way I cannot verify that it works unless I check out the branch locally.

PR #2495, opened 2020-11-10
feat(InputStepper): added readOnly prop

mainframev:
- added `readOnly` prop for `InputStepper`
- test also added<br/><br/><br/><url>LiveURL: https://orbit-feat-inputstepper-readonly.surge.sh</url>


packages/orbit-components/src/InputStepper/__tests__/index.test.js
```diff
@@ -27,12 +28,12 @@ describe("InputSteppe", () => {
       />,
     );
     const input = screen.getByRole("spinbutton", { name: /Label/ });
-    expect(input).toHaveValue(1);
```
silvenon, line 30: So this fails when it's `readOnly`?

silvenon (approved):

PR #2500, opened 2020-11-11
improve: preview changelog on release

mainframev:
Small script for showing commits from previous version till upcoming

I've checked lerna source code and etc, but found it more simple to use `git log <current tag>..HEAD`<br/><br/><br/><url>LiveURL: https://orbit-chore-release-version-prompting.surge.sh</url>


silvenon:
I updated the script to use `conventional-changelog` to preview the changelog in the same format as it's going to be written, this will make it easier to skim. Also, considering that `lerna publish` already is a prompt, I thought it was unnecessary to prompt twice, so I just output changelog before running `lerna publish`.

Screen Shot 2020-11-12 at 15 19 11

vepor (approved):

PR #2501, opened 2020-11-12
fix: flow granular types

mainframev:
[ORBIT-1069](https://jira.kiwi.com/browse/ORBIT-1069)<br/><br/><br/><url>LiveURL: https://orbit-fix-flow-granular.surge.sh</url>


silvenon (commented):
Summary from Slack:

In my ticket I was wrong, `import typeof MyComponent from "@kiwicom/orbit/lib/MyComponent"` does result in the desired type, which is `React.Component<Props>`, so it was a false report 😕

mainframev:
[Closing for now, seems like not needed change. 
](https://github.com/kiwicom/orbit/pull/2501#pullrequestreview-529377837)

PR #2503, opened 2020-11-13
ci: lint only changed files when safe

silvenon:
<br/><br/><br/><url>LiveURL: https://orbit-ci-improve-speed-eslint.surge.sh</url>


silvenon:
`fancy-log` is used for `gulp` output, that's why I'm using it.

silvenon:
There appears to be a problem with `@adeira/monorepo-utils`'s git command on CI, perhaps it's using a different version.

tasks/eslintCI.js
```diff
@@ -0,0 +1,47 @@
...
+  } else {
+    filesToLint = changedFiles.filter(fileName => fileName.match(/\.(jsx?|tsx?|flow)$/));
+  }
+
+  const results = await eslint.lintFiles(filesToLint);
+  const errorResults = ESLint.getErrorResults(results); // --quiet
+
+  if (results.errorCount > 0) {
+    const formatter = await eslint.loadFormatter();
+    console.log(formatter(errorResults));
```
mainframev, line 42: maybe here we can use `log.error(...)` from `fancy-log` as we already imported that module :) Or it does not make sense to use with `eslint.loadFormatter()`? 🤔 

Everything works good :)

mainframev (dismissed):

tasks/eslintCI.js
silvenon: Thanks for the suggestion, but I think it's best to use `fancy-log` for single line outputs, this will span multiple lines.

silvenon:
This works now, please review again because the PR had changed significantly.

mainframev (approved):

PR #2507, opened 2020-11-15
feat: Storybook docs

mainframev:
Added example of storybook docs usage

- left current examples as they are. Only had to change a bit their import, because it had some errors with Docs plugin and i turned it off for examples
- In order to not break current examples api, i created examples in `InputStepper` component and put the docs inside `__docs__` folder

Added that PR in order to show you how it might look, we can play a bit with imports/settings to make it look even better, but the idea is simple, we will have all examples under `Docs` tab in Storybook in every component, it will be much easier for users to interact with components behavior and docs. It also will be possible to run storybook only with docs flag or to export them. 

<br/><br/><br/><url>LiveURL: https://orbit-feat-storybook-controls.surge.sh</url>


mainframev:
Close. We decided to adopt new structure of stories.

PR #2513, opened 2020-11-18
[Card] show loading when card don't have children

straxico:
show loading when don't have children

- [ ] New Components has both `.flow` and `d.ts` files and are exprorted in `index.js.flow` and `index.d.ts`


mainframev:
Card can have only title and description, which should be shown, without loading message. Closing.

PR #2521, opened 2020-11-23
feat: upgrade styled-components to v5

silvenon:
## Why?

- without it we cannot identify type issues like #2520
- by using it ourselves we can confidently raise the version of peer dependency
  - yarn only prints a warning, but npm v7 will fail to install `@kiwicom/orbit-components` in a project that uses `styled-components` v5
- we would have better type checking for development

## Why not?

- this might be a breaking change for everyone using `styled-commponents` v4, but I'm not sure
- 300+ Flow errors that probably need to be fixed manually
  - I was able to automate the rest of the errors with search & replace and a codemod, see `transforms/styled-components-defaultProps.js`
- 500+ failing tests to fix

---

The current set of changes is far as I'm going to push it for now, this PR will be an ongoing effort.


silvenon:
The size change is incorrect because `compressed-size-action` currently counts comments, see preactjs/compressed-size-action#39.

mainframev:
Closing. There is #2865. 
@silvenon thanks for the codemod and etc. It was easier for me to push another PR.

PR #2524, opened 2020-11-23
test(Text): rewrite to Testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-text.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Text/__tests__/index.test.js
+++ b/packages/orbit-components/src/Text/__tests__/index.test.js
@@ -1,38 +1,32 @@
 // @flow
 import * as React from "react";
-import { shallow, mount } from "enzyme";
+import { screen, render } from "@testing-library/react";
 
 import Text from "../index";
-import SPACINGS_AFTER from "../../common/getSpacingToken/consts";
 import { SIZE_OPTIONS, TYPE_OPTIONS } from "../consts";
 import defaultTheme from "../../defaultTheme";
 
 describe("Text", () => {
-  const text = "Children text";
-  const dataTest = "test";
-  const spaceAfter = SPACINGS_AFTER.NORMAL;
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Text/__tests__/index.test.js
```diff
@@ -1,38 +1,50 @@
...
+      </Text>,
+    );
+
+    expect(screen.getByTestId(dataTest)).toHaveAttribute("id", id);
+    expect(screen.getByTestId(dataTest)).toHaveStyle({
+      fontSize: defaultTheme.orbit.fontSizeTextSmall,
+    });
+    expect(screen.getByTestId(dataTest)).toHaveStyle({
+      color: defaultTheme.orbit.colorTextPrimary,
+    });
```
silvenon, line 29:
```suggestion
    expect(screen.getByTestId(dataTest)).toHaveStyle({
      fontSize: defaultTheme.orbit.fontSizeTextSmall,
      color: defaultTheme.orbit.colorTextPrimary,
    });
```

silvenon (changes requested):
When rewriting tests, try to use this opportunity to make them more compact. So instead of having a single assertion per test, e.g.:

```js
it("should have data-test")
it("should have margin-bottom")
it("should have should have some attribute")
it("should have should have some other attribute")
it("should have should have children")
```

you can merge all of those into a single test called e.g. `should have expected DOM output`. Not only do we save space, but the tests are a bit faster because it doesn't have to render and cleanup so much.

Then if a component has a specific behavior we want to test, that can go in a separate test.

The same goes for `Table` tests. We can try to leave tests better than we found them, not only strictly migrate.

silvenon (approved):

PR #2528, opened 2020-11-24
fix: expose getScrollPosition in Modal

silvenon:
Reading scroll position was unintentionally possible when Modal was a class component because people could access Modal's elements through refs and read their `scrollTop`, but they had to duplicate Modal's logic for figuring out which element to read based on the viewport.

These changes implement official API for this called `getScrollPosition` which handles this internally.

Initially I thought of exposing `scrollPosition` value, but as we’ve discussed we would then have to read it all the time, which has performance implications, then I thought of `scrollingElement`, but it would be tedious for people to constantly check whether `scrollingElement.current` exists, and besides we only want to allow them to read scroll position and nothing more, so eventually I came up with `getScrollPosition`, which pairs very nicely with `setScrollPosition`.
<br/><br/><br/><url>LiveURL: https://orbit-fix-modal-scroll-position.surge.sh</url>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,12 @@ module.exports = {
       files: "*.ts?(x)",
       extends: ["./config/eslintTS"],
     },
+    {
+      files: "**/__typetests__/**",
+      rules: {
+        "@typescript-eslint/explicit-module-boundary-types": "off",
+      },
+    },
     // some ESLint rules fail in certain cases, so we're disabling them
     {
       files: ["packages/orbit-components/src/utils/**/*"],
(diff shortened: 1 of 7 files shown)


silvenon:
Pasting background info from Slack just in case:

Initially I thought of exposing `scrollPosition` value, but as we’ve discussed we would then have to read it all the time, which has performance implications, then I thought of `scrollingElement`, but it would be tedious for people to constantly check whether `scrollingElement.current` exists, and besides we only want to allow them to read scroll position and nothing more, so eventually I came up with `getScrollPosition`, which pairs very nicely with `setScrollPosition`.

vepor (approved):

PR #2530, opened 2020-11-25
Add scrollingElementRef prop to Modal and expose modalBody and modalContent as backup

silvenon:
When Modal was a class component all of its properties were exposed, regardless of whether or not that was the intention. It turned out that Search was relying on `modalBody` and `modalContent`, so after we converted Modal to a function component their code broke.

Changes introduced by this PR:

- add `scrollingElementRef` prop as an abstraction so that people don't have to know when scrolling element changes (specifically, media query)
  - now they can avoid replicating this logic when they need to access the element
  - they can also implement cleanup with `React.useEffect`, which AFAIK wasn't possible before
- expose `modalBody`and `modalContent` as backup

This is an example of using `scrollingElementRef` with cleanup:

```jsx
function App() {
  const [scrollingElement, setScrollingElement] = React.useState(null);
  const scrollingElementFn = React.useCallback((node) => {
    setScrollingElement(node);
  }, []);
  const handleScroll = React.useCallback(() => {
    // ...
  }, []);
  React.useEffect(() => {
    if (scrollingElement) {
      scrollingElement.addEventListener("scroll", handleScroll);
    }
    return () => {
      if (scrollingElement) {
        scrollingElement.removeEventListener("scroll", handleScroll);
      }
    };
  }, [scrollingElement, handleScroll]);
  return (
    <Modal scrollingElementRef={scrollingElementFn}>
      content
    </Modal>
  );
}
```

<br/><br/><br/><url>LiveURL: https://orbit-fix-modal-expose-scrolling-element.surge.sh</url>


silvenon:
Doesn't work properly, accessing `modalRef.current.scrollingElement.current` when `isLargeMobile` changes will access the _previous_ scrolling element, not the current one. I need to write an intergration test for this use case.

silvenon:
An example of using `scrollingElementRef` with a cleanup.

```jsx
function App() {
  const [scrollingElement, setScrollingElement] = React.useState(null);
  const scrollingElementFn = React.useCallback((node) => {
    setScrollingElement(node);
  }, []);
  const handleScroll = React.useCallback(() => {
    // ...
  }, []);
  React.useEffect(() => {
    if (scrollingElement) {
      scrollingElement.addEventListener("scroll", handleScroll);
    }
    return () => {
      if (scrollingElement) {
        scrollingElement.removeEventListener("scroll", handleScroll);
      }
    };
  }, [scrollingElement, handleScroll]);
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
      <Modal scrollingElementRef={scrollingElementFn} />
    </div>
  );
}
```

silvenon:
✅ Fixed and updated the PR description.

packages/orbit-components/src/Modal/index.d.ts
```diff
@@ -11,6 +11,7 @@ type Size = "extraSmall" | "small" | "normal" | "large";
 export interface Props extends Common.Global {
   readonly size?: Size;
   readonly children: React.ReactNode;
+  readonly scrollingElementRef?: React.Ref<HTMLElement>;
```
mainframev, line 14: do not forget to add that to `README.md` ;)

mainframev (approved):

mainframev (approved):

PR #2532, opened 2020-11-26
feat(Textarea): add required prop

mainframev:
Issue from our backlog [ORBIT-1043](https://jira.kiwi.com/browse/ORBIT-1043?workflowName=0K%3A+To+Do+%7C+In+Progress+%7C+Blocked+%7C+Done&stepId=3)<br/><br/><br/><url>LiveURL: https://orbit-feat-required-textarea.surge.sh</url>


packages/orbit-components/src/Textarea/Textarea.stories.js
```diff
@@ -121,6 +121,7 @@ storiesOf("Textarea", module)
       const maxLength = number("maxLength", undefined);
       const dataTest = text("dataTest", "test");
       const rows = number("rows", 3);
+      const required = boolean("requred", false);
```
silvenon, line 124:
```suggestion
      const required = boolean("required", false);
```

packages/orbit-components/src/Textarea/README.md
```diff
@@ -34,6 +34,7 @@ Table below contains all types of the props available in Textarea component.
 | resize      | [`enum`](#enum)            | `"vertical"` | The resize option for Textarea.                                                                                                                                |
 | rows        | `number`                   |              | Specifies the height of the text area (in lines).                                                                                                              |
 | readOnly    | `boolean`                  |              | Adds readOnly to html textarea element.                                                                                                                        |
+| required    | `boolean`                  |              | Optional prop, adds required to html textarea                                                                                                                  |
```
silvenon, line 37: This is not consistent with how we usually write, at some point lack of code tags can become confusing. In this case "required" is also a regular word, so in order to help it become perceived as the name of the HTML attribute, we can put it in code tags.

We should generally write nicer, for example "HTML" is an acronym so it should be uppercase, but never mind that now.
```suggestion
| required    | `boolean`                  |              | Optional prop, adds `required` to the `<textarea>` HTML element.                                                                                               |
```

silvenon (changes requested):

silvenon (approved):

PR #2533, opened 2020-11-26
LinkList changes

mainframev:
- fixed `Linklist` a11y issue. [ORBIT-1437](https://jira.kiwi.com/browse/ORBIT-1437)
- rewrote test to testing-library<br/><br/><br/><url>LiveURL: https://orbit-fix-LinkList-a11y.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/LinkList/__tests___/index.test.js
+++ b/packages/orbit-components/src/LinkList/__tests___/index.test.js
@@ -1,22 +1,26 @@
 // @flow
-import { mount } from "enzyme";
 import * as React from "react";
+import { screen, render } from "@testing-library/react";
 
 import LinkList from "..";
 
 describe("#LinkList", () => {
-  const wrapper = mount(
-    <LinkList direction="row" spacing="XXSmall">
-      <div>link</div>
-    </LinkList>,
-  );
+  it("should have expected dom output", () => {
+    render(
+      <LinkList dataTest="kek" spacing="XXLarge">
+        <div>link 1</div>
+        <div>link 2</div>
+        <div>link 3</div>
+      </LinkList>,
+    );
 
-  it("should have children", () => {
-    expect(wrapper.contains("link")).toBe(true);
-  });
-
-  it("should pass props to Stack", () => {
-    expect(wrapper.find("Stack").prop("direction")).toBe("row");
-    expect(wrapper.find("Stack").prop("spacing")).toBe("XXSmall");
+    expect(screen.getByText("link 1")).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/LinkList/__tests___/index.test.js
```diff
@@ -1,22 +1,26 @@
...
-    expect(wrapper.contains("link")).toBe(true);
-  });
-
-  it("should pass props to Stack", () => {
-    expect(wrapper.find("Stack").prop("direction")).toBe("row");
-    expect(wrapper.find("Stack").prop("spacing")).toBe("XXSmall");
+    expect(screen.getByText("link 1")).toBeInTheDocument();
+    expect(screen.getByTestId("kek")).toBeInTheDocument();
+    expect(screen.getAllByRole("listitem")[0]).toHaveStyle({
+      margin: "0px 0px 40px 0px",
```
silvenon, line 20: I would rather avoid making assumptions about all other margins, would this work? I think `toHaveStyle` makes assertions to computed styles, so they don't have to be exactly how we defined them, only the result.
```suggestion
      marginBottom: "40px",
```

packages/orbit-components/src/LinkList/__tests___/index.test.js
```diff
@@ -1,22 +1,26 @@
...
+    });
+    expect(screen.getAllByRole("listitem")[2]).toHaveStyle({
+      margin: "0px",
```
silvenon, line 23:
```suggestion
      marginBottom: "0px",
```

silvenon (changes requested):
Otherwise great! 👍

packages/orbit-components/src/LinkList/__tests___/index.test.js
mainframev: aha, did not know, that it will take it, thx

packages/orbit-components/src/LinkList/index.js
```diff
@@ -25,11 +27,30 @@ StyledLinkList.defaultProps = {
...
+    direction &&
+    String(getDirectionSpacingTemplate(direction)).replace(
+      "__spacing__",
+      getSpacing(props)[spacing],
+    );
+
+  return css`
+    margin: ${margin && rtlSpacing(margin)}!important;
+    &:last-child {
+      margin: 0 !important;
```
silvenon, line 42: I forgot to ask, is using `!important` here necessary?

silvenon (dismissed):

silvenon (approved):

PR #2534, opened 2020-11-26
fix(Hide): add aria-hidden if it's hidden

mainframev:
- added aria-hidden if it's hidden

Issue: [ORBIT-182](https://jira.kiwi.com/browse/ORBIT-182)<br/><br/><br/><url>LiveURL: https://orbit-fix-Hide.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Desktop/__tests__/index.test.js
+++ b/packages/orbit-components/src/Desktop/__tests__/index.test.js
@@ -44,6 +44,7 @@ describe("Desktop", () => {
       }
 
       <div
+        aria-hidden="false"
         class="c0"
       >
         kek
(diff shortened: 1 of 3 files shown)


silvenon:
Setting [aria-hidden])(https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute) to `true` is unnecessary when `display: none;` is being used, that's why frameworks like Bootstrap and Tailwind offer the [`sr-only`](https://tailwindcss.com/docs/screen-readers) utility, which visually hides the element, but not from screen readers.

Screen Shot 2020-11-27 at 11 34 55

mainframev:
Closed. Unnecessary change

PR #2535, opened 2020-11-26
test(List): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-list.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/List/__tests__/index.test.js
+++ b/packages/orbit-components/src/List/__tests__/index.test.js
@@ -1,58 +1,27 @@
 // @flow
 import * as React from "react";
-import { shallow, mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
 
 import List from "../index";
 import ListItem from "../ListItem";
 import Check from "../../icons/Check";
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/List/__tests__/index.test.js
```diff
@@ -1,58 +1,33 @@
...
+    expect(screen.getByRole("list")).toHaveStyle({ marginBottom: "12px" });
   });
 
-  it("should have children", () => {
-    children.forEach(node => {
-      expect(node.type()).toBe(ListItem);
-      expect(node.find("ListItem").render().prop("data-test")).toBe(dataTest);
-      expect(node.render().text()).toBe(content);
-    });
+  it("ListItem should have expected output", () => {
```
silvenon, line 28: You can remove this test because you already assert that `content` exists in the previous test, and `List` doesn't have special logic on outputting children, it just does `{children}`.

silvenon (changes requested):

silvenon (approved):

PR #2536, opened 2020-11-26
test(Loading): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-loading.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Loading/__tests__/index.test.js
+++ b/packages/orbit-components/src/Loading/__tests__/index.test.js
@@ -1,37 +1,38 @@
 // @flow
 import * as React from "react";
-import { shallow } from "enzyme";
+import { render, screen } from "@testing-library/react";
 
 import Loading from "../index";
 import TYPE_OPTIONS from "../consts";
 
 describe("Loading", () => {
-  const dataTest = "test";
-  it("should render type", () => {
-    const type = TYPE_OPTIONS.BOX_LOADER;
-    const component = shallow(<Loading type={type} />);
-    const loader = component.find("Loading__StyledLoader");
-    expect(component.find("Loading__StyledLoader").childAt(0).parent()).toEqual(loader);
-  });
-  it("should render type with text", () => {
+  const { BOX_LOADER, SEARCH_LOADER, INLINE_LOADER } = TYPE_OPTIONS;
+  it("should have expected DOM output", () => {
     const text = "Test test";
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Loading/__tests__/index.test.js
```diff
@@ -1,37 +1,38 @@
...
+        display: "flex",
+        justifyContent: "center",
+      });
+    },
+  );
+
+  it("should have spinner icon", () => {
+    render(<Loading dataTest="kek" type={TYPE_OPTIONS.PAGE_LOADER} />);
+    expect(screen.getByTestId("kek").firstChild).toHaveStyle({ width: "40px", height: "40px" });
+    expect(screen.getByTestId("kek").querySelector("svg")).toHaveAttribute("viewBox", "0 0 40 40");
```
silvenon, line 36: Tests should be as agnostic as possible and not have hard coded values if they are irrelevant. Here I think it's enough to just check whether an `svg` element exists, because it's the only one. If we happen to change the spinner illustration with a different `viewBox`, then the test won't break.
```suggestion
    expect(screen.getByTestId("kek").querySelector("svg")).toBeInTheDocument();
```

packages/orbit-components/src/Loading/consts.js
```diff
@@ -1,10 +1,10 @@
 // @flow
 const TYPE_OPTIONS = {
-  BUTTON_LOADER: "buttonLoader",
-  SEARCH_LOADER: "searchLoader",
   BOX_LOADER: "boxLoader",
-  PAGE_LOADER: "pageLoader",
+  SEARCH_LOADER: "searchLoader",
   INLINE_LOADER: "inlineLoader",
+  BUTTON_LOADER: "buttonLoader",
+  PAGE_LOADER: "pageLoader",
```
silvenon, line 7: It seems like you just reordered these, right? 🤔

silvenon (changes requested):

packages/orbit-components/src/Loading/consts.js
mainframev: yes, because first i did something like ```Object.values(TYPE_OPTIONS).slice(0, 3)```, later rewrite it in a different way but that change remained :)

silvenon (approved):

PR #2540, opened 2020-11-27
fix(Popover): popover bottom position issue

mainframev:
Issue in [plz-orbit](https://skypicker.slack.com/archives/CAMS40F7B/p1606426217336800)

appears only when actions prop is present, on mobile is ok, broken on desktop
<br/><br/><br/><url>LiveURL: https://orbit-fix-Popover.surge.sh</url>


silvenon (approved):
Great!

PR #2541, opened 2020-11-27
docs: create basic setup for new docs

silvenon:
The site is almost empty, this only concerns the basic setup.

- using TypeScript for type checking
- using Tailwind's [Preflight](https://tailwindcss.com/docs/preflight) as base styles
- processing guides with MDX, example available at `/guides/using-component-types`
  - being careful that we maintain same URLs
- syntax highlighting with Prism (I don't know what we're using for current docs)
- processing images using best practices with `gatsby-image`
<br/><br/><br/><url>LiveURL: https://orbit-docs-workspace.surge.sh</url>


vepor (approved):
LGTM!  🚀 👏 

I didn't find any issues locally and it works as described! Good job @silvenon 🙂

docs/src/components/RocketImage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const data: {
+    file: {
+      childImageSharp: {
+        fluid: FluidObject;
+      };
+    };
+  } = useStaticQuery(graphql`
+    query {
+      file(relativePath: { eq: "rocket.jpg" }) {
+        childImageSharp {
```
mainframev, line 16: can we add also something like: 
```
srcWebp
srcSetWebp
tracedSVG
```
?)

tsconfig.json
```diff
@@ -3,12 +3,13 @@
...
-    "jsx": "preserve",
+    "jsx": "react",
     "strict": true,
     "noImplicitAny": false
   },
   "include": [
     "packages/orbit-components/**/*",
-    "packages/orbit-design-tokens/**/*"
+    "packages/orbit-design-tokens/**/*",
+    "docs/**/*"
```
mainframev, line 13: maybe we should create completely different `tsconfig.json` for that workspace? inside docs, like `docs/tsconfig.json`, in order to not change our common and we can make it more strict with `noImpicitAny: true` and other rules 🤔

mainframev:
> LGTM! 🚀 👏
> 
> I didn't find any issues locally and it works as described! Good job @silvenon 🙂

Me too. LGTM 💪🏻 I've two questions, but nothing special, so good job 👍🏻

mainframev (approved):

docs/src/components/RocketImage.tsx
silvenon: I'm not a big fan of traced SVG, especially for this purpose, I prefer current blurred placeholder. But yes to WebP! Any image optimizations are welcome, as long as the loading experience is good. 👍

tsconfig.json
silvenon: What kind of changes do you think we might need? And do you think we should create separate configurations for other workspaces as well? If we happen to need different `compilerOptions` for `docs` later, e.g. Preact instead of React, then we'll need to split it and deal with any potential complexity that comes with that, but until then I see no need, unless you disagree.

docs/gatsby-node.js
```diff
@@ -0,0 +1,46 @@
...
+    reporter.panicOnBuild(`Error when querying guides.`);
+    return;
+  }
+  result.data.allMdx.nodes.forEach(guide => {
+    createPage({
+      path: `/guides/${path.basename(guide.slug)}`,
+      component: `${__dirname}/src/templates/Guide.tsx`,
+      context: {
+        title: guide.frontmatter.title,
+        body: guide.body,
```
CollierCZ, line 42: Why pass this through as page context rather than querying for it on a page basis? The pages will eventually be more complex (excerpts, images for social, etc.), so more querying would be done. It generally seems favored to do that in the templates and do less in gatsby-node.

docs/src/components/Code.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const InlineCode = styled.code`
+  padding: 0.25em 0.5em;
+  background: #ddd;
+  border-radius: 0.5em;
+  color: #000;
+  /* monospace typefaces appear a bit larger */
+  font-size: calc(1em - 2px);
+`;
+
+// copied from Prism's theme "Okaidia"
```
CollierCZ, line 12: If you don't mind a different theme, it seems like it'd be easier to offload this onto a package like https://github.com/FormidableLabs/prism-react-renderer

docs/src/guides/developers/using-component-types.mdx
```diff
@@ -1,4 +1,6 @@
-# How to use component types (Typescript | Flow)
+---
+title: Using component types
```
CollierCZ, line 2: This will expand with known fields, such as whatever we decide to replace the current `excerpt` with. I imagine it'd be good to have an image field for metadata purposes as well.

docs/src/mdx-components.tsx
```diff
@@ -0,0 +1,33 @@
...
+  TableCell,
+} from "@kiwicom/orbit-components";
+import { InlineCode, Pre } from "./components/Code";
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const p = Text;
+export const h1 = (props: Props) => <Heading as="h1" type="display" {...props} />;
```
CollierCZ, line 19: I don't think I'd support h1s in Markdown. I'd enforce titles in frontmatter and lint out any h1s. That makes it easier to use the data for other purposes.

docs/src/mdx-components.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const h6 = (props: Props) => <Heading as="h6" type="title5" {...props} />;
+export const table = Table;
+export const thead = TableHead;
+export const tbody = TableBody;
+export const tr = TableRow;
+export const td = TableCell;
+export const th = TableCell;
+export const pre = Pre;
+export const inlineCode = InlineCode;
+export const a = TextLink;
```
CollierCZ, line 33: It'd be good to handle internal links separately from external links. It'd be enough to handle it with an `onClick` event with `navigate` from Gatsby for internal links.

docs/src/components/Prose.tsx
```diff
@@ -0,0 +1,21 @@
+import styled from "styled-components";
+
+const Prose = styled.div`
+  font-family: ${({ theme }) => theme.orbit.fontFamily};
+  font-size: ${({ theme }) => theme.orbit.fontSizeTextNormal};
+  > * + * {
+    margin-top: ${({ theme }) => theme.orbit.spaceSmall};
+  }
+  h1 {
```
CollierCZ, line 9: What's the benefit of handling heading spacing and such here and not in the `mdx-components`? We have `spaceAfter` still, no?

I'm not certain the value of the basic font CSS in this component as opposed to more generally.

docs/gatsby-node.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+    }
+  `);
+  if (result.errors) {
+    reporter.panicOnBuild(`Error when querying guides.`);
+    return;
+  }
+  result.data.allMdx.nodes.forEach(guide => {
+    createPage({
+      path: `/guides/${path.basename(guide.slug)}`,
```
CollierCZ, line 38: This renders, e.g., `{base}/guides/using-component-types`. Shouldn't it be `{base}/developer/using-component-types`? Two things here:
1. I don't see the purpose of "guides" in the url.
2. I think the collection should be a part of the url.

As an aside, I prefer it with the trailing slash. Right now, I think both versions are being generated (because with trailing slash is done automatically), which can appear like duplicated content for SEO purposes (maybe?).

docs/src/components/Code.tsx
silvenon: Wow, this looks great! 😍 I'll definitely switch over.

docs/gatsby-node.js
silvenon: Oh, I think I understand, you mean that I could pass for example a node ID, and use that to retrieve the rest of the data in the template, right? That will definitely be much better 👍

docs/src/guides/developers/using-component-types.mdx
silvenon: Could you elaborate? Are you talking about two separate things? Is there something I should change here?

docs/src/mdx-components.tsx
silvenon: I agree, I'll remove this mapping.

docs/src/mdx-components.tsx
silvenon: Good idea, I can also use Gatsby's `Link` component considering that `TextLink` supports `asComponent` 💪 I should also remember to pass `external` to `TextLink` for external links!

docs/src/components/Prose.tsx
silvenon: My understanding is that the `spaceAfter` prop isn't necessarily going to be around forever. The only use case for it that I can think of is writing documents, which component libraries aren't really good at, they are intended for applications. We'll most probably want to add `margin-top` to headings, which is something that Orbit doesn't (and shouldn't) support, so I created this component to define in one place styles that define relations between elements in a document.

> I'm not certain the value of the basic font CSS in this component as opposed to more generally.

If by "more generally" you mean that I could apply `font-family` and `font-size` to the the entire HTML document, that's probably a better way to go, but I'll see how this goes.

docs/gatsby-node.js
silvenon: > This renders, e.g., `{base}/guides/using-component-types`. Shouldn't it be `{base}/developer/using-component-types`? 

The decision was based solely on the assumption that we want to retain the same URLs, see https://orbit.kiwi/guides/using-component-types/

> As an aside, I prefer it with the trailing slash. Right now, I think both versions are being generated (because with trailing slash is done automatically), which can appear like duplicated content for SEO purposes (maybe?).

Thanks for the note, I'll definitely look into that.

CollierCZ: I think we're going to restructure and I'd prefer the URLs to match the new structure. I'd add [`gatsby-redirect-from`](https://github.com/kremalicious/gatsby-redirect-from) to keep the old URLs functional, but I wouldn't necessarily stick to them for the future.

docs/gatsby-node.js
CollierCZ: Yes. Apparently ID is slightly faster than slug, which is what I'm used to using.

docs/src/guides/developers/using-component-types.mdx
CollierCZ: I think I was just trying to plan for the future. It's also possible to type the frontmatter in Markdown files in Gatsby, so I assumed it'd be good to think about it.

tsconfig.json
mainframev: yes, i meant that probably we will use different compiler options, since it's new workspace, we can make it more stricter etc, not like our main `tsconfig.json` which we use only for type checking of ts declarations. It's just suggestion

docs/src/guides/developers/using-component-types.mdx
silvenon: I still don't understand, sorry 😅 What do you mean by "type the frontmatter"? I think it will be really good to familiarize myself with the current process, how the documents are currently written so that I can replicate the useful parts of the process and discard or improve the rest.

docs/gatsby-node.js
silvenon: Got it! I wasn't aware that this was the plan.

docs/src/guides/developers/using-component-types.mdx
CollierCZ: When there are things like images in frontmatter, it can help to explicitly add that to the Gatsby schema. Like https://gitlab.skypicker.com/kiwi/handbook/-/blob/master/gatsby-node.js#L225

silvenon: Oh, cool! I'll definitely add that. 👍

tsconfig.json
silvenon: Ok, if it turns out we need it, I'll create one.

PR #2542, opened 2020-11-27
refactor(Pagination): useMediaQuery instead of Hide, title

mainframev:
- `useMediaQuery` instead of `Hide`
- added `title` (aria-label) when `hideLables` is true<br/><br/><br/><url>LiveURL: https://orbit-refactor-pagination.surge.sh</url>


silvenon (approved):
LGTM! Just to add for posterity that the point of this is to be able to mock the viewport size for easy testing. Specifically, for #2539.

PR #2549, opened 2020-11-30
test(Select): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-Select.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Select/__tests__/Select.test.js
+++ b/packages/orbit-components/src/Select/__tests__/Select.test.js
@@ -1,81 +1,82 @@
 // @flow
 
 import * as React from "react";
-import { shallow, mount } from "enzyme";
+import { render, screen, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import Select from "../index";
 import SPACINGS_AFTER from "../../common/getSpacingToken/consts";
 
-const mockChange = jest.fn();
-const placeholder = "Default placeholder";
-const dataTest = "test";
-const tabIndex = "-1";
-const name = "name";
-const id = "test-id";
-const objectOptions = [
-  { value: "1", label: "One" },
-  { value: "2", label: "Two" },
-  { value: "3", label: "Three" },
-  { value: "4", label: "Four" },
-  { value: "hidden-five", label: "Hidden Five" },
-  { disabled: true, value: "disabled-six", label: "Disabled Six" },
-];
-const spaceAfter = SPACINGS_AFTER.NORMAL;
-const dataAttrs = {
-  "data-recording-ignore": true,
-};
-
 describe("Select", () => {
-  const component = shallow(
-    <Select
-      id={id}
-      value="1"
-      name={name}
-      placeholder={placeholder}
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/Select/__tests__/Select.test.js
```diff
@@ -1,17 +1,21 @@
...
-const mockChange = jest.fn();
 const placeholder = "Default placeholder";
 const dataTest = "test";
+const label = "label";
 const tabIndex = "-1";
 const name = "name";
 const id = "test-id";
+const onBlur = jest.fn();
+const onFocus = jest.fn();
+const onChange = jest.fn();
```
silvenon, line 18: Move these mocks inside the first `it` block just to be safe, so that it's obvious that they are only being used in the first test. Every test should define their own mocks, and in general be as self-contained as possible.

packages/orbit-components/src/Select/__tests__/Select.test.js
```diff
@@ -26,56 +30,54 @@ const dataAttrs = {
...
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+    expect(select).toHaveAttribute("id", id);
+    expect(select).toHaveAttribute("readonly");
+    expect(select).toHaveAttribute("tabindex", "-1");
+    expect(select).toHaveAttribute("data-state", "ok");
+    expect(select).toHaveAttribute("data-recording-ignore");
+    expect(select).toHaveAttribute("name", name);
+    expect(screen.getByLabelText(label)).toBeInTheDocument();
+    expect(screen.getByText(placeholder)).toBeInTheDocument();
+    expect(document.querySelector("svg")).toBeInTheDocument();
```
silvenon, line 64: No need to assert that the icon exists, there's no logic involved in its visibility (no condition), and it's not a critical Select behavior.
```suggestion
```

packages/orbit-components/src/Select/__tests__/Select.test.js
```diff
@@ -26,56 +30,54 @@ const dataAttrs = {
...
-describe(`Required field`, () => {
-  const label = "Label";
-  const component = mount(
-    <Select options={objectOptions} label={label} required onChange={mockChange} />,
-  );
-  it("should render asterisk", () => {
-    expect(component.find("FormLabel__StyledAsterisk").exists()).toBe(true);
+  it("should be disabled", () => {
+    render(<Select options={objectOptions} disabled />);
+    expect(screen.getByRole("combobox")).toHaveAttribute("disabled");
```
silvenon, line 81:
```suggestion
    expect(screen.getByRole("combobox")).toBeDisabled();
```

silvenon (changes requested):
Otherwise very good! 👍

packages/orbit-components/src/Select/__tests__/Select.test.js
mainframev: ah, yes, good catch, not sure why i did it in the same way again

silvenon (approved):

PR #2554, opened 2020-12-01
test(Portal): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-Portal.surge.sh</url>


packages/orbit-components/src/Portal/__tests__/index.test.js
```diff
@@ -1,43 +1,22 @@
...
-    expect(component).toMatchSnapshot();
+    render(
+      <Portal renderInto="portal">
+        <p>Content</p>
+      </Portal>,
+    );
+
+    expect(screen.getByText("Content")).toBeInTheDocument();
+    expect(document.getElementById("portal")).toBeInTheDocument();
   });
```
silvenon, line 21: You need to finish with `document.body?.removeChild(target)` at the end because Testing Library's `cleanup` removes only what was added by `render`.

silvenon (changes requested):

packages/orbit-components/src/Portal/__tests__/index.test.js
mainframev: oh, yes, thanks :)

silvenon (approved):

PR #2557, opened 2020-12-01
test(PricingTable): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-PricingTable.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/PricingTable/__tests__/index.test.js
+++ b/packages/orbit-components/src/PricingTable/__tests__/index.test.js
@@ -1,188 +1,108 @@
 // @flow
 
 import * as React from "react";
-import { mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
+import useMediaQuery from "../../hooks/useMediaQuery";
 import Button from "../../Button";
 import Badge from "../../Badge";
 import PricingTableItem from "../PricingTableItem";
 import PricingTable from "../index";
-import PricingTableContext from "../PricingTableContext";
 
-const onClick = jest.fn();
-const onClick2 = jest.fn();
-const dataTest = "test";
-const name = "name";
-const price = "$749";
-const actions = <Button>Buy</Button>;
-const content = <ul>List</ul>;
-const image = <img alt="" />;
-const badge = <Badge>Badge</Badge>;
+jest.mock("../../hooks/useMediaQuery", () => jest.fn());
+const useMediaQueryMock: JestMockFn<[], $Shape<$Call<typeof useMediaQuery>>> = useMediaQuery;
 
-jest.mock("../../hooks/useMediaQuery", () => {
-  return () => {
-    return {
-      isTablet: false,
-    };
-  };
+afterEach(() => {
+  useMediaQueryMock.mockRestore();
 });
 
-describe("PricingTableItem mobile", () => {
-  const component = mount(
-    <PricingTableContext.Provider value={{ basis: 0 }}>
-      <PricingTableItem
-        dataTest={dataTest}
-        name={name}
-        price={price}
-        featureIcon={image}
-        badge={badge}
-        action={actions}
-        active
-        compact
-        onClick={onClick}
-      >
-        {content}
-      </PricingTableItem>
-    </PricingTableContext.Provider>,
-  );
+describe("PricingTableItem", () => {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/PricingTable/__tests__/index.test.js
```diff
@@ -1,188 +1,104 @@
...
+    const dataTest = "test";
+    const name = "name";
+    const price = "$749";
+    const actions = <Button>Buy</Button>;
+    const content = "content";
+    const image = <img alt="" />;
+    const badge = <Badge>Badge</Badge>;
+
+    render(
+      <PricingTableContext.Provider value={{ basis: 0 }}>
```
silvenon, line 36: `PricingTableContext` is an implementation detail that should not be tested directly, only indirectly via `PricingTable`. `PricingTableItem` should not be tested in isolation.

packages/orbit-components/src/PricingTable/__tests__/index.test.js
```diff
@@ -1,188 +1,104 @@
...
+      .join();
+
+    const boxShadowActive = trimBoxShadow.concat(
+      `,inset 0 0 0 2px ${theme.orbit.paletteBlueNormal}`,
+    );
+
+    expect(screen.getByTestId("test")).toBeInTheDocument();
+    expect(screen.getByText("kek")).toBeInTheDocument();
+    expect(screen.getByTestId("first")).toHaveStyleRule("box-shadow", boxShadowActive);
+    expect(screen.getByTestId("second")).toHaveStyleRule("box-shadow", trimBoxShadow);
```
silvenon, line 102: Box shadow is eye candy that could change later, breaking our test. A broken test should generally indicate a broken functionality, so tests should be written to be resilient. Unit tests should test only critical features, styles are mostly for visual regression tests.

silvenon (changes requested):
By taking a look at PricingTable component, I think the following functionalities need to be tested:

- stuff like content, price etc. should be visible in the DOM, and event handler should be called, you're already doing all of that correctly
- `PricingTableItem` with `active` should have the radio input checked
- on sizes lower than `desktop` _only_ the contents of active item should be visible
- props like `desktopRadio` and `mobileDescription` aren't being tested, see if there are other missing ones that you can test

I think that's it. These don't all have to be separate tests, decide for yourself how it's best to split this.

Btw, the `activeElement` prop of `PricingTable` is somewhat weird because you have to manually match it to the `active` prop of `PricingTableItem`, instead of `PricingTable` being able to figure that out automatically considering that it already loops through children. 🤷

packages/orbit-components/src/PricingTable/__tests__/index.test.js
```diff
@@ -1,188 +1,108 @@
...
-  it("should have price", () => {
-    expect(pricingTableItem.find("Text").at(1).children().text()).toBe(price);
-  });
-  it("should have featureIcon", () => {
-    expect(pricingTableItem.find("img").exists()).toBe(true);
-  });
-  it("should have badge", () => {
-    expect(pricingTableItem.find("PricingTableItem__StyledBadgeWrapperContent").exists()).toBe(
-      true,
+describe("PricingTableItem", () => {
```
silvenon, line 20: I don't think grouping tests into `PricingTableItem` and `PricingTable` makes sense anymore because your testing features of both components in both groups, so you can merge them together.

silvenon (changes requested):

packages/orbit-components/src/PricingTable/__tests__/index.test.js
```diff
@@ -1,188 +1,108 @@
...
+  it("should have expected DOM output", () => {
+    useMediaQueryMock.mockImplementation(() => ({ isTablet: false }));
```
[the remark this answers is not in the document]
mainframev: sorry, i did not understand what you mean, it checks with `mobileDescription` prop and that needs to be separated from the first one and also be lower than `isTablet`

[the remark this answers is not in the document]
silvenon: This test passes even if `isTablet` is `true` because the component only checks for `isDesktop`, which is `undefined` in this test, which is falsy so the test passes, but not because of the reason that this mock implies.

silvenon (approved):

PR #2560, opened 2020-12-02
test(SkipNavigation): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-SkipNavigation.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/SkipNavigation/__tests__/index.test.js
+++ b/packages/orbit-components/src/SkipNavigation/__tests__/index.test.js
@@ -1,28 +1,44 @@
 // @flow
 import * as React from "react";
-import { shallow } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import SkipNavigation from "..";
 
 describe("SkipNavigation", () => {
-  const actions = [
-    {
-      link: "https://www.kiwi.com/cz/pages/content/terms",
-      name: "Got to terms and condition",
-    },
-    {
-      name: "Add baggage",
-    },
-    {
-      name: "Reguest refund",
-    },
-  ];
-  const feedbackUrl = "https://www.kiwi.com/en/";
+  it("should have expected DOM output", () => {
+    const actions = [
+      {
+        link: "https://www.kiwi.com/cz/pages/content/terms",
+        name: "Got to terms and condition",
+      },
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/SkipNavigation/index.js
```diff
@@ -23,12 +23,14 @@ const StyledNavigation = styled.div`
...
+      clip: rect(0 0 0 0);
+      height: 1px;
+      margin: -1px;
       overflow: hidden;
+      padding: 0;
       position: absolute;
-      pointer-events: none;
-      opacity: 0;
-      width: 0;
+      width: 1px;
```
silvenon, line 33: Considering that you're changing this, could you also add `white-space: nowrap` just in case? Because it's included in `sr-only` utilities.

packages/orbit-components/src/SkipNavigation/__tests__/index.test.js
```diff
@@ -1,28 +1,46 @@
...
+          {
+            name: "Add baggage",
+          },
+        ]}
+        feedbackUrl="https://www.kiwi.com/en/"
+      />,
+    );
 
-  const component = shallow(<SkipNavigation actions={actions} feedbackUrl={feedbackUrl} />);
+    const wrapper = document.querySelectorAll("div")[1];
```
silvenon, line 40: I think this element is also `container` that's being returned from `render`, could you access it that way instead?

```js
const { container } = render(
```

silvenon (changes requested):

packages/orbit-components/src/SkipNavigation/__tests__/index.test.js
mainframev: i tried that and it did not work

packages/orbit-components/src/SkipNavigation/index.js
mainframev: i'll add it, but i used [this information]( https://accessibility.18f.gov/hidden-content/), so it's seems correct

silvenon: My bad, your source looks good 👍 Either way the `white-space` rule won't hurt, so it's fine to leave it.

silvenon (approved):

PR #2563, opened 2020-12-03
test(Stack): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-stack-to-testing-lib.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Stack/__tests__/index.test.js
+++ b/packages/orbit-components/src/Stack/__tests__/index.test.js
@@ -1,480 +1,167 @@
 // @flow
 import * as React from "react";
-import { mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
 
 import Stack from "../index";
-import defaultTheme from "../../defaultTheme";
+import theme from "../../defaultTheme";
 import InputField from "../../InputField";
 import Button from "../../Button";
-import { ALIGNS, DIRECTIONS, JUSTIFY, SPACINGS } from "../consts";
-import SPACINGS_AFTER from "../../common/getSpacingToken/consts";
-import { getBreakpointWidth } from "../../utils/mediaQuery";
+import { DIRECTIONS, SPACINGS } from "../consts";
 import { QUERIES } from "../../utils/mediaQuery/consts";
+import getSpacing from "../helpers/getSpacing";
 
-const defaultValues = {
-  align: "start",
-  basis: undefined,
-  direction: "column",
-  grow: true,
-  inline: false,
-  justify: "start",
-  shrink: false,
-  spaceAfter: undefined,
-  spacing: SPACINGS.MEDIUM,
-  wrap: false,
-};
+describe("Stack", () => {
+  it("should have expected DOM output", () => {
+    const dataTest = "test";
 
-describe("Default Stack", () => {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Stack/__tests__/index.test.js
```diff
@@ -1,480 +1,163 @@
...
-      media: getBreakpointWidth(QUERIES.DESKTOP, defaultTheme),
-    });
-    expect(stack).toHaveStyleRule("flex-basis", `calc(25% - ${defaultTheme.orbit.spaceLarge})`, {
-      media: getBreakpointWidth(QUERIES.DESKTOP, defaultTheme),
-    });
-    expect(stack).toHaveStyleRule("margin", "0 0 32px 0!important", {
-      media: getBreakpointWidth(QUERIES.DESKTOP, defaultTheme),
-      modifier: "& > *",
+
+  it.each(Object.entries(SPACINGS))("should change spacing", (name, spacing) => {
```
silvenon, line 119:
```suggestion
  it.each(Object.entries(SPACINGS))("should change spacing %s", (name, spacing) => {
```

packages/orbit-components/src/Stack/__tests__/index.test.js
```diff
@@ -1,480 +1,163 @@
...
+        <div>kek</div>
+        <div>bur</div>
+      </Stack>,
+    );
+
+    const getValue = (val: "start" | "end" | "between" | "center" | "around") => {
+      if (val === "around" || val === "between") return `space-${val}`;
+      if (val === "center") return "center";
+      return `flex-${val}`;
+    };
```
silvenon, line 97: To avoid logic like this, you can pass pairs of prop value and what result you expected to `it.each`:

```js
it.each([
  ["start", "flex-start"],
  ["end", "flex-end"],
  ["between", "space-between"],
  ["center", "center"],
  ["around", "space-around"],
])("should change justify %s", (justify, expected) => {
  // ...
})
```

packages/orbit-components/src/Stack/__tests__/index.test.js
```diff
@@ -1,480 +1,163 @@
...
+
+    const expected = {
+      mediumMobile: `${theme.orbit.spaceXXSmall}!important`,
+      largeMobile: `${theme.orbit.spaceXSmall}!important`,
+      tablet: `${theme.orbit.spaceSmall}!important`,
+      desktop: `${theme.orbit.spaceMedium}!important`,
+      largeDesktop: `${theme.orbit.spaceLarge}!important`,
+    };
+
+    expect(screen.getByTestId(dataTest)).toHaveStyle({ marginRight: expected[query] });
```
silvenon, line 161: The use of `it.each` here suggests different JSX for each iteration, which is not true (except for `dataTest`), so it looks like you can just do a `forEach` loop for these assertions instead?

silvenon (changes requested):
Testing Library puts us at slight disadvantage here because we're no longer able to assert particular media queries, but considering that we're unit testing every piece of `getViewportFlexStyles` (except for `getViewportFlexStyles` itself) it's not a huge loss.

silvenon (approved):

PR #2564, opened 2020-12-03
SkipLink changes

mainframev:
- rewrite a11y, buttonLabel replaced with `aria-label` on `nav` element, i think it's better approach 🤔 
- changed `sr-only` styles a bit
- rewrote to testing library<br/><br/><br/><url>LiveURL: https://orbit-test-SkipLink.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/SkipLink/README.md
+++ b/packages/orbit-components/src/SkipLink/README.md
@@ -13,7 +13,7 @@ After adding import into your project you can use it simply like:
   links={[
     {
       href: "#terms",
-      name: "Got to terms and condition",
+      name: "Go to terms and conditions",
     },
     {
       name: "Reguest refund",
@@ -23,14 +23,20 @@ After adding import into your project you can use it simply like:
 />
 ```
 
+Screen reader will read the content in the following order:
+
+- content of the link
+- aria-label of nav
+- navigation
+
 ## Props
 
 The table below contains all types of props available in the SkipLink component.
 
-| Name      | Type                | Description                            |
-| :-------- | :------------------ | :------------------------------------- |
-| label     | `string`            | Description for screen readers.        |
-| **links** | [`links[]`](#links) | An array specifying links to point to. |
+| Name        | Type                | Description                            |
+| :---------- | :------------------ | :------------------------------------- |
+| buttonLabel | `string`            | Description for screen readers.        |
+| **links**   | [`links[]`](#links) | An array specifying links to point to. |
 
 ## links
 
--- a/packages/orbit-components/src/SkipLink/README.md
+++ b/packages/orbit-components/src/SkipLink/README.md
@@ -27,10 +27,10 @@ After adding import into your project you can use it simply like:
 
 The table below contains all types of props available in the SkipLink component.
 
-| Name        | Type                | Description                            |
-| :---------- | :------------------ | :------------------------------------- |
-| buttonLabel | `string`            | Description for screen readers.        |
-| **links**   | [`links[]`](#links) | An array specifying links to point to. |
+| Name      | Type                | Description                            |
+| :-------- | :------------------ | :------------------------------------- |
+| label     | `string`            | Description for screen readers.        |
+| **links** | [`links[]`](#links) | An array specifying links to point to. |
 
 ## links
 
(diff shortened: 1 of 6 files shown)


packages/orbit-components/src/SkipLink/__tests__/index.test.js
```diff
@@ -1,29 +1,30 @@
...
-  const component = shallow(<SkipLink links={links} buttonLabel={buttonLabel} />);
-
-  it("Should contain description block", () => {
-    expect(component.find("SkipLink__StyledLabel").exists()).toEqual(true);
-  });
-
-  it("Should have 2 Links", () => {
-    expect(component.find("SkipLink__StyledLink").at(1).exists()).toEqual(true);
+    render(<SkipLink links={links} label="https://www.kiwi.com/en/" />);
+    expect(screen.getByRole("link")).toHaveAttribute("href", links[0].href);
```
silvenon, line 23: The prior test weren't conforming to our testing conventions regarding [hardcoding values](https://github.com/kiwicom/orbit/blob/6c25d955f2cbca59010f859bcda4233c1918c128/.github/contribution/testing-conventions.md#pass-literal-props-directly), but I didn't request changes because they were used in a way where type checking would catch mistakes. However, I think accessing array elements and properties like this is leaves some chance of making a mistake, so please either hardcode values:
```suggestion
    expect(screen.getByRole("link")).toHaveAttribute("href", "Go to terms and conditions");
```
or save those values into variables:
```suggestion
    expect(screen.getByRole("link")).toHaveAttribute("href", termsAndConditions);
```

packages/orbit-components/src/SkipLink/__tests__/index.test.js
```diff
@@ -1,29 +1,30 @@
...
-      href: "#link",
-    },
-  ];
-  const buttonLabel = "https://www.kiwi.com/en/";
+  it("should have expected DOM output", () => {
+    const onClick = jest.fn();
+    const links = [
+      {
+        href: "https://www.kiwi.com/cz/pages/content/terms",
+        name: "Got to terms and condition",
```
silvenon, line 14:
```suggestion
        name: "Go to terms and conditions",
```

silvenon (changes requested):
Copied from Slack for transparency: I like the new prop name much better, but please edit the commit message to include `BREAKING CHANGE: (...)` in the body so that it's emphasized in the changelog.

packages/orbit-components/src/SkipLink/README.md
```diff
@@ -27,10 +27,10 @@ After adding import into your project you can use it simply like:
 
 The table below contains all types of props available in the SkipLink component.
 
-| Name        | Type                | Description                            |
-| :---------- | :------------------ | :------------------------------------- |
-| buttonLabel | `string`            | Description for screen readers.        |
-| **links**   | [`links[]`](#links) | An array specifying links to point to. |
+| Name      | Type                | Description                            |
+| :-------- | :------------------ | :------------------------------------- |
+| label     | `string`            | Description for screen readers.        |
```
vepor, line 32: Tried it with screen reader.
The screen reader output changed. With applying the `label` to `aria-label` of `nav`. It will be read as follows:
- content of the link
- `aria-label` of `nav`
- navigation

Therefore it would make sense to document it or change the description of `label` so it's aligned with actual behaviour.

vepor (changes requested):
Besides small thing, LGTM 🙂

packages/orbit-components/src/SkipLink/index.js
```diff
@@ -42,31 +42,17 @@ StyledLink.defaultProps = {
...
+    <StyledNavigation aria-label={label}>
       {links &&
         links.map(({ href, name, onClick }, index) => {
           return (
             <StyledLink
               key={encodeURIComponent(name + index)}
               href={href}
-              tabIndex={href ? "" : "0"}
-              role={href ? "" : "Button"}
+              tabIndex={onClick && "0"}
```
silvenon, line 54: Aren't anchor elements focusable by default? What does this do?

mainframev: it's not for anchor, it's for button, when `onClick` is provided, there is change to `role="Button"` and it will not be focusable without `tabIndex="0"`

silvenon: Oh, I see!

silvenon (approved):

PR #2565, opened 2020-12-04
test(Slider): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-Slider.surge.sh</url>


silvenon (approved):
I'd split assertions for event handlers into a separate test called `should trigger event handlers` because they are outside of the scope of what I would consider to be "DOM output", they are related to behavior. But it's cool either way, and can be merged.

And omg, this component is _huge_! 😅 it's a great thing you reduced the size of this test. 👌

PR #2567, opened 2020-12-05
chore: add storybook-backgrounds

mainframev:
- Recently we had this question in [plz-orbit](https://skypicker.slack.com/archives/CAMS40F7B/p1607078251385500). So i think it would be nice to add that addon for such cases, when users need to check how component looks like on different bgs.

<br/><br/><br/><url>LiveURL: https://orbit-chore-storybook-bgs.surge.sh</url>

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@lerna/project": "^3.21.0",
     "@octokit/rest": "^15.13.1",
     "@storybook/addon-actions": "^6.0.28",
+    "@storybook/addon-backgrounds": "^6.1.10",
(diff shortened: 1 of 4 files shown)


silvenon (commented):
Nice! Is it possible to also automatically flip base text color to white? So that the text above is visible, too.

mainframev:
> Nice! Is it possible to also automatically flip base text color to white? So that the text above is visible, too.

sure it's possible, did it, check it :)

silvenon (approved):
Perfect!

silvenon (approved):

PR #2570, opened 2020-12-07
docs: extend schema

mainframev:
Just change and extened our schema a bit<br/><br/><br/><url>LiveURL: https://orbit-feat-graphql-schema.surge.sh</url>

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -44,13 +44,30 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
     reporter.panicOnBuild(`Error when querying guides.`);
     return;
   }
-  result.data.allMdx.nodes.forEach(guide => {
-    const slug = omitNumbers(guide.slug);
+  result.data.allMdx.nodes.forEach(({ id, slug }) => {
+    const cuttedSlug = omitNumbers(slug);
 
     createPage({
-      path: `/${slug}`,
+      path: `/${cuttedSlug}`,
       component: `${__dirname}/src/templates/Doc.tsx`,
-      context: { id: guide.id },
+      context: { id },
     });
   });
+};
+
+exports.createSchemaCustomization = ({ actions }) => {
+  actions.createTypes(
+    `
+    type Mdx implements Node {
+      frontmatter: MdxFrontmatter!
+    }
+
+    type MdxFrontmatter {
+      title: String!
+      date: Date @dateformat
+      type: String
+      excerpt: String
+    }
+  `,
+  );
 };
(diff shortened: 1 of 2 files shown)


mainframev:
oh, i have to fix it, because it recieves html, not mdx

docs/gatsby-node.js
```diff
@@ -17,14 +17,22 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 exports.createPages = async ({ graphql, actions, reporter }) => {
   const { createPage } = actions;
   const result = await graphql(`
-    query AllGuides {
+    query GuidesQuery {
       allMdx(filter: { fields: { collection: { eq: "guides" } } }) {
         nodes {
+          id
+          body
```
CollierCZ, line 24: It's not necessary to query for `body` when creating the node. I'd take it out.

docs/gatsby-node.js
```diff
@@ -17,14 +17,22 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
...
+          slug
           frontmatter {
             title
+            date
+            image {
+              id
+              absolutePath
+              name
+              size
+            }
```
CollierCZ, line 34: I'm not sure that these fields are necessary in page context either. They can be queried in the pages themselves.

docs/gatsby-node.js
```diff
@@ -33,14 +41,33 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+        ...frontmatter,
       },
     });
   });
 };
+
+exports.createSchemaCustomization = ({ actions }) => {
+  actions.createTypes(
+    `
+    type Mdx implements Node @infer {
```
CollierCZ, line 62: I think if you want to infer types, you don't need to say so explicitly.

Also, I think it might be better to not infer types. At the very least, I would think it'd be good not infer them for the frontmatter and only include things where we have explicitly typed the fields.

docs/gatsby-node.js
```diff
@@ -33,14 +41,33 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+      frontmatter: MdxFrontmatter!
+    }
+
+    type MdxFrontmatter @infer {
+      title: String!
+      image: File @fileByRelativePath
+      date: Date @dateformat
```
CollierCZ, line 69:
```suggestion
      date: Date @dateformat
      type: String
      excerpt: String
```

Other known fields

CollierCZ (changes requested):

CollierCZ (approved):

PR #2571, opened 2020-12-07
docs: add docs for Getting Started section

CollierCZ:
Added content from Orbit.kiwi for the section Getting Started.

It currently includes a few components not yet implemented:

- FancyLinks (for tiles that link to external resources)
- SocialLinks (for links to social media of contributors)
- GitHubContributors (for a list of contributors)

All have Jira tasks created.

@silvenon, this is how I was thinking of structure, with numbers at the start of file names to indicate position in the navigation. Please let me know if you have better ideas or this should be structured differently. 

I want to get an idea of if it will work before I add a lot of content.<br/><br/><br/><url>LiveURL: https://orbit-docs-add-content.surge.sh</url>


silvenon (commented):
I agree with numbering files, that is a common solution to the ordering problem.

After thinking about the structure more, and examining the navigation structure again, I think this is indeed the best strategy. My only suggestion would be to rename all `index.md` files to `meta.yml`, to eliminate any doubts that this file can hold anything other than metadata. If you and @mainframev agree, we can start implementing this in Gatsby.

mainframev:
@CollierCZ
agree with numbers, we can also sort by title, something like

something like
```
query guidesQuery {
allGuides(sort: {fields: frontmatter___title, order: ASC}) {
   title,
   id,
...etc
}
```

mainframev:
There is probably some misunderstanding. I thought that `guides` section in `docs` should represent `Guides` section of `Orbit.kiwi`, but as in this PR `Getting started` is placed inside `guides`, it seems that that dir is like common for everything. 

I assume we have to move it to separate dirs inside some common folder for it, like `documents`. Structure will be like: 
`src/documents/guides/`
`src/documents/getting-started/`
`src/documents/components`

and change path inside `gatsby-config.js`

silvenon:
`getting-started` directory also needs to be renamed to `01-getting-started`, no?

CollierCZ:
Comments should be addressed. I hope.

silvenon (approved):

PR #2572, opened 2020-12-07
Various improvements in docs setup

silvenon:
This was based on the review comments I received in #2541.

- docs: solve syntax highlighting better
- docs: fetch guide info via id in the template
- chore: install type declarations for @mdx-js/react
- docs: remove MDX mapping for `<h1>`
- docs: get the most out of MDX component mapping
- docs: configure testing
- docs: refine MDX anchor style and handle external
- docs: adjust color of `<code>` in links

One thing I left out is guides URL, we'll do this separately after we agree on a complete content strategy, based on #2571.
<br/><br/><br/><url>LiveURL: https://orbit-docs-review-improvements.surge.sh</url>

### DIFF
--- a/docs/src/mdx-components.tsx
+++ b/docs/src/mdx-components.tsx
@@ -16,7 +16,6 @@ interface Props {
 }
 
 export const p = Text;
-export const h1 = (props: Props) => <Heading as="h1" type="display" {...props} />;
 export const h2 = (props: Props) => <Heading as="h2" type="title1" {...props} />;
 export const h3 = (props: Props) => <Heading as="h3" type="title2" {...props} />;
 export const h4 = (props: Props) => <Heading as="h4" type="title3" {...props} />;
--- a/docs/src/mdx-components.tsx
+++ b/docs/src/mdx-components.tsx
@@ -11,22 +11,70 @@ import {
 } from "@kiwicom/orbit-components";
 import { InlineCode, CodeBlock } from "./components/Code";
 
-interface Props {
-  children: React.ReactNode;
-}
-
-export const p = Text;
-export const h2 = (props: Props) => <Heading as="h2" type="title1" {...props} />;
-export const h3 = (props: Props) => <Heading as="h3" type="title2" {...props} />;
-export const h4 = (props: Props) => <Heading as="h4" type="title3" {...props} />;
-export const h5 = (props: Props) => <Heading as="h5" type="title4" {...props} />;
-export const h6 = (props: Props) => <Heading as="h6" type="title5" {...props} />;
-export const table = Table;
-export const thead = TableHead;
-export const tbody = TableBody;
-export const tr = TableRow;
-export const td = TableCell;
-export const th = TableCell;
+export const p = ({ children }: React.HTMLAttributes<HTMLParagraphElement>) => (
+  <Text>{children}</Text>
+);
+export const h2 = ({ children }: React.HTMLAttributes<HTMLHeadingElement>) => (
+  <Heading as="h2" type="title1">
+    {children}
+  </Heading>
+);
+export const h3 = ({ children }: React.HTMLAttributes<HTMLHeadingElement>) => (
+  <Heading as="h3" type="title2">
+    {children}
+  </Heading>
+);
+export const h4 = ({ children }: React.HTMLAttributes<HTMLHeadingElement>) => (
+  <Heading as="h4" type="title3">
+    {children}
+  </Heading>
+);
+export const h5 = ({ children }: React.HTMLAttributes<HTMLHeadingElement>) => (
+  <Heading as="h5" type="title4">
+    {children}
+  </Heading>
+);
+export const h6 = ({ children }: React.HTMLAttributes<HTMLHeadingElement>) => (
+  <Heading as="h6" type="title5">
+    {children}
+  </Heading>
+);
+export const table = ({ children }: React.TableHTMLAttributes<HTMLTableElement>) => (
+  <Table>{children}</Table>
+);
+export const thead = ({ children }: React.HTMLAttributes<HTMLTableSectionElement>) => (
+  <TableHead>{children}</TableHead>
+);
+export const tbody = ({ children }: React.HTMLAttributes<HTMLTableSectionElement>) => (
+  <TableBody>{children}</TableBody>
+);
+export const tr = ({ children }: React.HTMLAttributes<HTMLTableRowElement>) => (
+  <TableRow>{children}</TableRow>
+);
+export const td = ({
+  children,
+  align,
+  valign,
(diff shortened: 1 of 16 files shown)


docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,100 @@ import {
...
+    align={align === "left" || align === "center" || align === "right" ? align : undefined}
+    verticalAlign={valign}
+  >
+    {children}
+  </TableCell>
+);
+export const th = ({ children, align }: React.ThHTMLAttributes<HTMLTableHeaderCellElement>) => (
+  <TableCell
+    as="th"
+    align={align === "left" || align === "center" || align === "right" ? align : undefined}
```
mainframev, line 72: why undefined? if default value for `TableCell` is `left` 🤔

docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,100 @@ import {
...
+      `}
+    >
+      <TextLink href={href} external>
+        {children}
+      </TextLink>
+    </span>
+  );
+};
+
+function useIsUrlExternal(url: string | void): boolean {
```
mainframev, line 99: i think it's better to use optional `url?: string` instead of `void` here

docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,100 @@ import {
...
+  return React.useMemo(() => {
+    if (!url) return false;
+    // https://stackoverflow.com/a/2911045/1247274
+    // we can't use URL for this because IE doesn't support it
```
mainframev, line 103: should we care about IE at all? i think we agreed, that will not support IE :) 

and it might be outdated, since the link is from 2010 and update there is in 2016 🤔

mainframev (changes requested):

docs/src/mdx-components.tsx
mainframev, line 99: is void make any difference here? i think not, you are still able to call function like that `useIsUrlExternal()` and ts will not complain about that. So probably make it optional will be better :)

mainframev (changes requested):

docs/src/mdx-components.tsx
silvenon: Passing `undefined` allows the default value to kick in, I didn't want to copy defaults here.

docs/src/mdx-components.tsx
silvenon: Considering that I cannot call this hook conditionally, otherwise I'd be breaking the rule of hooks, and that `url` can be `undefined`, I think this is the only way to satisfy type checking. What specifically did you have in mind?

docs/src/mdx-components.tsx
silvenon: Good question, I guess I didn't remember us declaring that we won't care about IE. In that case I'll use `URL`. And as for the solution possibly being outdated, I'm not _too_ worried about it because it appears to work fine, but I'll write tests for it to confirm.

docs/src/mdx-components.tsx
silvenon: Actually I realized I was wrong, fixed! ✅

docs/src/mdx-components.tsx
silvenon: I tried using `URL`, but I realized that I don't really understand how it works, so I sticked to using the anchor element method. I added tests for it to prove that it works well.

mainframev (dismissed):

docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,90 @@ import {
...
+  >
+    {children}
+  </TableCell>
+);
+export const code = CodeBlock;
 export const inlineCode = InlineCode;
-export const a = TextLink;
+export const a = ({ children, href }: React.AnchorHTMLAttributes<HTMLAnchorElement>) => {
+  const isExternal = useIsUrlExternal(href);
+  if (!href) return <a href={href}>{children}</a>;
```
vepor, line 82: If `href` evals to `false`, should we really to pass it to `<a href={href} />`? 🤔

vepor (commented):
Just a question, otherwise LGTM 🚀

docs/src/mdx-components.tsx
silvenon: `href` can be only `string` or `undefined`, so passing it to `href` wouldn't do any harm. But perhaps I should do this instead?

```tsx
export const a = ({ children, href, ...props }: React.AnchorHTMLAttributes<HTMLAnchorElement>) => {
  const isExternal = useIsUrlExternal(href);
  if (!href) return <a {...props}>{children}</a>;
```

so that it's a fallback entirely. (No sense in passing `href` because it's `undefined`.)

silvenon: @vepor I believe I fixed this part now by avoiding using `<a>` and `<Link>` entirely, I realized that I can simply use Gatsby's `navigate` if the link is local. The scenario where `href` is `undefined` is irrelevant and doesn't need any special handling.

vepor: OK, nice! 🙂

docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,96 @@ import {
...
+    >
+      <TextLink
+        href={href}
+        external={isExternal}
+        onClick={event => {
+          if (isExternal || !href) return;
+          event.preventDefault();
+          navigate(href);
+        }}
+      >
```
CollierCZ, line 99:
```suggestion
      <TextLink
        href={href}
        external={isExternal}
        iconRight={<NewWindow ariaLabel="Opens in new window" />}
        onClick={event => {
          if (isExternal || !href) return;
          event.preventDefault();
          navigate(href);
        }}
      >
```

I'd add this for accessibility/useability

docs/src/mdx-components.tsx
```diff
@@ -9,25 +9,96 @@ import {
...
+      css={css`
+        /* TextLink's line-height affects nested elements like <code> */
+        a {
+          line-height: normal;
+        }
+      `}
+    >
+      <TextLink
+        href={href}
+        external={isExternal}
```
CollierCZ, line 93:
```suggestion
        external={isExternal}
        iconRight={<NewWindow ariaLabel="Opens in new window" />}
```

I'd add this for accessibility/usability.

CollierCZ: Just noticed you changed it so it's not only external, so probably only with a check for that.

silvenon: Done. (I updated my branch, that's why all the new commits.)

mainframev:
rebase it with current master and we can merge it i think :)

mainframev (dismissed):

silvenon:
@mainframev rebased.

mainframev (approved):

PR #2573, opened 2020-12-08
improve(Docs): improve folder structure

mainframev:
- changed folder structure to `documents/<orbit-kiwi-section>/`
<br/><br/><br/><url>LiveURL: https://orbit-improve-folder-structure.surge.sh</url>


silvenon (commented):
Looks good, let's just wait until we agree exactly on how to proceed.

docs/gatsby-node.js
```diff
@@ -10,7 +8,7 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
   createNodeField({
     node,
     name: "collection",
-    value: parent.sourceInstanceName,
+    value: parent.relativeDirectory.split("/")[0],
```
CollierCZ, line 11: I'm not sure of the point of this change? 

It might help if I were more certain of the point of the collection field in general that @silvenon added. Are we planning on sourcing different MDX? When is the filtering planned and why is a regex by file path more awkward? Right now, no filtering is necessary that I can see.

docs/gatsby-config.js
```diff
@@ -7,8 +7,8 @@ module.exports = {
     {
       resolve: "gatsby-source-filesystem",
       options: {
-        name: "guides",
-        path: `${__dirname}/src/guides`,
+        name: "documents",
+        path: `${__dirname}/src/documents`,
```
CollierCZ, line 11: I'd strongly prefer `docs` to `documents`.

PR #2574, opened 2020-12-08
feat(docs): Navigation bar, Layout

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-feat-docs-navigation.surge.sh</url>


mainframev:
@silvenon could you look on it? It's still have to be polished a bit, but can be reviewed now already.
There are some things i am not sure about:

- What search we will have, search page + search input and dropdown list
- Probably we will need different mediaqueries, than we have in orbit currently.

docs/src/components/SearchInput/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </StyledInnerWrapper>
+        {dropdown && results.length > 0 && opened && (
+          <StyledDropdown role="listbox" aria-label="search-dropdown" ref={dropdownRef}>
+            {results.map(({ title, excerpt, id, slug }) => (
+              <StyledDropdownItem key={id} tabIndex={-1}>
+                <Link to={slug}>{title}</Link>
+                <p>{excerpt}</p>
+              </StyledDropdownItem>
+            ))}
+          </StyledDropdown>
```
silvenon, line 115: I think it’s best to use [Downshift](https://www.downshift-js.com), or more specifically, [`useCombobox`](https://www.downshift-js.com/use-combobox). That way you get all the functionality and a11y.

docs/src/components/SearchInput/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+        )}
+      </StyledWrapper>
+    );
+  },
+);
+
+Input.displayName = "Input";
```
silvenon, line 122: Instead of doing this, I recommend passing a named function to `React.forwardRef` instead, so it can set `displayName` automatically to `ForwardRef(SearchInput)`, I think this should work:

```js
const Input = React.forwardRef(function SearchInput() { /* ... */ })
```

https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

docs/src/components/SearchInput/primitives/StyledDropdown.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  &:nth-child(even) {
+    background: ${({ theme }) => theme.orbit.paletteCloudNormal};
+  }
+  outline: none;
+  > p {
+    font-size: calc(1em - 2px);
+    color: ${({ theme }) => theme.orbit.colorTextPrimary};
+  }
+`;
```
silvenon, line 34: Lots of interpolated functions here, I recommend saving some effort (and performance) with `css`:

```js
const StyledDropdownItem = styled.li`
  ${({ theme }) => css`
    ...
  `}
`
```

then inside you can put all the styles and interpolate `theme` values directly, instead of functions.

docs/src/components/SearchInput/primitives/StyledInputContainer.ts
```diff
@@ -0,0 +1,19 @@
...
+  display: flex;
+  position: relative;
+  box-shadow: ${({ theme }) => theme.orbit.boxShadowAction};
+  padding: 1em 1.5em;
+  border-radius: 22px;
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  box-sizing: border-box;
+  height: 44px;
```
silvenon, line 13: How did you come up with exactly `44px`?

silvenon (changes requested):
Also delete `.DS_Store` and add it to `.gitignore`.

docs/src/components/SearchInput/primitives/StyledDropdown.ts
mainframev: i know about that, thanks

docs/src/components/Navbar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const isHomePage = `${window.location.origin}/` === window.location.href;
+  const { isTablet } = useMediaQuery();
+
+  return (
+    <StyledWrapper isHomePage={isHomePage}>
+      <Link to="/">
+        <Logo height={40} />
+      </Link>
+      {!isHomePage && <Input />}
+      <Stack
```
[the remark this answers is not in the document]
mainframev: yes :)

docs/src/components/SearchInput/index.tsx
mainframev: yes, downshift will be a good option, thanks

docs/src/components/Navbar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  backdrop-filter: blur(5px);
+  z-index: 10;
+`;
+
+interface Props {
+  onBookmarks?: () => void;
+  onSideNav?: () => void;
+}
+
+const Navbar: React.FC<Props> = ({ onBookmarks, onSideNav }) => {
```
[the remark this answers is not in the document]
mainframev: Seems like no difference, i prefer that way, [it has even some advantages](https://github.com/typescript-cheatsheets/react)

silvenon (dismissed):

PR #2575, opened 2020-12-09
feat(Modal): extended size

mainframev:
added new size value, as we agreed before. Slack discussion [here](https://skypicker.slack.com/archives/CAMS40F7B/p1606485777342700)<br/><br/><br/><url>LiveURL: https://orbit-feat-modal.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Modal/__examples__/SIZES.js
+++ b/packages/orbit-components/src/Modal/__examples__/SIZES.js
@@ -13,6 +13,8 @@ export default {
     const [showModalSmall, setShowModalSmall] = React.useState(false);
     const [showModalNormal, setShowModalNormal] = React.useState(false);
     const [showModalLarge, setShowModalLarge] = React.useState(false);
+    const [showModalExtraLarge, setShowModalExtraLarge] = React.useState(false);
+
     return (
       <>
         {showModalExtraSmall && (
@@ -62,6 +64,18 @@ export default {
             </ModalSection>
           </Modal>
         )}
+        {showModalExtraLarge && (
+          <Modal
+            size="extraLarge"
+            onClose={() => {
+              setShowModalExtraLarge(false);
+            }}
+          >
+            <ModalSection>
+              Orbit is an open source design system for your next travel project.
+            </ModalSection>
+          </Modal>
+        )}
         <Stack>
           <Button
             onClick={() => {
@@ -90,6 +104,13 @@ export default {
             }}
           >
             Show large modal
+          </Button>
+          <Button
+            onClick={() => {
+              setShowModalExtraLarge(true);
+            }}
+          >
+            Show extra large modal
           </Button>
         </Stack>
       </>
--- a/packages/orbit-components/transforms/Modal-size.js
+++ b/packages/orbit-components/transforms/Modal-size.js
@@ -1,0 +1,24 @@
+// @noflow
+/* eslint-disable no-param-reassign */
+function transformModalSizeProp(fileInfo, api) {
+  const j = api.jscodeshift;
+  const root = j(fileInfo.source);
+
+  return root
+    .find(j.JSXOpeningElement, { name: { name: "Modal" } })
+    .find(j.JSXAttribute, { name: { name: "size" } })
+    .forEach(path => {
+      const { loc, type } = path.node.value;
+      if (type === "StringLiteral" || type === j.Literal.name) {
+        path.node.value.value = "extraLarge";
+      } else {
+        console.warn(
+          "Unable to transform Modal size value because it's not a string literal, please verify whether large size is being used and use extraLarge instead.",
+          `${fileInfo.path}:${loc.start.line}:${loc.start.column}`,
+        );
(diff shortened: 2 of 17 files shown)


packages/orbit-components/src/Modal/__examples__/SIZES.js
```diff
@@ -62,6 +64,18 @@ export default {
             </ModalSection>
           </Modal>
         )}
+        {showModalExtraLarge && (
+          <Modal
+            size="extraLarge"
+            onClose={() => {
+              showModalExtraLarge(false);
```
silvenon, line 71:
```suggestion
              setShowModalExtraLarge(false);
```

silvenon (changes requested):

packages/orbit-components/src/Modal/__examples__/SIZES.js
```diff
@@ -91,6 +105,13 @@ export default {
           >
             Show large modal
           </Button>
+          <Button
+            onClick={() => {
+              setShowModalExtraLarge(true);
+            }}
+          >
+            Show large modal
```
silvenon, line 113:
```suggestion
            Show extra large modal
```

silvenon (changes requested):

silvenon (dismissed):

packages/orbit-components/transforms/Modal-size.js
```diff
@@ -0,0 +1,16 @@
...
+function transformModalSizeProp(fileInfo, api) {
+  const j = api.jscodeshift;
+  const root = j(fileInfo.source);
+
+  return root
+    .find(j.JSXOpeningElement, { name: { name: "Modal" } })
+    .find(j.JSXAttribute, { name: { name: "size" }, value: { value: "large" } })
+    .forEach(path => {
+      path.node.value.value = "extraLarge";
+    })
```
silvenon, line 12: This is the happy path, but for example in `frontend/frontend` I found an instance with `size={size}`, so in cases where the value is not a string literal I think the codemod should output a warning like: "Unable to transform Modal size value because it's not a string literal, please verify whether `large` size is being used and use `extraLarge` instead.", along with the absolute file path with location information (line and column number) in a format that can be opened with Cmd+click in VS Code. You can extract line and column number from the AST (AST Explorer hides location data by default, but you can uncheck the checkbox above.)

For example:

```suggestion
    .find(j.JSXAttribute, { name: { name: "size" } })
    .forEach(path => {
      if (path.node.value.type === j.Literal) {
        path.node.value.value = "extraLarge";
      } else {
        // output warning
      }
    })
```

packages/orbit-components/transforms/__tests__/Modal-size.test.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+const { defineTest } = require("jscodeshift/dist/testUtils");
+
+defineTest(__dirname, "Modal-size");
```
silvenon, line 4: It's best to test codemods with both Flow and TypeScript parser just to be sure that there is no funny business. You can see how to do this in `Stack-spacing` tests.

silvenon (changes requested):
Please also amend the commit message with the info that a codemod exists and add command how to run it (including `--parser` info), linking to the codemod on GitHub `master` branch, just like we did for previous ones. That way we can amend the codemod if we find mistakes and people still get to access the latest version.

packages/orbit-components/transforms/Modal-size.js
mainframev: it will not work, because `j.Literal` is an object. `j.Literal.name` will work

mainframev: but anyway, thanks for suggestion

packages/orbit-components/transforms/Modal-size.js
```diff
@@ -0,0 +1,23 @@
...
+    .find(j.JSXOpeningElement, { name: { name: "Modal" } })
+    .find(j.JSXAttribute, { name: { name: "size" } })
+    .forEach(path => {
+      if (path.node.value.type === j.Literal.name) {
+        path.node.value.value = "extraLarge";
+      } else {
+        console.warn(
+          "Unable to transform Modal size value because it's not a string literal, please verify whether large size is being used and use extraLarge instead.",
+          fileInfo.path,
+        );
```
silvenon, line 17: Almost, this is what I meant:
```suggestion
        const { loc } = path.node.value;
        console.warn(
          "Unable to transform Modal size value because it's not a string literal, please verify whether large size is being used and use extraLarge instead.",
          `${fileInfo.path}:${loc.start.line}:${loc.start.column}`,
        );
```
Considering that we have location data available, why not help people as much as possible so people can just straight to the location of the `size` value. 😉

mainframev: oh, yes, thanks a lot

vepor (dismissed):

silvenon (approved):

PR #2582, opened 2020-12-14
test(Translate): test React component directly

silvenon:
The tests were testing lower level API instead of the component itself. Now we're testing `pureTranslate` indirectly through the Translate component because that's the API that people are using.

Mocking `window.matchMedia` was necessary because `ThemeProvider` requires it.<br/><br/><br/><url>LiveURL: https://orbit-testing-library-translate.surge.sh</url>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,12 @@ module.exports = {
       extends: ["./config/eslintTS"],
     },
     {
+      files: "**/__tests__/**",
+      rules: {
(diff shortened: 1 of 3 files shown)


mainframev:
LGTM :) 

Probably we can rewrite this component to use `useDictionary` hook and mock that hook in test 🤔

silvenon:
@mainframev done.

mainframev (approved):

PR #2586, opened 2020-12-14
test(textarea): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-textarea.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Textarea/__tests__/index.test.js
+++ b/packages/orbit-components/src/Textarea/__tests__/index.test.js
@@ -1,108 +1,73 @@
 // @flow
 import * as React from "react";
-import { shallow } from "enzyme";
+import { render, screen, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import Textarea from "../index";
-import { SIZE_OPTIONS, RESIZE_OPTIONS } from "../consts";
 import SPACINGS_AFTER from "../../common/getSpacingToken/consts";
 
-describe(`Textarea with help`, () => {
-  const size = SIZE_OPTIONS.NORMAL;
-  const name = "name";
-  const label = "Label";
-  const value = "value";
-  const placeholder = "placeholder";
-  const tabIndex = "-1";
-  const dataTest = "test";
-  const maxLength = 200;
-  const fullHeight = true;
-  const onChange = jest.fn();
-  const onFocus = jest.fn();
-  const onBlur = jest.fn();
-  const spaceAfter = SPACINGS_AFTER.NORMAL;
+describe("Textarea", () => {
+  it("should have expected DOM output", () => {
+    const name = "name";
+    const label = "Label";
+    const value = "value";
+    const placeholder = "placeholder";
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/Textarea/__tests__/index.test.js
```diff
@@ -1,108 +1,70 @@
...
-    expect(component.find(`FormFeedback`).exists()).toBe(true);
-  });
-  it("should execute onChange method", () => {
-    area.simulate("change");
-    expect(onChange).toHaveBeenCalled();
-  });
-  it("should execute onFocus method", () => {
-    area.simulate("focus");
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+    expect(textarea).toHaveAttribute("tabindex", "0");
```
silvenon, line 45: You can leave setting `tabIndex` explicitly to `-1`, just to assert that setting it works as intended, like the Enzyme test did before, then split event handlers into a separate test without setting `tabindex` specifically (defaulting to `0`).

packages/orbit-components/src/Textarea/index.js
```diff
@@ -128,7 +128,7 @@ const Textarea = React.forwardRef<Props, HTMLElement>(
       name,
       rows,
       readOnly,
-      tabIndex,
+      tabIndex = 0,
```
silvenon, line 131: I'm not against this, but it's the same as `undefined`, no? Because `textarea` is focusable by default.

silvenon (changes requested):

packages/orbit-components/src/Textarea/__tests__/index.test.js
```diff
@@ -1,108 +1,75 @@
...
+    expect(screen.getByText("Something useful.")).toBeInTheDocument();
+    expect(textarea).toHaveAttribute("maxlength", maxLength.toString());
+    expect(textarea).toHaveAttribute("rows", "4");
+    expect(textarea).toHaveAttribute("name", name);
+    expect(textarea.parentElement).toHaveStyle({ marginBottom: "12px" });
+    expect(textarea).toHaveStyle({ padding: "12px" });
+
+    fireEvent.blur(textarea);
+    expect(onBlur).toHaveBeenCalled();
+    // userEvent type does not fire onChange, not sure why, use fireEvent
```
silvenon, line 58: It didn't work because you passed `readOnly` to `TextArea`. 😄 I started thinking it was a bug with `@testing-library/user-event` and `textarea`, but then I finally saw it.

silvenon (changes requested):

packages/orbit-components/src/Textarea/__tests__/index.test.js
mainframev: ah, great, thanks :)

silvenon (approved):
Now we don't test `readOnly`, my thought was to do assertions with `userEvent.type` in another test, but never mind now.

PR #2588, opened 2020-12-14
test(TooltipPrimitive): small fixes

silvenon:
I went to check whether `TooltipPrimitive` tests needed to be refactored to Testing Library, they were already using it but I noticed a couple of things.

- `jest-styled-components` needed to be imported before because the
component was being rendered outside the test block

- instead of mocking `useMemo` implementation, which is very bad, we're
now mocking the `randomID` util, which was the purpose of the `useMemo`
mock

### DIFF
--- a/packages/orbit-components/src/primitives/TooltipPrimitive/__tests__/index.test.js
+++ b/packages/orbit-components/src/primitives/TooltipPrimitive/__tests__/index.test.js
@@ -2,22 +2,20 @@
 import * as React from "react";
 import { render } from "@testing-library/react";
 
-// TODO: it is failing without it, seems like @testing-library does not see it in global config
-import "jest-styled-components";
 import Tooltip from "../index";
 import Airplane from "../../../icons/Airplane";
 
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/primitives/TooltipPrimitive/__tests__/index.test.js
```diff
@@ -2,22 +2,20 @@
...
+describe("Tooltip", () => {
   it("it should match snapshot", () => {
+    const content = "Write some message to the user";
+    const { container } = render(
+      <Tooltip content={content}>
+        <Airplane />
+      </Tooltip>,
+    );
+
     expect(container.firstChild).toMatchSnapshot();
```
mainframev, line 19: can we get rid of snapshot testing here? or it will be a part of another PR?

silvenon: I wanted to get rid of it at first, too, but there's quite a bit of functionality to cover 😕 so I think it's best to do it outside of this epic.

mainframev (approved):

PR #2591, opened 2020-12-15
test(MobileDialogPrimitive): rewrite to Testing Library

silvenon:
I tried to test `aria-describedby`, but there doesn't appear to be a Testing Library query for this.
<br/><br/><br/><url>LiveURL: https://orbit-testing-library-mobile-dialog-primitive.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
+++ b/packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
@@ -1,29 +1,37 @@
 // @flow
 import * as React from "react";
-import { mount } from "enzyme";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import MobileDialog from "../index";
-import Button from "../../../Button";
+
+jest.mock("../../../utils/randomID", () => name => name);
+
+let dialogs: HTMLElement;
+
+beforeAll(() => {
+  dialogs = document.createElement("div");
+  dialogs.setAttribute("id", "dialogs");
+  document.body?.appendChild(dialogs);
+});
+
+afterAll(() => {
+  document.body?.removeChild(dialogs);
+});
 
 describe("MobileDialogPrimitive", () => {
-  const content = "My text link";
-  const tabIndex = "1";
-  const dataTest = "test";
-
-  const component = mount(
-    <MobileDialog tabIndex={tabIndex} dataTest={dataTest} content={content}>
-      <Button>Button</Button>
-    </MobileDialog>,
-  );
-
-  const dialog = component.find("MobileDialog");
-  it("should contain a children", () => {
-    expect(dialog.children().exists()).toBe(true);
-  });
-  it("should have data-test", () => {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
```diff
@@ -1,29 +1,37 @@
...
+import userEvent from "@testing-library/user-event";
 
 import MobileDialog from "../index";
-import Button from "../../../Button";
 
-describe("MobileDialogPrimitive", () => {
-  const content = "My text link";
-  const tabIndex = "1";
-  const dataTest = "test";
+jest.mock("../../../utils/randomID", () => name => name);
```
mainframev, line 8: we have it already mocked in `enzymeConfig.js`, but since it supposed to be removed later, we will move it to common jest config or will mock it every time like that?

packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
```diff
@@ -1,29 +1,37 @@
...
+  dialogs.setAttribute("id", "dialogs");
+  document.body?.appendChild(dialogs);
+});
+
+afterAll(() => {
+  document.body?.removeChild(dialogs);
+});
+
+describe("MobileDialogPrimitive", () => {
+  it("should have expected DOM output", async () => {
```
mainframev, line 23: seems like redundant `async`, there is no `await` anywhere 🤔

mainframev (changes requested):

mainframev (dismissed):

packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
mainframev, line 23: seems like redundant `async`, there is no `await` inside and test passes without it

packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
silvenon: If you're referring to this:

```js
const mockMath = Object.create(global.Math);
mockMath.random = () => 0.5;
global.Math = mockMath;
```

I don't think that's a good way because we might be later be using `Math.random` for something else, and might not necessarily want to mock it. I'm a little uncomfortable about mocking it globally like that, I think it's best to always mock `randomID` explicitly in tests where that's needed. Alternatively, we can move this `jest.mock` call to the setup file, I just want to avoid mocking `Math.random`.

mainframev: okay, i just decided to ask to be sure :) thanks

packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.js
silvenon: You're right, it remained from something I previously tried.

mainframev (approved):

PR #2593, opened 2020-12-15
test(Tile): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-tile.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Tile/components/TileWrapper/index.js
+++ b/packages/orbit-components/src/Tile/components/TileWrapper/index.js
@@ -59,8 +59,8 @@ const TileWrapper = ({
     as={as}
     tabIndex={tabIndex}
     role={role}
-    ariaExpanded={ariaExpanded}
-    ariaControls={ariaControls}
+    aria-expanded={ariaExpanded}
+    aria-controls={ariaControls}
     id={id}
     title={htmlTitle}
   >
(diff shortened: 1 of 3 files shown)


packages/orbit-components/src/Tile/components/TileWrapper/index.js
```diff
@@ -59,8 +59,8 @@ const TileWrapper = ({
     as={as}
     tabIndex={tabIndex}
     role={role}
-    ariaExpanded={ariaExpanded}
-    ariaControls={ariaControls}
+    aria-expanded={ariaExpanded}
+    aria-controls={ariaControls}
```
silvenon, line 63: Nice! 🥇

packages/orbit-components/src/Tile/index.js
```diff
@@ -53,7 +53,7 @@ const Tile = ({
       onKeyDown={handleKeyDown(onClick)}
       as={href ? "a" : "div"}
       tabIndex={!href ? "0" : undefined}
-      role={!href ? "button" : undefined}
+      role={href ? "link" : "button"}
```
silvenon, line 56: `TileWrapper` will be `<a>` if it has `href`, so setting `role` to `link` is redundant and should remain `undefined`, tests pass regardless of this change.

silvenon (changes requested):

silvenon (approved):

PR #2598, opened 2020-12-16
test(Tag): rewrite to testing library

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-test-tag.surge.sh</url>


packages/orbit-components/src/Tag/index.js
```diff
@@ -170,6 +171,7 @@ const Tag = (props: Props) => {
         >
           <StyledClose
             tabIndex="0"
+            aria-label="close"
```
silvenon, line 174: It would be better if this was a separate commit also.

silvenon (approved):

PR #2603, opened 2020-12-16
improve: stop enforcing commit scope case

silvenon:
I assume this was intended for committing changes to React components, but we also have hooks, helpers, etc. Having to commit the scope as PascalCase doesn't make sense, so we're disabling this rule.

This has been driving me crazy 😅 <br/><br/><br/><url>LiveURL: https://orbit-improve-commitlint-liberal-scope.surge.sh</url>

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -2,7 +2,7 @@
 module.exports = {
   extends: ["@commitlint/config-conventional"],
   rules: {
-    "scope-case": [2, "always", "pascal-case"],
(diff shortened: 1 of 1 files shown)


silvenon:
But I have to say that I don't understand [these examples](https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md), how does `[0, "always", 72]` compare to just `[0]`…?

mainframev (approved):

PR #2607, opened 2020-12-16
test(Grid): rewrite to testing library

mainframev:
`toHaveStyle` did not work with some styles, i think it's ok to have `toHaveStyleRule` since it will work with `testing-library` too, mediaQueries also did not work without `toHaveStyleRule`. 🤔 <br/><br/><br/><url>LiveURL: https://orbit-test-grid.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/utils/Grid/__tests__/index.test.js
+++ b/packages/orbit-components/src/utils/Grid/__tests__/index.test.js
@@ -1,401 +1,43 @@
 // @flow
 import * as React from "react";
-import { mount } from "enzyme";
+import { render } from "@testing-library/react";
 
-import theme from "../../../defaultTheme";
-import { QUERIES } from "../../mediaQuery/consts";
-import { getBreakpointWidth } from "../../mediaQuery/index";
 import Grid from "..";
 
-describe("Grid with basic props", () => {
-  const inline = true;
-  const maxWidth = "1440px";
-  const width = "100%";
-  const columns = "repeat(2, 1fr)";
-  const columnGap = "10px";
-  const rows = "repeat(2, minmax(30px, 100px))";
-  const rowGap = "20px";
-  const dataTest = "test";
-  const component = mount(
-    <Grid
-      inline={inline}
-      maxWidth={maxWidth}
-      width={width}
-      columns={columns}
-      columnGap={columnGap}
-      rows={rows}
-      rowGap={rowGap}
-      dataTest={dataTest}
-    >
-      <div />
-      <div />
-      <div />
-      <div />
-    </Grid>,
-  );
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/utils/Grid/__tests__/index.test.js
```diff
@@ -1,401 +1,110 @@
...
-    expect(grid).toHaveStyleRule(
-      "-ms-grid-columns",
-      "minmax(100px,1fr) 20px minmax(100px,1fr) 20px minmax(100px,1fr) 20px minmax(100px,1fr)",
-      {
-        media: getBreakpointWidth(QUERIES.DESKTOP, theme),
-      },
+});
+
+describe("Grid, different mediaQueries", () => {
+  it.each(Object.entries(QUERIES))("should have styles for each mediaquery: %s", (name, query) => {
```
silvenon, line 79: As I said the last time you used this pattern: when I see `it.each` I expect that the component will be rendered differently in every test, but this is not true, the only thing that changes in this test is the assertion at the bottom. Why don't you just do this instead?

```js
  it("should have styles for each media query", () => {
    const dataTest = "test";

    render(
      <Grid
        dataTest={dataTest}
        mediumMobile={{ maxWidth: "600px" }}
        largeMobile={{ maxWidth: "700px" }}
        tablet={{ maxWidth: "800px" }}
        desktop={{ maxWidth: "900px" }}
        largeDesktop={{ maxWidth: "1000px" }}
      >
        <div>1</div>
        <div>2</div>
        <div>3</div>
        <div>4</div>
      </Grid>,
    );

    const expected = {
      mediumMobile: "600px",
      largeMobile: "700px",
      tablet: "800px",
      desktop: "900px",
      largeDesktop: "1000px",
    };

    // eslint-disable-next-line no-restricted-syntax
    for (const query of Object.keys(expected)) {
      expect(screen.getByTestId(dataTest)).toHaveStyleRule("max-width", expected[query], {
        media: getBreakpointWidth(query, theme),
      });
    }
  });
```

silvenon (changes requested):
I think the fact that `toHaveStyle` doesn't support this is a jsdom problem, because `getComputedStyle(grid)` doesn't contain those styles either.

Instead of this large amount of assertions, I suggest merging the props of both tests (so that the initial test contains media props as well) and using a single snapshot test, this seems like a good use case for snapshots.

packages/orbit-components/src/utils/Grid/__tests__/index.test.js
mainframev: yes, makes sense

silvenon (dismissed):

silvenon (approved):

PR #2609, opened 2020-12-16
test: remove Enzyme

silvenon:
We use `raf/polyfill` in our tests even though we didn't have `raf`
installed directly. Apparently Enzyme was depending on it because after
removing it Jest was no longer able to find `raf`, so now it's installed
directly.
<br/><br/><br/><url>LiveURL: https://orbit-remove-enzyme.surge.sh</url>


silvenon:
These changes appear to have revealed some legit type checking problems, not sure why yet. 🤔

silvenon:
Done ✅

mainframev:
could you separate this changes on two commits? there are some changes to components, which probably should not be related to `test` scope

silvenon:
@mainframev done. 👍

PR #2613, opened 2020-12-18
fix(docs): window and document SSR

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-fix-docs-window.surge.sh</url>


silvenon (commented):
I suggest continuing the early return pattern instead, so you don't have to change the return type:

```js
if (!url || typeof document === "undefined") return false;
```

This way you avoid the type problem you're having in CI.

silvenon (approved):

PR #2616, opened 2020-12-18
fix(Box): missing box-sizing

mainframev:
<br/><br/><br/><url>LiveURL: https://orbit-fix-box.surge.sh</url>


silvenon (approved):

mainframev:
We are missing it, since we provide props, such as `maxHeight`, `maxWidth`, `padding`, we should have set `box-sizing: border-box`.

PR #2621, opened 2020-12-21
feat(docs): sidenav, bookmarks

mainframev:
- added `Sidenav` component
- `NavigationLinks` on mobile inside Sidenav
- added  `Bookmarks` sidebar, stores bookmarks in localstorage
- added  `Bookmark` component, to add bookmarks from pages 
- added checkbox input component (toggling developer mode) inside `Sidenav` 

<br/><br/><br/><url>LiveURL: https://orbit-feat-sidenav.surge.sh</url>


docs/src/components/Bookmarks/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+`;
+
+const Bookmarks = () => {
+  const [checked, setChecked] = React.useState(false);
+
+  React.useEffect(() => {
+    if (load("devMode") === "enabled") setChecked(true);
+    else setChecked(false);
+  }, [load, setChecked]);
```
silvenon, line 24: In #2631 I enabled ESLint rules for React hooks, it will help us see which dependencies we actually need, and we don't need these, ESLint won't complain.
```suggestion
  }, []);
```

docs/src/components/Bookmarks/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (load("devMode") === "enabled") {
+      setChecked(false);
+      save("devMode", "disabled");
+    } else {
+      setChecked(true);
+      save("devMode", "enabled");
+    }
+  };
+
+  const bookmarks = load("bookmarks");
```
silvenon, line 36: This should be inside a `useEffect`.

docs/src/components/Navigation/Mobile.tsx
```diff
@@ -0,0 +1,101 @@
...
+      background: ${theme.orbit.paletteCloudLight};
+    }
+  `}
+`;
+
+interface Props {
+  name: string;
+}
+
+export default function Mobile({ name }: Props) {
```
silvenon, line 49: For better display name in developer tools.
```suggestion
export default function MobileNavigation({ name }: Props) {
```

docs/src/components/Sidenav/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      max-width: ${width}px;
+    `)};
+  `}
+`;
+
+const StyledOpenButton = styled(({ className, children, ariaLabel, onClick }) => (
+  <button className={className} aria-label={ariaLabel} onClick={onClick} type="button">
+    {children}
+  </button>
+))`
```
silvenon, line 43: I believe you can handle these situations better with [`.attrs`](https://styled-components.com/docs/api#attrs), then it's instantly clear that you're only converting `ariaLabel` to `aria-label` and adding `type="button"`.

silvenon (commented):
We're mixing `ComponentName.tsx` with `ComponentName/index.tsx` structure 😬 But for components like `SearchInput` it totally makes sense. Should we go all in then? Just rename the rest of `ComponentName.tsx` to `ComponentName/index.tsx`? Not here, of course.

docs/src/components/Sidenav/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const StyledAsideWrapper = styled.aside<WrapperProps>`
+  ${({ width, theme, shown }) => css`
+    background: ${theme.orbit.paletteWhite};
+    padding: 10px 20px;
+    z-index: 100;
+    margin: 0;
+    display: block;
+    position: absolute;
+    box-sizing: border-box;
```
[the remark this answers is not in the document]
mainframev: ah, yes, thanks, it's kinda habit already, so thanks for mentioning it

silvenon:
Regarding `eslintCI.js` failure, before we give up and disable it, I just want to use this opportunity to try to debug it. Could you add the following lines in the `eslintCI` function? 🙏

```js
console.log('ALL:\n')
console.log(JSON.parse(await fsx.readFile(`${process.env.HOME}/files.json`)).join('\n'))
console.log('MODIFIED:\n')
console.log(JSON.parse(await fsx.readFile(`${process.env.HOME}/files_modified.json`)).join('\n'))
console.log('ADDED:\n')
console.log(JSON.parse(await fsx.readFile(`${process.env.HOME}/files_added.json`)).join('\n'))
console.log('DELETED:\n')
console.log(JSON.parse(await fsx.readFile(`${process.env.HOME}/files_deleted.json`)).join('\n'))
console.log('RENAMED:\n')
console.log(JSON.parse(await fsx.readFile(`${process.env.HOME}/files_renamed.json`)).join('\n'))
```

Just so I can see what it contains. Afterwards I'll probably give up.

silvenon (commented):
Haven't tested bookmarks yet, but one minor note.

docs/src/components/Bookmarks/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }
+      toggleIcon={<StarEmpty ariaHidden />}
+    >
+      {bookmarks && (
+        <Stack direction="column">
+          <h3>Your bookmarks</h3>
+          {Object.entries(JSON.parse(bookmarks)).map(([key, slug]) => (
+            // @ts-expect-error entries
+            <StyledLink key={key} to={slug}>
+              {key}
```
silvenon, line 54: I'd expect the name of the bookmark for e.g. `/getting-started/for-designers/kiwi/` to be something like "Getting started / For designers / For Kiwi.com employees", or just "For Kiwi.com employees"according to the current orbit.kiwi, but definitely not "01-getting-started/01-for-designers/01-kiwi".

silvenon (changes requested):
Gatsby builds fine (after #2655), so I think the `localStorage` `useEffect` problem is fixed, the bookmark seems to lead to the correct place, and the slug is now saved to `localStorage` instead of title, but I still need to refresh the page to see new bookmarks, and considering that navigating the site doesn't refresh the page I think it's important to have that.

silvenon:
Why does this PR increase the bundle size…?

silvenon (changes requested):

PR #2623, opened 2020-12-24
feat(seat): new version of seat component

mainframev:
Added changes to `Seat` component, faced some issues, don't know why, but after copying svgs path etc, they did not work as expected, so i changed them a bit. In order to add text, i used svg's text, aligned them somehow, hope that will be enough. 

<br/><br/><br/><url>LiveURL: https://orbit-feat-new-seat.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Seat/__examples__/LABEL.js
+++ b/packages/orbit-components/src/Seat/__examples__/LABEL.js
@@ -1,0 +1,12 @@
+// @flow
+import * as React from "react";
+
+import Seat from "../index";
+
+export default {
+  Example: () => <Seat label="MZ" />,
+  info: {
+    title: "Seat with label",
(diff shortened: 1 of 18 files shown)


packages/orbit-components/src/Seat/__examples__/LABEL.js
```diff
@@ -0,0 +1,12 @@
...
+
+import Seat from "../index";
+
+export default {
+  Example: () => <Seat label="MZ" />,
+  info: {
+    title: "Seat with label",
+    description: "Seat can have a label text inside",
+  },
+};
```
silvenon, line 12: New examples should be imported in `examples.js` to be visible in the Storybook.

silvenon (approved):

PR #2624, opened 2020-12-25
feat(docs): added breadcrumbs

mainframev:
Orbit.kiwi: https://orbit-docs-feat-docs-breadcrumbs.surge.sh
 Storybook: https://orbit-feat-docs-breadcrumbs.surge.sh

### DIFF
--- a/docs/src/components/Breadcrumbs.tsx
+++ b/docs/src/components/Breadcrumbs.tsx
@@ -1,0 +1,59 @@
+import React from "react";
+import styled from "styled-components";
+import { WindowLocation } from "@reach/router";
+import { Link } from "gatsby";
+
+import { capitalize } from "../utils/common";
+
+const StyledList = styled.ul`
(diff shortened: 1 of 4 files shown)


silvenon (commented):
I visited `/getting-started/for-designers/kiwi/` and saw "orbit.kiwi /
for-kiwi.com-employees". According to this design:

Screen Shot 2021-01-11 at 08 59 29

I expected to see "orbit.kiwi / Getting started / For designers". So:

- "Getting started" and "For designers" is missing
- tab names should apparently not be a part of breadcrumbs (which makes sense)
- crumb names are not lowercase and there are no hyphens

silvenon (changes requested):
Almost, but dashes between "Getting started" and "For designers" need to go as well. 😄 

Screen Shot 2021-02-19 at 00 45 49

silvenon (approved):

PR #2625, opened 2020-12-25
chore: init eslint-plugin-orbit-components in monorepo

mainframev:
What've done:

- Moved eslint orbit plugin to our monorepo
- Moved it from `js` to `ts`, but parser will remain as `babel-eslint`
- Adjust existing code with `babel-types`

The purpose of these RP is to have it in our monorepo and prepare for further changes, we'd like to extend our plugin with more rules, especially with a11y. 

<br/><br/><br/><url>LiveURL: https://orbit-chore-eslint-orbit-monorepo.surge.sh</url>


packages/eslint-plugin-orbit-components/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../config/eslintTS.js",
+  "rules": {
+    "@typescript-eslint/prefer-readonly-parameter-types": "off",
+    "@typescript-eslint/explicit-function-return-type": "off"
+  }
+}
```
silvenon, line 7: Put these rules in `overrides` in root `.eslintrc.js` instead, and delete this file. The `eslintTS.js` file is incidental and we'll probably merge it with `.eslintrc.js` later.

packages/eslint-plugin-orbit-components/.gitignore
```diff
@@ -0,0 +1,4 @@
+.idea
+node_modules
+dist
+yarn-error.log
```
silvenon, line 4: We only need `dist`.

packages/eslint-plugin-orbit-components/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ["@babel/preset-env", "@babel/preset-typescript"],
+};
```
silvenon, line 3: If the only reason why we're using Babel is to have ES imports, we should convert imports to requires and get rid of the build step. Thoughts?

packages/eslint-plugin-orbit-components/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "eslint-plugin-orbit-components",
+  "version": "0.2.0",
+  "description": "Collection of ESLint rules to enforce the best usage practices of @kiwicom/orbit-components",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/kiwicom/eslint-plugin-orbit-components"
```
silvenon, line 7:
```suggestion
    "url": "https://github.com/kiwicom/orbit.git",
    "directory": "packages/eslint-plugin-orbit-components"
```

packages/eslint-plugin-orbit-components/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "eslintplugin",
+    "eslint-plugin",
+    "orbit-components",
+    "kiwicom"
+  ],
+  "author": "Kiwi.com",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "rm -fr ./dist && babel ./src --extensions '.ts' --out-dir ./dist",
+    "test": "jest tests"
```
silvenon, line 20: We don't need the `test` script.

packages/eslint-plugin-orbit-components/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "eslint",
+    "eslintplugin",
+    "eslint-plugin",
+    "orbit-components",
+    "kiwicom"
+  ],
+  "author": "Kiwi.com",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "rm -fr ./dist && babel ./src --extensions '.ts' --out-dir ./dist",
```
silvenon, line 19: Setting aside the question whether to have the build step at all, if we're going to have it we also need a `prepublishOnly` script which runs `build` because the root `publish-packages` script runs `publish` in every package, so if some packages need to run anything before publishing, they need to define it for themselves.

packages/eslint-plugin-orbit-components/yarn.lock
```diff
@@ -0,0 +1,5714 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
silvenon, line 2: This `yarn.lock` needs to be deleted.

.eslintignore
```diff
@@ -1,6 +1,7 @@
 es
 umd
 flow-typed
+dist
```
silvenon, line 4: It's best to be specific, so we know what's it referring to.

```suggestion
packages/eslint-plugin-orbit-components/dist
```

.eslintrc.js
```diff
@@ -27,6 +27,14 @@ module.exports = {
         "@typescript-eslint/explicit-module-boundary-types": "off",
       },
     },
+    {
+      files: "packages/eslint-plugin-orbit-components/**/*",
```
silvenon, line 31: Best to be specific, to ensure that ESLint doesn't fail for JS files where `@typescript-eslint` plugin isn't enabled and those rules don't exist.
```suggestion
      files: "packages/eslint-plugin-orbit-components/**/*.ts?(x)",
```

packages/eslint-plugin-orbit-components/babel.config.js
```diff
@@ -0,0 +1,17 @@
...
+const browserslist = require("@kiwicom/browserslist-config");
+
+module.exports = {
+  presets: [
+    [
+      "@babel/preset-env",
+      {
+        targets: {
+          node: true,
+          browsers: browserslist,
```
silvenon, line 11: This code won't be running in the browser, so it's not necessary to convert it for browsers. I recommend adding a `.browserslistrc` file in this workspace containing `current node`. That way `@babel/preset-env` should automatically detect it and use it, and hopefully not merge that config with our root `.browserslistrc` file 🤞 But it doesn't matter if that happens, I'll fix it later.

packages/eslint-plugin-orbit-components/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "eslint",
+    "eslintplugin",
+    "eslint-plugin",
+    "orbit-components",
+    "kiwicom"
+  ],
+  "author": "Kiwi.com",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "rm -fr ./dist && babel ./src --extensions '.ts' --out-dir ./dist"
```
silvenon, line 20: Please also add `"prepublishOnly": "build"`.

silvenon (approved):

PR #2627, opened 2020-12-29
refactor: layout components

mainframev:
Unified same logic in `Stack`, `Inline`, `Box`
<br/><br/><br/><url>LiveURL: https://orbit-refactor-layouts.surge.sh</url>


packages/orbit-components/src/Inline/Inline.stories.js
```diff
@@ -31,7 +31,7 @@ storiesOf("Inline", module)
   .add(
     "Default",
     () => {
-      const spacing = select("spacing", SPACING, SPACING[1]);
+      const spacing = select("spacing", Object.values(SPACING), SPACING[1]);
```
vepor, line 34: `SPACING[1]` is `undefined`, need to use e.g. `SPACING.XXXSMALL`. It applies to all other same assertions in this file.

packages/orbit-components/src/Box/normalize.js
```diff
@@ -50,52 +43,46 @@ const normalizeSize = (val, property): string => {
...
+    maxWidth: formatCSS("max-width", val),
+    maxHeight: formatCSS("max-height", val),
+  };
+
+  if (!all[key]) return formatCSS(key, val);
+
+  return all[key];
+};
+
 const normalize = (object: MediaQueryObject) => ({ theme }: ThemeProps) => {
```
vepor, line 80: Do we have tests for this normalizer? We could easily introduce bugs without them, especially if we perform refactoring 🙂

vepor (changes requested):
Good job! Just some small things 🙂

packages/orbit-components/src/Box/normalize.js
mainframev: it's covered via normal tests, no need to test normalizer, if output is correct - it means it works fine 🤔

vepor (approved):

PR #2633, opened 2021-01-06
feat(popover): offset

mainframev:
[ORBIT-1436](https://jira.kiwi.com/browse/ORBIT-1436)<br/><br/><br/><url>LiveURL: https://orbit-feat-popover-offset.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Popover/index.js
+++ b/packages/orbit-components/src/Popover/index.js
@@ -18,7 +18,7 @@ const StyledPopoverChild = styled.div`
 
 const Popover = ({
   children,
-  offset,
+  offset = { top: 0, left: 0 },
   content,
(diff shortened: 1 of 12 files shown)


packages/orbit-components/src/Popover/index.js
```diff
@@ -135,6 +136,7 @@ const Popover = ({
         <Portal renderInto="popovers">
           <PopoverContentWrapper
             shown={shown}
+            offset={offset}
```
silvenon, line 139: This way people would always have to specify both `top` and `left`. Do we want to allow them to omit one of those, and let it default to `0`? In that case I think you could do something like this instead:

```suggestion
            offset={{ top: 0, left: 0, ...offset }}
```

silvenon (approved):

PR #2645, opened 2021-01-08
feat: add more content to the landing page

silvenon:
I added buttons bellow the title, some of the cards, linked them to documents etc. There are many things still missing from the landing page, but it's much better to merge bit by bit than everything at once, especially in this early stage where we have to write a lot of code.

- icons in cards are missing because we don't have the SVGs yet
- the description of the Patterns card is missing, in Figma it's the same as the one in Components card
- some of the cards aren't linked because they don't have the corresponding docunment added to the codebase yet
- the "Show all items" button doesn't do anything yet
<br/><br/><br/><url>LiveURL: https://orbit-feat-landing-page.surge.sh</url>

### DIFF
--- a/docs/src/components/ButtonLink.tsx
+++ b/docs/src/components/ButtonLink.tsx
@@ -18,7 +18,7 @@ type Props<T extends React.ElementType = typeof DefaultComponent> = PolymorphicP
 >;
 
 const StyledComponent = styled(DefaultComponent)<Props>`
-  ${({ theme, type, size = "medium" }) => css`
+  ${({ theme, type, size }) => css`
     display: inline-flex;
     align-items: center;
     ${size === "medium" &&
@@ -70,6 +70,8 @@ const StyledComponent = styled(DefaultComponent)<Props>`
 
 export default function ButtonLink<T extends React.ElementType = typeof DefaultComponent>({
   as: asProp,
+  type = "primary",
+  size = "medium",
   iconLeft,
   iconRight,
   children,
@@ -79,7 +81,7 @@ export default function ButtonLink<T extends React.ElementType = typeof DefaultC
 
   return (
     // @ts-expect-error react-polymorphic-types is currently not compatible with styled-components
-    <StyledComponent as={Component} {...restProps}>
+    <StyledComponent as={Component} type={type} size={size} {...restProps}>
       {iconLeft}
       <span>{children}</span>
       {iconRight}
--- a/docs/src/components/ButtonLink.tsx
+++ b/docs/src/components/ButtonLink.tsx
@@ -5,7 +5,7 @@ import { PolymorphicPropsWithoutRef } from "react-polymorphic-types";
 const DefaultComponent = "a";
 
 interface OwnProps {
-  type: "primary" | "secondary";
+  type?: "primary" | "secondary";
   size?: "medium" | "large";
   iconLeft?: React.ReactNode;
   iconRight?: React.ReactNode;
--- a/docs/src/components/ButtonLink.tsx
+++ b/docs/src/components/ButtonLink.tsx
@@ -1,0 +1,88 @@
+import React from "react";
+import styled, { css } from "styled-components";
+import { PolymorphicPropsWithoutRef } from "react-polymorphic-types";
+
+const DefaultComponent = "a";
+
+interface OwnProps {
+  type: "primary" | "secondary";
+  size?: "medium" | "large";
+  iconLeft?: React.ReactNode;
+  iconRight?: React.ReactNode;
+  children: React.ReactNode;
+}
+
+type Props<T extends React.ElementType = typeof DefaultComponent> = PolymorphicPropsWithoutRef<
+  OwnProps,
+  T
+>;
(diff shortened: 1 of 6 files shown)


docs/src/components/ButtonLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+  children: React.ReactNode;
+}
+
+type Props<T extends React.ElementType = typeof DefaultComponent> = PolymorphicPropsWithoutRef<
+  OwnProps,
+  T
+>;
+
+const StyledComponent = styled(DefaultComponent)<Props>`
+  ${({ theme, type, size = "medium" }) => css`
```
mainframev, line 21: I think that way will be better, like we do in orbit usually

```
{
   const Component = asProp || DefaultComponent;
    return (
      <Component {...props}>
        {props.children}
      </Component>
    );
  },
)`
```

docs/src/components/ButtonLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+    display: inline-flex;
+    align-items: center;
+    ${size === "medium" &&
+    css`
```
mainframev, line 25: here is no need for `css`, because there is no interpolation inside

docs/src/pages/index.tsx
```diff
@@ -12,17 +27,155 @@ export default function Home() {
...
+          }
         `}
       >
-        <Heading type="display">
+        <div>
+          <Heading type="display">
+            <div
+              css={css`
+                max-width: 20ch;
+                font-size: 48px;
```
mainframev, line 44: noticed, somewhere in code you use `em` for font-size, why here `48px` and not `3em`?

docs/src/components/ButtonLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+      padding: 1.25rem 2rem;
+    `}
+    border-radius: 100px; /* value safely greater than button height to achieve pill effect */
+    font-weight: 500;
+
+    > * + * {
+      margin-left: 10px;
+    }
+
+    ${type === "primary" &&
```
mainframev, line 39: set `primary` as default and you can omit that condition, plus will not be needed to pass this prop, if type is not secondary.

docs/src/components/ButtonLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+      background: ${theme.orbit.paletteProductLight};
+      color: ${theme.orbit.paletteProductDark};
+      &:hover {
+        background: ${theme.orbit.paletteProductLightHover};
+      }
+      &:active {
+        background: ${theme.orbit.paletteProductLightActive};
+      }
+      &:focus {
+        background: ${theme.orbit.paletteProductLightHover};
```
mainframev, line 65: i think it would be better with inline ternary conditions, less lines of code

mainframev (changes requested):

docs/src/components/ButtonLink.tsx
silvenon: You mean `type === "primary" ? <primary color> : <secondary color>`? I would rather not assume we'll only have two types.

docs/src/components/ButtonLink.tsx
silvenon: Actually, I thnk this will not help me achieve my type checking goal, which is to allow the props of whichever component I pass to `as`, for example for Gatsby's `Link` i wanted the prop `to` to be available. So I don't want to inline the components to not cause any more complexity regarding that because I was barely able to achieve this 😅

docs/src/components/ButtonLink.tsx
silvenon: I don't really see how extracting the styles to `resolve*` functions will make this nicer, I see it only as moving the logic somewhere else. I like to be able to write the code in a way that can be read from top to bottom, and only extract it if it's complicated, and readability could be improved with variable/function naming, or if the file is too large.

mainframev (approved):

docs/src/pages/index.tsx
```diff
@@ -12,17 +27,154 @@ export default function Home() {
...
+            css={css`
+              margin-top: 3rem;
+            `}
+          >
+            <Inline spacing="small">
+              <ButtonLink
+                size="large"
+                as={Link}
+                iconRight={<ArrowRight />}
+                to="/getting-started/for-designers/kiwi"
```
CollierCZ, line 62: It was my understanding that the paths would be similar to how they are now. That would mean that the first tab would not have a separate path. So this path would be `/getting-started/for-designers/` (without `kiwi` -- like: https://orbit.kiwi/getting-started/for-designers/).

Is that not the plan, @silvenon ? It applies many other places (e.g., with "guidelines" below), so it would help for linking in general to know the rule to follow.

CollierCZ (changes requested):

docs/src/pages/index.tsx
silvenon: Will fix this in separate PR, thanks!

PR #2649, opened 2021-01-11
improve: build JSON files when changing examples

silvenon:
By using lint-staged's function syntax we can avoid passing file names
as arguments that usually happens behind the scenes in object syntax, so
we can be sure that lint-staged will run the script exactly as we see
it, because we don't need to pass arguments.
<br/><br/><br/><url>LiveURL: https://orbit-improve-auto-build-examples.surge.sh</url>


silvenon:
Ideally we would be able to automatically stage updated JSON files, but this should be enough for now.

mainframev (approved):

PR #2652, opened 2021-01-12
feat: contributors component

mainframev:
- Added `Contributors` component [ORBIT-1537](https://jira.kiwi.com/browse/ORBIT-1537)
- Has `compact` version for other contributors and bigger one for our main contributors with data from `data.json` + github api.
<br/><br/><br/><url>LiveURL: https://orbit-feat-contributors-component.surge.sh</url>


docs/src/components/Contributors/Contributor.tsx
```diff
@@ -0,0 +1,112 @@
...
+    </Component>
+  );
+})`
+  width: ${({ size }) => resolveSize(size)};
+`;
+
+const StyledHeading = styled.h4`
+  color: ${({ theme }) => theme.orbit.colorHeading};
+  font-size: calc(1em + 2px);
+  font-weight: 800;
```
silvenon, line 31: It would probably be better to use the `fontWeightBold` token for this because the designs are in Circular Pro, and I'm not sure if that typeface has this font weight. We should try to [include Circular Pro](https://orbit.kiwi/foundation/typography/circular-pro/) to soon in a separate PR, but I'm not sure if it's as simple as it seems.

silvenon (changes requested):

docs/src/components/Contributors/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useStaticQuery, graphql } from "gatsby";
+import Contributor from "./Contributor";
+
+const CURRENT_CONTRIBUTORS = [
+  "Victor Genaev",
+  "Luděk Vepřek",
+  "Aaron Collier",
+  "Matija Marohnić",
+  "William Kolmačka",
+  "Milan Seitler",
```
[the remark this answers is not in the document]
mainframev: nope, i think we should solve it on github somehow 🤔 . We cant fetch will because he does not have name, only login. Milan is not in contributors at all and that is probably because he did not contribute to project on github.

docs/src/components/Contributors/Contributor.tsx
mainframev: not sure, if we can't include `Circular Pro`, because of licence issue

silvenon (commented):
For posterity: we decided to have two separate queries — one for fetching the members of the Orbit team (for this we need to update the team members), and another to fetch contributors, from which we should exclude the Orbit team. This is a more stable solution than asking designers to contribute to the code just so they get included in the contributors query.

silvenon (commented):
Otherwise looks good 👍 I'll approve after adding Will & Milan, let me know if you need help with that.

silvenon (approved):

PR #2657, opened 2021-01-13
feat: deprecate old Seat

mainframev:
return old seat into deprecated, [plz-orbit request ](https://skypicker.slack.com/archives/CAMS40F7B/p1610549529003800)<br/><br/><br/><url>LiveURL: https://orbit-chore-deprecated-seat.surge.sh</url>


silvenon (changes requested):
The code looks good, but change the title to something like `feat: deprecate old Seat component` because the changes add something new to the codebase, and it adds a new entry to Babel plugin, so not only will a `chore` commit not be visible in the changelog, but also Babel plugin won’t be published.

PR #2659, opened 2021-01-13
fix(Seat): randomize id to prevent namespace errors

mainframev:
Thanks to `Loki` found one bug, fixed it. There was namespace collisions when several Seat components on page. <br/><br/><br/><url>LiveURL: https://orbit-fix-seat-ids.surge.sh</url>


silvenon (approved):
Awesome! Looks good 👍

PR #2660, opened 2021-01-14
feat(docs): add options for redirects

CollierCZ:
Both client- and server-side

The redirects should work in production builds (e.g., `yarn build && yarn serve`), but not necessarily development builds.

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -43,5 +43,12 @@ module.exports = {
         },
       },
     },
+    {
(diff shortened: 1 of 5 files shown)


CollierCZ:
The failing test seems unrelated to this change.

silvenon (approved):
Resolving the `yarn.lock` conflict is as easy as running `yarn`.

silvenon (approved):

PR #2663, opened 2021-01-16
docs: establish clear URL strategy for documents

silvenon:
- The custom `slug` field has been renamed to `path` to communicate that it can be used for linking, but also because `gatsby-plugin-mdx` also provides a field called `slug`, which I assume is what already caused a confusion in `createPages`, where `slug` was being used instead of `fields.slug`.

- All local URL paths (i.e. the values of `Link`'s `to` prop) should have a trailing slash, to avoid potential SEO problems with the same page treated as two different pages

- First tabs are treated as the default view for the URL, the effect of this change is illustrated by the URL changed on the landing page. This was achieved by reading the `type` in the `meta.yml` file when constructing the path. I was able to easily test this without actually reading and writing to the filesystem by using [`memfs`](https://www.npmjs.com/package/memfs).

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -33,12 +33,12 @@ module.exports = {
     {
       resolve: `@gatsby-contrib/gatsby-plugin-elasticlunr-search`,
       options: {
-        fields: ["title", "excerpt", "slug"],
+        fields: ["title", "excerpt", "path"],
         resolvers: {
           Mdx: {
             title: n => n.frontmatter.title,
             excerpt: n => n.frontmatter.excerpt,
-            slug: n => n.fields.slug,
+            path: n => n.fields.path,
           },
         },
       },
--- a/docs/utils/__tests__/document.test.js
+++ b/docs/utils/__tests__/document.test.js
@@ -1,0 +1,42 @@
+const fsx = require("fs-extra");
+const path = require("path");
+const dedent = require("dedent");
+
+const { getDocumentUrlPath } = require("../document");
+
+const ROOT = path.resolve(__dirname, "../../src/documentation");
+
+jest.mock("fs", () => require("memfs").fs);
+
+describe("document utils", () => {
+  describe(getDocumentUrlPath.name, () => {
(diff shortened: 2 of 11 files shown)


docs/utils/__tests__/document.test.js
```diff
@@ -0,0 +1,42 @@
...
+      );
+      expect(getDocumentUrlPath("/01-getting-started/02-for-designers/01-kiwi.mdx")).toBe(
+        "/getting-started/for-designers/",
+      );
+      expect(getDocumentUrlPath("/01-getting-started/02-for-designers/02-open-source.mdx")).toBe(
+        "/getting-started/for-designers/open-source/",
+      );
+    });
+  });
+});
```
silvenon, line 42: @CollierCZ this test file should hopefully be self-documenting how the URL strategy will work. Let me know if this matches what you were talking about. 🙏

docs/gatsby-config.js
```diff
@@ -33,12 +33,12 @@ module.exports = {
     {
       resolve: `@gatsby-contrib/gatsby-plugin-elasticlunr-search`,
       options: {
-        fields: ["title", "excerpt", "slug"],
+        fields: ["title", "excerpt", "path"],
```
CollierCZ, line 36: If there's not a field named `slug`, plugins such as [gatsby-redirect-from](https://github.com/kremalicious/gatsby-redirect-from) won't work, I believe. It assumes there's a field known as `slug`. The benefit of that plugin in my opinion outweighs a slight confusion between `slug` and `fields.slug`.

CollierCZ (changes requested):

docs/gatsby-config.js
silvenon: Great catch! I fixed this by renaming `fields.path` to `fields.slug`. I didn't rename anything else, `path` above, because it's not necessary, this `fields` option just has an unfortunate name, it's unrelated to `fields` in GraphQL query.

mainframev (approved):

CollierCZ (approved):

PR #2664, opened 2021-01-16
fix(IllustrationPrimitive): allow empty alt

mainframev:
[from AAAron ](https://skypicker.slack.com/archives/GSGN9BN6Q/p1610723763003800)<br/><br/><br/><url>LiveURL: https://orbit-fix-illustration-empty-string.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/primitives/IllustrationPrimitive/__tests__/index.test.js
+++ b/packages/orbit-components/src/primitives/IllustrationPrimitive/__tests__/index.test.js
@@ -24,11 +24,21 @@ describe("IllustrationPrimitive", () => {
 
     const img = screen.getByRole("img");
 
-expect(img).toHaveStyle({ maxHeight: "90px" });
+    expect(img).toHaveStyle({ maxHeight: "90px" });
     expect(screen.getByAltText("Alternative text")).toBeInTheDocument();
     expect(screen.getByTestId("test")).toBeInTheDocument();
     expect(img).toHaveAttribute("src", URL);
     expect(img).toHaveAttribute("srcset", expect.stringContaining(URL_RETINA));
     expect(img).toHaveStyle({ marginBottom: defaultTheme.orbit.spaceSmall });
+  });
+
+  it("should have empty alt", () => {
+    render(<IllustrationPrimitive name="Accommodation" alt="" />);
+    screen.getByAltText("");
+  });
+
+  it("should have name in alt", () => {
+    render(<IllustrationPrimitive name="Accommodation" />);
+    screen.getByAltText("Accommodation");
   });
 });
--- a/packages/orbit-components/src/primitives/IllustrationPrimitive/index.js
+++ b/packages/orbit-components/src/primitives/IllustrationPrimitive/index.js
@@ -48,14 +48,16 @@ const IllustrationPrimitive = ({
   size = SIZE_OPTIONS.MEDIUM,
   dataTest,
   spaceAfter,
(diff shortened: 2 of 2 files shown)


packages/orbit-components/src/primitives/IllustrationPrimitive/__tests__/index.test.js
```diff
@@ -24,11 +24,21 @@ describe("IllustrationPrimitive", () => {
...
     expect(screen.getByAltText("Alternative text")).toBeInTheDocument();
     expect(screen.getByTestId("test")).toBeInTheDocument();
     expect(img).toHaveAttribute("src", URL);
     expect(img).toHaveAttribute("srcset", expect.stringContaining(URL_RETINA));
     expect(img).toHaveStyle({ marginBottom: defaultTheme.orbit.spaceSmall });
   });
+
+  it("should have empty alt", () => {
+    render(<IllustrationPrimitive name="Accommodation" alt="" />);
+    expect(screen.getByAltText("")).toBeInTheDocument();
```
silvenon, line 37: Reminder: you don't have to do the `expect(...).toBeInTheDocument()` when you're using `GetBy` because `GetBy` queries themselves already throw when they don't find the element. You only need to use it when you're using `QueryBy`.

We have a ton of these in our codebase and it's nothing something you need to change here, we can later clean them up with a codemod, I'm just saying this so that you know for future.

packages/orbit-components/src/primitives/IllustrationPrimitive/index.js
```diff
@@ -48,14 +48,17 @@ const IllustrationPrimitive = ({
...
-  <StyledImage
-    illustrationName={name}
-    alt={alt || name}
-    size={size}
-    data-test={dataTest}
-    spaceAfter={spaceAfter}
-  />
-);
+}: Props) => {
+  const getAltText = alt === "" || (alt && alt.length > 0) ? alt : name;
```
silvenon, line 52: I'm having a hard time reading this logic. Could you write a `getAltText` function and use `if` statements to make the intent clearer?

silvenon (changes requested):

packages/orbit-components/src/primitives/IllustrationPrimitive/index.js
```diff
@@ -48,14 +48,22 @@ const IllustrationPrimitive = ({
...
-    alt={alt || name}
-    size={size}
-    data-test={dataTest}
-    spaceAfter={spaceAfter}
-  />
-);
+}: Props) => {
+  const getAltText = () => {
+    if (alt === "") return alt;
+    if (alt && alt.length > 0) return alt;
```
silvenon, line 54:
```suggestion
    if (typeof alt === "string") return alt;
```

packages/orbit-components/src/primitives/IllustrationPrimitive/index.js
```diff
@@ -48,14 +48,22 @@ const IllustrationPrimitive = ({
...
+    return name;
+  };
+
+  return (
+    <StyledImage
+      illustrationName={name}
+      alt={getAltText()}
```
silvenon, line 61: Looks like the logic could be simplified to this:

```jsx
alt={typeof alt === "string" ? alt : name}
```

It's absolutely unnecessary to hold this PR any further, I'm just saying this for educational purposes.

silvenon (approved):

PR #2671, opened 2021-01-18
feat(eslint-orbit): add unique-ids rule

mainframev:
I added new rule `unique-id`, for some [svg references ](https://developer.mozilla.org/en-US/docs/Web/SVG/Element). Did not include all, but most used. If you think we should add them more - can be discussed here. Also did not write test to check them all, not sure if it's needed. <br/><br/><br/><url>LiveURL: https://orbit-feat-new-eslint-rules.surge.sh</url>


packages/eslint-plugin-orbit-components/src/rules/svgUniqueId.ts
```diff
@@ -0,0 +1,88 @@
+import { Rule } from "eslint";
+import { types as t } from "@babel/core";
+
+export default {
+  create: (context: Rule.RuleContext) => {
```
silvenon, line 5: I also recommend adding the `meta` object describing the rule's `type` (IMO `"problem"`) and `docs`.

packages/eslint-plugin-orbit-components/src/rules/svgUniqueId.ts
```diff
@@ -0,0 +1,88 @@
...
+          if (!n.openingElement) return;
+          n.openingElement.attributes.forEach(attr => {
+            if (t.isJSXAttribute(attr) && t.isLiteral(attr.value)) {
+              const { value } = attr.value;
+              const count = IDS.filter(id => id === value).length;
+              if (count >= 2) {
+                context.report({
+                  // @ts-expect-error node
+                  node: n,
+                  message: `${attr.name.name}="${value}", should have unique id`,
```
silvenon, line 79: I think that the error message should recommend using the `useRandomID` hook that you can create as part of this PR, as described in the Jira ticket.

silvenon (commented):
I recommend two more changes:

1. try to extract this rule into a new internal package (`"private": true` in `package.json`) called something like `@kiwicom/eslint-plugin-orbit-internal`, however I don't know if this will work as easily as I hope 🤞
2. make the rule less sophisticated, it doesn't even need to be SVG-oriented because we want it to work for other elements as well, I recommend starting by checking `id` attributes of _any_ element or component, and you don't need to count them, the problematic part should be if the value of `id` is a `Literal`

To elaborate the 2nd suggestion: AFAICT, this rule wouldn't have caught the problems you ran into because the duplicate `id`s weren't in the same file, they were in the same HTML document because the same component was being rendered more than once, correct? This is why I suggest that the rule simply complains if the `id` value is a `Literal`, because statically speaking you cannot possibly know whether the component in that file is being rendered more than once, so you don't even have to try.

mainframev:
@silvenon resolved everything, seems to work fine

packages/eslint-plugin-orbit-internal/src/configs/recommended.ts
```diff
@@ -0,0 +1,12 @@
+const recommended = {
+  plugins: ["orbit-internal"],
+  rules: {
+    "orbit-internal/unique-id": "error",
+  },
+  parserOptions: {
+    sourceType: "module",
+    ecmaVersion: 2018,
+  },
```
silvenon, line 9: It doesn't seem like it should be the concern of this plugin's recommended config to set this. It might overwrite settings of a plugin before it.

packages/eslint-plugin-orbit-internal/src/rules/unique-id.ts
```diff
@@ -0,0 +1,42 @@
...
+      JSXAttribute(node: t.JSXAttribute) {
+        if (node.value) {
+          if (node.name.name === "id") {
+            JSXAttributes.push(node);
+          }
+        }
+      },
+
+      "Program:exit": () => {
+        JSXAttributes.forEach(n => {
```
silvenon, line 27: Just curious, would it be different if you put the logic that's inside this loop in the place of the `JSXAttributes.push(node)` above?

packages/eslint-plugin-orbit-internal/yarn.lock
```diff
@@ -0,0 +1,2093 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
silvenon, line 1: We should have a single `yarn.lock` file.

packages/orbit-components/src/SmartPassIllustrations/illustrations/SmartPassV1.js
```diff
@@ -325,7 +330,7 @@ const SmartPassV1 = ({ primary = "white", secondary = "black", ...props }: Props
       </g>
       <defs>
         <linearGradient
-          id="paint0_linear"
+          id={randomIDLinear}
```
silvenon, line 333: Wonderful, it already caught some problems! 🏆

silvenon (changes requested):
Awesome work! 💯

silvenon (approved):

PR #2679, opened 2021-01-22
feat(Tag): add forwardRef

mainframev:
[request in plz-orbit ](https://skypicker.slack.com/archives/CAMS40F7B/p1611232617003800)<br/><br/><br/><url>LiveURL: https://orbit-feat-tag-ref.surge.sh</url>

### DIFF
--- a/packages/orbit-components/src/Tag/index.js.flow
+++ b/packages/orbit-components/src/Tag/index.js.flow
@@ -18,6 +18,6 @@ export type Props = {|
   ...Globals,
 |};
 
-declare export default React.ComponentType<Props>;
+declare export default React.AbstractComponent<Props, HTMLDivElement>;
 
 declare export var StyledTag: ReactComponentStyled<any>;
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/Tag/index.js.flow
```diff
@@ -14,6 +14,7 @@ export type Props = {|
   +icon?: React.Node,
   +selected?: boolean,
   +size?: Size,
+  +ref?: Ref,
```
silvenon, line 17: Instead of this it's better to export a `React.AbstractComponent`, see `InputField/index.js.flow` as an example. There's no need for the `Ref` utility at all.

silvenon (changes requested):

packages/orbit-components/src/Tag/index.js.flow
mainframev: it's bad example :D there is spread Ref type, we should fix that

PR #2681, opened 2021-01-24
chore: use AbstractComponent instead of Ref type spread

mainframev:
Removed `...Ref` from `index.js.flow` files, where `AbstractComponent` is possible to use<br/><br/><br/><url>LiveURL: https://orbit-chore-remove-ref.surge.sh</url>


silvenon (approved):
Nice!

PR #2695, opened 2021-01-27
docs: rename Card component to Tile

silvenon:
To align with the Jira ticket, and it's a better name.
<br/><br/><br/><url>LiveURL: https://orbit-docs-rename-card-to-tile.surge.sh</url>


mainframev:
@silvenon we already have `Tile` component in `src` :)

mainframev (approved):

PR #2700, opened 2021-01-28
build: clean before building design tokens

silvenon:
This prevents mixing files between branches.

I recommend cherry-picking this to `dev/tokens` as well, that way switching back and forth won't cause unexpected problems. I had an error when I tried to open the documentation, but then remembered the possibility that the new and old files are mixed together, and deleting the directories worked.
<br/><br/><br/><url>LiveURL: https://orbit-build-design-tokens-clean.surge.sh</url>


vepor (approved):
I will rebase `dev/tokens` to master 😉   Thanks @silvenon

PR #2703, opened 2021-01-28
feat(tokens): add border radius tokens

vepor:
Added border radius tokens for foundation and global level.

### DIFF
--- a/packages/orbit-design-tokens/src/dictionary/transforms/values/index.ts
+++ b/packages/orbit-design-tokens/src/dictionary/transforms/values/index.ts
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import { Property } from "style-dictionary";
 
-import { isSpacing, isColor } from "../../utils/is";
+import { isSpacing, isColor, isBorderRadius } from "../../utils/is";
 import { errorTransform } from "../../utils/errorMessage";
 
 /*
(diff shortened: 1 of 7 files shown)


packages/orbit-design-tokens/src/dictionary/utils/is.ts
```diff
@@ -2,11 +2,14 @@ import { Property } from "style-dictionary";
...
+const isTypeOf = ({ type, attributes: { category } }: Property, typeName: string): boolean =>
+  typeName === type || typeName === category;
 
-export const isSpacing = ({ type, attributes: { category } }: Property): boolean =>
-  type === "spacing" || category === "spacing";
+export const isColor = (prop: Property): boolean => isTypeOf(prop, "color");
+
+export const isSpacing = (prop: Property): boolean => isTypeOf(prop, "spacing");
+
+export const isCornerRadius = (prop: Property): boolean => isTypeOf(prop, "corner-radius");
```
silvenon, line 12: Add test for `isCornerRadius`.

silvenon (changes requested):

packages/orbit-design-tokens/src/dictionary/transforms/values/index.ts
```diff
@@ -27,6 +27,22 @@ const stringJavascript = {
   },
 };
 
+/*
+  Transforms corner radius value to Javascript compatible format with pixels.
+ */
+const cornerRadiusJavascript = {
```
silvenon, line 33: Did you plan to rename this variable?

packages/orbit-design-tokens/src/dictionary/utils/__tests__/is.test.ts
```diff
@@ -1,9 +1,9 @@
-import { isColor, isSpacing, isInternal, isNotInternal } from "../is";
+import { isColor, isSpacing, isInternal, isNotInternal, isBorderRadius } from "../is";
 
-const typeColor = {
-  type: "color",
+const tokenPlaceholder = ({ type = "color", internal = true }) => ({
```
silvenon, line 3: It looks like you don't need these default values, and in tests it's better to be explicit about what you're testing.
```suggestion
const tokenPlaceholder = ({ type, internal }) => ({
```

silvenon (changes requested):

silvenon (approved):

PR #2716, opened 2021-02-04
feat(docs): add linting

CollierCZ:
Added linting of markdown code, prose, and links.

Some notes:
* Because [ESLint doesn't support async rules](https://github.com/mdx-js/eslint-mdx/issues/203), the check for validating links is separate from the other remark rules.
* That check still runs on staged files because it can detect missing assets and so on quickly before committing.
* Because @silvenon's method for link checking seems to be quick enough, I've used that instead of a GitHub action or other docker image. If it starts to seem slow (maybe with more content), we can switch or run it only on changed files (with a list from [conditional page builds](https://www.gatsbyjs.com/docs/conditional-page-builds/)).
* I tried using rehype-parse and the rest of the unified system for link checking (since we're already using it a bunch), but it didn't parse the HTML accurately (got empty results). Didn't think it was worth time trying to figure out if I could fix that.
* Going with Vale for prose because using remark linters would force errors when prose linting isn't 100% accurate.
 Orbit.kiwi: https://orbit-docs-docs-linting.surge.sh
 Storybook: https://orbit-docs-linting.surge.sh


.eslintrc.js
```diff
@@ -179,6 +179,44 @@ module.exports = {
         "@typescript-eslint/explicit-module-boundary-types": "off",
       },
     },
+    {
+      files: "*.mdx",
+      extends: ["plugin:mdx/recommended"],
+      rules: {
+        "react/jsx-filename-extension": ["error", { extensions: [".mdx"] }],
+        "mdx/no-unescaped-entities": "OFF",
```
silvenon, line 187: I also suggest setting `mdx/remark` to `"error"` ("recommended" config sets it to `"warn"`) because we want to prevent code with violations from being merged.

package.json
```diff
@@ -120,6 +121,13 @@
     "react-dom": "^16.14.0",
     "react-element-to-jsx-string": "^14.3.2",
     "react-test-renderer": "^16.14.0",
+    "remark-frontmatter": "^3.0.0",
```
silvenon, line 124: > This plugin is affected by the new parser in remark ([`micromark`](https://github.com/micromark/micromark), see [`remarkjs/remark#536`](https://github.com/remarkjs/remark/pull/536)). Use version 2 while you’re still on remark 12. Use version 3 for remark 13+.
>
> —`remark-frontmatter`'s [readme](https://github.com/remarkjs/remark-frontmatter#readme)

Our version of `remark` evaluations to 10.

```suggestion
    "remark-frontmatter": "^2.0.0",
```

silvenon (changes requested):

.eslintrc.js
CollierCZ: If it's set globally like that, I'd probably not use the prose linting that way. With Vale, we can set the error level for individual rules (some prose rules are more reliable than others, I've found).

silvenon: Sounds good 👍

lint-staged.config.js
```diff
@@ -2,7 +2,8 @@
 
 module.exports = {
   "*.{js,js.flow,ts}": "eslint --fix --report-unused-disable-directives",
-  "*.md": "prettier --write",
+  "*.{md,mdx}": "prettier --write",
```
silvenon, line 5: I'm pretty sure that our Prettier ESLint integration should apply to MDX, too. Try making a Prettier mistake in an `.mdx` file, then run `eslint --fix` on it. If it doesn't work correctly, you might need to set `parser` to `"markdown"` in the `prettier/prettier` rule for `*.mdx` files.
```suggestion
  "*.md": "prettier --write",
```

lint-staged.config.js
```diff
@@ -2,7 +2,8 @@
...
+  "*.mdx": "eslint --report-unused-disable-directives",
```
silvenon, line 6: To apply Prettier (if indeed it does work):
```suggestion
  "*.mdx": "eslint --fix --report-unused-disable-directives",
```

silvenon (commented):
I just remembered one more thing.

lint-staged.config.js
silvenon: Actually, `parser` should be `"mdx"`, not `"markdown"`. Hopefully that happens automatically, though. 🤞

lint-staged.config.js
CollierCZ: The reason I'd not done this was cause I wasn't sure that the fixes from remark-lint would be applied properly. I guess we can try it out and see.

lint-staged.config.js
CollierCZ: It worked from the beginning. Only reason to avoid it was as below, not being sure about fixes on remark rules. The parser seemed automatic.

docs/src/documentation/05-development/01-guides/04-no-exporting-components.mdx
```diff
@@ -77,14 +77,14 @@ and would require a lot of work in users' repositories.
...
+This, in the end, could lead to inconsistent behavior or visual style for our applications,
 even though we can't prohibit all possibilities when it comes to custom CSS overrides.
 
-### Extending functionality to styled-components requires running more JS code
+### Extended functionality requires running more JS code
 
 Given the nature of how styled-components work,
 allowing extension could lead to unnecessary complexity in UI components
-and also more time-consuming runtimes.
+and also more time-consuming run times.
```
silvenon, line 87: "Runtime" is a valid programming term, although I'm not sure about the use of plural.

docs/src/documentation/06-kiwi-use/02-content/04-specific-areas/social-media/02-visual-treatment.mdx
```diff
@@ -14,12 +15,15 @@ Our photography should always have a sense of optimism.
...
-Lighting, color and tone should be **unfiltered and realistic \#NoFilter!**
+Lighting, color, and tone should be **unfiltered and realistic \#NoFilter!**
 
 Each image should establish an emotional connection with our audience,
 showing them the world Kiwi.com can offer.
 
+<!-- prettier-ignore-start -->
+<!-- eslint-disable-next-line -->
 ## Our destinations...
+<!-- prettier-ignore-end -->
```
silvenon, line 26: There are several more instances of this:
```suggestion
<!-- eslint-disable -->
## Our destinations...
<!-- eslint-enable -->
```

docs/src/documentation/06-kiwi-use/02-content/04-specific-areas/social-media/02-visual-treatment.mdx
```diff
@@ -14,12 +15,15 @@ Our photography should always have a sense of optimism.
...
 
 **ARE** real, authentic and full of wanderlust.
```
silvenon, line 28: Vale warnings complain about words like "ARE" and "NOT", and considering that they are already in bold, no need for all caps, but your choice 😉

docs/src/documentation/06-kiwi-use/02-content/04-specific-areas/technical-content/04-language.mdx
```diff
@@ -149,18 +162,20 @@ They make it clear what steps need to be done when.
 
 Don't weaken this guidance by numbering other things where the order doesn't matter.
 For example, headings in an article should follow a clear semantic structure.
-That structure should be enough, so the headings should not be numbered.
+That structure should be enough, so the headings should't be numbered.
```
silvenon, line 165: Or [shorn't](https://youtu.be/jboRRpAti6Q).
```suggestion
That structure should be enough, so the headings shouldn't be numbered.
```

silvenon (changes requested):

.github/contribution/docs.md
```diff
@@ -0,0 +1,113 @@
...
+Now you can make a change to a file in `docs/src/` (such as the docs in `docs/src/documentation/`)
+and your site updates automatically.
+
+## Code style
+
+To keep the files used to build our docs clear, we enforce standard Markdown code styling.
+
+We use [ESLint](https://eslint.org/) to enforce the style for our MDX files
+(just as we do for our JavaScript and TypeScript files).
+The specific rules are part of an extension for `plugin:mdx`,
```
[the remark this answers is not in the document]
CollierCZ: The idea was connected to the next line -- to help people find the settings specific to docs in the very long ESLint settings file. But I can make that more clear.

.github/contribution/docs.md
```diff
@@ -0,0 +1,113 @@
...
+
+But we don't enforce a strict limit to line length.
+If we did, automatic tools would rearrange lines at any change.
+Then reviews would be more difficult:
+`git diff` would show changes based on formatting rather than meaning.
+
+So we try to break lines at semantic breaks.
+This means having new sentences on new lines.
+This should help keep the differences for review about semantics,
+and not formatting.
```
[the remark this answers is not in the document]
CollierCZ: I didn't want something that even reported violations, which is why I stayed away from enforcing anything. I'd be OK with warnings, but errors seemed unnecessary for a rule that should be based on semantics (which simple parsing isn't great at). And I don't think we can get warnings this way

.github/contribution/docs.md
```diff
@@ -0,0 +1,113 @@
...
+we also lint to ensure the docs follow our [style guide](https://orbit.kiwi/content/voice-tone/)
+as much as possible.
+
+For this, we use [Vale](https://docs.errata.ai/vale/about).
+Our specifics can be found in the [styles directory](../styles),
+with rules in the `Kiwi` folder and accepted spellings in [accept.txt](../styles/Vocab/Kiwi/accept.txt).
+
+To check your prose yourself, [install Vale](https://docs.errata.ai/vale/install).
+Or you can create a pull request
+and the Vale GitHib Action adds comments pointing out places that could be improved.
```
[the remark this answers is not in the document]
CollierCZ: Seems like linting should have helped with that. 🤦

.github/workflows/lint-prose.yml
```diff
@@ -0,0 +1,25 @@
+name: Lint prose
+
+on:
+  push:
+    branches-ignore:
+      - main
+  pull_request:
+    branches:
+      - main
```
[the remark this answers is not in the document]
CollierCZ: I was going to suggest something like paths even for builds. Because we don't need to be rebuilding docs everytime a component changes, nor rebuilding components for docs. But I can update this one for sure too. Would probably also include md files in other paths, as they can be included in docs and linting.

docs/src/documentation/05-development/01-guides/04-no-exporting-components.mdx
CollierCZ: Didn't seem like it in plural, so I opted for something that passed. But I can shift it back to singular.

.github/styles/Kiwi/Acronyms.yml
```diff
@@ -0,0 +1,64 @@
...
+  - TCP
+  - TODO
+  - URI
+  - URL
+  - USB
+  - UTF
+  - XML
+  - XSS
+  - YAML
+  - ZIP
```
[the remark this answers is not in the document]
CollierCZ: Some of those I think should be clear to the target audience. And some I thought I ignored already (because they're examples of what _not_ to do. It's partly that the suggestion is only a suggestion, so I'm OK with ignoring the suggestion when it makes sense to me. 

Also, you missed BCE 😁

docs/src/documentation/06-kiwi-use/02-content/04-specific-areas/social-media/02-visual-treatment.mdx
CollierCZ: I know I fixed that at some point. Must have rebased my way out of it. 😞

docs/src/documentation/06-kiwi-use/02-content/04-specific-areas/social-media/02-visual-treatment.mdx
CollierCZ: I also disliked the warnings (and the style before even seeing the warnings). But I didn't create the text, so I felt bad adjusting it. But if Vale says it, it must be true, so I I guess I should change it.

silvenon (approved):

.eslintrc.js
```diff
@@ -190,6 +190,46 @@ module.exports = {
         "@typescript-eslint/prefer-readonly-parameter-types": "off",
       },
     },
+    {
+      files: "*.mdx",
+      extends: ["plugin:mdx/recommended"],
+      rules: {
+        "react/jsx-filename-extension": ["error", { extensions: [".mdx"] }],
+        "mdx/no-unescaped-entities": "OFF",
```
JounQin, line 198: Hi, I’m the author of `eslint-mdx`, so glad that you’re using it.

Just to confirm, is there any problem with this rule or you’d just want to ignore it?

.eslintrc.js
```diff
@@ -190,6 +190,46 @@ module.exports = {
...
+        "import/extensions": [
+          "error",
+          "ignorePackages",
+          {
+            tsx: "never",
+          },
+        ],
+      },
+      settings: {
+        "import/resolver": {
```
JounQin, line 209: [`eslint-import-resolver-typescript`](https://github.com/alexgorbatchev/eslint-import-resolver-typescript) should be preferred for ts files.

.github/contribution/docs.md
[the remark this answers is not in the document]
JounQin: @CollierCZ `eslint-plugin-mdx` supports `.md` files at the same time with `remark-lint` under the hood.

lint-staged.config.js
```diff
@@ -3,6 +3,10 @@
 module.exports = {
   "*.{js,js.flow,ts}": "eslint --fix --report-unused-disable-directives",
   "*.md": "prettier --write",
+  "*.mdx": [
```
JounQin, line 6: `.md` can be merged into this one

silvenon: First of all, you're awesome for reviewing this PR. ❤️ 

What do you mean by this, though? We use Prettier through ESLint, so it`*.mdx` doesn't need to be prettified additionally, unlike `*.md` files which _only_ use Prettier.

.eslintrc.js
silvenon: I agree even our `*.ts?(x)` files don't seem to have the resolver set up, I can't remember what was the problem I ran into… 🤔

lint-staged.config.js
JounQin: Why `remark-lint` related rules are not enabled for `.md` files?

.eslintrc.js
CollierCZ: Hi! Thanks so much for having a look at our configuration. When the rule is turned off, I get a lot of warnings about apostrophes and the like being able to be escaped. I don't want to escape things in markdown.

These cases are markdown inside components ([example](https://github.com/kiwicom/orbit/pull/2716/files#diff-97ba94fb83d933c8f2c77510e1b46bcb1d49f4376e86aa275c195a25301a9b2eR24). I turned it off to get only relevant warnings.

JounQin: That's interesting, I thought `mdx` as `jsx` more than `md`, but users like you may think oppositely. So I'm considering disable it by default in the recommended config.

CollierCZ: I don't want to speak for others, but in my case I definitely think of it as `md` with extra features. But I'm a writer first and a programmer second.

lint-staged.config.js
CollierCZ: That's just a temporary thing, probably. We were focusing on the new docs that we're adding, which are all `.mdx`. We can then go over the older docs we have and see how the rules and everything works. Didn't want to fix a bunch of errors without knowing if we liked the workflow. 🙂

PR #2719, opened 2021-02-04
Orbit kiwi deployment

mainframev:
Added deploy of Orbit.kiwi on surge, same as we have it with storybook. Updated `Octokit`, because it had update with 3 major versions with breaking changes. Found bug with `readme` inside `Popover`. 
 Orbit.kiwi: https://orbit-docs-chore-deploy-action.surge.sh
 Storybook: https://orbit-chore-deploy-action.surge.sh


.github/workflows/test.yml
```diff
@@ -56,17 +56,26 @@ jobs:
...
+        run: |
+          yarn orbit-components deploy:surge https://orbit-${BRANCH_URL}.surge.sh --token ${{ secrets.SURGE_TOKEN }}
+          yarn orbit-kiwi deploy:surge https://orbit-docs-${BRANCH_URL}.surge.sh --token ${{ secrets.SURGE_TOKEN }}
 
-      - name: Update PR description
-        run: yarn orbit-components deploy:updateURL ${BRANCH_NAME} https://orbit-${BRANCH_URL}.surge.sh ${{ secrets.OCTO_TOKEN }}
+      - name: Update description
+        run: |
+          yarn orbit-components deploy:updateURL ${PR_NUMBER} https://orbit-docs-${BRANCH_URL}.surge.sh ${{ secrets.OCTO_TOKEN }} DOCS
+          yarn orbit-components deploy:updateURL ${PR_NUMBER} https://orbit-${BRANCH_URL}.surge.sh ${{ secrets.OCTO_TOKEN }}
```
silvenon, line 81:
```suggestion
          yarn orbit-components deploy:updateURL ${PR_NUMBER} https://orbit-${BRANCH_URL}.surge.sh ${{ secrets.OCTO_TOKEN }} Storybook
```

silvenon (dismissed):

docs/src/documentation/03-components/04-overlay/popover/02-react.mdx
```diff
@@ -2,6 +2,6 @@
 title: React
 ---
 
-import PopoverReadMe from "@kiwicom/orbit-components/src/Popover/README.md"
+import PopoverReadMe from "@kiwicom/orbit-components/src/Popover/README.md";
```
CollierCZ, line 5: Why only change this? It was the pattern everywhere to use `src` and it works fine for each page. (In part because of how files are included in the monorepo, so we don't need to source them from the published package.) If it's going to be changed, it should be changed everywhere.

PR #2724, opened 2021-02-08
refactor(tokens): implementing new namespace/object/variant structure

vepor:


packages/orbit-design-tokens/src/dictionary/transforms/attributes/index.ts
```diff
@@ -3,73 +3,63 @@ import { Attributes, Property } from "style-dictionary";
...
-  For other types:
   attributes: {
-    category: "base",
-    name: "space",
-    type: "XXS",
-    // default is not present
+    namespace: "foundation" (or "global", "variant"),
+    object: "palette" (or e.g. "border-radius"),
+    variant: "blue" (or e.g. "small"),
+    subVariant: "light" (nullish for most properties),
```
silvenon, line 14: Would this be more correct?
```suggestion
    subVariant: "light" (nullish for most variants),
```

silvenon (approved):

packages/orbit-design-tokens/src/dictionary/transforms/attributes/index.ts
vepor: Yes it would! 🙂 Thanks for the suggestion.

silvenon (approved):

PR #2726, opened 2021-02-08
feat(docs): internal github contributors plugin

mainframev:
Not sure how to solve it differently. I tried to explicitly type schema, but it's remote schema, so it did not help 🤔 . Tried to wrap remote schema with `graphql-tools`, but without any result for now. Here is another approach, it's possible to fetch github users/contributors without any token at all, but number of allowed requests is limited, so it will cause error anyway, i deleted the plugin we used for fetching contributors, replaced it with custom small fetching internal plugin and provided fallback for schema, it works if token is not correct (401). 

 Orbit.kiwi: https://orbit-docs-fix-contributors-gql.surge.sh
 Storybook: https://orbit-fix-contributors-gql.surge.sh


docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,56 @@
...
+          content: JSON.stringify(fallback),
+          contentDigest: createContentDigest(fallback),
+        },
+      });
+    } else {
+      console.error(error);
+    }
+  }
+  return undefined;
+};
```
silvenon, line 56: Instead of all of this this custom plugin could serve for providing an empty array as the GraphQL response that `gatsby-source-graphql` provides. Then in `gatsby-config.js` you'd check whether `GH_TOKEN` exists, if it does configure `gatsby-source-graphql`, if it doesn't configure the custom plugin. (Then in the component you'd check whether the array is empty etc. etc.)

What do you think?

silvenon (changes requested):
Instead of people having to define an empty `GH_TOKEN=` in `.env`, also in the future other environment variables that we don't need for this can fail, I recommend catching `dotenv-safe`'s error and checking its [`missing` property](https://github.com/rolodato/dotenv-safe/blob/1ab395c2045bc8a2737da1a683cb6d56605a08f4/MissingEnvVarsError.js#L15):

```js
try {
  require("dotenv-safe").config({
    example: path.resolve(__dirname, `../.env.example`),
    path: path.resolve(__dirname, `../.env`),
  });
} catch (err) {
  if (err.missing.includes("GH_TOKEN")) {
    // print out warning in the console that it's not a big deal,
    // but that a few components might not be visible, and link to the GitHub docs
    // about how to create a token and define which scope is needed
    // (the same thing you'll print out in the Alert component)
  }
}
```

Another thing, for some reason I can't see avatars anymore, could you check out why? 🤔

docs/plugins/contributors/fallback.json
```diff
@@ -0,0 +1,34 @@
...
+  "hireable": "",
+  "bio": "",
+  "twitter_username": "",
+  "public_repos": 5,
+  "public_gists": 1,
+  "followers": 9,
+  "following": 6,
+  "created_at": "2014-04-10T18:16:16Z",
+  "updated_at": "2021-01-27T10:58:04Z"
+}
```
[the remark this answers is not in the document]
mainframev: it's needed, you can't run gatsby website with that error, because it will fail on graphql query, which will try to use non-existing schema 🤔

docs/plugins/contributors/gatsby-node.js
mainframev: i don't think we should have them both, custom plugin or `gatsby-source-graphql`, it's same logic twice, custom plugin can be run without providing token at all, that is the difference with `gatsby-source-graphql`, which needs that token in `gatsby-config.js`. 🤔

silvenon (commented):
Also, I noticed that the `Image` component in the console complains about the wrong `fluid` value, could you see why that happens? 🙏

docs/plugins/contributors/fallback.json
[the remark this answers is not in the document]
CollierCZ: Dummy data is necessary if you aren't going to customize the schema. Otherwise, GraphQL won't be able to run the query. I don't think we need any fields that we aren't querying for, though.

But if you're going to go to the trouble of creating the dummy data, it could easily be replicated in a schema customization. Then if there's a null value/empty array, you can use the warning Alert. I think that kind of experience would be easier for developers.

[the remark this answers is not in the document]
mainframev: @CollierCZ you mean typing it explicitly how you suggested before? how we can achieve that with remote schema?

CollierCZ:
Separate question, but should we standardize where data is stored? This component keeps it in the same folder as the TypeScript files, while others use the `data` directory. Should it be consistent or doesn't it matter?

mainframev:
> Separate question, but should we standardize where data is stored? This component keeps it in the same folder as the TypeScript files, while others use the `data` directory. Should it be consistent or doesn't it matter?

I think it should be stored inside that component, because it's not supposed to be used somewhere else in the project, only that component needs that

docs/plugins/contributors/fallback.json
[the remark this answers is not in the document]
mainframev: yep, my fault, i tried to extend external graphql schema, so it did not work, but when i switched to rest api, i forgot to customize it and choosed that way with creating dummy node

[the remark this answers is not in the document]
CollierCZ: I'm still unsure why you can't define the schema directly, as we know what fields we're looking for. And I'm pretty sure [external schema get merged](https://www.gatsbyjs.com/docs/schema-generation/#6-merging-in-third-party-schemas) into the Gatsby schema. But if it works, it works.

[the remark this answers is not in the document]
mainframev: Did you check the last commit? It's explicitly defined, everything is ok, but it's not external schema anymore, since we don't use `graphql-api`, we use rest api now, that is why it works, i think, i did same before with external and it was not working.

CollierCZ:
> > Separate question, but should we standardize where data is stored? This component keeps it in the same folder as the TypeScript files, while others use the `data` directory. Should it be consistent or doesn't it matter?
> 
> I think it should be stored inside that component, because it's not supposed to be used somewhere else in the project, only component needs that

I can understand that. Just viewed it as potentially a separation of concerns, so the content would be in one place. That way it could be updated separately from the code more easily. But it's not a big deal either way. I just like consistency.

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,70 @@
...
+        internal: {
+          type: NODE,
+          content: JSON.stringify(u),
+          contentDigest: createContentDigest(u),
+        },
+      });
+    });
+  } catch (error) {
+    if (error.missing.includes("GH_TOKEN")) {
+      console.warn(`token is missign, please create one:
```
CollierCZ, line 62:
```suggestion
      console.warn(`Missing an access token for GitHub. Please create one:
```

docs/src/components/Github/index.tsx
```diff
@@ -1,47 +1,67 @@
...
 
-  return (
-    <>
-      {github.repository.collaborators.nodes.map(({ id, ...info }) => (
+  const githubDocs = (
+    <a
+      href="https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token"
+      aria-label="creating-a-personal-access-token"
+    >
+      please, create one
```
CollierCZ, line 48:
```suggestion
      Please create one
```

docs/src/components/Github/index.tsx
```diff
@@ -1,47 +1,67 @@
...
+    <Grid>
+      {allContributor.nodes.map(({ id, ...info }) => (
         <Contributor key={id} id={id} {...info} />
       ))}
-    </>
+    </Grid>
+  ) : (
+    <Alert type="warning">
+      <Text>
+        Contributors component is not shown, because you are missing github personal access token,{" "}
```
CollierCZ, line 61:
```suggestion
        Contributors are not shown because you're missing an access token for GitHub.{" "}
```

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,70 @@
...
+const NODE = `contributor`;
+
+exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }, { repo, owner }) => {
+  const { createNode, createTypes } = actions;
+
+  createTypes(
+    `
+    type Contributor implements Node {
+      id: ID!
+      login: String
```
CollierCZ, line 13: Why include fields we're not querying for?

mainframev: it does not make any perfomance difference, if token is present, there are all fields there, i've added all possible we might use

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,78 @@
...
+        "Token is needed to display Contributors component, which is based on github api",
+        "How to create github personal token: https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token",
+        "Do not forget to include required token scopes: all repo scopes and read:org",
+        "After you will create token in github, do not forget to create .env file in the root folder with GH_TOKEN=<YOUR TOKEN>",
+      );
+    } else {
+      console.error(error);
+      console.info(
+        "It is possible that you forgot to include repo and read:org scopes in your personal github token",
+      );
```
silvenon, line 73: Wow, this is a great addition, thanks for this! 🥇

silvenon (approved):
From my perspective it's all good, but Aaron should probably approve as well.

docs/plugins/contributors/gatsby-node.js
CollierCZ: My concern would be that they could change the fields which would cause conflicts and breakage for things we don't need. I don't suppose it's all that likely, it just seemed unnecessary. 🤷

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,78 @@
...
+        internal: {
+          type: NODE,
+          content: JSON.stringify(u),
+          contentDigest: createContentDigest(u),
+        },
+      });
+    });
+  } catch (error) {
+    if (error.missing.includes("GH_TOKEN")) {
+      console.warn("Missing an access token for GitHub. Please create one");
```
CollierCZ, line 62:
```suggestion
      console.warn("Missing an access token for GitHub. Please create one.");
```

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,78 @@
...
+      console.info(
+        "Token is needed to display Contributors component, which is based on github api",
+        "How to create github personal token: https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token",
+        "Do not forget to include required token scopes: all repo scopes and read:org",
+        "After you will create token in github, do not forget to create .env file in the root folder with GH_TOKEN=<YOUR TOKEN>",
```
CollierCZ, line 67:
```suggestion
        "The token is needed to display the Contributors component, which is based on the GitHub API.",
        "Create a personal access token: https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token",
        "Don't forget to include the required scopes for the token: all repo scopes and read:org.",
        "After creating a token in GitHub, add it to a .env file in the root folder with GH_TOKEN=<YOUR TOKEN>",
```

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,78 @@
...
+      );
+    } else {
+      console.error(error);
+      console.info(
+        "It is possible that you forgot to include repo and read:org scopes in your personal github token",
```
CollierCZ, line 72:
```suggestion
        "You may have forgotten to include the repo and read:org scopes for your GitHub access token.",
```

docs/plugins/contributors/gatsby-node.js
```diff
@@ -0,0 +1,78 @@
...
+        children: [],
+        internal: {
+          type: NODE,
+          content: JSON.stringify(u),
+          contentDigest: createContentDigest(u),
+        },
+      });
+    });
+  } catch (error) {
+    if (error.missing.includes("GH_TOKEN")) {
```
CollierCZ, line 61: Not sure if it's a use case we want to cover, but if an _empty_ GH_TOKEN is included, this fails and does not give any extra information (though I'm directed there by the Alert).

docs/src/components/Github/index.tsx
```diff
@@ -1,47 +1,68 @@
...
         <Contributor key={id} id={id} {...info} />
       ))}
-    </>
+    </Grid>
+  ) : (
+    <Alert type="warning">
+      <Text>
+        Contributors component is not shown, because you are missing github personal access token
+        {githubDocs}
+        Check the terminal for more details
```
CollierCZ, line 63:
```suggestion
        Contributors component is not shown, because you are missing github personal access token.{" "}
        {githubDocs}{" "}
        Check the terminal where you started the site for more details.
```

CollierCZ (changes requested):
A private plugin feels inherently more fragile to me, but it seems to be working fine. I made some suggestions for the wording and possibly to consider another case. But otherwise seems good to me.

docs/plugins/contributors/gatsby-node.js
CollierCZ: Are these definitely the scopes needed? I thought I gave my token only read:org. Did you test it?

mainframev:
> A private plugin feels inherently more fragile to me, but it seems to be working fine. I made some suggestions for the wording and possibly to consider another case. But otherwise seems good to me.

It all started after the token problem, which is not a problem at all in my opinion, it's enough for us to add some description in docs about how to use token and which right scopes should be enabled, we are using it for a quite long time already (for releases) and it can be problem only with newcomer developer. 🤔  

Thank you for review :) It was helpful. In the end small internal plugin is not bad decision, we covered additional use cases and it works fine

PR #2730, opened 2021-02-11
feat(tokens): rename spacing scale to proportion based

vepor:
Rename spacing tokens to proportion based format.

I had the change the proposed schema, because it will be very limiting to use `2-x` (or camelCased variant `2x`) as a token name. 
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Identifier_after_number

The problem would be mainly noticeable in the foundation usage, e.g. if somebody would like to change the spacings token via `createTheme` function he would have to pass these properties in different way – which is not that common when comes to usage of object in Javascript. It would be also limitation for us. See further code examples.

The definition for spacing token would like this:

```js
const space = {
  "2x": "value",
  "4x": "value",
  ...
}
```

Then also the mapping from foundation to global/component-specific tokens would be a bit tricky, all the time we would have to ensure that the value is properly referenced in the Javascript format, e.g.:

```js
// you can't use `space2x: foundation.space.2x`
const tokens = {
  space2x: foundation.space["2x"],
  space4x: foundation.space["4x"],
  ...
}
```

**Therefore it would be better to use verbal values instead of numbers for the multiplier or fraction, e.g. `one-x`, `half-x`**

The last change I have made was to split the `attribute/nov` camelCase string conversion to specific attribute transform, so it doesn't do this side effect.

### DIFF
--- a/packages/orbit-design-tokens/src/dictionary/definitions/foundation/space.json
+++ b/packages/orbit-design-tokens/src/dictionary/definitions/foundation/space.json
@@ -1,44 +1,44 @@
 {
   "foundation": {
     "space": {
-      "XXS": {
+      "half-x": {
+        "type": "spacing",
+        "value": 2,
+        "internal": true
+      },
+      "one-x": {
         "type": "spacing",
         "value": 4,
         "internal": true
       },
-      "XS": {
+      "two-x": {
         "type": "spacing",
         "value": 8,
         "internal": true
       },
-      "S": {
+      "three-x": {
         "type": "spacing",
         "value": 12,
         "internal": true
       },
-      "M": {
+      "four-x": {
         "type": "spacing",
         "value": 16,
         "internal": true
       },
-      "L": {
+      "six-x": {
         "type": "spacing",
         "value": 24,
         "internal": true
       },
-      "XL": {
+      "eight-x": {
         "type": "spacing",
         "value": 32,
         "internal": true
       },
-      "XXL": {
+      "ten-x": {
         "type": "spacing",
         "value": 40,
-        "internal": true
-      },
-      "XXXL": {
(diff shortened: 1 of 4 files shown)


packages/orbit-design-tokens/src/dictionary/definitions/foundation/space.json
```diff
@@ -1,45 +1,45 @@
...
         "internal": true
       },
-      "XL": {
+      "eight-x": {
         "type": "spacing",
         "value": 32,
         "internal": true
       },
-      "XXL": {
+      "ten-x": {
```
silvenon, line 39: [10x developer](https://twitter.com/search?q=10x%20developer&src=typed_query)! 💪

packages/orbit-design-tokens/src/dictionary/definitions/foundation/space.json
```diff
@@ -1,45 +1,45 @@
...
         "type": "spacing",
         "value": 40,
         "internal": true
-      },
-      "XXXL": {
-        "type": "spacing",
-        "value": 52,
-        "internal": true
```
silvenon, line 42: Why is this one being removed? 🤔

silvenon (dismissed):
I don’t like how this error dictates what the API will be 😕  But I can see that it could be a very uncomfortable problem.

packages/orbit-design-tokens/src/dictionary/definitions/foundation/space.json
vepor: For spacings the `52px` is not being used that much. Talked with @williamkolmacka and we already had plan to remove it from spacing scale.

PR #2732, opened 2021-02-11
chore: gatsby-styled-components

mainframev:
Orbit.kiwi: https://orbit-docs-chore-gatsby-svg-plugin.surge.sh
 Storybook: https://orbit-chore-gatsby-svg-plugin.surge.sh


silvenon (commented):
@mainframev could you just expand the description of this PR with what problem it is solving? 🙏

PR #2735, opened 2021-02-12
fix: scope internal packages to avoid security bug

silvenon:
Our internal workspace `eslint-plugin-orbit-internal` is vulnerable to the npm dependency confusion attack, so the current solution is to scope the package name (along with `orbit.kiwi` just in case), because other people cannot publish `@kiwicom` packages.

ESLint apparently doesn't support adding recommended config of scoped plugins (I wasn't able to confirm it yet, though), so I had to work around it.

 Orbit.kiwi: https://orbit-docs-fix-scope-internal-packages.surge.sh
 Storybook: https://orbit-fix-scope-internal-packages.surge.sh


mainframev:
You have to update name of the rule inside config too :)

package.json
```diff
@@ -91,7 +91,7 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jest": "^24.1.3",
     "eslint-plugin-jsx-a11y": "^6.4.1",
-    "eslint-plugin-orbit-internal": "^0.1.0",
+    "@kiwicom/eslint-plugin-orbit-internal": "^0.1.0",
```
RobinCsl, line 94: You could also use to make sure there would be no version mismatch: if it finds the package locally, it will install it, regardless of the version (whereas it might look for it on NPM if it can't find the right version locally).
```suggestion
    "@kiwicom/eslint-plugin-orbit-internal": "*",
```

vepor (approved):
LGTM 💪

PR #2736, opened 2021-02-12
docs: add FancyLink component

CollierCZ:
and add as shortcode.

I was going to add svgs for the other services, but they were getting to be a lot and I wasn't sure about the legality of hosting them under our license, so I just added a package for it.

 Orbit.kiwi: https://orbit-docs-docs-fancylink.surge.sh
 Storybook: https://orbit-docs-fancylink.surge.sh


docs/src/components/FancyLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+import styled from "styled-components";
+import { Figma, Github, ReactJs, Spectrum, Storybook } from "@icons-pack/react-simple-icons";
+import { ColorPicker, Download, FlightDirect, Google, Tips } from "@kiwicom/orbit-components/icons";
+import { Tile } from "@kiwicom/orbit-components";
+
+import useIsUrlExternal from "../hooks/useIsUrlExternal";
+
+interface FancyLinkProps {
+  download: boolean;
+  icon: string | React.ReactNode;
```
mainframev, line 11: `React.ReactNode` will be sufficient here, because it has `string`

docs/src/components/FancyLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+      case "storybook":
+        usedIcon = <Storybook />;
+        break;
+      case "lightbulb":
+        usedIcon = <Tips />;
+        break;
+      default:
+        usedIcon = <FlightDirect />;
+    }
+  }
```
mainframev, line 59: could you rewrite it to a function with switch/case inside? and instead of breaks just return the icon

docs/src/components/FancyLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+  display: block;
+`;
+
+const FancyLink = ({ download, href, icon, title }: FancyLinkProps) => {
+  const isExternal = useIsUrlExternal(href);
+
+  let usedIcon = icon || <FlightDirect />;
+
+  // If not passing an icon
+  if (typeof icon === "string") {
```
mainframev, line 27: in the interface you have `string | React.Node` i suggest you to left it only as `React.Node` or event better will be to type these string cases as union and have it something like that `Icons | React.Node`.

docs/src/components/FancyLink.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import styled from "styled-components";
+import { Figma, Github, ReactJs, Spectrum, Storybook } from "@icons-pack/react-simple-icons";
+import { ColorPicker, Download, FlightDirect, Google, Tips } from "@kiwicom/orbit-components/icons";
+import { Tile } from "@kiwicom/orbit-components";
+
+import useIsUrlExternal from "../hooks/useIsUrlExternal";
+
+interface FancyLinkProps {
```
mainframev, line 9: no need to name it specifically `FancyLinkProps`, there is only one component here, so just `Props` as we usually do with components :)

docs/src/components/FancyLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+  download: boolean;
```
mainframev, line 10: i suggest you to have `download` as optional, because it's better to add it when it's needed, rather than add `false` to prop explicitly ;)

mainframev (commented):
LGTM. Just a few comments :)

docs/src/components/FancyLink.tsx
CollierCZ: Certainly better. Was intended. 🙈

mainframev (approved):

PR #2737, opened 2021-02-12
feat(eslint-plugin-orbit): default-theme rule

mainframev:
Added new rule, called it `default-theme`, as we discussed it earlier, that we should prevent some cases of wrong defaultTheme usage

 Orbit.kiwi: https://orbit-docs-feat-defaultTheme-eslint.surge.sh
 Storybook: https://orbit-feat-defaultTheme-eslint.surge.sh

### DIFF
--- a/packages/eslint-plugin-orbit-components/__tests__/defaultTheme.test.ts
+++ b/packages/eslint-plugin-orbit-components/__tests__/defaultTheme.test.ts
@@ -1,0 +1,93 @@
+import ruleTester from "../ruleTester";
+import defaultTheme, { ERRORS } from "../src/rules/defaultTheme";
+
+describe("defaultTheme", () => {
+  ruleTester.run("defaultTheme", defaultTheme, {
+    valid: [
+      {
+        code: `
+          import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";\
+
+          const Wrapper = styled.div\`
+            display: \${({ theme }) => theme.orbit.paletteWhite};
+          \`;
+
+          Wrapper.defaultProps = {
+            theme: defaultTheme,
+          }
+        `,
+      },
+      {
+        code: `
+          import { defaultTheme } from "@kiwicom/orbit-components";\
+
+          const Wrapper = styled.div\`
+            display: \${({ theme }) => theme.orbit.paletteWhite};
+          \`;
+
+          Wrapper.defaultProps = {
+            theme: defaultTheme,
+          }
+        `,
+      },
+      {
+        code: `
+          import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";\
+
+          <ThemeProvider theme={defaultTheme}>
+            <BrandProvider value={brand}>
+              <FetchedProvider value={fetched}>
+                <LogProvider value={{ log: (e, p) => cuckoo.nitro(e, p) }}>
+                  <InitRelayEnvironment clientID="frontend">{children}</InitRelayEnvironment>
+                </LogProvider>
+              </FetchedProvider>
+          </BrandProvider>
+        </ThemeProvider>
+        `,
+      },
+    ],
+    invalid: [
+      {
+        code: `
+        import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";\
+
+          const failedStatus = {
+            isActive: true,
+            activeColor: defaultTheme.orbit.paletteOrangeNormal /* #f9971e */,
+            node: <CloseCircle color="warning" />,
+            content: getContentPerStatus(activeStatus, true),
+          }
+        `,
+        errors: [ERRORS.variableDeclaration],
+      },
+      {
+        code: `
+          import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";\
+
+          const Wrapper = styled.div\`
+            color: $\{defaultTheme.orbit.paletteProductNormal};
+          \`
+        `,
+        errors: [ERRORS.styled],
+      },
+      {
+        code: `import theme from "@kiwicom/orbit-components/lib/defaultTheme";`,
+        errors: [ERRORS.import],
+      },
+      {
+        code: `
+          import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";
+
+          const {
+            borderWidthCard,
+            borderStyleCard,
+            borderColorCard,
+            borderRadiusNormal,
+            spaceMedium,
+          } = defaultTheme.orbit;
+        `,
+        errors: [ERRORS.destructured],
+      },
+    ],
+  });
+});
--- a/packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
+++ b/packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
@@ -1,0 +1,64 @@
+# default-theme
+
+The aim of this rule is to prevent a wrong usage of the defaultTheme object from @kiwicom/orbit-components.
+
+## Rule details
+
+The following patterns are considered errors:
+
+```jsx
+import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";
+
+const failedStatus = {
+  isActive: true,
+  activeColor: defaultTheme.orbit.paletteOrangeNormal,
+  node: <CloseCircle color="warning" />,
+  content: getContentPerStatus(activeStatus, true),
+};
+```
+
+```jsx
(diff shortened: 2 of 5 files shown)


packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
```diff
@@ -0,0 +1,70 @@
+# default-theme
+
+The aim of this rule is to prevent the wrong usage or defaultTheme object from Orbit.
```
vepor, line 3:
```suggestion
The aim of this rule is to prevent a wrong usage of the defaultTheme object from @kiwicom/orbit-components.
```

packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
```diff
@@ -0,0 +1,70 @@
...
+## Rule details
+
+The following patterns are considered errors:
+
+```jsx
+import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";
+
+const failedStatus = {
+  isActive: true,
+  activeColor: defaultTheme.orbit.paletteOrangeNormal
```
vepor, line 14:
```suggestion
  activeColor: defaultTheme.orbit.paletteOrangeNormal,
```

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
```diff
@@ -0,0 +1,125 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+export const ERRORS = {
+  variableDeclaration: "Do not use defaultTheme as value, use useTheme() hook or ThemeConsumer",
```
vepor, line 5: I think in the most cases you can consume it with the `ThemeConsumer`, but also by passing down the `theme` object, so it would be nice to specify that.

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
```diff
@@ -0,0 +1,125 @@
...
+  styled: "Do not use defaultTheme directly inside styled components, create defautlProps",
```
vepor, line 6:
```suggestion
  styled: "Do not use defaultTheme directly inside styled components, either create defautlProps or consume the theme from with arrow function expression.",
```

vepor (changes requested):

packages/eslint-plugin-orbit-components/__tests__/defaultTheme.test.ts
```diff
@@ -0,0 +1,81 @@
+import ruleTester from "../ruleTester";
+import defaultTheme, { ERRORS } from "../src/rules/defaultTheme";
+
+describe("defaultTheme", () => {
+  // @ts-expect-error TODO
```
silvenon, line 5: You can avoid a lot of these error suppression comments here and in the rule file itself, along with having to specify argument types, by simply setting type of the rule export itself to `Rule.RuleModule` type, which is what `ruleTester.run` expects:

```ts
// defaultTheme.ts
import * as t from "@babel/types";
import type { Rule } from "eslint";

// ...

const defaultThemeRule: Rule.RuleModule = {
  // ...
};

export default defaultThemeRule;
```

TypeScript will helpfully point out all suppression comments that are no longer needed, so you don't have to worry about missing something.

Unfortunately, one thing I wasn't able to figure out is the `never` problem in the `VariableDeclaration` function, it doesn't make sense to me, but maybe you'll be able to figure it out! 😉 🤞

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
```diff
@@ -0,0 +1,126 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+export const ERRORS = {
+  variableDeclaration: "Do not use defaultTheme as value, use useTheme() hook or ThemeConsumer",
+  styled:
+    "Do not use defaultTheme directly inside styled components, either create defautlProps or consume the theme from with arrow function expression.",
+  import: "Do not name defaultTheme import as theme, please rename it to defaultTheme/themeDefault",
+  desctructured: "Do not destructure defaultTheme, use useTheme() hook or ThemeConsumer",
```
silvenon, line 9: Also update in the test file.
```suggestion
  destructured: "Do not destructure defaultTheme, use useTheme() hook or ThemeConsumer",
```

packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+
+"orbit-components/default-theme": <enabled>
+
+```
+
+```
+
+```
```
silvenon, line 70: There appears to be an empty block, and some additional whitespace. Is it on purpose?

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
```diff
@@ -0,0 +1,126 @@
...
+            if (t.isImportDefaultSpecifier(s)) {
+              if (s.local.name === "theme") {
+                context.report({
+                  // @ts-expect-error TODO
+                  node,
+                  message: ERRORS.import,
+                });
+              } else if (t.isImportDefaultSpecifier(s)) {
+                specifier = s.local.name;
+              }
```
silvenon, line 38: Would it be too much trouble to allow this as well? 🙏 

```js
import { defaultTheme } from "@kiwicom/orbit-components";
```

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
```diff
@@ -0,0 +1,126 @@
...
+      ImportDeclaration(node: t.ImportDeclaration) {
+        if (node.source.value === "@kiwicom/orbit-components/lib/defaultTheme") {
+          node.specifiers.forEach(s => {
+            if (t.isImportDefaultSpecifier(s)) {
+              if (s.local.name === "theme") {
+                context.report({
+                  // @ts-expect-error TODO
+                  node,
+                  message: ERRORS.import,
+                });
```
silvenon, line 35: I would personally avoid this part, or at least making it optional, considering that it's a matter of personal preference how you name it 🤔 I can imagine this not being perceived well and resulting in quite a few errors without it pointing to an actual problem, unlike the rest of the rules.

packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
```diff
@@ -0,0 +1,70 @@
...
+import defaultTheme from "@kiwicom/orbit-components/lib/defaultTheme";
+
+const {
+  borderWidthCard,
+  borderStyleCard,
+  borderColorCard,
+  borderRadiusNormal,
+  spaceMedium,
+} = defaultTheme.orbit;
+```
```
silvenon, line 42: Could you provide a short explanation and an example proper usage after each bad pattern you listed?

For example, I don't understand why destructuring is problematic, although I assume it's because `defaultTheme` should be used only as a whole, and destructuring makes it harder for us to track all these variables and print out warnings for them, so we just mark destructuring itself as a problem?

silvenon (changes requested):

packages/eslint-plugin-orbit-components/src/rules/defaultTheme.ts
mainframev: I would left it as it is, do not think it will cause any problem, the reason why I did it is that theme imported from `defaultTheme` should be used only inside SC defaultProps and Provider. `theme` is for consumer, for instance. I think we should have consistent names for that import.

packages/eslint-plugin-orbit-components/__tests__/defaultTheme.test.ts
mainframev: I was too lazy to solve it :D thanks for it

packages/eslint-plugin-orbit-components/docs/rules/default-theme.md
mainframev: Yes, I can, but in a few words: it's an anti-pattern, in our FE repo I saw that this example is copied from there, users tend to destructure it and pass it later to SC.

silvenon: Ok, I understand.

silvenon (approved):

PR #2738, opened 2021-02-13
feat(eslint-plugin-orbit): custom-styles

mainframev:
Added eslint rule to check custom written styles, which should use our tokens. Did not provide individual error message for each property, so feel free to advice if you think it's needed. Currently checks next properties: `box-shadow`, `color`, `background`, `color`, `font-size`, `line-height`, `border-radius`, `z-index`. I think for `left` and `right` we should have separate rule.

 Orbit.kiwi: https://orbit-docs-feat-custom-colors-rule.surge.sh
 Storybook: https://orbit-feat-custom-colors-rule.surge.sh

### DIFF
--- a/packages/eslint-plugin-orbit-components/docs/rules/custom-colors.md
+++ b/packages/eslint-plugin-orbit-components/docs/rules/custom-colors.md
@@ -1,0 +1,23 @@
+# custom-colors
+
+Prevents inconsistencies between Orbit and custom colors
+
+## Rule details
+
+The following patterns are considered warnings:
+
+```jsx
+const StyledWrapper = styled.div`
+  color: #252a31;
+  background: #fff;
+`;
+```
+
+The following patterns are **not** considered warnings:
+
+```jsx
+const StyledWrapper = styled.div`
+  color: ${({theme}) => theme.orbit.colorTextPrimary};
+  background: ${({theme}) => theme.orbit.paletteWhite};
+`,
+```
--- a/packages/eslint-plugin-orbit-components/__tests__/customColors.test.ts
+++ b/packages/eslint-plugin-orbit-components/__tests__/customColors.test.ts
@@ -1,0 +1,32 @@
+import ruleTester from "../ruleTester";
+import customColors from "../src/rules/customColors";
+
+describe("custom-colors", () => {
+  // @ts-expect-error TODO
+  ruleTester.run("custom-colors", customColors, {
+    valid: [
+      {
+        code: `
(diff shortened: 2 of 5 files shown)


vepor:
I generally agree that this kind of rule would be very handy to restrict the use of custom and hardcoded values. Sadly I don't think we should introduce it globally for such big set of css properties.

I think that maybe we could focus right now only on colors (`color`, `background-color`) and typography (`font-family`, `line-height`, `font-size`). Could you split this into subset of two rules? 🙏 

Also I can imagine that in some cases you would like to enable/disable one of them specifically for one file (or a codebase). We don't support every possible value, and won't be. Something you have the need to use e.g. `font-family: inherit` and this is something that doesn't have any reason to be stored as a token. Would consider it as a warning, not an error.

vepor (changes requested):
I generally agree that this kind of rule would be very handy to restrict the use of custom and hardcoded values. Sadly I don't think we should introduce it globally for such big set of css properties.

I think that maybe we could focus right now only on colors (`color`, `background-color`) and typography (`font-family`, `line-height`, `font-size`). Could you split this into subset of two rules? 🙏 

Also I can imagine that in some cases you would like to enable/disable one of them specifically for one file (or a codebase). We don't support every possible value, and won't be. Something you have the need to use e.g. `font-family: inherit` and this is something that doesn't have any reason to be stored as a token. Would consider it as a warning, not an error.

packages/eslint-plugin-orbit-components/docs/rules/custom-colors.md
```diff
@@ -0,0 +1,23 @@
...
+Prevents inconsistencies between Orbit and custom colors
+
+## Rule details
+
+The following patterns are considered warnings:
+
+```jsx
+const StyledWrapper = styled.div`
+  color: #808080;
+  background: #f1f1f1;
```
silvenon, line 12: With `#f1f1f1` it looks like it should be replaced with `theme.orbit.paletteWhite`, but that color is not white.
```suggestion
  background: #fff;
```

silvenon (changes requested):

packages/eslint-plugin-orbit-components/docs/rules/custom-colors.md
vepor: I would just the documentation example here, but definitely not bother with similar colors. We might implement color token suggestions later 🙂

silvenon (approved):

PR #2739, opened 2021-02-14
feat(eslint-plugin-orbit): prefer-single-destructure

mainframev:
Rule which counts usage of arrow functions inside SC, suggests to destructure it in one single function and pass it down via css helper function from SC. Throws warning from config, not error. 

 Orbit.kiwi: https://orbit-docs-feat-arrow-func-styled.surge.sh
 Storybook: https://orbit-feat-arrow-func-styled.surge.sh


packages/eslint-plugin-orbit-components/docs/rules/arrow-functions-count.md
```diff
@@ -0,0 +1,34 @@
+# arrow-functions-count
+
```
vepor, line 2: Using too many arrow functions in interpolations can have a negative impact on performance, because they have to be evaluated with execution context. This is done internally by wrapping all functions into `css` helper from `styled-components` . In most cases it's far more better to use one single arrow function and then use nested conditions that returns `css` if necessary.

packages/eslint-plugin-orbit-components/src/rules/arrowFunctionsCount.ts
```diff
@@ -0,0 +1,49 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+const LIMIT = 5;
+export const ERROR =
+  "You are using too many arrow function expressions. You can desctructure all props in single function and pass it down via css helper function";
```
vepor, line 6: You are using too many arrow function expressions in one single styled component. This can negatively impact performance. Consider using one single function to desctructure all props and return one single css helper function if possible.

packages/eslint-plugin-orbit-components/docs/rules/arrow-functions-count.md
vepor: Use ☝️  as the description of this rule.

vepor (changes requested):

packages/eslint-plugin-orbit-components/src/rules/arrowFunctionsCount.ts
```diff
@@ -0,0 +1,48 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+const LIMIT = 5;
+export const ERROR =
+  "You are using too many arrow function expressions in one single styled component. This can negatively impact performance. Consider using one single function to desctructure all props and return one single css helper function if possible.";
+export default {
```
silvenon, line 7: Try using `Rule.RuleModule` here 🙏

packages/eslint-plugin-orbit-components/src/configs/recommended.ts
```diff
@@ -3,6 +3,7 @@ const recommended = {
   rules: {
     "orbit-components/button-has-title": "error",
     "orbit-components/unnecessary-text": "error",
+    "orbit-components/arrow-functions-count": "warn",
```
silvenon, line 6: `warn` is a great choice 👍

packages/eslint-plugin-orbit-components/src/rules/arrowFunctionsCount.ts
```diff
@@ -0,0 +1,48 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+const LIMIT = 5;
+export const ERROR =
+  "You are using too many arrow function expressions in one single styled component. This can negatively impact performance. Consider using one single function to desctructure all props and return one single css helper function if possible.";
```
silvenon, line 6: Nice! I would phrase it a little differently to sound less accusatory: 
```suggestion
  "Using many arrow function expressions in a single styled component can negatively impact performance. Consider using one single function to destructure all props and return one single css helper function if possible.";
```

packages/eslint-plugin-orbit-components/src/rules/arrowFunctionsCount.ts
```diff
@@ -0,0 +1,48 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+const LIMIT = 5;
```
silvenon, line 4: I wonder if the limit of five is too arbitrary 🤔  Maybe we should be more opinionated and simply warn against more than one?

vepor: Having it set up for just one will be very limiting. The common use case is to have some styles together with media queries. If we would count also with more complex interpolations I think that limiting it to `~5` is fine. Alternatively we could have an option for the rule with defining the `schema` in meta.

.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
     "airbnb",
     "plugin:react-hooks/recommended",
     "plugin:prettier/recommended",
+    "plugin:@kiwicom/orbit-internal/recommended",
```
vepor, line 11: I would not add for the internal plugin now. We have been talking with @silvenon that we might merge these two plugins and we could use `eslint-plugin-orbit-components` internally. We'll discuss it more tomorrow.

silvenon (dismissed):

PR #2740, opened 2021-02-14
feat(eslint-plugin-orbit): rtl-utils rule

mainframev:
The rule which should help enforce usage of our RTL utility functions and avoid mistakes like `$({theme}) => theme.rtl ? "right: " : "left: "}) 8px`

 Orbit.kiwi: https://orbit-docs-feat-rtl-rule.surge.sh
 Storybook: https://orbit-feat-rtl-rule.surge.sh

### DIFF
--- a/packages/eslint-plugin-orbit-components/__tests__/rtlUtils.test.ts
+++ b/packages/eslint-plugin-orbit-components/__tests__/rtlUtils.test.ts
@@ -1,0 +1,113 @@
+import ruleTester from "../ruleTester";
+import rtlUtils, { RightOrLeftError, SpacingError } from "../src/rules/rtlUtils";
+
+describe("theme-rtl", () => {
+  ruleTester.run("rtl-utils", rtlUtils, {
+    valid: [
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
+              \${right}: 0;
+            \`
+          `,
+      },
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
+              \${({ someCustomProperty }) => someCustomProperty ? "right" : "left"};
+            \`
+          `,
+      },
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
+              margin-\${right}: \${({ theme }) => theme.orbit.spaceSmall};
+            \`
+          `,
+      },
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
+              margin-\${right}: \${({ theme }) => theme.orbit.spaceSmall};
+            \`
+          `,
+      },
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-orbit-components/src/rules/themeRTL.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+      },
+
+      TaggedTemplateExpression(node: t.TaggedTemplateExpression) {
+        if (t.isMemberExpression(node.tag)) {
+          if (t.isIdentifier(node.tag.object) && node.tag.object.name === specifier) {
+            if (t.isTemplateLiteral(node.quasi)) {
+              node.quasi.expressions.forEach(e => {
+                if (t.isArrowFunctionExpression(e)) {
+                  if (t.isConditionalExpression(e.body)) {
```
vepor, line 34: We should be more specific when comes to the conditional expression, otherwise we would prohibit the usage of arrow functions that returns conditional expression completely.

The rule will also report this case:

```jsx
const StyledDiv = styled.div`
  ${({ someCustomProperty }) => someCustomProperty ? "right" : "left"};
`
```

I would try to implement it only for conditional expression that uses `theme.rtl` or `rtl` (if it's destructurized) as the  `test` property of `ConditionalExpression`.

Also please add some generic use case that uses conditional expressions to the valid set of tests 🙏

packages/eslint-plugin-orbit-components/__tests__/themeRTL.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
+
+            const StyledWrapper = styled.div\`
+              margin-\${right}: \${({ theme }) => theme.orbit.spaceSmall};
+            \`
+          `,
+      },
+    ],
+    invalid: [
```
vepor, line 39: Would also test a case when the conditional expression returns different spacing value, e.g.:

```js
padding: ${({ theme }) => theme.rtl ? `0 ${theme.orbit.spaceSmall} 0 ${theme.orbit.spaceXSmall}` : `0 ${theme.orbit.spaceXSmall} 0 ${theme.orbit.spaceSmall}`};
```

This is a case that should be covered by use of `rtlSpacing` not `right/left` directly.

packages/eslint-plugin-orbit-components/src/rules/themeRTL.ts
```diff
@@ -0,0 +1,63 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+
+export const ERROR =
+  "Do not use theme.rtl conditions. Import RTL utility functions from Orbit and use them instead";
```
vepor, line 5: Would have the the error more specific.

```
Do not use theme.rtl as the test value of conditional expression. Consider importing either left or right function from @kiwicom/orbit-components. See more on https://orbit.kiwi/utilities/right-to-left-languages/.
```

vepor (changes requested):

silvenon (commented):
I'm also learning about Orbit anti-patterns from this! 😃

packages/eslint-plugin-orbit-components/__tests__/themeRTL.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+            const StyledWrapper = styled.div\`
+              \${right}: 0;
+            \`
+          `,
+      },
+      {
+        code: `
+          import styled from 'styled-components';
+          import { right } from "@kiwicom/orbit-components/lib/utils/rtl";
```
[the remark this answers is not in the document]
mainframev: cmd+c -> cmd+v approach :D

packages/eslint-plugin-orbit-components/__tests__/themeRTL.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+            const StyledWrapper = styled.div\`
+              margin-\${right}: \${({ theme }) => theme.orbit.spaceSmall};
+            \`
+          `,
+      },
```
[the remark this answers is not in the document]
mainframev: good, did not notice

PR #2747, opened 2021-02-16
feat(docs): add IconsList

CollierCZ:
And use in foundation/icons/ for now (will add to component page later). Not the prettiest list ever, but seems functional.

 Orbit.kiwi: https://orbit-docs-docs-iconslist.surge.sh
 Storybook: https://orbit-docs-iconslist.surge.sh

### DIFF
--- a/docs/src/components/IconsList.tsx
+++ b/docs/src/components/IconsList.tsx
@@ -1,0 +1,144 @@
+import React, { useEffect, useState } from "react";
+import styled from "styled-components";
+import {
+  Button,
+  ButtonLink,
+  Card,
+  CardSection,
+  Coupon,
+  Heading,
+  InputField,
+  Stack,
+  Text,
+  Tile,
+} from "@kiwicom/orbit-components";
+import { Link as LinkIcon, Search } from "@kiwicom/orbit-components/icons";
+import iconsList from "@kiwicom/orbit-components/lib/data/icons.json";
+import { CopyToClipboard } from "react-copy-to-clipboard";
+
+import { CodeBlock } from "./Code";
+import { copyTimeout, pascalize } from "../utils/common";
+
+const Grid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(280px, 1fr));
+  grid-gap: ${({ theme }) => theme.orbit.spaceXSmall};
+`;
+
+const StyledAnchorWrapper = styled.div`
+  svg {
+    visibility: hidden;
+  }
+
+  &:hover svg {
+    visibility: visible;
+  }
+`;
+
+interface IconObjectShape {
+  character: string;
+  description: string;
+  svg: string;
+  url: string;
+}
+
+interface IconProps {
+  iconName: string;
+  opened: boolean;
+  setOpenedIcon: React.Dispatch<React.SetStateAction<string>>;
+}
+
+const Icon = ({ iconName, opened, setOpenedIcon }: IconProps) => {
+  const [copied, setCopied] = useState(false);
+  useEffect(() => {
+    copyTimeout(copied, setCopied);
+  }, [copied, setCopied]);
+
+  const iconObject: IconObjectShape = iconsList[iconName];
+  const pascalName = pascalize(iconName);
+
+  return (
+    <div id={iconName}>
+      <Tile onClick={() => setOpenedIcon(opened ? "" : iconName)}>
+        <Stack justify="between">
+          <img src={`data:image/svg+xml;utf8,${iconObject.svg}`} alt="" />
+          <Coupon>{iconObject.character}</Coupon>
+        </Stack>
+        <Text type="secondary">{iconName}</Text>
+      </Tile>
+      {opened && (
+        <Card>
+          <CardSection>
+            <Stack>
+              <StyledAnchorWrapper>
+                <Stack flex spacing="XXXSmall" align="center">
+                  <Heading as="h3" type="title2" spaceAfter="medium">
+                    {iconName}
+                  </Heading>
+                  <ButtonLink
+                    iconLeft={<LinkIcon />}
+                    href={`#${iconName}`}
+                    type="secondary"
+                    compact
+                  />
+                </Stack>
+              </StyledAnchorWrapper>
+              {iconObject.description && <Text>{iconObject.description}</Text>}
+              <CodeBlock className="language-jsx">{`import { ${pascalName} } from "@kiwicom/orbit-components/icons"\n\n<${pascalName} />`}</CodeBlock>
+              <Stack direction="row" justify="between">
+                <Button size="small" type="secondary" href={iconObject.url} external>
+                  Download SVG
+                </Button>
+                <CopyToClipboard text={iconObject.svg}>
+                  <Button
+                    size="small"
+                    type="secondary"
+                    disabled={copied}
+                    onClick={() => setCopied(true)}
+                  >
+                    {copied ? "Copied source!" : "Copy SVG source"}
+                  </Button>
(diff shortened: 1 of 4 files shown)


docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { CodeBlock } from "./Code";
+import { copyTimeout, pascalize } from "../utils/common";
+
+const Grid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(280px, 1fr));
+  grid-gap: ${({ theme }) => theme.orbit.spaceXSmall};
+`;
+
+const StyledAnchorWrapper = styled.a`
```
mainframev, line 27: In the console you have `index.js:2177 Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.`  Consider to have it like `div` or layout it differently.

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+  a {
+    float: left;
+    padding-top: 4px;
+    padding-right: 2px;
+    margin-left: -22px;
```
mainframev, line 32: i tried it without that negative margin and it looks better. Why you need that? 🤔

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+              <StyledAnchorWrapper>
+                <Heading as="h3" type="title2" spaceAfter="medium">
+                  <a href={`#${iconName}`}>
+                    <LinkIcon />
+                  </a>
+                  {iconName}
+                </Heading>
+              </StyledAnchorWrapper>
+              {iconObject.description && <Text>{iconObject.description}</Text>}
+              <CodeBlock className="language-jsx">{`import { ${pascalName} } from "@kiwicom/orbit-components/icons"\n\n<${pascalName} />`}</CodeBlock>
```
mainframev, line 90: when i opened it locally, code aligned on the right side, should'n it be aligned on the left? 🤔

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const [copied, setCopied] = useState(false);
+  useEffect(() => {
+    copyTimeout(copied, setCopied);
+  }, [copied, setCopied]);
+
+  const iconObject: IIconObject = iconsList[iconName];
+  const pascalName = pascalize(iconName);
+
+  return (
+    <div id={iconName}>
```
mainframev, line 69: do we need `id` here? we are trying to avoid usage of IDs, only when there is specific reason like svg masks.

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const filteredIconNames = allIconNames.filter(name => name.includes(filter.toLowerCase()));
+
+  return (
+    <Stack spacing="large">
+      <InputField
+        prefix={<Search />}
+        placeholder="Find an icon"
+        value={filter}
+        onChange={event => {
+          if (event.target instanceof HTMLInputElement) {
```
mainframev, line 132: you can remove that condition and rewrite it like `onChange={ev => setFilter(ev.currentTarget.value)}`

docs/src/utils/common.ts
```diff
@@ -5,3 +5,25 @@ export const omitNumbers = (str: string) =>
...
 export const capitalize = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
+
+export const pascalize = (str: string) =>
+  str
+    .replace(/-(.)/g, (_match, chartacter: string) => {
+      return chartacter.toUpperCase();
+    })
+    .replace(/^(.)/, ($1: string) => {
+      return $1.toUpperCase();
+    });
```
mainframev, line 16: here you can omit return and the curly braces, so it will look more shorter :)

mainframev (changes requested):

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+  useEffect(() => {
+    copyTimeout(copied, setCopied);
+  }, [copied, setCopied]);
+
+  const iconObject: IIconObject = iconsList[iconName];
+  const pascalName = pascalize(iconName);
+
+  return (
+    <div id={iconName}>
+      <Tile onClick={() => setOpenedIcon(opened ? "" : iconName)}>
```
mainframev, line 70: You can also consider to use `Slide` component, so it will appear more smoothly, but it's up to you :)

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+    width: 20px;
+    visibility: hidden;
+  }
+
+  &:hover svg {
+    visibility: visible;
+  }
+`;
+
+interface IIconObject {
```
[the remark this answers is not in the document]
CollierCZ: I just copied that from https://github.com/kiwicom/orbit/blob/master/docs/src/components/Github/index.tsx#L14, so I thought that was the standard here. But I can certainly change it.

docs/src/components/IconsList.tsx
CollierCZ: Without the negative margin, the header is pushed to the right and doesn't align with the elements below it. Because the anchor link will mostly be hidden, I thought that was more important.

docs/src/components/IconsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const Grid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(280px, 1fr));
+  grid-gap: ${({ theme }) => theme.orbit.spaceXSmall};
+`;
+
+const StyledAnchorWrapper = styled.a`
+  a {
+    float: left;
```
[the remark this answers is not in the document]
CollierCZ: I can use `position: absolute`.

docs/src/components/IconsList.tsx
CollierCZ: I see it left aligned, both locally and at https://orbit-docs-docs-iconslist.surge.sh/foundation/icons/#clock. What browser are you using?

docs/src/components/IconsList.tsx
CollierCZ: Is there a better way to generate anchors on the page?

docs/src/components/IconsList.tsx
CollierCZ: It is slightly nicer that way, but not so much more, though I might just have been doing it wrong. I like keeping the data off the page until needed.

docs/src/components/IconsList.tsx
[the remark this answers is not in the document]
mainframev: Ofc Chrome.  I think it's because i used it on the wrong place on the main page to look on it, will check it again

docs/src/components/IconsList.tsx
mainframev: Ofc Chrome. I think it's because i used it on the wrong place on the main page to look on it (it could be affected by Stack or other wrapper), will check it again

docs/src/components/IconsList.tsx
CollierCZ: Moved to a ButtonLink on the right of the head, so this is all unnecessary.

docs/src/components/IconsList.tsx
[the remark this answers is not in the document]
CollierCZ: Moved to a ButtonLink on the right of the head, so this is all unnecessary.

mainframev (approved):

PR #2750, opened 2021-02-18
fix(docs): navbar

mainframev:
I had to turn off check on CI for dead links, because in navbar we are still have dead links, we can turn on it later, when all links will be ready :) Also it had missing token, so added it on CI from env for that task.
 
 Orbit.kiwi: https://orbit-docs-fix-replace-hook-ssr.surge.sh
 Storybook: https://orbit-fix-replace-hook-ssr.surge.sh

### DIFF
--- a/docs/utils/check-links.js
+++ b/docs/utils/check-links.js
@@ -87,7 +87,7 @@ const checkForDeadUrls = async () => {
         .map(url => `${url} status: ${results[url].statusCode}`)
         .join("\n")}`,
     );
-    // TODO: uncomment
+    // TODO: uncomment when the dead links in Navbar have been resolved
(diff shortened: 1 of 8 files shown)


docs/utils/check-links.js
```diff
@@ -86,7 +87,8 @@ const checkForDeadUrls = async () => {
         .map(url => `${url} status: ${results[url].statusCode}`)
         .join("\n")}`,
     );
-    process.exit(1);
+    // TODO: uncomment
```
silvenon, line 90:
```suggestion
    // TODO: uncomment when the dead links in Navbar have been resolved
```

silvenon (approved):

PR #2752, opened 2021-02-19
improve: cleanup Babel setup

silvenon:
The root Babel configuration was mostly applying to `@kiwicom/orbit-components`, so now it has been moved there, so that it's responsible for a single package. Also, considering that Storybook stories are importing the source code it made a lot of sense to reuse the same Babel configuration in Storybook's webpack configuration, instead of maintaining a separate Babel configuration there.

`@adeira`'s Babel preset uses Flow, and considering that we plan to gradually switch to TypeScript, I broke it down into plugins that we actually (might) need, and it turned out that the list is pretty simple.  Now we know exactly what our Babel configuration does and it's much more flexible.

 Orbit.kiwi: https://orbit-docs-improve-babel-cleanup.surge.sh
 Storybook: https://orbit-improve-babel-cleanup.surge.sh


.github/contribution/testing-conventions.md
```diff
@@ -130,23 +130,21 @@ This part of testing is often abandoned by people, but we are aware of how impor
...
+- iOS Safari 14.0-14.3
+- iOS Safari 13.4-13.7
 - iOS Safari 12.2-12.4
-- Opera 68
+- Opera 72
+- Safari 14
 - Safari 13.1
+- Samsung Internet 13.0
 - Samsung Internet 12.0
-- Samsung Internet 11.1-11.2
```
silvenon, line 150: I didn't realize this part changed, I'll have to figure out why, but ignore it for now.

docs/babel.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  presets: ["gatsby"],
+  presets: ["gatsby", "@babel/preset-typescript"],
```
vepor, line 2: We don't have `@babel/preset-typescript` anywhere as dependency.

vepor (changes requested):
As we discussed already, so just you won't forget.

.github/contribution/testing-conventions.md
silvenon: For posterity: this turned out to be a result of `browserslist` update in db33af97ed4b890fbdf22c378e3224e0f15d901d, but our setup for automatically updating the this Markdown file isn't sophisticated enough to catch that.

vepor (approved):

PR #2778, opened 2021-02-23
feat(docs): add illustration list

CollierCZ:
and refactor iconlist to abstract shared code

 Orbit.kiwi: https://orbit-docs-docs-illustrationlist.surge.sh
 Storybook: https://orbit-docs-illustrationlist.surge.sh


docs/src/components/VisualList.tsx
```diff
@@ -0,0 +1,148 @@
...
+    copyTimeout(copied, setCopied);
+  }, [copied, setCopied]);
+
+  const visualObject: VisualObjectShape = list[visualName];
+
+  return (
+    <div id={visualName}>
+      <Tile onClick={() => setOpenedVisual(opened ? "" : visualName)}>
+        <Stack justify="between">
+          <img src={getImgSource(visualObject)} alt="" />
```
vepor, line 75: @CollierCZ I would create small styled components and attach some styles, so we are able to resolve the problematic case with TIcketFlexi illustration. It will at least work for now, until we will have different solution with this exceptional illustration.

```jsx
const StyledImage = styled.img`
  max-height: 180px;
  max-width: 100%;
`
```

Add the same time I would apply `justify="center"` on the wrapping Stack.

mainframev (approved):

docs/src/components/VisualList.tsx
silvenon: My suggestion to use [aspect ratio box](https://css-tricks.com/aspect-ratio-boxes/):

```jsx
const StyledRatioContainer = styled.div`
  position: relative;
  padding-bottom: calc(400 / 677 * 100%); /* height / width = ratio */
`
const StyledImage = styled.img`
  position: absolute;
  width: 100%;
  height: 100%;
  object-fit: contain;
`

// ...

<StyledRatioContainer>
  <StyledImage src={getImgSource(visualObject)} alt="" />
</StyledRatioContainer>
```

mainframev (approved):

PR #2783, opened 2021-02-26
feat(docs): add guidelines components

CollierCZ:
I went with parsing children rather than passing images as props for a couple of reasons:

* It seems in keeping with syntax elsewhere.
* It seems easier to then include FigmaImage components as the images. When that component is implented, the checks here for images would probably have to be changed.

But I'm not sure of a stable type checker for the images at the moment.
 Orbit.kiwi: https://orbit-docs-docs-guidelines.surge.sh
 Storybook: https://orbit-docs-guidelines.surge.sh


docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }`};
+    padding: ${theme.orbit.spaceXSmall} ${theme.orbit.spaceLarge} ${theme.orbit.spaceXSmall}
+      ${theme.orbit.spaceMedium};
+    display: flex;
+  `}
+`;
+
+const ContentContainer = styled.div`
+  ${({ theme }) => css`
+    padding-left: ${theme.orbit.spaceXSmall};
```
mainframev, line 46: `padding-${left}`, for RTL :)

docs/src/components/GuidelinesSideBySide.tsx
```diff
@@ -0,0 +1,72 @@
...
+const Wrapper = styled.div<GuidelineType>`
+  ${({ theme, type }) => css`
+    border-top: 4px solid
+      ${type === "do" ? theme.orbit.paletteGreenNormal : theme.orbit.paletteRedNormal};
+    padding: ${theme.orbit.spaceLarge} ${theme.orbit.spaceMedium};
+  `}
+`;
+
+interface StyledContainerProps {
+  sizeCheck: boolean | null | undefined;
```
mainframev, line 23:
```suggestion
  sizeCheck?: boolean | null;
```

docs/src/components/HeaderWithLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+    visibility: hidden;
+  }
+
+  &:hover svg {
+    visibility: visible;
+  }
+`;
+
+interface Props {
+  children: string;
```
mainframev, line 21: do not call this props `children` if it's supposed to be string only, call it `href` for instance, because `children` usually means `React.ReactNode` that can be confusing :)

docs/src/components/HeaderWithLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const StyledAnchorWrapper = styled.div`
+  margin-top: -10px;
+
+  svg {
+    visibility: hidden;
+  }
+
+  &:hover svg {
+    visibility: visible;
```
mainframev, line 16: i would probably solve it via `opacity` and also add some transition, so the link will appear more smoothly :)

docs/src/components/GuidelinesSideBySide.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <CloseCircle color="critical" ariaLabel="Don't" />
+    )}
+    <div>{children}</div>
+  </Stack>
+);
+
+const GuidelineContainer = ({ children, type }) => {
+  const { isTablet } = useMediaQuery();
+  return (
+    <StyledContainer sizeCheck={isTablet}>
```
mainframev, line 49: probably will be better to rename `sizeCheck` to `isTablet`, then it's easier to understand when looking inside SC.

docs/src/components/GuidelineImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+import styled, { css } from "styled-components";
+import { Stack } from "@kiwicom/orbit-components";
+import useMediaQuery from "@kiwicom/orbit-components/lib/hooks/useMediaQuery";
+
+import { DoDontHeader, GuidelineType } from "./Guideline";
+
+const GuidelineContainer = styled.div`
+  ${({ theme }) => css`
+    padding: ${theme.orbit.spaceXLarge};
+    padding-top: 0;
```
mainframev, line 11: also, shorthand here is better :)

docs/src/components/GuidelineImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+    background-color: ${theme.orbit.paletteCloudLight};
+    border-radius: ${theme.orbit.spaceMedium};
+  `}
+`;
+
+interface SizeCheckProps {
+  sizeCheck: boolean | null | undefined;
```
mainframev, line 18:
```suggestion
  sizeCheck?: boolean | null;
```

docs/src/components/GuidelineImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+const GuidelineWithImage = ({ children, type }: GuidelineWithImageProps) => {
+  const { isMediumMobile } = useMediaQuery();
+  const image = children[0];
+  const content = children.slice();
+  content.shift();
+  return (
+    <GuidelineContainer>
+      <Border type={type}>
+        <Stack>
+          <ImageContainer sizeCheck={isMediumMobile}>{image}</ImageContainer>
```
mainframev, line 54:
```suggestion
          <ImageContainer isMediumMobile={isMediumMobile}>{image}</ImageContainer>
```

docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+interface ImageContainerProps {
+  noLeftPadding?: boolean;
+  middleAlign?: boolean;
+  sizeCheck?: boolean | null | undefined;
+}
+
+const ImageContainer = styled.div<ImageContainerProps>`
+  ${({ noLeftPadding, sizeCheck, theme }) => css`
+    padding: ${noLeftPadding ? theme.orbit.spaceMedium : theme.orbit.spaceXLarge};
+    ${noLeftPadding && "padding-left: 0;"}
```
mainframev, line 59: shorthand for padding here :)

docs/src/components/GuidelinesSideBySide.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Wrapper type={type}>
+        {(Array.isArray(children.props.children) &&
+          children.props.children.map(child => (
```
mainframev, line 52: use `React.Children.Map` here

mainframev (changes requested):
I think it will be better to create folder `Guideline` inside `components` and move there all three: 
- `Guideline` -> `index.tsx`
- `GuidelineImages`
- `GuidelineSidebySide`

docs/src/components/GuidelineImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+    height: 100%;
+  `}
+`;
+
+interface GuidelineWithImageProps extends GuidelineType, ImageProps {}
+
+const GuidelineWithImage = ({ children, type }: GuidelineWithImageProps) => {
+  const { isMediumMobile } = useMediaQuery();
+  const image = children[0];
+  const content = children.slice();
```
mainframev, line 48:
```suggestion
  const content = children.slice(1, children.length);
```

docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+    border-radius: ${theme.orbit.spaceMedium};
+  `}
+`;
+
+const ImageBorder = styled.div<GuidelineType>`
+  ${({ theme, type }) => css`
+    border-left: 4px solid
+      ${type === "do" ? theme.orbit.paletteGreenNormal : theme.orbit.paletteRedNormal};
+    padding: ${theme.orbit.spaceMedium};
+    padding-left: ${theme.orbit.spaceXLarge};
```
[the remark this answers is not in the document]
CollierCZ: Why do you have this preference? I like seeing explicitly which one is being set. Plus I don't like repeating the same value three times. I'm fine to change, just wanted to understand.

docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+      return child;
+    });
+  } else if (checkIfImageAndAddToArray(children)) {
+    content = ""; // only should apply for guidelines with only an image
+  }
+
+  const typeOpposite = type === "do" ? "dont" : "do";
+
+  return (
+    <StyledComponent id={sluggify(title) || undefined}>
```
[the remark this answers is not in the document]
CollierCZ: I have no memory of why I did that. 🙈

docs/src/components/GuidelinesSideBySide.tsx
CollierCZ: Good point. I renamed them in all the files. It came from early testing when I was using different checks for different parts, but in the end, each file has just one check.

docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+      ${type === "do" ? theme.orbit.paletteGreenNormal : theme.orbit.paletteRedNormal};
+    padding: ${theme.orbit.spaceMedium};
+    padding-left: ${theme.orbit.spaceXLarge};
+    height: 100%;
+  `}
+`;
+
+const DoDontHeaderWrapper = styled.div`
+  ${({ theme }) => css`
+    padding-left: ${theme.orbit.spaceLarge};
```
mainframev, line 78:
```suggestion
    padding-left: ${({theme}) => theme.orbit.spaceLarge};
```

docs/src/components/Guideline.tsx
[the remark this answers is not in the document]
CollierCZ: It was to handle cases without a title, but I can put a check for it directly in the function instead.

docs/src/components/Guideline.tsx
```diff
@@ -0,0 +1,163 @@
...
+  if (Array.isArray(children)) {
+    content = children.map(child => {
+      if (checkIfImageAndAddToArray(child)) {
+        return "";
+      }
+      return child;
+    });
+  } else if (checkIfImageAndAddToArray(children)) {
+    content = ""; // only should apply for guidelines with only an image
+  }
```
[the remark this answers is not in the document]
CollierCZ: I made at least a slight improvement.

docs/src/components/Guidelines/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+          return "";
+        }
+        return child;
+      });
+    }
+    if (checkIfImageAndAddToArray(object)) {
+      return ""; // only should apply for guidelines with only an image
+    }
+    return object;
+  };
```
mainframev, line 114: I am not happy with this check and functions, looks like it can be done more simple, I'll check it more right now and will put a suggestion here :) 

something like that can be done: 
```
  const isImage = object => {
    if (object.props?.children?.props?.className === imageWrapperClass) return true;
    return false;
  };

  interface Content {
    images: React.ReactNode[];
    content: React.ReactNode;
  }

  const getContent = data => {
    return React.Children.toArray(data).reduce(
      (acc: Content, cur) => {
        if (isImage(cur)) acc.images.push(cur);
        else acc.content = cur;

        return acc;
      },
      { images: [], content: null },
    );
  };

  const { images, content } = getContent(children);
```

mainframev (changes requested):

mainframev (approved):

PR #2784, opened 2021-02-28
fix(docs): add ComponentStatus to shortcodes

CollierCZ:
This way the build doesn't fail if you put in the wrong component name. Not sure another elegant way to handle it.

Orbit.kiwi: https://orbit-docs-docs-components.surge.sh
Storybook: https://orbit-docs-components.surge.sh


silvenon (approved):
Nice!

silvenon (approved):

PR #2791, opened 2021-03-02
fix(CallOutBanner): illustration type

mainframev:
There is an error inside tsx file when `CallOutBanner` has `illustration` prop. Noticed during the creation of examples. 

 Orbit.kiwi: https://orbit-docs-fix-calloutbanner-ts-type.surge.sh
 Storybook: https://orbit-fix-calloutbanner-ts-type.surge.sh


mainframev:
Screenshot 2021-03-02 at 10 52 07

silvenon (approved):
Nice!

PR #2795, opened 2021-03-02
fix(Slider): add TypeScript declaration for calculateOf util

mainframev:
added declaration export, this util is used inside of one of the examples
 Orbit.kiwi: https://orbit-docs-fix-add-export-for-slider-util.surge.sh
 Storybook: https://orbit-fix-add-export-for-slider-util.surge.sh

### DIFF
--- a/packages/orbit-components/src/Slider/utils/calculateCountOf.d.ts
+++ b/packages/orbit-components/src/Slider/utils/calculateCountOf.d.ts
@@ -1,0 +1,6 @@
+import { Value, Data } from "..";
+
+// eslint-disable-next-line @typescript-eslint/prefer-readonly-parameter-types
+declare const CalculateCountOf: (data: Data, value: Value, min: number) => [number, number];
+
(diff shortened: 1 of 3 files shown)


packages/orbit-components/src/Slider/utils/calculateCountOf.d.ts
```diff
@@ -0,0 +1,6 @@
+import { Value, Data } from "..";
+
+// eslint-disable-next-line @typescript-eslint/prefer-readonly-parameter-types
+declare const CalculateCountOf: (data: Data, value: Value, min: number) => [number, number];
+
+export default CalculateCountOf;
```
silvenon, line 6: FYI: the Flow declaration also exports the `CalculateCountOf` as a named export, but that's pointless because you can always get it with `typeof`, so I don't think it's worth mirroring that here.

silvenon (approved):

PR #2808, opened 2021-03-03
feat(playground): examples

mainframev:
The scope for `react-live` is grabbed via `ast` and passed as node  scope with name, path, and if it’s default or named export. 

 Orbit.kiwi: https://orbit-docs-docs-playground-examples.surge.sh
 Storybook: https://orbit-docs-playground-examples.surge.sh


docs/src/components/Example/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+      ...acc,
+      [name]: Components[name],
+    };
+  }, {});
+
+  return (
+    <LiveProvider
+      code={fields.example}
+      scope={{ ...modules, defaultTheme, styled, css }}
+      theme={dracula}
```
silvenon, line 74: I love this theme! 😍

silvenon (approved):

docs/src/mdx-components.tsx
```diff
@@ -162,6 +183,9 @@ export const a = function Anchor({
     </span>
   );
 };
+
 export const Callout = ({ icon = true, ...props }) => (
   <Alert icon={icon} spaceAfter="large" {...props} />
 );
+
+export const Examples = ({ exampleId }) => <Example exampleId={exampleId} />;
```
CollierCZ, line 191: Given that the prop is exactly the same, I don't see a benefit of including it in this file. It can be included like the other components that are just copies in DocLayout. Then you don't have to get into the confusion of having different naming (`Examples`).

docs/src/__examples__/Accordion/ACTIONS.tsx
```diff
@@ -0,0 +1,126 @@
...
+        </AccordionSection>
+      </Accordion>
+    );
+  },
+  info: {
+    title: "Section actions",
+    description:
+      "To override the open button and add your own actions in the section heading, use the <code>actions</code> property of the section.",
+  },
+};
```
CollierCZ, line 126: Is there a reason to be including these in the file but not the example?

docs/plugins/sandbox/gatsby-node.js
```diff
@@ -0,0 +1,35 @@
...
+
+    createNodeField({
+      node,
+      name: "example",
+      value: example ? example[0] : "",
+    });
+
+    createNodeField({
+      node,
+      name: "example-id",
```
CollierCZ, line 31: This gets changed into `example_id`, so I'd change it here.

docs/src/components/Example/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+              onClick={() => setOpenEditor(prev => !prev)}
+              type="secondary"
+              ariaExpanded={isOpened}
+              iconRight={isOpened ? <ChevronUp /> : <ChevronDown />}
+            >
+              Code
+            </ButtonLink>
+          </Stack>
+          <Stack inline justify="end" align="center">
+            <CopyToClipboard text={fields.example}>
```
CollierCZ, line 92: This leaves me with copying the id, not the code.

docs/src/components/Example/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+                default
+              }
+            }
+          }
+        }
+      }
+    `,
+  );
+
+  const { fields } = allFile.nodes.find(n => n.fields.example_id === exampleId);
```
CollierCZ, line 54: If an incorrect ID is used, it will break the whole build. Could you please add a check for that? Return a note about not having that example yet or something.

CollierCZ (changes requested):
In the placeholders, this component was called ReactExample, and I think that'd be better naming. There's a chance we'll have mobile components or whatever so making this more specific might help. If not, at least adjust the globals in `.eslintrc.js` and all of the examples that use it in content.

Also, why is it in a folder?

I'm not a fan of copying over the all-caps naming convention. I didn't like it so much before and I'm not sure I see the value of it now. Not sure even the underscores are really necessary, but I don't dislike them as much. Given that the ID comes from the name, I would really like more clear naming. It'd be nice to have it clear what the ID is from a quick glance. Even better would be being able to just copy the file name.

Please add usage examples to `src/pages/mdx-examples.mdx`.

In the redesign, it's part of a card with raised elevation. Could you add that? Otherwise, it isn't clear. In some versions, the actions bar at the bottom in the redesign had a background. Right now the actions and everything blend into the page background.

The displayed examples don't include the import statements. We wanted the copy and paste to be as effortless as possible. The data is there

mainframev:
> The displayed examples don't include the import statements. We wanted the copy and paste to be as effortless as possible. The data is there

It should not include imports, because it will not have any affect on source code inside react-live. It's in the design, but from usage perspective it's useless :)

CollierCZ:
> It's in the design, but from usage perspective it's useless :)

What perspective of usage? I imagine copying and pasting it into an app. Then having the imports as part of what's copied is useful. Like what we have now at Orbit.kiwi. Is that not useful? 🤔

mainframev:
> > It's in the design, but from usage perspective it's useless :)
> 
> What perspective of usage? I imagine copying and pasting it into an app. Then having the imports as part of what's copied is useful. Like what we have now at Orbit.kiwi. Is that not useful? 🤔

Ah, misunderstood it. Yep, makes sense

docs/src/__examples__/Accordion/ACTIONS.tsx
mainframev: it was moved from current examples, did not include it by myself.

docs/src/components/Example/index.tsx
mainframev: nope, it copies the code :)

PR #2809, opened 2021-03-04
feat(tokens): deprecated tokens eslint rule

vepor:


packages/eslint-plugin-orbit-components/src/rules/noDeprecatedToken.ts
```diff
@@ -0,0 +1,96 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+/*
```
vepor, line 3: cc @silvenon

packages/eslint-plugin-orbit-internal/package.json
```diff
@@ -8,7 +8,8 @@
     "postinstall": "yarn build"
   },
   "dependencies": {
-    "@babel/types": "=7.12.10"
+    "@babel/types": "=7.12.10",
+    "eslint-plugin-orbit-components": "^0.3.0"
```
vepor, line 12: @silvenon I would like to have it also used internally, but currently we don't build the `eslint-plugin-orbit-components` locally for development purposes, but it would be required. Other option would be to have duplicate implementation, or to use publish `@kiwicom/eslint-plugin-orbit-internal` and have it use it from there.

silvenon: I vote for building `eslint-plugin-orbit-components` on `postinstall`; we'll have more and more rules that we'll want to use in our own codebase, and compiling the plugin is fast. No need to re-export it from `eslint-plugin-orbit-internal`.

Assuming you agree, when implementing building on `postinstall` don't forget to run `pinst` on `prepublishOnly` and `postpublish` just like in `@kiwicom/orbit-components`. Also, in root `.eslintrc.js` you can add `plugin:orbit-components/recommended` to `extends`. And while you're at it, you could also add `plugin:@kiwicom/orbit-internal/recommended`, too, I guess that works now (it didn't work for me before), and remove its rules and remove it from `plugins` because that's all included in the recommended configuration.

packages/eslint-plugin-orbit-components/src/rules/noDeprecatedToken.ts
silvenon: For me it totally makes sense that `@kiwicom/orbit-design-tokens` is a dependency of `eslint-plugin-orbit-components` because of this rule, at least for those versions of the ESLint plugin which will have this rule. What would you gain by `@kiwicom/orbit-components` re-exporting it…? I'm wondering what is the dilemma.

packages/eslint-plugin-orbit-internal/package.json
silvenon: I vote for building `eslint-plugin-orbit-components` on `postinstall`; we'll have more and more rules that we'll want to use in our own codebase, and compiling the plugin is fast. No need to re-export it from `eslint-plugin-orbit-internal`.

Assuming you agree, when implementing building on `postinstall` don't forget to run `pinst` on `prepublishOnly` and `postpublish` just like in `@kiwicom/orbit-components`. Also, in root `.eslintrc.js` you can add `plugin:orbit-components/recommended` to `extends`. And while you're at it, you could also add `plugin:@kiwicom/orbit-internal/recommended`, too, I guess that works now (it didn't work for me before), and remove its rules and remove it from `plugins` because that's all included in the recommended configuration.

packages/eslint-plugin-orbit-components/src/rules/noDeprecatedToken.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const reportIfDesignToken = (node: t.MemberExpression) => {
+      if (isOrbitDesignToken(node) && t.isIdentifier(node.property)) {
+        reportError(node.property.name, node.property);
+      }
+    };
+    return {
+      MemberExpression(node: t.MemberExpression) {
+        reportIfDesignToken(node);
+      },
```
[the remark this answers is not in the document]
vepor: I was thinking about that. The implementation would be definitely more simple, but on the other hand I'm not sure if it would be more performant, because generally there's a ton of `Identifier(s)` in every file 🤔 

For instance I took random file from `orbit-components` and it contains 159 identifiers (types) compared 29 of member expressions.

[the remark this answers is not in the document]
vepor: And my assumption is that if we will access 5-6x times more nodes it will be less performant even with the fact we check for e.g. `isThemeOrbit` on every `MemberExpression` node.

packages/eslint-plugin-orbit-components/src/rules/noDeprecatedToken.ts
vepor: I was just not sure if we want to add another dependency to the whole cycle. `@kiwicom/orbit-design-tokens` are direct dependency of `@kiwicom/orbit-components`, but since it won't be peer dependency (for components and the eslint plugin) it makes sense to simply add it 🙂 Thanks for your input @silvenon

packages/eslint-plugin-orbit-components/src/rules/noDeprecatedToken.ts
```diff
@@ -0,0 +1,96 @@
+import * as t from "@babel/types";
+import type { Rule } from "eslint";
+/*
+  TODO: Either we should add dependency @kiwicom/orbit-design-tokens for the eslint-plugin,
+  or we can re-export it inside @kiwicom/orbit-components.
+ */
+import deprecatedTokens from "@kiwicom/orbit-design-tokens/lib/deprecated-tokens.json";
+
+export const mockTokens = {
```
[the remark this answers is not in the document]
vepor: Oh yeah, didn't realize 🙂

packages/eslint-plugin-orbit-internal/package.json
vepor: I will create separate PR for this, if you agree. This PR will be merged into `dev/tokens`, but we might benefit from it on master already and the scope would be outside of this PR.

vepor: I will add the rule just into `eslint-plugin-orbit-components` for now. We don't have to bother we the build cycle currently.

silvenon (approved):

PR #2811, opened 2021-03-04
improve: make compiling ESLint plugins faster

silvenon:
[SWC](https://swc.rs/) is faster than Babel, and in our case it improves the speed of compiling each plugin by about 500ms, because why not. We could also use it for new design tokens and see what would be the difference.

 Orbit.kiwi: https://orbit-docs-improve-swc.surge.sh
 Storybook: https://orbit-improve-swc.surge.sh


silvenon:
The speed improvement is insignificant, and with Babel it was already ~1s so this is definitely premature optimization.

PR #2812, opened 2021-03-05
ci/separate deploy workflows

silvenon:
- ci: speed up measuring size and improve naming
- ci: split deployment into manual workflows


silvenon:
Unfortunately, [manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) only works in the default branch, i.e. `master`, so it wouldn't solve our problem. We could probably get around it, but it would require a significant amount of work and wouldn't be that convenient.

PR #2813, opened 2021-03-05
feat(Tooltip): accommodate full-width children

silvenon:
Tooltip children like Button with `fullWidth` prop weren't taking up the entire width of the container because Tooltip's children wrapper was always inline. The new `fullWidth` prop allows choosing whether the wrapper will be inline or block.


packages/orbit-components/src/Tooltip/Tooltip.stories.js
```diff
@@ -93,6 +93,21 @@ storiesOf("Tooltip", module)
         "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
     },
   )
+  .add(
+    "Full-width",
```
mainframev, line 97:
```suggestion
    "FullWidth",
```

packages/orbit-components/src/Tooltip/Tooltip.stories.js
```diff
@@ -93,6 +93,21 @@ storiesOf("Tooltip", module)
...
+      const content = text("content", "Write your text here.");
+      return (
+        <Tooltip fullWidth content={content}>
+          <Button fullWidth>Full-width</Button>
+        </Tooltip>
+      );
+    },
+    {
+      info:
+        "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
```
mainframev, line 108: This description is related to the `playground` story, not to the current one.

packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
```diff
@@ -17,6 +17,7 @@ const MobileDialog = ({
   content,
   stopPropagation = false,
   removeUnderlinedText,
+  fullWidth = false,
```
mainframev, line 20: why it should be `false` by default? it will not have any difference with `undefined`

silvenon: Almost, but this way we're being explicit. E.g. if at some point we decide that TooltipPrimitive's  `fullWidth` defaults to `true`, that won't break Tooltip.

packages/orbit-components/src/Tooltip/Tooltip.stories.js
silvenon: I just copy-pasted the info from other Tooltip stories, but I'll remove this because it doesn't make sense.

packages/orbit-components/src/Tooltip/Tooltip.stories.js
silvenon: I want to write this in natural language. I thought that "full-width" was the correct way to write it, but "full width" is correct, so I fixed it.

mainframev (dismissed):

vepor (commented):
Do we really want to have it named `fullWidth` when it's connected to `display` attribute?

vepor (approved):

PR #2818, opened 2021-03-08
improve: check only needed environment vars

silvenon:
Always check only for variables needed for performing that particular task, instead of checking for unnecessarily failing if someone doesn't have a variable needed for a task they never happen to perform.

 Orbit.kiwi: https://orbit-docs-improve-loading-env-variables.surge.sh
 Storybook: https://orbit-improve-loading-env-variables.surge.sh

### DIFF
--- a/docs/utils/warnings.js
+++ b/docs/utils/warnings.js
@@ -7,7 +7,7 @@ const warnMissingAccessToken = error => {
       "Don't forget to include the required scopes for the token: all repo scopes and read:org.",
       "After creating a token in GitHub, add it to a .env file in the root folder with GH_TOKEN=<YOUR TOKEN>",
     );
-  } else {
+  } else if (typeof error.missing === "undefined") {
     console.error(error);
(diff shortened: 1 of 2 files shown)


mainframev (approved):

docs/utils/warnings.js
```diff
@@ -7,7 +7,7 @@ const warnMissingAccessToken = error => {
       "Don't forget to include the required scopes for the token: all repo scopes and read:org.",
       "After creating a token in GitHub, add it to a .env file in the root folder with GH_TOKEN=<YOUR TOKEN>",
     );
-  } else {
+  } else if (typeof error.missing === "undefined") {
```
mainframev, line 10: i think there is no need for that `else if` block, you can just add it as `if ((typeof error.missing === "undefined")` and that can be the first check before `GH_TOKEN` then it makes more sense I think  :)

mainframev (approved):

PR #2820, opened 2021-03-10
ci: use cache and artifacts to speed up workflow

CollierCZ:
It would also be possible to set up separate workflows for building/testing and deploying, but they would require a non-official action: https://github.com/dawidd6/action-download-artifact

Also, it doesn't seem currently possible to [update cache in GitHub workflows](https://github.com/actions/cache/issues/342), so the previous caching was doing nothing after any update to yarn.lock. My hope is this new version will restore from previous caches and then save to a unique one for the branch, saving time for any updates to that branch.

 Orbit.kiwi: https://orbit-docs-ci-workflow.surge.sh
 Storybook: https://orbit-ci-workflow.surge.sh


CollierCZ:
First run with no cache ~20 minutes not including publishing. Second run with cache ~10 minutes not including publishing.

silvenon (approved):

PR #2821, opened 2021-03-10
feat(tokens): add deprecated tokens for already defined replacements

vepor:
Added old – deprecated tokens to all existing replacements.
There's still a lot of tokens to be defined. 

Also added test for deprecated – to check if all old tokens returned by `getTokens` exist in the `createTheme`. The test case will fail currently.

Later I will also add tests to test if the deprecated and replacement token has the same value.


packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
```diff
@@ -2,13 +2,20 @@ import { Property } from "style-dictionary";
 
 import { getValue } from "../utils/get";
 
+const getDeprecatedToken = deprecatedToken => {
+  if (deprecatedToken) {
+    return deprecatedToken.name;
```
silvenon, line 7: Just to confirm: aside from extracting this into a function, this also changed from `deprecatedToken.value.name` to `deprecatedToken.name`, right?

packages/orbit-design-tokens/src/js/__tests__/deprecatedTokens.test.ts
```diff
@@ -0,0 +1,643 @@
...
+    paletteSocialFacebookHover: theme.palette.social.facebookHover,
+    paletteSocialFacebookActive: theme.palette.social.facebookActive,
+  };
+};
+
+describe("All deprecated tokens", () => {
+  const oldTokensObject = oldTokens();
+  const newTokensObject = createTheme();
+  it.each(Object.keys(oldTokensObject))("%s should exist in the return of createTheme", key => {
+    expect(key in newTokensObject).toBe(true);
```
silvenon, line 641: Generally avoid using `in` because it checks inherited properties as well, which we don't want.
```suggestion
    expect(newTokensObject).toHaveProperty(key);
```

silvenon (approved):

packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
vepor: Yes, on purpose. We want to always return the name of the token with `deprecatedToken.name`. If the value is aliased, the result would be something from the foundation. I will also implement a bit smarter replacement values, so if there's not any available replacement, the eslint-plugin would also suggest to use some global value.

Issue #2789, opened 2021-03-02
bug: useMediaQuery doesn't clean up after itself


mapsgeek:
**Is your feature request related to a problem? Please describe.**

useMediaQuery hook breaks SSR on nextjs, producing mismatch errors and unexpected rendering

**Describe the solution you'd like**
i'm very beginner in react but with some search i found this answer here
_Originally posted by @lfades in https://github.com/vercel/next.js/discussions/17443#discussioncomment-87097_

so i used react useEffect as a wrapper to update a local state value on useMediaQuery changes.
so i can achieve the navigationbar example turns into drawer on mobile
   ```
  const [queryLargeMobile, setqueryLargeMobile] = useState(true);
	const { isLargeMobile } = useMediaQuery();

	useEffect(() => {
		setqueryLargeMobile(isLargeMobile);
		return () => {};
	}, [isLargeMobile]);
```
**Describe alternatives you've considered**
i'm not sure if my solution is good but would be nice if useMediaQuery support SSR.

CollierCZ:
What result would you expect from `useMediaQuery` in SSR? You can use `useEffect` or read about [other approaches](https://material-ui.com/components/use-media-query/#server-side-rendering), but as it notes there, SSR and client-side queries aren't so compatible.

mapsgeek:
when i use `useMediaQuery` directly from `@orbit` i get mismatch error and the rendering of components gets missed up
so i did use a wrapper in `useEffect` as i show in code above but either with the wrapper or the native `useMediaQuery` i know get
`Can't perform a React state update on an unmounted component.` on routing change and tried many work arounds i found on search like adding `isMounted` hook on the clean up function but it just didn't work!
i understand that this would be not something `Orbit` library specific and more related to React but, any hints would be really appreciated.

mapsgeek:
the issue is apparently in the `useMediaQuery` hook, regardless of the SSR compatibility, the usage of the hook in a `Navbar` component on a route then moving to another page that doesn't have access or use the `Navbar` doesn't stop the hook from being called causing the warning `Can't perform a React state update on an unmounted component.`

using this hook [use-media](https://github.com/streamich/use-media), exact same code setup and usage scenario doesn't cause this issue because since the `Navbar` is unmounted the hook is simply stops updating the `mediaQuery `value.
However this UI library has great styles and good components but the support and replies on issues even for a react beginner
seems to be so worrying to depend on.

CollierCZ:
[...]

PR #2822, opened 2021-03-10
fix(useMediaQuery): remove listeners on cleanup

silvenon:
Orbit.kiwi: https://orbit-docs-fix-use-media-query-cleanup.surge.sh
 Storybook: https://orbit-fix-use-media-query-cleanup.surge.sh


CollierCZ:
Closes #2789

PR #2823, opened 2021-03-10
feat(tokens): add type variant tokens for alert button and social button

vepor:
Added component-specific tokens for alert-button and social-button components also with deprecated tokens that are semantically related to this two components.

### DIFF
--- a/packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
+++ b/packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
@@ -2,13 +2,20 @@ import { Property } from "style-dictionary";
 
 import { getValue } from "../utils/get";
 
+const getDeprecatedToken = deprecatedToken => {
+  if (deprecatedToken) {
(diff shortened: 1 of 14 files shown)


packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
```diff
@@ -2,13 +2,20 @@ import { Property } from "style-dictionary";
 
 import { getValue } from "../utils/get";
 
+const getDeprecatedToken = deprecatedToken => {
```
silvenon, line 5: Maybe it should be typed to define what interface it expects?

silvenon (approved):

packages/orbit-design-tokens/src/dictionary/formatters/json-tokens.ts
vepor: I have still some on going small issues with typeof of the style-dictionary, so I'll try to resolve them a bit later 🙂

PR #2826, opened 2021-03-11
feat(docs): add tabs

CollierCZ:
I'm not sure if it's not too much information to be loading gatsby-node. Could consider memozing, I suppose as well. But given that the whole process took 10 seconds, I wasn't sure it was necessary to try to optimize it. Also not super happy with the extra div to hide the box shadow bottom, but didn't find an easy other solution quickly enough.

 Orbit.kiwi: https://orbit-docs-docs-add-tabs.surge.sh
 Storybook: https://orbit-docs-add-tabs.surge.sh


docs/src/components/DocLayout.tsx
```diff
@@ -56,30 +81,39 @@ export default function DocLayout({ children, location, path, title }: Props) {
...
-                  GuidelinesSideBySide,
-                  Do,
-                  Dont,
-                }}
-              >
-                {children}
-              </MDXProvider>
+                {description && (
+                  <Box padding={{ left: "XXLarge" }}>
+                    {" "}
```
silvenon, line 93: Looks like this space is unnecessary.

docs/src/components/Tabs.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export interface TabObject {
+  frontmatter: {
+    title: string;
+  };
+  fields: {
+    slug: string;
+    tabCollection: string | null;
+  };
+}
```
silvenon, line 15: It seems that the shape of this interface is affected by how this data is received, could you flatten it?

docs/src/components/Tabs.tsx
```diff
@@ -0,0 +1,95 @@
...
+  location: string;
+}
+
+const Tabs = ({ location, tabs }: TabsProps) => {
+  const [activeTab, setActiveTab] = React.useState(location);
+  React.useEffect(() => setActiveTab(location), [location]);
+  return (
+    <Box padding={{ left: "medium" }} margin={{ top: "none" }}>
+      {tabs.map(tab => (
+        <Tab key={tab.frontmatter.title} tab={tab} isActive={activeTab === tab.fields.slug} />
```
silvenon, line 89: It's not necessary to have state, you can just do this:
```suggestion
        <Tab key={tab.frontmatter.title} tab={tab} isActive={location === tab.fields.slug} />
```
Also maybe rename the prop itself to `activeTab`, the name `location` strongly suggests the location object.

silvenon (changes requested):

docs/utils/document.js
```diff
@@ -7,18 +7,31 @@ const omitNumbers = filePath => filePath.replace(rePathNumbers, "$1");
 
 const ROOT = path.resolve(__dirname, "../src/documentation");
 
-function getDocumentUrlPath(filePath) {
-  const { dir, base, name } = path.parse(filePath);
+function getMetaPath(dir) {
+  return path.join(ROOT, dir, "meta.yml");
+}
+
+function getMetaFile(dir) {
```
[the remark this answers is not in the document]
CollierCZ: I was worried about confusing that with metadata in general and where that overlaps. So I went with getMetaFileData. Hope that makes sense.

docs/src/components/Tabs.tsx
CollierCZ: Flattened in Doc layout (which is different than DocLayout 🙊 ).

docs/src/components/Prose.tsx
```diff
@@ -3,19 +3,9 @@ import styled from "styled-components";
...
     margin-bottom: ${({ theme }) => theme.orbit.spaceLarge};
   }
-  > h2,
-  > h3,
-  > h4,
-  > h5,
-  > h6 {
-    margin-top: ${({ theme }) => theme.orbit.spaceLarge};
-  }
 `;
```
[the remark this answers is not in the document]
CollierCZ: I just remember you added `Prose` and had future plans for it. But I can remove it.

docs/src/components/Tabs.tsx
CollierCZ: Removed state, definitely agree with that. Given that the prop is a boolean, the naming with `is` still makes more sense to me. We have something similar [in VisualList](https://github.com/kiwicom/orbit/blob/master/docs/src/components/VisualList.tsx#L156), where the equivalent of `activeTab` (`openedVisual` in that case) is used for the wrapper, and the boolean is `opened`. I could use just `active` here.

docs/src/components/Prose.tsx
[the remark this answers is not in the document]
silvenon: Who knows anymore 😆 If we happen to need it in the future we can always extract it again.

docs/src/components/Tabs.tsx
CollierCZ: Right, so now that I figured out what you meant, I've changed it to `activeTab` 🙈

silvenon (approved):

PR #2838, opened 2021-03-16
chore(svgr): update to v5

mainframev:
Orbit.kiwi: https://orbit-docs-chore-svgr5.surge.sh
 Storybook: https://orbit-chore-svgr5.surge.sh


docs/src/components/ButtonLink.tsx
```diff
@@ -163,6 +163,7 @@ export default function ButtonLink<T extends React.ElementType = typeof DefaultC
   const Component = asProp || DefaultComponent;
 
   return (
+    // @ts-expect-error react-polymorphic-types is currently not compatible with styled-components
```
vepor, line 166: Is this in the scope of the PR? Just curious why it was necessary to use suppress comment right now.

mainframev: because it caused error in the pipeline 🤔

vepor: 👌

vepor (approved):

PR #2841, opened 2021-03-20
docs: polish navigation design

silvenon:
- remove navigation links, we don't have plans for them yet, but we might bring them back later
- make navigation responsive
- align Sidenav header with navigation
- add maximum width to navigation and document layout
  - the design suggests that navigation bar should take the entire viewport width, but that would be too wide on larger screens, we need to constrain it

 Orbit.kiwi: https://orbit-docs-docs-remove-navigation-links.surge.sh
 Storybook: https://orbit-docs-remove-navigation-links.surge.sh


docs/src/hooks/useCollections.ts
```diff
@@ -1,41 +0,0 @@
-import { useStaticQuery, graphql } from "gatsby";
```
CollierCZ, line 1: If we're not using this anymore, should we get rid of the field in gatsby-node?

CollierCZ (commented):
Overall, looks nice. Left a couple of small things to think about. 

Also, it's a little sub-optimal on very large screens when the content is constrained in the middle but then the drawer opens on the side. But there's probably no need to optimize for such edge cases.

docs/src/hooks/useCollections.ts
CollierCZ: Nevermind

docs/src/components/DocLayout.tsx
```diff
@@ -27,10 +28,14 @@ const StyledWrapper = styled.div`
 `;
 
 const StyledMain = styled.main`
+  position: relative;
   display: flex;
   flex-direction: column;
   overflow: hidden;
-  padding: 2rem;
+  padding: 2rem ${CONTENT_PADDING};
```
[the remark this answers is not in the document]
silvenon: Because `CONTENT_PADDING` refers only to horizontal padding, but it seemed verbose to name it that way. I wanted to keep the fact that vertical and horizontal padding are the same incidental.

docs/src/components/Layout.tsx
```diff
@@ -19,7 +20,7 @@ const StyledMain = styled.main`
   display: flex;
   flex-direction: column;
   overflow: hidden;
-  padding: 2rem;
+  padding: 2rem ${CONTENT_PADDING};
```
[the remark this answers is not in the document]
silvenon: Explained above.

docs/src/components/DocLayout.tsx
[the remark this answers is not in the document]
CollierCZ: I know naming is hard, but I definitely didn't understand that from the name. Do you anticipate the vertical and horizontal padding being different in the future?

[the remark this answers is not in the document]
silvenon: Now I use it for both paddings. ✅

CollierCZ (dismissed):

docs/src/components/Sidenav/index.tsx
```diff
@@ -13,7 +13,7 @@ interface WrapperProps {
 const StyledAsideWrapper = styled.aside<WrapperProps>`
   ${({ width, theme, shown }) => css`
     background: ${theme.orbit.paletteWhite};
-    padding: 10px 20px;
+    padding: 0 20px;
```
mainframev, line 16: it's a bit uncelar to me, when we should use `rem` and `px`, i understand cases when the css props are different, but now we have paddings in `px` and `rem` :) I think it's my part of the code, which was written before we started to use `rem` values in a project, but if you fixed it here, i think it will be better to replace as `rem` here too

mainframev (commented):
The star icon has now huge padding and box-shadow on active does not look right now. Before navigation was made with `Hide` components, that's was made like that because it does not cause flickering of the content on initial render, with `useMediaQueries` it happens, we have to do something with that. 

Just curious, where all the links? by design they should be in the navigation bar, something changed? it will be moved to sidenav?

docs/src/components/Sidenav/index.tsx
silvenon: For stuff like padding it doesn't matter. In my CSS I prefer using rems when I don't know the exact value, or when it doesn't matter that much, and saying `1rem` is nicer to me than `16px` because using pixels makes me think that it has to be exactly this, but that's just my own personal preference.

I saw no reason to change your padding to rems, and we probably aren't ever going to change the root font size, so it both can be treated the same. But if you have strong opinions whether to use one or the other I'll follow your choice 😃

mainframev: okay, that's maybe just my approach, usually when i start to use `rem` for paddings/margins, i have all values or almost all set as `rems`, because it's convenient

PR #2842, opened 2021-03-21
docs: move search to a modal

silvenon:
The main change is moving search to a modal and polishing how it works and looks. By having search in a modal we don't have to worry about layout issues, I was inspired by Docusaurus. Also, now search results contain the entire breadcrumbs, so you can actually know what you got, and search also takes into consideration this while searching. The searching is still imperfect due to my lack of understanding of elasticlunr, for example "passionat" works, but "passionate" doesn't.

Disregard how the search bar looks in the navigation bar, this will be fixed in a separate PR.


CollierCZ:
If it's going to open in a modal, could you make the trigger look more like a button and less like placeholder text on an input field? Right now it looks like it will be an input field and having to jump to a different place of the screen confuses me at first. I think it's both the color of the text and the length that give me that impression.

silvenon:
It's designed like an input on purpose to make it clear that you're going end up in an input element, having it look like a button might be confusing to some people as well. I'll think about how to make it less confusing, maybe with a better hover effect.

CollierCZ:
I think it should only look like an input if it's going to stay like an input. A button that opens a modal where there is an input seems like a familiar pattern to me. An input changing position does not. Can ask the designers. 🤷

docs/src/components/SearchInput/primitives/StyledWrapper.ts
```diff
@@ -1,10 +1,13 @@
...
+  Large,
+}
 
 const StyledWrapper = styled.div<{ size: Size }>`
   position: relative;
-  min-width: 24em; /* so that the placeholder fits in */
+  max-width: 24em; /* so that the placeholder fits in */
   ${({ size }) =>
     size === Size.Large &&
     css`
```
mainframev, line 13:
```suggestion
   `
```

docs/src/components/SearchInput/primitives/StyledWrapper.ts
```diff
@@ -1,10 +1,13 @@
...
-import { Size } from "..";
+export enum Size {
+  Regular,
+  Large,
+}
 
 const StyledWrapper = styled.div<{ size: Size }>`
   position: relative;
-  min-width: 24em; /* so that the placeholder fits in */
+  max-width: 24em; /* so that the placeholder fits in */
```
mainframev, line 10: i think for that kind of values we have to use pixels

docs/src/components/SearchInput/primitives/StyledDropdown.ts
```diff
@@ -1,38 +1,46 @@
...
-    }
+    border-radius: ${theme.orbit.borderRadiusLarge};
     outline: none;
     > p {
       font-size: calc(1em - 2px);
-      color: ${theme.orbit.colorTextPrimary};
+    }
+
+    ${type === "interactive" &&
+    css`
```
mainframev, line 19:
```suggestion
   `
```

docs/src/components/SearchInput/primitives/StyledDropdown.ts
```diff
@@ -1,38 +1,46 @@
...
+      &:active {
+        color: ${theme.orbit.paletteProductNormalActive};
+      }
+      > p {
+        color: ${theme.orbit.colorTextPrimary};
+      }
+    `}
+
+    ${type === "disabled" &&
+    css`
```
mainframev, line 36:
```suggestion
   `
```

docs/src/components/SearchInput/index.tsx
```diff
@@ -1,119 +1,66 @@
...
+`;
+const StyledTogglePrefix = styled(StyledPrefix)`
+  pointer-events: none;
+  position: absolute;
+  left: 1.5em;
+  top: 50%;
+  transform: translateY(-50%);
+`;
+const StyledSearchToggle = styled.button.attrs({ type: "button" })<{ size: Size }>`
+  ${({ theme, size }) => css`
```
mainframev, line 27:
```suggestion
  ${({ theme, size }) => `
```

packages/babel-plugin-orbit-components/index.js
```diff
@@ -55,6 +55,7 @@ const pathOverwrites = {
   TimelineStep: "Timeline/TimelineStep",
   SeatLegend: "Seat/components/SeatLegend",
   WizardStep: "Wizard/WizardStep",
+  useMediaQuery: "hooks/useMediaQuery",
```
mainframev, line 58: great! :)

mainframev (commented):
LGTM :)

docs/src/components/SearchInput/primitives/StyledWrapper.ts
silvenon: I used ems because I based it on placeholder, which is character count, so it makes sense to me. If there's a wider typeface, I'd like this value to be wider.

packages/babel-plugin-orbit-components/index.js
silvenon: Our tests caught this! 😄 💪

Strajk:
Keyboard navigation behaves weirdly 🤓

Where can I send invoice for testing? :D

silvenon:
- fix the bug reported above by @Strajk, thanks!
- apply design changes
- reimplement searching functionality with [fuzzaldrin-plus](https://www.npmjs.com/package/fuzzaldrin-plus) rather than elasticlunr, now it's simpler and it supports fuzzy searching
- implementing recent search visits is postponed for stable, multiple decisions need to be made for that feature

silvenon:
Note to self: I forgot about this, but it's not critical for code review:

Screen Shot 2021-04-14 at 18 48 22

docs/src/components/Search/SearchModal.tsx
```diff
@@ -0,0 +1,211 @@
...
+  `)}
+`;
+const StyledSearchWrapper = styled.div`
+  margin-bottom: 1rem;
+  font-size: 1rem;
+`;
+const StyledSearchInput = styled(StyledInput)`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
mainframev, line 63: these styled don't have any effect on input, i suppose your intention was to truncate the text, but it does not, behaves in a same way with and without :)

docs/src/components/Search/primitives/StyledMenu.ts
```diff
@@ -0,0 +1,41 @@
+import styled from "styled-components";
+
+export const StyledMenu = styled.ul<{ hasResults?: boolean }>`
+  ${({ hasResults = false }) => `
```
mainframev, line 4: why it's set by default as false? it's false even without default parameter

mainframev (changes requested):

docs/src/components/Search/primitives/StyledMenu.ts
silvenon: Just a coding pattern I was trying out, as by default it's not `false`, it's `undefined`, but it's not important.

docs/src/components/Search/SearchModal.tsx
```diff
@@ -0,0 +1,211 @@
+import React from "react";
+import styled, { css } from "styled-components";
+import { useStaticQuery, graphql, navigate } from "gatsby";
+import capitalize from "capitalize";
```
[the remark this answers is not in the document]
silvenon: Oh wow, there was also no `capitalize` in `docs/packge.json` 😅

docs/src/components/Search/primitives/StyledMenu.ts
mainframev: yes, by false i meant falsy value, not false itself.

mainframev (approved):

PR #2847, opened 2021-03-24
fix(ButtonPrimitive): remove redundant size property

vepor:
Removed redundant property `size` from the ButtonPrimitive component. It didn't have any effect on the styles, because we use `fontSize` and `height` properties to setup CSS attributes related to actual size.
 Orbit.kiwi: https://orbit-docs-fix-buttonprimitive-size.surge.sh
 Storybook: https://orbit-fix-buttonprimitive-size.surge.sh

### DIFF
--- a/packages/orbit-components/src/primitives/ButtonPrimitive/README.md
+++ b/packages/orbit-components/src/primitives/ButtonPrimitive/README.md
@@ -54,20 +54,11 @@ Table below contains all types of the props available in ButtonPrimitive compone
 | padding          | `string`                   |            | The inner padding of the ButtonPrimitive.                                                                                                                      |
 | rel              | `string`                   |            | Specifies the rel of an element. [See Functional specs](#functional-specs)                                                                                     |
 | role             | `string`                   |            | Specifies the role of an element.                                                                                                                              |
-| size             | [`enum`](#enum)            | `"normal"` | The size of the ButtonPrimitive.                                                                                                                               |
 | spaceAfter       | `enum`                     |            | Additional `margin-bottom` after component. [See this docs](https://github.com/kiwicom/orbit/tree/master/packages/orbit-components/src/common/getSpacingToken) |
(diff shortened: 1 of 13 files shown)


packages/orbit-components/src/primitives/ButtonPrimitive/README.md
```diff
@@ -54,21 +54,12 @@ Table below contains all types of the props available in ButtonPrimitive compone
...
 | width            | `string`                   |            | The width of the ButtonPrimitive. Can be any string - `100px`, `20%`.                                                                                          |
 
-### enum
-
-| size       |
-| :--------- |
-| `"small"`  |
-| `"normal"` |
-| `"large"`  |
-
```
silvenon, line 71: `size` needs to be removed from other READMEs of Button, ButtonLink and SocialButton as well.

packages/orbit-components/src/primitives/ButtonPrimitive/index.d.ts
```diff
@@ -5,6 +5,8 @@ import * as Common from "../../common/common";
 type inexactString = string | null | undefined;
 type functionReturningString = () => string;
 
+export type Size = "small" | "normal" | "large";
```
silvenon, line 8: Why is this being exported?

silvenon (changes requested):

packages/orbit-components/src/primitives/ButtonPrimitive/README.md
vepor: I can't actually delete that @silvenon. The `size` is still available on `Button`, `ButtonLink` and `SocialButton`. It was deleted only for the ButtonPrimitive, because it doesn't have any effect on the visual style. The related attributes to the `size` are calculated on the e.g. `Button` level.

packages/orbit-components/src/primitives/ButtonPrimitive/index.d.ts
vepor: It's consumed in e.g. `ButtonLink/index.d.ts`

silvenon (approved):

PR #2848, opened 2021-03-24
feat(docs): add table of contents

CollierCZ:
Currently sits on the right side of the main content in tablet and larger screens and as a collapse inside the content in smaller screens.

There's a height and a scroll on it because I'd like it to be sticky, but it currently can't be because the aside from SideNav requires `overflow: hidden` to have its transitions work. Possible solutions:
* Institute complicated JS rules to change the CSS for the ToC to replicate `position: sticky` behavior (see Sticky component -- the logic could be reused, but it's a lot to maintain)
* Change the animation of SideNav to something React-based like http://reactcommunity.org/react-transition-group/css-transition and then only have it in the DOM when it's shown then remove the `overflow: hidden`), but requires another dependency, etc.
* Not have it be sticky and be slightly sad
* Something more clever I haven't come up with

Other problem: I don't see an easy way to include headings generated by components. So, for example, `/component-status/` will seem to not have headings. Using rehype plugins didn't seem to work. 🤷 
 Orbit.kiwi: https://orbit-docs-docs-add-toc.surge.sh
 Storybook: https://orbit-docs-add-toc.surge.sh


docs/src/components/HeadingWithLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+  &:hover svg {
+    opacity: 100%;
+  }
+`;
+
+interface Props {
+  headingText: string;
+  headingLevel?: 2 | 3 | 4 | 5 | 6;
+  noId?: boolean;
+  spaceAfter?: "smallest" | "small" | "normal" | "medium" | "none";
```
mainframev, line 26: can you add export of `spaceAfter` type inside main package and use it here?

mainframev (commented):
When you make changes to our main package together with orbit.kiwi changes, add it to a separate commit, I noticed you added a small change, it would be nice to have it as separate commit in the changelog :)

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,78 @@
...
+    font-size: ${level <= 2 ? theme.orbit.fontSizeTextNormal : theme.orbit.fontSizeTextSmall};
+    text-indent: -${theme.orbit.spaceXLarge};
+    padding-left: ${theme.orbit.spaceXLarge};
+    display: inline-block;
+
+    ::before {
+      display: inline-block;
+      content: "";
+      border-top: 2px solid;
+      width: ${level === 1 ? "4px" : level === 2 ? "8px" : "12px"};
```
mainframev, line 30: it's nested ternary, can you rewrite it to a function? :)

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,78 @@
...
+    ${!alwaysVisible &&
+    `${StyledAnchor} {
+    visibility: hidden;
+  }`}
+
+    :hover ${StyledAnchor},:focus ${StyledAnchor},:active ${StyledAnchor} {
+      visibility: visible;
+    }
+  `}
+`;
```
mainframev, line 54: here it will be better to have it either inline because if the prop value is evaluated to false, it will not be rendered, but according to our last changes in `eslint-plugin-orbit` i suggest to have it like that: 

```
const StyledTocList = styled.ul<StyledTocListprops>`
   ${({ alwaysVisible }) => `
     max-height: 100vh;
     overflow-y: auto;
     visibility: ${!alwaysVisibile && `hidden`};
     ... etc
`
`
```

mainframev: plus you don need the `css` function, if you do not use function interpolation inside

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const getTocList = (array: TocItemObject[], level: Levels) => {
+  const nextLevel: Levels = ++level;
```
mainframev, line 57: eslint will definitely complain about it, you are assigning to a functional parameter

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,78 @@
...
+    visibility: hidden;
+  }`}
+
+    :hover ${StyledAnchor},:focus ${StyledAnchor},:active ${StyledAnchor} {
+      visibility: visible;
+    }
+  `}
+`;
+
+const getTocList = (array: TocItemObject[], level: Levels) => {
```
mainframev, line 56:
```suggestion
const getTocList = (array: TocItemObject[], level: number) => {
```

the function receives number, after `Levels.First`

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ));
+};
+
+interface TableOfContentsProps extends StyledTocListProps {
+  items: TocItemObject[];
+}
+
+const TableOfContents = ({ alwaysVisible, items }: TableOfContentsProps) => (
+  <nav>
+    <StyledTocList alwaysVisible={alwaysVisible}>{getTocList(items, Levels.First)}</StyledTocList>
```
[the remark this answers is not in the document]
CollierCZ: See if you still think so after a check when there's no ToC. 🤷

docs/src/components/DocLayout.tsx
```diff
@@ -96,25 +130,36 @@ export default function DocLayout({ children, description, location, path, tabs,
...
+              <TocWrapper>
+                <Hide on={["smallMobile", "mediumMobile", "largeMobile"]}>
+                  <TableOfContents items={tableOfContents} />
+                </Hide>
+              </TocWrapper>
+              <ContentContainer padding="XLarge" elevation="raised">
+                <Hide on={["tablet", "desktop", "largeDesktop"]}>
+                  <Collapse label="Table of contents">
+                    <TableOfContents items={tableOfContents} alwaysVisible />
+                  </Collapse>
```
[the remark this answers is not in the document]
CollierCZ: But it appears in different places in the DOM: once outside ContentContainer and once inside it. That's why I put it here. I could add conditional rendering to it, but I thought that might be complicated.

docs/src/components/HeadingWithLink.tsx
CollierCZ: It already was exported, so I just used it.

docs/src/hooks/useMdxPageTableofContents.ts
```diff
@@ -0,0 +1,25 @@
+import { useStaticQuery, graphql } from "gatsby";
+
+const useMdxPageTableofContents = (pagename: string) => {
+  const { mdxPages } = useStaticQuery(
+    graphql`
+      query MdxPagesQuery {
+        mdxPages: allMdx(filter: { fields: { collection: { eq: "pages" } } }) {
+          nodes {
+            tableOfContents(maxDepth: 4)
```
silvenon, line 9: Like you said, this can only detect titles statically, in order to detect _all_ headings we would need to analyze the resulting HTML. The easiest way might be to do it on the client using [rehype-dom](https://github.com/rehypejs/rehype-dom).

docs/src/pages/component-status.mdx
```diff
@@ -5,19 +5,28 @@ description: A detailed breakdown of our components and status of their implemen
 
 import { ComponentStatusList } from "../components/ComponentStatus";
 import Layout from "../components/Layout";
+import useMdxPageTableofContents from "../hooks/useMdxPageTableofContents.ts";
```
silvenon, line 8:
```suggestion
import useMdxPageTableofContents from "../hooks/useMdxPageTableofContents";
```

silvenon (changes requested):
I'll mess with sticky behavior locally and get back to you.

docs/src/components/DocLayout.tsx
```diff
@@ -65,16 +76,39 @@ const ContentContainer = styled(Box)`
...
 `;
 
+const TocWrapper = styled.div`
+  ${mediaQueries.tablet(css`
+    order: 2;
+  `)}
+
+  padding-left: ${({ theme }) => theme.orbit.spaceSmall};
+  > * {
+    position: sticky;
```
silvenon, line 86: I can't figure out why this doesn't work, I'll need an explanation once again how it's connected to `overflow: hidden`.

docs/src/components/DocLayout.tsx
```diff
@@ -65,16 +76,39 @@ const ContentContainer = styled(Box)`
...
+    top: 0;
+    transition: top ${({ theme }) => theme.orbit.durationNormal} ease-in-out;
```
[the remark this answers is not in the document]
CollierCZ: I think I copied this from kiwi.wiki, which uses the Orbit NavigationBar, which reappears on scrolling up. It's a useful effect. 🤷

docs/src/components/HeadingWithLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+      case 2:
+        return ["h2", "title1"];
+      case 4:
+        return ["h4", "title3"];
+      case 5:
+        return ["h5", "title4"];
+      case 6:
+        return ["h6", "title5"];
+      default:
+        return ["h3", "title2"];
```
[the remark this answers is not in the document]
CollierCZ: Though without a default case, typescript complains that there's no return from the function.

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const getWidth = (level: number) => {
+  if (level === 0) return "4px";
+  if (level === 1) return "8px";
+  return "12px";
+};
+
+const StyledAnchor = styled.a<StyledAnchorProps>`
+  ${({ level, theme }) => css`
+    color: ${level === 0 ? theme.orbit.paletteInkNormal : theme.orbit.paletteInkLight};
```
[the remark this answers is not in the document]
CollierCZ: Ah yes, I knew I forgot something. A whole scrollspy and everything. 🤦

docs/src/components/TableOfContents.tsx
```diff
@@ -0,0 +1,74 @@
...
+    font-size: ${level <= 1 ? theme.orbit.fontSizeTextNormal : theme.orbit.fontSizeTextSmall};
+    text-indent: -${theme.orbit.spaceXLarge};
+    padding-left: ${theme.orbit.spaceXLarge};
+    display: inline-block;
+
+    ::before {
+      display: inline-block;
+      content: "";
+      border-top: 2px solid;
```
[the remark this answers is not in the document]
CollierCZ: Just setting the background to currentColor is enough. It inherits from its element.

docs/src/pages/component-status.mdx
CollierCZ: As I noted to Victor in a different PR, the eslint rules you set up expect extensions to be tsx. So when the files are ts, it complains. Should I 1. change the file extension, 2. change the rule, or 3. leave it as is?

docs/src/components/DocLayout.tsx
CollierCZ: It's solved with #2856

docs/src/hooks/useMdxPageTableofContents.ts
CollierCZ: You also brought up the idea of React Context. 🤷

docs/src/components/TableOfContents.tsx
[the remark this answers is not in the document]
silvenon: According to the design at 2nd level the line is _not_ `currentColor`, it's lighter than text. Text is "Text / Normal Medium", while line is "Ink / Lighter".

docs/src/pages/component-status.mdx
silvenon: Sorry about the ESLint mistake 🤦 it should be changed to allow `.ts`

docs/src/components/HeadingWithLink.tsx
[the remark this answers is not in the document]
silvenon: Sounds good 👍

docs/src/components/DocLayout.tsx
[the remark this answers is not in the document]
silvenon: As long as it does something, I didn't notice anything, that's why I was asking.

docs/src/components/TableOfContents.tsx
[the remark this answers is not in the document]
CollierCZ: That was a good catch!  I definitely didn't notice that. But I found out it was a mistake. So `currentColor` is fine. 🙂

silvenon (commented):
The `Sidenav` Portal PR has been merged, so do whatever you need to do to make it sticky 🤞  and a few more comments and this can be merged. We can refine it in a separate PR, I removed the commit I pushed here.

docs/src/hooks/useMdxPageTableofContents.ts
silvenon: I'll deal with this in a separate PR.

docs/src/components/HeadingWithLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Link as LinkIcon } from "@kiwicom/orbit-components/icons";
+import { SpaceAfter } from "@kiwicom/orbit-components/lib/common/common";
+
+import { getTextFromChildren, slugify } from "../utils/common";
+
+export const StyledAnchorWrapper = styled.div`
+  margin-top: -10px;
+
+  svg {
+    opacity: 0%;
```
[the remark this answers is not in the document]
CollierCZ: "A <number> in the range 0.0 to 1.0, inclusive, or a <percentage> in the range 0% to 100%, inclusive" ([source](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#values)). Not universal support though, so I can change it.

CollierCZ:
It's sticky now. Just needed to remove the hiding of the overflows.

CollierCZ:
Since I addressed that comment and I'd like this for another feature, I'm going to go ahead and merge it in (based on the "merge fast" strategy you proposed). If there's anything that needs fixing, let me know and I'll open a PR for it.

PR #2852, opened 2021-03-24
chore(stories): update all stories to csf format

mainframev:
I decided to update all stories to CSF format, because in order to update to SC5 I have to solve flow errors, it will be easier to update flow first, to reduce the amount of error (i checked, some errors disappear after update) and we still have to update flow, so why not to update both. 

 Orbit.kiwi: https://orbit-docs-chore-update-all-stories-to-csf.surge.sh
 Storybook: https://orbit-chore-update-all-stories-to-csf.surge.sh

### DIFF
--- a/packages/orbit-components/src/Icon/Icon.stories.js
+++ b/packages/orbit-components/src/Icon/Icon.stories.js
@@ -1,68 +1,78 @@
 // @flow
 
 import * as React from "react";
-import { storiesOf } from "@storybook/react";
 import { select, text, boolean } from "@storybook/addon-knobs";
 
-import * as Icons from "../icons";
 import IconList from "./IconList";
+import * as Icons from "../icons";
 import { ICON_SIZES, ICON_COLORS } from "./consts";
 import RenderInRtl from "../utils/rtl/RenderInRtl";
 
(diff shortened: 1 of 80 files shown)


packages/orbit-components/src/Icon/Icon.stories.js
```diff
@@ -1,68 +1,70 @@
...
-          <Icon reverseOnRtl />
-        </RenderInRtl>
-      );
-    },
-    {
-      info:
-        "We use icons to draw attention to specific actions in our products. Visit Orbit.Kiwi for more detailed guidelines.",
-    },
-  )
-  .lokiSkip("List of all icons", () => <IconList />);
```
silvenon, line 68: Let me know about this.

silvenon:
Otherwise looks good ✅

packages/orbit-components/src/Icon/Icon.stories.js
mainframev: fixed manually ;)

mainframev: updated manually

silvenon (dismissed):

PR #2856, opened 2021-03-25
docs: render Sidenav in Portal

silvenon:
When scrolling while Sidenav is open, Sidenav would scroll along the content, so in order to make it fixed we need to render it in a portal.

Let me know if I'm overlooking something.

 Orbit.kiwi: https://orbit-docs-docs-sidenav-portal.surge.sh
 Storybook: https://orbit-docs-sidenav-portal.surge.sh

### DIFF
--- a/docs/src/components/Sidenav/index.tsx
+++ b/docs/src/components/Sidenav/index.tsx
@@ -1,56 +1,10 @@
 import * as React from "react";
 import styled, { css } from "styled-components";
-import { Portal, mediaQueries } from "@kiwicom/orbit-components";
+import { Portal, Drawer } from "@kiwicom/orbit-components";
 import Menu from "@kiwicom/orbit-components/lib/icons/MenuHamburger";
-import Close from "@kiwicom/orbit-components/lib/icons/Close";
 import mq from "@kiwicom/orbit-components/lib/utils/mediaQuery";
-import useClickOutside from "@kiwicom/orbit-components/lib/hooks/useClickOutside";
 
 import { CONTENT_PADDING } from "../../consts";
-
-interface WrapperProps {
-  width: number;
-  shown: boolean;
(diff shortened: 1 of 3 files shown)


docs/src/components/Sidenav/index.tsx
```diff
@@ -1,23 +1,40 @@
...
+
 interface WrapperProps {
   width: number;
   shown: boolean;
 }
 
+export const HEADER_HEIGHT = "52px"; // safely above the search input height
+export const paddingMixin = css`
+  padding-top: 0.5rem;
+  padding-bottom: 0.5rem;
```
mainframev, line 19: could you `padding: 0.5rem 0` instead? I see wrong `padding-left` was removed below

CollierCZ:
Is there a reason not to use a Drawer for it? Either way, it removes the need for `overflow: hidden` and so resolves the issue with the sticky ToC.

docs/src/components/Sidenav/index.tsx
silvenon: I don't think it's a good practice for mixins, especially responsive ones. Imagine that I set left and right padding to the element that I'm applying the mixin to, that means that the mixin would reset that padding on larger screens, which would be unexpected because the function of the padding is only to share vertical padding between sidenav and navigation (the name sucks, though). I think having robust styles is worth this small DRY sacrifice.

mainframev: in my opinion, there is no need for that padding mixin at all. Especially when you export it from `SideNav` and use it inside `Navbar`, for such kind of mixins should be then some common folder and it's used only on two places

silvenon:
Using Drawer now 💯

mainframev (approved):

PR #2857, opened 2021-03-26
fix(TextLink): called twice when stopPropagation specified

### LINKED ISSUE #2853: textlink onClick when stopPropagation call two time
textlink onClick when stopPropagation enable call two time

need use else

```
  const onClickHandler = ev => {
    if (stopPropagation) {
      ev.stopPropagation();
      if (onClick) onClick(ev);
    }
    if (onClick) onClick(ev);
  };
```
https://github.com/kiwicom/orbit/blob/7b1152bb9372d9ab2824000c59e9181ce9a9f65d/packages/orbit-components/src/TextLink/index.js#L124

vepor:
Fixed #2853 
 Orbit.kiwi: https://orbit-docs-fix-textlink-click.surge.sh
 Storybook: https://orbit-fix-textlink-click.surge.sh

### DIFF
--- a/packages/orbit-components/src/TextLink/__tests__/index.test.js
+++ b/packages/orbit-components/src/TextLink/__tests__/index.test.js
@@ -25,6 +25,17 @@ describe("#TextLink", () => {
     expect(onClick).toHaveBeenCalled();
   });
 
+  it("should execute onClick method once when stopPropagation", () => {
+    const onClick = jest.fn();
+    render(
+      <TextLink onClick={onClick} stopPropagation>
+        {title}
+      </TextLink>,
+    );
+    userEvent.click(screen.getByText(title));
+    expect(onClick).toHaveBeenCalledTimes(1);
+  });
+
   it("should render with props", () => {
     const dataTestLeftIcon = "leftIcon";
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/TextLink/__tests__/index.test.js
```diff
@@ -25,6 +25,17 @@ describe("#TextLink", () => {
...
 
+  it("should execute onClick method once when stopPropagation", () => {
+    const onClick = jest.fn();
+    render(
+      <TextLink onClick={onClick} stopPropagation>
+        {title}
+      </TextLink>,
+    );
+    userEvent.click(screen.getByText(title));
+    expect(onClick).toHaveBeenCalledTimes(1);
```
mainframev, line 36: not a big deal, but in that case `toHaveBeenCalled` is enough i think :)

silvenon: @mainframev this PR fixes the bug where the handler was being called more than once. Using `toHaveBeenCalled` would pass even if the bug was still present, so using `toHaveBeenCalledTimes(1)` is the right thing to do.

mainframev (approved):

PR #2872, opened 2021-04-06
fix: orbit.kiwi layout wrapper

mainframev:
fixed the issue of overflowed content, check current master, how it looks like

 Orbit.kiwi: https://orbit-docs-fix-overflow.surge.sh
 Storybook: https://orbit-fix-overflow.surge.sh


silvenon (approved):
Awesome!

PR #2873, opened 2021-04-06
refactor(Footer): small changes to footer

mainframev:
I've refactored a bit footer component, I will have to put a `Vercel` logo there soon. Noticed, that it can be done a bit more simply via `Stack`. Now it much the initial design, since the height of both parts of footer are equal to the height of the proposed design from Will. 

 Orbit.kiwi: https://orbit-docs-feat-verce-banner.surge.sh
 Storybook: https://orbit-feat-verce-banner.surge.sh

### DIFF
--- a/docs/src/components/Footer.tsx
+++ b/docs/src/components/Footer.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import styled, { css } from "styled-components";
-import { Box, Text, Desktop, mediaQueries } from "@kiwicom/orbit-components";
+import { Text, Desktop, Stack, mediaQueries } from "@kiwicom/orbit-components";
 import { Link } from "gatsby";
 
 import Switch from "./Switch";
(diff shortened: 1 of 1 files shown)


mainframev:
@silvenon could you check 🙏🏻 ? if you are agree with the changes

docs/src/components/Footer.tsx
```diff
@@ -107,96 +98,72 @@ export default function Footer() {
...
-              <Box
-                display="flex"
-                align="center"
-                padding={{ top: "XSmall", bottom: "XSmall" }}
-                tablet={{ padding: { top: "XSmall", bottom: "XSmall", left: "XXLarge" } }}
-                desktop={{ padding: { left: "XXLarge" } }}
-                largeDesktop={{ padding: { top: "large", bottom: "large", left: "XXLarge" } }}
-              >
-                <Box padding={{ right: "XLarge" }} desktop={{ shrink: 0 }}>
+              <Stack inline grow={false} spacing="none">
```
silvenon, line 138: You need to add spacing between label and the switch. This is how the footer looked at 367 px before:

Screen Shot 2021-04-07 at 12 16 15

and after your change:

Screen Shot 2021-04-07 at 12 15 47

Plus, the switch is overflowing the footer on the right (the screenshot is taken full-width).

You can also notice that the "Built & maintained by Kiwi.com" part is no longer centered, not sure if that was on purpose.

silvenon (changes requested):

silvenon (approved):
Looks good now 👍

PR #2875, opened 2021-04-08
fix(useBoundingRect): fix TypeScript declaration

silvenon:
BREAKING CHANGE: In TypeScript `useBoundingRect` now requires a type parameter based on which HTML element it measures:

```tsx
function App() {
  const [dimensions, ref] = useBoundingRect<HTMLDivElement>();
  return <div ref={ref} />;
}
```

 Orbit.kiwi: https://orbit-docs-fix-ts-use-bounding-rect.surge.sh
 Storybook: https://orbit-fix-ts-use-bounding-rect.surge.sh


silvenon:
The TypeScript test was failing otherwise.

mainframev (approved):

PR #2876, opened 2021-04-09
fix: use-rtl error messages

mainframev:
Orbit.kiwi: https://orbit-docs-fix-use-rtl.surge.sh
 Storybook: https://orbit-fix-use-rtl.surge.sh

### DIFF
--- a/packages/eslint-plugin-orbit-components/src/rules/useRtl.ts
+++ b/packages/eslint-plugin-orbit-components/src/rules/useRtl.ts
@@ -2,13 +2,13 @@ import * as t from "@babel/types";
 import { Rule } from "eslint";
 
 export const ERROR_RTL_SPACING =
-  "Values for right and left edges are different. Use Orbit's rtlSpacing utility.";
+  "Use Orbit's rtlSpacing utility because values for right and left edges are different. https://orbit.kiwi/utilities/right-to-left-languages/#rtlspacing";
 
 export const ERROR_BORDER_RADIUS =
   "border-radius edge values are different. Use borderRadius utility.";
 
-export const errorBasic = (prop, value) =>
-  `Values for ${prop} and ${value} edges are different. Use Orbit's rtlSpacing utility. https://orbit.kiwi/utilities/right-to-left-languages/#rtlspacing`;
+export const errorBasic =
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-orbit-components/src/rules/useRtl.ts
```diff
@@ -2,13 +2,13 @@ import * as t from "@babel/types";
...
-  "Values for right and left edges are different. Use Orbit's rtlSpacing utility.";
+  "Use Orbit's rtlSpacing utility because values for right and left edges are different. https://orbit.kiwi/utilities/right-to-left-languages/#rtlspacing";
 
 export const ERROR_BORDER_RADIUS =
   "border-radius edge values are different. Use borderRadius utility.";
 
-export const errorBasic = (prop, value) =>
-  `Values for ${prop} and ${value} edges are different. Use Orbit's rtlSpacing utility. https://orbit.kiwi/utilities/right-to-left-languages/#rtlspacing`;
+export const errorBasic =
+  "Values for right and left edges are different. Use Orbit's rtl utility. https://orbit.kiwi/utilities/right-to-left-languages/#right";
```
silvenon, line 11: I would rather not have `#right` then, it seems arbitrary.
```suggestion
  "Values for right and left edges are different. Use Orbit's rtl utility. https://orbit.kiwi/utilities/right-to-left-languages";
```

packages/eslint-plugin-orbit-components/src/rules/useRtl.ts
```diff
@@ -2,13 +2,13 @@ import * as t from "@babel/types";
 import { Rule } from "eslint";
 
 export const ERROR_RTL_SPACING =
-  "Values for right and left edges are different. Use Orbit's rtlSpacing utility.";
+  "Use Orbit's rtlSpacing utility because values for right and left edges are different. https://orbit.kiwi/utilities/right-to-left-languages/#rtlspacing";
```
silvenon, line 5: This change causes the style of error messages to be inconsistent again, not sure if that's on purpose.

mainframev: why incosistent? i changed it, because after the last change it was showing `rtlSpacing` error where supposed to be `right/left` :)

PR #2890, opened 2021-04-15
ci: restore Vale check

CollierCZ:
The original issue seems fixed: https://github.com/errata-ai/vale-action/issues/33
 Orbit.kiwi: https://orbit-docs-fix-vale-version.surge.sh
 Storybook: https://orbit-fix-vale-version.surge.sh


silvenon (approved):
🎉 📜 🖋️

PR #2891, opened 2021-04-15
fix(docs): external icon

CollierCZ:
The fix for the icon being inline is maybe a little hacky. 🤷
 Orbit.kiwi: https://orbit-docs-fix-external-icon.surge.sh
 Storybook: https://orbit-fix-external-icon.surge.sh

### DIFF
--- a/docs/src/mdx-components.tsx
+++ b/docs/src/mdx-components.tsx
@@ -152,6 +152,10 @@ export const dt = ({ children }: React.HTMLAttributes<HTMLElement>) => (
 
 export const inlineCode = InlineCode;
 
+const isKeyboardEvent = (e): e is React.KeyboardEvent => {
+  return (e as React.KeyboardEvent).getModifierState !== undefined;
+};
+
 export const a = function Anchor({
   children,
   href,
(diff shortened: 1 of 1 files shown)


docs/src/mdx-components.tsx
```diff
@@ -165,15 +170,28 @@ export const a = function Anchor({
...
-        onClick={event => {
+        iconRight={useExternalIcon && <NewWindow ariaLabel="Opens in new window" />}
+        onClick={(event: React.SyntheticEvent<HTMLLinkElement>) => {
           if (isExternal || !href) return;
+
+          // Allows opening in a new tab with ctrl/cmd + click
+          if (isKeyboardEvent(event)) {
+            if (event.metaKey || event.ctrlKey) return;
+          }
+
```
silvenon, line 194: There's a way in which can avoid that `onClick` handler. 💡 We can create a component that adjust's Gatsby's Link component to TextLink's `asComponent` prop:

```tsx
function LinkForOrbitTextLink({ href, ...props }: { href: string }) {
  return <Link to={href} {...props} />;
}
```

Then just add a conditional `asComponent` to `TextLink`:

```tsx
<TextLink
  // @ts-expect-error type declaration is not permissive enough
  asComponent={isExternal ? "a" : LinkForOrbitTextLink}
  // ...
```

TypeScript will probably complain because of the dumb type declaration which only allows `className` prop for `asComponent` components, which is why I added the `@ts-expect-error`.

silvenon (changes requested):

docs/src/mdx-components.tsx
CollierCZ: Seems interesting. I remembered you saying you were going to use Gatsby's Link and `asComponent`, so I figured there was a reason it wasn't there. But this way seems more elegant.

silvenon (approved):

PR #2892, opened 2021-04-15
docs(sandbox): examples adjustments

mainframev:
Storybook: https://orbit-feat-sandbox-ruler.surge.sh


CollierCZ:
Can the iframe be optional? With the slow loading on Surge, you can see how space for the component isn't created on first load (I thought the Loading would do that, but it doesn't seem to).

Also, it'd be nice if the iframe could have an option for a set height. Example: https://orbit-docs-feat-sandbox-ruler.surge.sh/components/overlay/modal/#structure-content If that's are going to be in the iframe, it's unusable at the current height. Or in that case, it might be OK without the iframe. 🤷‍♂️

mainframev:
> Can the iframe be optional? With the slow loading on Surge, you can see how space for the component isn't created on first load (I thought the Loading would do that, but it doesn't seem to).
> 
> Also, it'd be nice if the iframe could have an option for a set height. Example: https://orbit-docs-feat-sandbox-ruler.surge.sh/components/overlay/modal/#structure-content If that's are going to be in the iframe, it's unusable at the current height. Or in that case, it might be OK without the iframe. 🤷‍♂️

it's a draft PR yet :D 
It has `maxHeight` property :) 

Not sure about optionality, because we will have to have it twice and for the `full-page` will be opened a copy of the example. Now I'm checking what I can do to speed it up, but unfortunately, iframes are slower in general.

CollierCZ:
Sorry, missed the draft status. I blame dark mode. 🙈

I meant `minHeight` rather than `maxHeight`.

If it can't be optional, it might be OK. It's just the first thing on many pages and having it be slow loading there isn't a great user experience. But I'm sure you'll figure something out. 💪

CollierCZ:
Some thoughts without looking much at the code:

* Looks smooth. I also like that the playground and code buttons are combined into one. (At least for now)
* The full width is wider than content can be. That's a problem we had even with the old version, but we should maybe think about it at some point?
* The switching between viewports doesn't work ideally
  * Clicking mobile small does nothing
  * Click on any viewport smaller than the current one loads mobile small
  * They show widths that are different than our breakpoints (e.g., mobile large shows up to 700)
* It still should probably have the ability to set minHeight, so the [modal examples](https://f0e65e95.orbit.pages.dev/components/overlay/modal/) can show up

mainframev:
>   * They show widths that are different than our breakpoints (e.g., mobile large shows up to 700)

it's supposed to be like that, I did it on purpose because it takes the initial size and shows in the range of viewport sizes. It should not show exactly the same viewport size as we have, but just under which viewport size it is.

mainframev:
>   * They show widths that are different than our breakpoints (e.g., mobile large shows up to 700)

it's supposed to be like that, I did it on purpose because it takes the initial size and shows in the range of viewport sizes. It should not show exactly the same viewport size as we have, but just under which viewport size it is. 

LargeMobile is everything between 576 and 768 px, so it looks correct

mainframev:
>   * Click on any viewport smaller than the current one loads mobile small

it's probably because initially it had different logic, smallMobile was the first viewport and default, i will change it

CollierCZ:
> it's supposed to be like that, I did it on purpose because it takes the initial size and shows in the range of viewport sizes. It should not show exactly the same viewport size as we have, but just under which viewport size it is.

My issue is that it displays the number next to the viewport name, which makes it look like the number is the size for the viewport. 700 isn't a breakpoint for LargeMobile. The upper number is also not how we give breakpoints [elsewhere in the docs](https://orbit.kiwi/utilities/media-queries/)

mainframev:
> > it's supposed to be like that, I did it on purpose because it takes the initial size and shows in the range of viewport sizes. It should not show exactly the same viewport size as we have, but just under which viewport size it is.
> 
> My issue is that it displays the number next to the viewport name, which makes it look like the number is the size for the viewport. 700 isn't a breakpoint for LargeMobile. The upper number is also not how we give breakpoints [elsewhere in the docs](https://orbit.kiwi/utilities/media-queries/)

I will remove that number, I think it's enough to show only the current breakpoint name.

PR #2897, opened 2021-04-19
feat(docs): add redirects to old doc structure

CollierCZ:
Orbit.kiwi: https://orbit-docs-docs-redirects.surge.sh
 Storybook: https://orbit-docs-redirects.surge.sh


silvenon (changes requested):
For most components, other than accessibility and utilities, you can solve redirects programmatically using [`createRedirect`](https://www.gatsbyjs.com/docs/reference/config-files/actions/#createRedirect), because they follow the same pattern.

CollierCZ:
> For most components, other than accessibility and utilities, you can solve redirects programmatically using [`createRedirect`](https://www.gatsbyjs.com/docs/reference/config-files/actions/#createRedirect), because they follow the same pattern.

I could. I just thought that going forward it might help to have redirects for each page in a single place. So if a component changed its location, you'd see all redirects to in in the same file, rather than in different files. I was thinking of it from a content management perspective. But if you don't think that'd be useful in the future, I can change it.

CollierCZ:
> For most components, other than accessibility and utilities, you can solve redirects programmatically using [`createRedirect`](https://www.gatsbyjs.com/docs/reference/config-files/actions/#createRedirect), because they follow the same pattern.

I could. I just thought that going forward it might help to have redirects for each page in a single place. So if a component changed its location, you'd see all redirects to in in the same file, rather than in different files. I was thinking of it from a content management perspective. But if you don't think that'd be useful in the future, I can change it.

silvenon (approved):
I was in a dilemma whether to post that review comment exactly because of that reason, I totally understand the benefit of keeping it here. Repetition is not always bad 👍  besides, it's not like we're going to change these URLs.

PR #2899, opened 2021-04-19
docs: fix front page layout

mainframev:
fixed visual bugs on the front page, now it should look better. For icons inside `Tile/BrandedTile` set `display: none` on smallMobile and mediumMobile, I suppose it just takes the space and we can turn it off for such small viewports. 

Another thing, that can be discussed is usage of relative values, we have a lot of `rems`, but we do not change the base value, it can be useful for reducing the margin/padding size for narrow viewports, i doubt that we have to have `32px` padding for Tile on `smallMobile`

 Orbit.kiwi: https://orbit-docs-fix-front-page.surge.sh
 Storybook: https://orbit-fix-front-page.surge.sh


silvenon (approved):
Looks good!

PR #2900, opened 2021-04-21
feat(docs): add image container

CollierCZ:
Orbit.kiwi: https://orbit-docs-docs-image-container.surge.sh
 Storybook: https://orbit-docs-image-container.surge.sh

### DIFF
--- a/docs/src/components/ImageContainer.tsx
+++ b/docs/src/components/ImageContainer.tsx
@@ -10,24 +10,22 @@ interface Props {
 }
 
 const ImageContainer = ({ align = "center", border, size = "large", children }: Props) => {
-  const getContainerWidth = containerSize => {
-    if (containerSize === "small") return "200px";
-    if (containerSize === "medium") return "400px";
+  const getContainerWidth = () => {
+    if (size === "small") return "200px";
+    if (size === "medium") return "400px";
     return "650px";
   };
-  const containerWidth = getContainerWidth(size);
 
-  const getContainerMargin = containerAlignment => {
-    if (containerAlignment === "right") return "0 0 0 auto";
-    if (containerAlignment === "left") return "0";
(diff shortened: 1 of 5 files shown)


docs/src/components/ImageContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  size?: "small" | "medium" | "large";
+}
+
+const ImageContainer = ({ align = "center", border, size = "large", children }: Props) => {
+  const getContainerWidth = containerSize => {
+    if (containerSize === "small") return "200px";
+    if (containerSize === "medium") return "400px";
+    return "650px";
+  };
+  const containerWidth = getContainerWidth(size);
```
silvenon, line 18: It's a little confusing that `getContainerWidth` and `getContainerMargin` take an argument, so either move them outside of the component or use whatever value you're passing inside the function, e.g.:

```js
  const getContainerWidth = () => {
    if (size === "small") return "200px";
    if (size === "medium") return "400px";
    return "650px";
  };
  const containerWidth = getContainerWidth();
```

docs/src/components/ImageContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+    if (containerAlignment === "right") return "0 0 0 auto";
+    if (containerAlignment === "left") return "0";
+    return "0 auto";
+  };
+  const containerMargin = getContainerMargin(align);
+  return (
+    <div
+      css={css`
+        margin: ${containerMargin};
+        width: ${containerWidth};
```
silvenon, line 30: No need to store the return value to a variable:
```suggestion
        margin: ${getContainerMargin()};
        width: ${getContainerWidth()};
```

silvenon (changes requested):

silvenon (approved):

PR #2904, opened 2021-04-22
chore: remove docs deploy on surge

mainframev:
Storybook: https://orbit-chore-remove-docs-deploy.surge.sh

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -28,17 +28,6 @@ jobs:
             node-modules-${{ github.ref}}-
             node-modules-
 
-      - name: Restore Gatsby cache
-        uses: actions/cache@v2
-        with:
-          path: |
-            docs/.cache
-            docs/public
-          key: gatsby-${{ github.ref}}-${{ hashFiles('docs/gatsby-config.js', 'docs/gatsby-node.js') }}
-          restore-keys: |
-            gatsby-${{ github.ref}}-
-            gatsby-
-
       - name: Install dependencies
         run: yarn install --frozen-lockfile && yarn bootstrap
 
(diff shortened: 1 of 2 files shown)


silvenon (dismissed):

mainframev:
@silvenon any idea why it fails? seems like it can't apply artifact 🤔

.github/workflows/test_and_deploy.yml
```diff
@@ -48,15 +37,13 @@ jobs:
       - name: Build
         env:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
-        run: npx lerna run build
+        run: npx lerna run --ignore orbit.kiwi build
```
CollierCZ, line 40: This does not leave out the docs build. I think using `@kiwicom/orbit.kiwi` would work, but haven't checked it.

.github/workflows/test_and_deploy.yml
```diff
@@ -71,8 +58,7 @@ jobs:
           yarn prettier:test
 
       - name: Test
-        run: |
-          yarn test-ci
+        run: yarn test-ci
 
       - name: Check docs links
```
CollierCZ, line 63: This step requires a build of the docs. Removing the build means that the check isn't done.

CollierCZ (changes requested):

CollierCZ:
> @silvenon any idea why it fails? seems like it can't apply artifact 🤔

I think it's because the artifact is being saved as a top-level folder, so that's how it's extracted. Before, when it was saving two folders, it saved them in separate folders that were unpacked correctly. Now it's unpacking to the root instead of `packages/orbit-components/.out`.

You have to use the path option for the step: https://github.com/actions/download-artifact#download-a-single-artifact

mainframev:
> > @silvenon any idea why it fails? seems like it can't apply artifact 🤔
> 
> I think it's because the artifact is being saved as a top-level folder, so that's how it's extracted. Before, when it was saving two folders, it saved them in separate folders that were unpacked correctly. Now it's unpacking to the root instead of `packages/orbit-components/.out`.
> 
> You have to use the path option for the step: https://github.com/actions/download-artifact#download-a-single-artifact

Ah, yes, look like it works. Did not notice that little thing, thanks a lot :)

silvenon (approved):

PR #2910, opened 2021-04-26
feat(tokens): update deprecated white palette shades

vepor:
It goes against the normal definition of other colors, but we need a way how to define all deprecated old names.


vepor:
Mind the commit scope. Change it when merge & squash.

PR #2911, opened 2021-04-26
docs: implement a more accurate trail via Node API

silvenon:
- add missing meta files
- implement a more accurate trail via Node API

 Orbit.kiwi: https://orbit-docs-docs-trail.surge.sh
 Storybook: https://orbit-docs-trail.surge.sh

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -1,73 +1,159 @@
 const path = require("path");
+const fs = require("fs");
+const yaml = require("js-yaml");
 
 const { createFilePath } = require(`gatsby-source-filesystem`);
-const { doesPageHaveTabs, getDocumentUrlPath, getMetaFileData } = require("./utils/document");
+const {
+  omitNumbers,
+  getDocumentUrl,
+  metaFileDataMap,
+  getParentUrl,
+  getDocumentTrail,
+} = require("./utils/document");
 
-exports.onCreateNode = ({ node, getNode, actions }) => {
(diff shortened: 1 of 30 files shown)


silvenon:
@CollierCZ @mainframev checks have passed, let me know what you think.

docs/gatsby-node.js
```diff
@@ -1,73 +1,159 @@
...
-    node,
-    name: "headerLink",
-    value: getLinkForComponents() || node.frontmatter.headerLink || metaFile.headerLink,
-  });
+      createNodeField({
+        node,
+        name: "trail",
+        value: documentTrail,
+      });
+    } else {
```
mainframev, line 149: i suppose you do not need that `else` block.
You have to create node field on every node, otherwise gatsby will fail, but you can omit the `if` in that case and just have something like that 🤔 
```
createNodeField({
        node,
        name: "trail",
        value: documentTrail || [],
      });
```

or even better to rewrite it in a function, without usage of let, more declarative.

mainframev (commented):
looks good, i will look at it more later, but it seems to work fine, covered with a test, so good job :)

mainframev (approved):

CollierCZ:
I didn't really understand this PR at first. The naming wasn't clear to me. As I mentioned in #2918 , I see some drawbacks to implementing things through Gatsby node. But it might be a factor related to the meta files or something else.

PR #2914, opened 2021-04-28
feat(docs): add design tokens list and inline token

vepor:
Storybook: https://orbit-feat-docs-tokens-list.surge.sh


CollierCZ:
It was having trouble picking up the types for adeira/js, so I switched the estlint import resolver to TypeScript, but maybe you have a better solution @silvenon (or will when you're feeling better) 💪

CollierCZ:
It seems like InlineTokens maybe should also offer the option to copy the name or value. What do you think, @silvenon?

silvenon:
@CollierCZ not sure, currently double clicking is ok for copying the name, I think that's usually what people will want, but I can't really tell, and what is the way to incorporate both functionalities.

CollierCZ:
Double clicking doesn't work when the alternate name is shown and it's not possible to click in the Tooltip. Maybe it'd just be an enhancement for the future, just seemed unfortunate to not be able to copy.

silvenon:
Yes, it's unfortunate because it looks interactive. Considering that this is not a trivial addition, and that this isn't possible in current orbit.kiwi so it's not a downgrade, I'd leave for later. I created a ticket for it. What do you think of the other changes? Do you think we can merge this?

CollierCZ (approved):

PR #2915, opened 2021-04-28
feat(TextLink): add status types

vepor:
Orbit.kiwi: https://orbit-docs-feat-texlink-status.surge.sh
 Storybook: https://orbit-feat-texlink-status.surge.sh

### DIFF
--- a/packages/orbit-components/src/TextLink/TextLink.stories.js
+++ b/packages/orbit-components/src/TextLink/TextLink.stories.js
@@ -5,7 +5,9 @@ import { action } from "@storybook/addon-actions";
 import { text, boolean, select } from "@storybook/addon-knobs";
 
 import { TYPE_OPTIONS, SIZE_OPTIONS } from "./consts";
+import { TYPE_OPTIONS as TEXT_TYPES } from "../Text/consts";
 import * as Icons from "../icons";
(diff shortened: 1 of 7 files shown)


packages/orbit-components/src/TextLink/TextLink.stories.js
```diff
@@ -138,6 +140,30 @@ export const Playground = (): React.Node => {
...
+      elit dui, pellentesque a, faucibus vel, interdum nec, diam.
+    </Text>
+  );
+};
+
+TextLinkInText.story = {
+  name: "TextLink inside paragraph",
+  parameters: {
+    info:
+      "It inherits the visual style of the parent paragraph. Visit Orbit.Kiwi for more detailed guidelines.",
```
silvenon, line 163: Just to double check, link is supposed to be green only when the text is `primary`, right?

silvenon (dismissed):

packages/orbit-components/src/TextLink/TextLink.stories.js
vepor: Yes, because the TextLink is green/product only for the `primary` type.

silvenon (approved):

PR #2916, opened 2021-04-28
feat: introduce Switch component

vepor:

- [x] New Components has both `.flow` and `d.ts` files and are exprorted in `index.js.flow` and `index.d.ts`

 Storybook: https://orbit-feat-switch-component.surge.sh


silvenon (changes requested):
Using a checkbox is a neat CSS trick, but it doesn't follow best practices for a11y, instead I suggest:

- making the entire area clickable instead of just the circle, meaning that the focus ring would go around the whole switch as well (right now the cursor turns into `pointer` in areas which aren't clickable anyway)
- (just continuing from previous note) wrapping everything in `<button type="button" role="switch" aria-checked={checked}>` instead of `StyledSwitchButton`
- expose a `title` prop (or whatever you want to call it) which would be rendered inside the button, but visually hidden, so it would describe what the switch does for screen readers
- wrapping the icon into an element with `aria-hidden="true"`
- making sure that Switch can be toggled with the space bar as well as with mouse click

References:

- [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role)
- [Switch](https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/src/components/switch/switch.tsx) from `@headlessui/react`

mainframev:
> Using a checkbox is a neat CSS trick, but it doesn't follow best practices for a11y, instead I suggest:
> 
> * making the entire area clickable instead of just the circle, meaning that the focus ring would go around the whole switch as well (right now the cursor turns into `pointer` in areas which aren't clickable anyway)
> * (just continuing from previous note) wrapping everything in `<button type="button" role="switch" aria-checked={checked}>` instead of `StyledSwitchButton`
> * expose a `title` prop (or whatever you want to call it) which would be rendered inside the button, but visually hidden, so it would describe what the switch does for screen readers
> * wrapping the icon into an element with `aria-hidden="true"`
> * making sure that Switch can be toggled with the space bar as well as with mouse click
> 
> References:
> 
> * [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role)
> * [Switch](https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/src/components/switch/switch.tsx) from `@headlessui/react`

with current implementation is possible to toggle it with a spacebar ;) Thanks for the review, your points make sense, i will apply them

packages/orbit-components/src/Switch/index.js
```diff
@@ -0,0 +1,136 @@
...
+          onChange={!disabled ? onChange : undefined}
+          onFocus={!disabled ? onFocus : undefined}
+          type="checkbox"
+          data-test={dataTest}
+        />
+        <StyledSwitchButton
+          checked={checked}
+          disabled={disabled}
+          hasCustomIcon={!!icon}
+          role="button"
```
silvenon, line 125: I think you can remove this `role="button"`, considering that there's no event and it's not focusable.

packages/orbit-components/src/Switch/__tests__/index.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it("should be disabled", () => {
+    const onChange = jest.fn();
+    render(<Switch checked onChange={onChange} disabled />);
+
+    userEvent.click(screen.getByRole("button"));
+    expect(onChange).not.toHaveBeenCalled();
+    expect(screen.getByRole("switch")).toHaveAttribute("disabled");
+  });
+});
```
silvenon, line 34: Pressing <kbd>⮐</kbd> no longer works, could add a test for it as well as space bar?

packages/orbit-components/src/Switch/index.js
```diff
@@ -0,0 +1,136 @@
...
+  position: absolute;
+  opacity: 0;
+  width: 100%;
+  height: 100%;
+  top: 0;
+  left: 0;
+  margin: 0;
+  padding: 0;
+
+  focus + ${StyledSwitchButton} {
```
silvenon, line 81:
```suggestion
  &:focus + ${StyledSwitchButton} {
```

silvenon (changes requested):

packages/orbit-components/src/Switch/__tests__/index.test.js
mainframev: fixed

silvenon (dismissed):
Everything working correctly now.

mainframev (dismissed):

PR #2918, opened 2021-04-29
docs: implement side navigation

silvenon:
I postponed bookmarks redesign to another PR, so that part is unfinished, but side navigation both desktop and mobile should work.

This contains work from #2911, so that part doesn't need to be reviewed.
 Storybook: https://orbit-docs-sidenav.surge.sh


docs/src/images/orbit-logo.svg
```diff
@@ -0,0 +1,50 @@
...
+                </g>
+                <g transform="matrix(1.31771,0,0,2.56818,49.8333,-98.1818)">
+                    <path d="M210.175,53.329C209.827,53.33 209.545,53.614 209.544,53.965C209.545,54.316 209.827,54.6 210.175,54.601C210.524,54.6 210.806,54.316 210.806,53.965C210.806,53.614 210.524,53.33 210.175,53.329Z" style="fill:rgb(95,115,140);"/>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
silvenon, line 50: Logos are now updated, they are circular and the glyph is SVG rather than PNG.

docs/src/mdx-components.tsx
```diff
@@ -92,6 +92,7 @@ export const td = ({
   <TableCell
     as="td"
     align={align === "left" || align === "center" || align === "right" ? align : undefined}
+    whiteSpace="normal"
```
silvenon, line 95: The tables where horizontally overflowing in `DocLayout.tsx` because table cells by default don't wrap, not sure why.

docs/src/documentation/01-getting-started/01-for-designers/01-kiwi.mdx
```diff
@@ -1,7 +1,5 @@
 ---
 title: For Kiwi.com employees
-redirect_from:
-  - /getting-started/for-designers/
```
silvenon, line 4: This page appears to be redirect to itself, which was causing problems.

docs/src/components/Tabs.tsx
```diff
@@ -83,12 +72,15 @@ interface Props {
...
+const Tabs = ({ activeTab, tabs }: Props) => {
+  if (tabs.length === 0) return null;
+  return (
+    <Container>
+      {tabs.map(tab => (
+        <Tab key={tab.title} tab={tab} isActive={activeTab === tab.slug} />
+      ))}
+    </Container>
+  );
+};
```
silvenon, line 84: These changes attempt to make the tab shadow trick more stable, the prior implementation was interfering with examples.

silvenon (commented):
A couple of notes to help with review.

silvenon:
It's failing for the same reason as #2911 was previously failing, but that will be fixed once that's merged and I rebase, so it's fine to review.

docs/src/components/DocLayout/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+                        {title}
+                      </Heading>
+                    </Stack>
+                    {description && (
+                      <Box padding={{ left: "XXLarge" }}>
+                        <Text>{description}</Text>
+                      </Box>
+                    )}
+                  </Box>
+                  <StyledMobileOutdent>
```
mainframev, line 104: here is a visual bug, when the icon is not placed on the bottom
Screenshot 2021-04-30 at 13 20 47

mainframev (changes requested):

docs/src/components/DocNavigation/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+function DocNavigationItem({
+  devMode,
+  currentUrl,
+  level,
+  item,
+}: {
+  devMode: boolean;
+  currentUrl: string;
+  level: number;
+  item: NavigationItem;
```
mainframev, line 80: can you create an interface and pass it here? should look better :) 
you can create a separate file inside the folder `DocNavigation` called `records` or `interfaces` and pass all types there, so this file will look more cleaner, just a tip.

docs/src/components/DocNavigation/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+}) {
+  const [expanded, setExpanded] = React.useState(
+    level === 1 &&
+      item.type === "branch" &&
+      item.items.some(it =>
+        it.type === "leaf"
+          ? currentUrl.startsWith(it.url)
+          : it.items.some(i => i.type === "leaf" && currentUrl.startsWith(i.url)),
+      ),
```
mainframev, line 89: this looks a bit complicated for a default state 🤔  
it should be assigned to some variable at least

docs/src/components/DocNavigation/primitives/Collapse.tsx
```diff
@@ -0,0 +1,54 @@
...
+          }
+        `}
+  `};
+`;
+
+interface Props extends React.ComponentProps<typeof OrbitCollapse> {
+  hasCategories: boolean;
+}
+
+export default function Collapse({ expanded, label, hasCategories, children, onClick }: Props) {
```
mainframev, line 42: it's inside folder `primitives`, in the primitives folder, we usually place only SC primitives, without react components, it's not bad, but it's kinda against our habit, when i open the component inside primitives folder i expect to see only SC components there

so the correct way will be to have separate `Collapse` component and import there these primitives.

mainframev:
**_here are few visual things I noticed, that can be quickly fixed:_**
1. Bookmarks modal and Search modal now appear in the middle, it should be placed as before, about 30% from the top. 
2. Bookmarks modal does not have the same border-radius as the Search modal.
3. When I'm clicking on the sidenav item, the whole content is jumping, because of the main content width, it depends on the amount of content inside, so it should have some min-width, I suppose, to prevent that kind of jumping content.
4. Sidenav item should have a smaller arrow and the gap between text and arrow should be smaller. In Figma, it looks a bit different 🤔 
Screenshot 2021-04-30 at 13 43 11

mainframev (commented):
here are few visual things I noticed, that can be quickly fixed:

Bookmarks modal and Search modal now appear in the middle, it should be placed as before, about 30% from the top.
Bookmarks modal does not have the same border-radius as the Search modal.
When I'm clicking on the sidenav item, the whole content is jumping, because of the main content width, it depends on the amount of content inside, so it should have some min-width, I suppose, to prevent that kind of jumping content.
Sidenav item should have a smaller arrow and the gap between text and arrow should be smaller. In Figma, it looks a bit different 🤔
Screenshot 2021-04-30 at 13 43 11

docs/src/components/DocNavigation/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function getItemKey(item: NavigationItem) {
+  if (item.type === "branch") {
+    return `${item.type}-${item.name}`;
+  }
+  return item.url;
+}
+
+function DocNavigationItem({
```
mainframev, line 71: could you also split the `DocNavigationItem` and `DocNavigation` into two different files?

mainframev (approved):

CollierCZ:
Hmm. Now the changes propagated immediately, so it was probably just that I needed to clear the cache before trying out this PR. So you can ignore that part. I still get the list twice.

silvenon (commented):
Done with changes, hopefully fixed all the issues.

docs/src/documentation/03-components/02-structure/card/meta.yml
```diff
@@ -1,3 +1,3 @@
-title: Separator
-description: Separates content into sections.
+title: Card
+description: Separates content into sections. Hi
```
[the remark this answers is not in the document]
CollierCZ: That was from testing the caching issue and seeing if it applied to description as well. Guess I forgot to get rid of it. 🙈

CollierCZ (dismissed):
Other than the failing test, looks good to me!

PR #2922, opened 2021-05-04
feat(BaggageStepper): init

mainframev:
`BaggageStepper` for Baggage. Core functionality is copied from the original `Stepper`, adjusted with custom button via primitive, because of the size/box-shadow

dev: [ORBIT-1760](https://jira.kiwi.com/browse/ORBIT-1760)
design: [ORBIT-1745](https://jira.kiwi.com/browse/ORBIT-1745) 
slack discussion: [here](https://skypicker.slack.com/archives/CAMS40F7B/p1619443241298700?thread_ts=1618819835.251800&cid=CAMS40F7B)

 Storybook: https://orbit-feat-kulibabamstepper.surge.sh

### DIFF
--- a/packages/orbit-components/src/BaggageStepper/Stepper/index.d.ts
+++ b/packages/orbit-components/src/BaggageStepper/Stepper/index.d.ts
@@ -9,7 +9,6 @@ import { SharedProps, Event } from "../index";
 declare module "@kiwicom/orbit-components/lib/StepperStateless";
 
 type InputEvent = Common.Event<React.KeyboardEvent<HTMLInputElement>>;
-
 export interface Props extends SharedProps {
   readonly value?: number | string | (() => string);
   readonly disabledIncrement?: boolean;
@@ -18,6 +17,7 @@ export interface Props extends SharedProps {
   readonly onDecrement?: Event;
   readonly onIncrement?: Event;
   readonly onChange?: InputEvent;
+  readonly selected?: boolean;
 }
 
 declare const StepperStateless: React.FunctionComponent<Props>;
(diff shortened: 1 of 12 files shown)


packages/orbit-components/src/BaggageStepper/Stepper/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as React from "react";
+
+import * as Common from "../../common/common";
+import { SharedProps, Event } from "../index";
+
+declare module "@kiwicom/orbit-components/lib/StepperStateless";
+
+type InputEvent = Common.Event<React.KeyboardEvent<HTMLInputElement>>;
+
+// TODO: fix types, onChange is not being used inside
```
silvenon, line 13: Why is it defined then?

silvenon (changes requested):
This is how Stepper looks disabled:

Screen Shot 2021-05-05 at 11 15 28

and this is how BaggageStepper looks disabled:

Screen Shot 2021-05-05 at 11 16 18

I don't know how disabled state is supposed to look, but maybe connect it somehow with negative margins so that it's one solid piece? Just be careful that buttons need to stay on top `z-index`-wise, so that they are fully clickable.

packages/orbit-components/src/BaggageStepper/Stepper/index.d.ts
mainframev: i think negative margin will not help, because of the shape of the buttons 🤔

mainframev: Will said, i can remove that gray bg of disabled input

PR #2923, opened 2021-05-04
fix(Popover): fix bug with Tooltip inside Popover on mobile

mainframev:
[Request in plz-orbit.
](https://skypicker.slack.com/archives/CAMS40F7B/p1618235742201200 )
[ORBIT-1721](https://jira.kiwi.com/browse/ORBIT-1721)
 Storybook: https://orbit-fix-popover.surge.sh

### DIFF
--- a/packages/orbit-components/src/Popover/components/ContentWrapper.js
+++ b/packages/orbit-components/src/Popover/components/ContentWrapper.js
@@ -228,7 +228,9 @@ const PopoverContentWrapper = ({
     };
   }, []);
 
-  useClickOutside(popover, onClose, !isTablet);
+  useClickOutside(popover, ev => {
+    if (isTablet) onClose(ev);
+  });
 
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/Popover/components/ContentWrapper.js
```diff
@@ -226,7 +228,7 @@ const PopoverContentWrapper = ({
     };
   }, []);
 
-  useClickOutside(popover, onClose);
+  useClickOutside(popover, onClose, !isTablet);
```
silvenon, line 231: Instead of extending the functionality of `useClickOutside` my idea was to simply wrap the handler you're passing.
```suggestion
  useClickOutside(popover, () => { if (isTablet) onClose(); });
```

silvenon (changes requested):

silvenon (approved):
I'll assume that you tried this locally and it works.

PR #2928, opened 2021-05-10
docs: add page title & description

silvenon:
Storybook: https://orbit-docs-page-meta.surge.sh

### DIFF
--- a/docs/src/components/DocLayout/index.tsx
+++ b/docs/src/components/DocLayout/index.tsx
@@ -11,6 +11,7 @@ import {
 } from "@kiwicom/orbit-components";
 import { MDXProvider } from "@mdx-js/react";
 import { WindowLocation } from "@reach/router";
+import { Helmet } from "react-helmet";
 
 import defaultTheme from "../../theme";
 import * as components from "../../mdx-components";
@@ -39,6 +40,7 @@ import StyledDocNavigationWidth from "./primitives/StyledDocNavigationWidth";
 import StyledDocNavigationWrapper from "./primitives/StyledDocNavigationWrapper";
 import StyledTocWrapper from "./primitives/StyledTocWrapper";
 import StyledProse from "./primitives/StyledProse";
+import { getDocumentPageTitle } from "../../utils/document";
 
 interface Props {
   children: React.ReactNode;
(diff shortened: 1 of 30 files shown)


docs/src/components/Layout.tsx
```diff
@@ -26,18 +26,25 @@ const StyledMain = styled.main`
...
 
-export default function Layout({ children, location }: Props) {
+export default function Layout({ children, location, title, description }: Props) {
   return (
-    <ThemeProvider theme={defaultTheme}>
-      <DevModeProvider>
+    <>
+      <Helmet>
+        <meta charSet="utf-8" />
+        <title>{title}</title>
```
CollierCZ, line 38: Seems like the site title should go here, rather than be part of the titles of individual documents. Can make an exception for home page.

docs/src/pages/component-status.mdx
```diff
@@ -14,7 +14,7 @@ export default ({ children }) => {
   return (
     <Layout
       description="A detailed breakdown of our components and status of their implementation."
-      title="Component status 🚀"
+      title="Component status 🚀 | Orbit"
```
CollierCZ, line 17: If the site title is in Layout, this change is unnecessary. This seems like a less programmatic solution than adding in an SEO title field.

docs/src/components/DocLayout/index.tsx
```diff
@@ -71,120 +73,131 @@ export default function DocLayout({
...
-                        )}
-                      </Box>
-                    )}
-                    <Grid columns="1fr" tablet={{ columns: `${tocHasItems ? "80% 20%" : "100%"}` }}>
-                      {tocHasItems && (
-                        <StyledTocWrapper>
-                          <Hide on={["smallMobile", "mediumMobile", "largeMobile"]}>{Toc}</Hide>
-                        </StyledTocWrapper>
+    <>
+      <Helmet>
```
CollierCZ, line 77: For proper SEO and accessibility, we'll want more items in the helmet (like language for accessibility). It'd make sense to me to make it its own component (then you don't have to repeat the similar code as in Layout, and won't leave things out in one place but not the other (like the meta charset, unless there's a reason for that difference?).

CollierCZ (changes requested):
GitHub makes it seem like the part with the table of contents was changed, but I don't see a change there. Was something changed there?

docs/src/components/DocLayout/index.tsx
```diff
@@ -71,120 +73,131 @@ export default function DocLayout({
...
+        {title && (
+          <title>{getDocumentPageTitle(title, trail ? trail.map(t => t.name) : [])} | Orbit</title>
+        )}
+        {description && <meta name="description" content={description} />}
```
CollierCZ, line 81: I think it'd be good to have a fallback description for Orbit in general.

CollierCZ: Good point. But we also shouldn't have pages without descriptions, especially not important pages like the home page. I think that was the problem I saw. The solution is to add descriptions in more places.

docs/src/components/Head/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface Props {
+  title?: string;
+  hasSiteName?: boolean;
+  description?: string;
+  path: string;
+}
+
+export default function Head({ title, hasSiteName, description, path }: Props) {
+  const pageTitle = hasSiteName ? `${title} | Orbit` : title
```
CollierCZ, line 12: I think this is the opposite logic that we want. This is set to `isHome` for the index page, which returns true. Then it here adds ` | Orbit`. From the name `hasSiteName`, I would think it looks if the title already has the name in it. Then I would not add the name to it.

If that's true, then I'd change the logic in DocLayout. If I'm wrong, then the logic of `isHome` should change instead. 

Right now it's adding ` | Orbit` everywhere, which I think is not desired behavior.

docs/src/components/DocLayout/index.tsx
```diff
@@ -71,120 +74,131 @@ export default function DocLayout({
...
-                    )}
-                    <Grid columns="1fr" tablet={{ columns: `${tocHasItems ? "80% 20%" : "100%"}` }}>
-                      {tocHasItems && (
-                        <StyledTocWrapper>
-                          <Hide on={["smallMobile", "mediumMobile", "largeMobile"]}>{Toc}</Hide>
-                        </StyledTocWrapper>
+    <>
+      <Head
+        title={title ? getDocumentPageTitle(title, trail ? trail.map(t => t.name) : []) : "Orbit"}
+        hasSiteName={Boolean(title)}
```
CollierCZ, line 80: I think this should be reversed. See comment on Head.

CollierCZ (changes requested):

docs/src/components/DocLayout/index.tsx
silvenon: Should be fine, because we don't want `Orbit | Orbit` when the title is missing.

docs/src/components/Head/index.tsx
silvenon: True, I flipped mapping of `isHome` and `hasSiteName` — `hasSiteName={!isHome}`.

CollierCZ (approved):

PR #2930, opened 2021-05-12
docs: add 404 page

silvenon:
I'll add the image later, positioning images on responsive is time-consuming.

 Storybook: https://orbit-docs-404.surge.sh


docs/src/pages/404.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Stack flex direction="row">
+        <Stack spacing="XLarge">
+          <Stack spacing="small">
+            <Heading as="h1" type="display">
+              Page wasn&apos;t found. It probably doesn&apos;t exist.
+            </Heading>
+            <p>Try our search to find the content you&apos;re looking for.</p>
+          </Stack>
+          <Stack flex direction="row" spacing="XSmall">
+            <Button circled type="primary" iconRight={<TripIcon />}>
```
mainframev, line 23: i think you should use button based on gatsby link component here

docs/src/components/DocLayout/index.tsx
```diff
@@ -102,97 +104,95 @@ export default function DocLayout({
...
-                          }
-                          elevation={noElevation ? undefined : "raised"}
+                      </Box>
+                      <StyledMobileOutdent>
+                        {(tabs || headerLink) && (
+                          <Box
+                            display="flex"
+                            align="end"
+                            justify={tabs && tabs.length > 0 ? "between" : "end"}
+                            tablet={{ maxWidth: "80%" }}
```
mainframev, line 139: this component causes bug, with github icon, when layout changes, it's remains on 80%, when it should be on 100%. (When you click on guidelines)
Screenshot 2021-05-13 at 11 52 04

mainframev (changes requested):
The `Return to the home` button does not work. I wrote about it below

mainframev (dismissed):

PR #2934, opened 2021-05-13
fix(popover): close on overlay

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1620918962337700)
I've created the story, [that you can check if it works](https://orbit-fix-popover-split-renders.surge.sh/?path=/story/popover--with-tooltip)
 Storybook: https://orbit-fix-popover-split-renders.surge.sh

### DIFF
--- a/packages/orbit-components/src/Popover/components/ContentWrapper.js
+++ b/packages/orbit-components/src/Popover/components/ContentWrapper.js
@@ -183,6 +183,7 @@ const PopoverContentWrapper = ({
   const { isInsideModal } = React.useContext(ModalContext);
   const { isTablet } = useMediaQuery();
   const popover: {| current: React.ElementRef<*> |} = React.useRef(null);
+  const overlay: {| current: React.ElementRef<*> |} = React.useRef(null);
   const content: {| current: React.ElementRef<*> |} = React.useRef(null);
   const intervalRef = React.useRef(null);
   const position = calculatePopoverPosition(preferredPosition, preferredAlign);
@@ -216,6 +217,15 @@ const PopoverContentWrapper = ({
     [actions],
   );
 
+  const handleClick = React.useCallback(
+    ev => {
+      if (onClose && overlay.current.contains(ev.target)) {
+        onClose(ev);
+      }
+    },
+    [onClose],
+  );
+
   React.useEffect(() => {
     const timer = setTimeout(() => {
       if (popover.current) {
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/Popover/components/ContentWrapper.js
```diff
@@ -238,7 +250,7 @@ const PopoverContentWrapper = ({
 
   return (
     <>
-      <StyledOverlay shown={shown} isInsideModal={isInsideModal} />
+      <StyledOverlay shown={shown} isInsideModal={isInsideModal} ref={overlay} />
```
silvenon, line 253: Use `onMouseDown` on `StyledOverlay` instead of attaching and removing the event listener, that way we can test it. ✨

silvenon (changes requested):
The fix works, but considering that we broke such a critical functionality, we need to proceed with testing fixes. Here's how we can test this one specifically:

```jsx
  it("with tooltip", async () => {
    render(
      <Popover dataTest="popover" content={<Tooltip content="Content">Tooltip</Tooltip>}>
        <Button>Open popover</Button>
      </Popover>,
    );
    userEvent.click(screen.getByRole("button"));
    const overlay = (await screen.findByTestId("popover")).previousElementSibling;
    expect(overlay).toBeInTheDocument();
    if (overlay) userEvent.click(overlay);
    await waitForElementToBeRemoved(screen.queryByTestId("popover"));
  });
```

packages/orbit-components/src/Popover/components/ContentWrapper.js
mainframev: thanks a lot for the test :) <3

packages/orbit-components/src/Popover/components/ContentWrapper.js
```diff
@@ -216,6 +217,15 @@ const PopoverContentWrapper = ({
     [actions],
   );
 
+  const handleClick = React.useCallback(
+    ev => {
+      if (onClose && overlay.current.contains(ev.target)) {
```
silvenon, line 222: I forgot to add that after attaching `onMouseDown` to `StyledOverlay`, this check is now unnecessary because clicks outside of that element won't trigger this handler.

silvenon (approved):
Great! One minor comment with a superfluous check, but it's totally fine either way.

PR #2937, opened 2021-05-14
feat: introduce BadgeList component

vepor:

- [x] New Components has both `.flow` and `d.ts` files and are exprorted in `index.js.flow` and `index.d.ts`

 Storybook: https://orbit-feat-badgelist.surge.sh


packages/orbit-components/src/BadgeList/index.js
```diff
@@ -0,0 +1,19 @@
...
+
+const StyledBadgeList = styled.ul`
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+`;
+
+const BadgeList = ({ children, dataTest }: Props): React.Node => {
+  return <StyledBadgeList data-test={dataTest}>{children}</StyledBadgeList>;
```
mainframev, line 15: wdy about applying `role="presentation"` on the list?

packages/orbit-components/src/BadgeList/__tests__/index.test.js
```diff
@@ -0,0 +1,24 @@
...
+    render(
+      <BadgeList dataTest={dataTest}>
+        <BadgeListItem icon={<Airplane />} dataTest={childrenDataTest}>
+          {childrenText}
+        </BadgeListItem>
+      </BadgeList>,
+    );
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
+    expect(screen.getByTestId(childrenDataTest)).toBeInTheDocument();
+    expect(screen.getByTestId(childrenDataTest)).toHaveTextContent(childrenText);
```
mainframev, line 22:
```suggestion
    expect(screen.getByText(childrenText)).toBeInTheDocument();
```

packages/orbit-components/src/BadgeList/BadgeListItem/index.js
```diff
@@ -0,0 +1,110 @@
...
+  };
+  return tokens[$type];
+};
+
+const getIconColor = type => {
+  if (type === TYPE_OPTIONS.NEUTRAL) return ICON_COLORS.SECONDARY;
+  return type;
+};
+
+const StyledBadgeListItem = styled.li`
```
mainframev, line 30: hm, probably you do not need here custom styled component, it can be replaced with: 
```
    <Stack as="li" spaceAfter="smallest" align="center" spacing="none" data-test={dataTest}>
       ... content
    </Stack>
```

mainframev (changes requested):

packages/orbit-components/src/BadgeList/index.js
vepor: Doesn't using `role="presentation"` hide the element and children of it from readers?

packages/orbit-components/src/BadgeList/BadgeListItem/index.js
vepor: If we gonna use Stack for this use case, we need to also reset/normalize the styles for the `li` element, because it will have the default user agent styles. That's the reason why I've chosen this approach. Don't want to add yet another complexity to Stack 🙂

packages/orbit-components/src/BadgeList/index.js
mainframev: i think it should affect `ul` and `li`, but not the content of them. `[role="presentation"] negates the implicit 'list'
and 'listitem' role semantics but does not affect the contents.`. But i just asked, I'm not sure about it

silvenon (approved):

packages/orbit-components/src/BadgeList/BadgeList.stories.js
```diff
@@ -0,0 +1,104 @@
+// @flow
+import * as React from "react";
+import { action } from "@storybook/addon-actions";
+import { text, select } from "@storybook/addon-knobs";
+
+import * as Icons from "../icons";
```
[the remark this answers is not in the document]
vepor: Generally that's true, but for stories should be okay, because this module is not being imported (outside storybook). In the static bundle for Storybook they are present there anyway already 🙂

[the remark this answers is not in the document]
silvenon: Oh, right.

silvenon (approved):

PR #2950, opened 2021-05-19
ci: improve speed of size GitHub workflow

silvenon:
This builds strictly what is necessary for measuring size, seems to reduce the duration by about 4m.

 Storybook: https://orbit-ci-faster-compressed-size.surge.sh


mainframev (approved):
great! :)

PR #2954, opened 2021-05-20
feat(Modal): add disableAnimation prop

silvenon:
This prop is an escape hatch because Modal animation is currently implemented by transitioning the `top` CSS property, which results in a bad Cumulative Layout Shift score.

I wasn't sure whether to document it, as it's an escape hatch which we'll probably get rid of.

- [the original suggestion](https://skypicker.slack.com/archives/C9NJQS970/p1619528876094700?thread_ts=1619527814.093300&cid=C9NJQS970)

 Storybook: https://orbit-feat-modal-disable-animation.surge.sh

### DIFF
--- a/packages/orbit-components/src/Modal/index.js.flow
+++ b/packages/orbit-components/src/Modal/index.js.flow
@@ -37,6 +37,7 @@ export type Props = {|
   +isMobileFullPage?: boolean,
   +preventOverlayClose?: boolean,
   +hasCloseButton?: boolean,
+  +disableAnimation?: boolean,
   ...Globals,
 |};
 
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/Modal/index.js.flow
```diff
@@ -37,6 +37,7 @@ export type Props = {|
   +isMobileFullPage?: boolean,
   +preventOverlayClose?: boolean,
   +hasCloseButton?: boolean,
+  +disableAnimation?: boolean,
```
vepor, line 40: Don't forget to add it also into `index.d.ts` file 🙏

vepor (changes requested):

packages/orbit-components/src/Modal/index.js.flow
silvenon: It exists already. 🤔

PR #2955, opened 2021-05-20
Improve pipeline speed

silvenon:
Cutting down the longest workflow from ~20m to ~10m, not counting Cloudflare Pages for which there's no easy fix that doesn't include heavier rethinking, so that's still a bottleneck but at least we'll get feedback from other actions sooner.

- optimizing building orbit.kiwi by building only necessary parts of design tokens and components and turning on experimental [conditional page builds](https://www.gatsbyjs.com/docs/reference/release-notes/v3.0/#incremental-builds-in-oss) (will be the default in Gatsby v3)
- splitting deployment of Storybook into a separate workflow
- fully building design tokens and components in size workflow (even though it's unnecessary for measuring size), as a sanity check whether scripts are working properly
- with building out of the way tests can start earlier
- using Node v15 because it has npm v7, which has npxm, so that we can run `npx lerna bootstrap` and that way avoid running `postinstall` twice

I also made sure that scripts fail on CI if expected environment variables and directories are missing (e.g. `check-links` needs `GH_TOKEN` and `docs/public` to be available).

Storybook: https://orbit-ci-faster-build-test.surge.sh


CollierCZ (changes requested):
I like the idea of splitting Storybook with its deploy into a separate workflow (I'm assuming it doesn't require a build of the components). 

Not building the docs pages means that the job to check links in the docs won't work (it depends on built files being present). So I like the idea of avoiding it, but I'm not sure it'll work ideally at the moment.

As far as I know, Cloudflare Pages doesn't use caching or anything for Gatsby, so the builds there will always take longer and I'm not sure it'll save time overall. And not having logs from the builds isn't great, but I hope that'll be solved soon.

.github/workflows/storybook.yml
```diff
@@ -12,35 +12,32 @@ env: # Set environment variables for every job and step in this workflow
...
-  build_test:
+  build:
     runs-on: ubuntu-latest
-    name: Build and test
+    name: Build
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "15"
```
mainframev, line 26: upgrade also `.node-version` file

silvenon:
Ok, now I build and test documentation in a separate job in `test.yml`. I updated the PR description to reflect the current state. Now the savings are not spectacular, but are still significant.

mainframev (dismissed):

CollierCZ (changes requested):
The 12 minutes is without any caching for the docs site (because it's a new job and so no cache key existed), so my guess is it'll be less on future builds.

The build logs say they don't include a GitHub access key. That means not all links will be checked because some should come from that component.

The Cloudflare Pages job is still taking close to 20 minutes, which is going to be the bottleneck going forward.

CollierCZ:
Also, until we upgrade to Gatsby 3 (with webpack 5), I realized we're not taking advantage of the caching we have. i think we should be using incremental builds with the `GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES` environment variable set to `true`.

.github/workflows/test.yml
```diff
@@ -91,8 +91,9 @@ jobs:
           FIGMA_TOKEN: ${{ secrets.FIGMA_TOKEN }}
           STREAMLINE_SECRET: ${{ secrets.STREAMLINE_SECRET }}
           STREAMLINE_FAMILIES: ${{ secrets.STREAMLINE_FAMILIES }}
-          GH_TOKEN: ${{ secrets.GH_TOKEN }}
```
CollierCZ: I think both of the jobs need the token. If the build doesn't have the token, it won't build the contributors and the links to their profile pages won't get checked later.

silvenon:
I made sure that scripts fail in CI if expected environment variables are missing, and checking links should fail if `docs/public` is missing. (I forgot to also make it fail if `GH_TOKEN` is missing, I'll do that now.)

CollierCZ (dismissed):
The incremental builds don't seem to be working. It might be because they want us to use Gatsby Cloud for building/deploys. Or it might be that the hashes of the files are based on absolute paths and those might change between jobs. 🤷‍♂️ Either way, it still is faster, but it's clearing the cache between even content changes.

The GitHub contributors isn't working: https://cca79f77.orbit.pages.dev/getting-started/support/team/#react-components But maybe that's an issue with that component and so outside the scope of this one.

The changes look good in general.

silvenon:
> The incremental builds don't seem to be working.

Too bad 😕  hopefully we'll be able to switch to Gatsby v3 soon.

> The GitHub contributors isn't working

Yup, it doesn't work on master either, so we need to fix that in a separate PR.

silvenon:
Nothing new here, just squashing some commits together to tidy up the history. Will merge after passing CI.

PR #2961, opened 2021-05-25
docs(sandbox): fix components with Portals inside sandbox

mainframev:
We faced the problem inside our examples which are rendered inside a sandbox and rendered via `Portal`. Portals ignore the iframe document and render it inside the parent, normal document, but we need to show it inside Iframe. 

My first intention was to pass down the inner document to the `Portal`, so it will take the correct document, created the context `PortalProvider`, and tried to use the consumer inside `Portal` or components. Did not work. Seems like `react-live`   does not allow it, because other components did not have such an issue, only which are rendered as examples. 

so here is another solution, to add a prop to all components with `Portal`, which will turn off rendering in Portals. 
 Storybook: https://orbit-docs-portals-fix.surge.sh

### DIFF
--- a/packages/orbit-components/src/Dialog/README.md
+++ b/packages/orbit-components/src/Dialog/README.md
@@ -22,6 +22,7 @@ Table below contains all types of the props available in Dialog component.
 | Name              | Type                    | Description                                                                                          |
 | :---------------- | :---------------------- | :--------------------------------------------------------------------------------------------------- |
 | dataTest          | `string`                | Optional prop for testing purposes.                                                                  |
+| insidePortal      | `boolean`               | Optional prop, if set to `false` component is not rendered inside Portal, default is `true`          |
 | description       | `React.Node`            | Optional description of the main action that Dialog performs.                                        |
 | illustration      | `React.Node`            | Optional illustration of the Dialog.                                                                 |
 | **primaryAction** | `React.Node`            | Primary and required action that user can do with the Dialog.                                        |
--- a/packages/orbit-components/src/Dialog/README.md
+++ b/packages/orbit-components/src/Dialog/README.md
@@ -19,13 +19,13 @@ After adding import into your project you can use it simply like:
 
 Table below contains all types of the props available in Dialog component.
 
-| Name              | Type                    | Description                                                                                          |
-| :---------------- | :---------------------- | :--------------------------------------------------------------------------------------------------- |
-| dataTest          | `string`                | Optional prop for testing purposes.                                                                  |
-| insidePortal      | `boolean`               | Optional prop, if set to `false` component is not rendered inside Portal, default is `true`          |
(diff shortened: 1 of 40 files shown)


packages/orbit-components/src/Dialog/index.js
```diff
@@ -154,7 +155,7 @@ const Dialog = ({
 
   const dialogID = React.useMemo(() => randomID("dialog"), []);
 
-  return (
+  return insidePortal ? (
```
silvenon, line 158: So "insidePortal" is short for "render this inside portal"? Initially I thought that it meant "it's already inside portal, so don't render another one". Instead I recommend alternatives like `renderInPortal`, `withPortal` or `hasPortal`.

packages/orbit-components/src/Dialog/README.md
```diff
@@ -22,6 +22,7 @@ Table below contains all types of the props available in Dialog component.
 | Name              | Type                    | Description                                                                                          |
 | :---------------- | :---------------------- | :--------------------------------------------------------------------------------------------------- |
 | dataTest          | `string`                | Optional prop for testing purposes.                                                                  |
+| insidePortal      | `boolean`               | Optional prop, if set to `false` component is not rendered inside Portal, default is `true`          |
```
silvenon, line 25: I recommend telling people why they would want to turn this off, instead of explaining what it does. Let me know what you think.

```suggestion
| insidePortal      | `boolean`               | Optional prop, set it to `false` if you're rendering Dialog inside a custom portal, defaults to `true`          |
```

Same in readmes of other components, but without "optional prop" and "defaults to" parts.

packages/orbit-components/src/Dialog/index.js
```diff
@@ -184,6 +185,34 @@ const Dialog = ({
...
+            direction="column-reverse"
+            spacing="XSmall"
+            largeMobile={{ direction: "row", justify: "end" }}
+          >
+            {secondaryAction && <StyledAction>{secondaryAction}</StyledAction>}
+            <StyledAction>{primaryAction}</StyledAction>
+          </Stack>
+        </StyledDialogContent>
+      </StyledDialogCenterWrapper>
+    </StyledDialog>
```
silvenon, line 215: If all of this added code is the same as in the portal, don't copy this here. You can do something like this:

```jsx
const dialog = (
  // JSX code
)

return insidePortal ? (
  <Portal renderInto="modals">
    {dialog}
  </Portal>
) : dialog;
```

Same suggestion for other components.

silvenon (changes requested):

silvenon (approved):
Looks good! If there's a way to be able to scroll up completely, that would be perfect, I can only scroll up to this point:

Screen Shot 2021-05-25 at 18 35 35

Scrolling downwards works all the way, so it's only upwards that doesn't.

PR #2962, opened 2021-05-25
docs/design review fixes

silvenon:
- fix: increase tabs padding and don't rely on JS
- fix: add type definition for react-scrollspy
- chore: add missing Figma images
- fix: apply fixes after [design review](https://docs.google.com/document/d/10x3Bc77yHw4xXgHcVOTgBMFZ273t1goUoDHHsMQv2OY)
- fix: fix search results case ([ORBIT-1789](https://jira.kiwi.com/browse/ORBIT-1789))

 Storybook: https://orbit-docs-design-review-fixes.surge.sh


docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
```diff
@@ -36,7 +36,9 @@ check out our [interactive guide on action components](/guides/using-button-type
 
 > "If everything looks important, then nothing is important."
 >
-> ---Patrick Lencioni
+> <p>
+>   <span className="author">—Patrick Lencioni</span>
+> </p>
```
CollierCZ, line 41: I really wish there was a more semantic way to do this. It seems to be going away from the benefits of using MDX to get a small decorative benefit.

In terms of semantics, it would have to be a figcaption inside a figure wrapping the blockquote, which also seems suboptimal. Would it be possible to do it without the `p` tags at least?

docs/src/components/BrandedTile.tsx
```diff
@@ -21,16 +21,21 @@ interface Props {
   children: React.ReactNode;
 }
 
-const StyledWrapper = styled.div<{ primary: string }>`
+const StyledWrapper = styled.a<{ primary: string }>`
```
CollierCZ, line 24: This change affects other tiles that aren't links, like those on the [Roadmap](https://8af2731c.orbit.pages.dev/roadmap/). Could the element be decided based on props or something else?

CollierCZ: Also, this leads to invalid DOM nesting with `<a>` elements inside other `<a>` elements. So if this is an `<a>` element, the link in the bottom right of the tile should change.

docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
```diff
@@ -36,7 +36,9 @@ check out our [interactive guide on action components](/guides/using-button-type
 
 > "If everything looks important, then nothing is important."
```
CollierCZ, line 37: If we're going to the trouble of styling this completely as a quote, it doesn't require the quotation marks as much.
```suggestion
> If everything looks important, then nothing is important.
```

docs/src/components/BlockQuote/index.tsx
```diff
@@ -0,0 +1 @@
+export { default } from "./BlockQuote";
```
CollierCZ, line 1: Just for my education, why is this separate file with the export necessary?

docs/src/components/HeaderLink.tsx
```diff
@@ -39,15 +44,11 @@ const FancyLink = ({ href }: Props) => {
     }
   };
   return (
-    <ButtonLink
-      external
-      href={href}
-      iconLeft={getUsedIcon()}
-      iconRight={<NewWindow ariaLabel="Opens in new window" />}
```
CollierCZ, line 46: Without this, the accessibility of the link decreases -- there's no warning it will open a new window (see [WCAG recommendations](https://www.w3.org/TR/WCAG20-TECHS/G201.html)). That was the point of the icon, not just that it's an external site. I made a note in the design review doc about it too.

docs/src/components/ReactExample/index.tsx
```diff
@@ -16,10 +16,9 @@ interface Props {
 }
 
 const StyledExampleWrapper = styled.div`
-  ${({ theme }) => `
-    box-shadow: ${theme.orbit.boxShadowRaised};
-    border-radius: ${theme.orbit.borderRadiusLarge};
-  `}
+  box-shadow: rgb(37 42 49 / 8%) 0px 4px 8px 0px, rgb(37 42 49 / 16%) 0px 8px 24px 0px;
```
CollierCZ, line 19: This is `boxShadowRaisedSubtle`, so I think it'd make sense to use that token

docs/src/components/ReactExample/index.tsx
```diff
@@ -16,10 +16,9 @@ interface Props {
...
+  border-radius: 12px;
```
CollierCZ, line 20: This is `borderRadiusBadge`, but I'm not sure it's as important to use that token.

docs/src/components/Tabs.tsx
```diff
@@ -27,24 +27,24 @@ const StyledTabLink = styled(Link)`
 
 const StyledTab = styled.div`
   ${({ theme }) => css`
-    border-top-left-radius: ${theme.orbit.borderRadiusLarge};
-    border-top-right-radius: ${theme.orbit.borderRadiusLarge};
+    border-top-left-radius: 12px;
+    border-top-right-radius: 12px;
```
CollierCZ, line 31: As above, we have a token for this, if you'd like to use it.

docs/src/components/Tile.tsx
```diff
@@ -34,6 +34,36 @@ function TileTitle({ children }: { children: React.ReactNode }) {
...
+const StyledWrapper = styled.a<StyledContainerProps>`
+  ${({ theme, fullWidth, hasContent }) => `
+    display: block;
+    padding: 2rem;
+    border-radius: 1rem;
+    background: ${theme.orbit.paletteWhite};
+    box-shadow: ${theme.orbit.boxShadowRaisedSubtle};
+    transition: box-shadow ${theme.orbit.durationFast};
+    display: flex;
+    width: ${fullWidth && "100%"};
```
CollierCZ, line 50: Not sure it matters really, but wouldn't it make sense as this?
```suggestion
    ${fullWidth && "width: 100%;"}
```

CollierCZ (changes requested):
Seems pretty good overall. I don't like the emphasis of the Table of Contents headings, but that's a question for design. The implementation seems clear.

One small thing that I couldn't comment on inline (since you didn't change it):

```
<Text as="p" size="large">
```
I think this (in line 180 of `docs/src/components/Search/SearchModal.tsx`) should be downsized based on the review doc. In the next line, I think the text should be `We found <b>{results.length} results</b> matching your search`. Or I'm not sure why there's "generic" there, but the second "results" should definitely be "search".

docs/src/components/Tile.tsx
mainframev: it does not make any difference, if it's evaluated to `false`, the property will not be rendered.

CollierCZ: I was thinking in terms of reading the code rather than the result, but it probably makes no difference.

docs/src/components/BlockQuote/index.tsx
silvenon: It's not, I assumed it's how we're writing other components, but I misunderstood our convention, I should've put all code inside of `BlockQuote/index.tsx`.

docs/src/components/ReactExample/index.tsx
silvenon: I was also thinking that, but I'd rather leave it as a coincidence, considering that it's not a badge so it might be weird to link them like that.

docs/src/components/Tile.tsx
silvenon: @mainframev I recently checked, it _does_ produce the CSS declaration, but it's invalid CSS, so browsers ignore it. Ultimately it does produce the desired result, but in an uncomfortable way, so I don't think we should keep doing this.

Screen Shot 2021-05-27 at 16 14 15

silvenon: I'll assume that it works the way you describe when using `css`, then.

mainframev (dismissed):

docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
silvenon: I overengineered this like a pro 😎  now the blockquote component detects whether the quote ends with em dash, and considers that to be the attribution, and applies the necessary styles, so now you can write in MDX normally without any JSX.

docs/src/components/BrandedTile.tsx
silvenon: Fixed both problems. The second one by turning nested links and buttons into click-through spans and divs.

docs/src/components/HeaderLink.tsx
silvenon: I agree, so I brought back the icon.

CollierCZ (approved):
LGTM!

PR #2963, opened 2021-05-25
fix(docs): set image widths and allow side by side guidelines

CollierCZ:
Storybook: https://orbit-docs-image-widths.surge.sh


silvenon (commented):
This is still a bit crowded, so maybe snap to this view in the next breakpoint. But then again, the ToC is too narrow in that view, too, so maybe we should snap to side ToC later as well, making a little more room, feel free to do that as well as part of this PR.

Screen Shot 2021-05-26 at 11 42 26

CollierCZ:
I went with hiding the navigation sooner because it seemed to place more emphasis on the content, which I like, and looked smooth. I'd be happy to change it the other way if you'd prefer.

It's not great that we have the logic for when all those different parts have to be hidden or not in so many places. Hope I got them all.

silvenon (commented):
Yeah, organizing responsive design is quite tricky, when we start cleaning up we should see how it's best to do it because we don't want premature abstractions. But you got almost all of them 🏆 just one margin which should be set later, once the side navigation appears:

Screen Shot 2021-05-26 at 16 26 20

CollierCZ:
Didn't expect that one. Should be OK now.

silvenon (approved):
All good now!

PR #2964, opened 2021-05-26
feat(HorizontalScroll): add scroll-snap-type

mainframev:
Added `scroll-snap-type` to `HorizontalScroll`. Two props, to turn it on and the type for the property. 
 Storybook: https://orbit-feat-horizontalscroll-snap.surge.sh

### DIFF
--- a/packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.js
+++ b/packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from "react";
+import { select, number } from "@storybook/addon-knobs";
 
 import Box from "../Box";
 import Text from "../Text";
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/HorizontalScroll/README.md
```diff
@@ -19,12 +19,24 @@ After adding import into your project you can use it simply like:
...
-| children  | `React.ReactNode`     | :heavy_check_mark: |         | content of HorizontalScroll           |
+| Name       | Type                        | Required           | Default       | Description                               |
+| ---------- | --------------------------- | ------------------ | ------------- | ----------------------------------------- |
+| minHeight  | `number`                    |                    |               | set minimal height                        |
+| dataTest   | `string`                    |                    |               | prop for testing purposes                 |
+| spacing    | [`Spacing`](#Spacing)       |                    | "small"       | the spacing between children elements     |
+| children   | `React.ReactNode`           | :heavy_check_mark: |               | content of HorizontalScroll               |
+| useSnap    | `boolean`                   |                    |               | enable `scroll-snap-type` property        |
+| scrollSnap | [`ScrollSnap`](#ScrollSnap) |                    | "x mandatory" | set value for `scroll-snap-type` property |
```
silvenon, line 29: I think it's unnecessary to have two props for scroll snap, I'm wondering what could be a reasonable alternative 🤔  How about allowing `scrollSnap` to also be a boolean? So that if it's `true`, then it's `x mandatory`, and if it's `false` (default), then it's `none`. And besides boolean it can also take all those other values that are specified below.

What do you think?

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.js
```diff
@@ -24,10 +25,52 @@ export const Default = (): React.Node => {
           height="full"
         >
           <Text size="large" weight="bold" as="div">
-            <div style={{ height: "50px", display: "flex", alignItems: "center" }}>{key}</div>
+            <div
+              style={{
+                height: "50px",
+                display: "flex",
+                alignItems: "center",
+                scrollSnapAlign: "center",
```
silvenon, line 33: If I'm not mistaken, `scrollSnapAlign` doesn't do anything for this story because scroll snapping is not enabled, no?

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.js
```diff
@@ -24,10 +25,52 @@ export const Default = (): React.Node => {
...
+            </div>
           </Text>
         </Box>
       ))}
     </HorizontalScroll>
   );
 };
+
+export const withScrollSnap = (): React.Node => {
+  const scrollSnapAlign = text("scrollSnapAlign", "start");
```
silvenon, line 46:
```suggestion
  const scrollSnapAlign = select("scrollSnapAlign", ["none", "start", "end", "center"], "start");
```

silvenon (changes requested):

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.js
mainframev: yes, forgot to remove

silvenon (dismissed):
After addressing (or not addressing) these minor comments you can merge.

PR #2965, opened 2021-05-27
chore: update husky to v6

mainframev:
Husky v6 has some breaking changes compared to our version. 
 Storybook: https://orbit-chore-update-husky.surge.sh


.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install commitlint --edit ""
```
silvenon, line 4: Not sure what's the new version of `HUSKY_GIT_PARAMS`, but based on [husky's own `commit-msg`](https://github.com/typicode/husky/blob/a6fa2b1579e262478558d194a226b2dfb6c8e4b7/.husky/commit-msg) it seems to be `$1`.
```suggestion
npx --no-install commitlint --edit "$1"
```

silvenon (changes requested):

PR #2970, opened 2021-06-01
refactor(contributors): changes to contributors component

mainframev:
refactored the contributor components, so the error message is passed from the `gatsby-node` to the client. JSON with hardcoded users is also parsed now inside `gatsby-node`. 

 Storybook: https://orbit-fix-contributors.surge.sh


docs/src/documentation/01-getting-started/04-support/03-team.mdx
```diff
@@ -5,23 +5,10 @@ redirect_from:
...
-
-From small reviews or additions to contributions of whole components.
-We really appreciate all your efforts when helping us to make Orbit better
-👏
-
-### React components
-
 <GithubContributors />
-
-Thanks to all these folks for caring, your contribution helps to all Orbit consumers.
```
silvenon, line 27: I would like to avoid moving MDX to JSX as much as possible. Is there a way to get what you wanted, but also keep this content in MDX?

silvenon (changes requested):

silvenon:
Btw, now displaying all contributors appears to work! Does that mean you found the problem? 🤔  I'll assume that the error message works because I expected to be able to see it in the deployment.

docs/src/documentation/01-getting-started/04-support/03-team.mdx
mainframev: I thought for this particular case there is no need to have it inside `MDX`, it was in `MDX` because it had two separate components for `OrbitTeam` and `Contributors`. Now it's one component because it's fetching all data from graphql, not like before from json and github.

docs/plugins/contributors/gatsby-node.js
```diff
@@ -63,6 +73,26 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }, { r
...
+      dribbble: String
+      github: String
+      avatar_url: String
+      twitter: String
+      website: String
+      active: Boolean
+    }
+  `;
 
-  return [];
```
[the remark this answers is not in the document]
mainframev: right

mainframev:
> Btw, now displaying all contributors appears to work! Does that mean you found the problem? 🤔 I'll assume that the error message works because I expected to be able to see it in the deployment.

i think i will also add tests for that, so it will be covered, to make sure it works properly

silvenon (approved):
Seems to work great!

PR #2977, opened 2021-06-07
chore(dependabot): forbidden access to tokens

mainframev:
Dependabot can't have access anymore on pull_request for tokens, that is why it's always failing.
Link to github issue: https://github.com/dependabot/dependabot-core/issues/3253

[I think this comment well describes the problem](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-799447007).


mainframev:
postponed.

PR #2980, opened 2021-06-08
fix: store URL metadata in Gatsby cache

silvenon:
This (hopefully) fixes the problem of error caused by missing keys in `metaFileDataMap`, the assumption being that the data didn't get stored because those nodes got cached, and therefore skipped, so we're using the same API to store metadata, too. This assumption is based on the fact that the problem goes away after clearing Gatsby cache.

 Storybook: https://orbit-fix-metaFileDataMap-cache.surge.sh

### DIFF
--- a/docs/gatsby-node.js
+++ b/docs/gatsby-node.js
@@ -3,15 +3,9 @@ const fs = require("fs");
 const yaml = require("js-yaml");
 
 const { createFilePath } = require(`gatsby-source-filesystem`);
-const {
-  omitNumbers,
-  getDocumentUrl,
-  metaFileDataMap,
-  getParentUrl,
-  getDocumentTrail,
-} = require("./utils/document");
+const { omitNumbers, getDocumentUrl, getParentUrl, getDocumentTrail } = require("./utils/document");
 
-exports.onCreateNode = ({ node, getNode, actions, reporter }) => {
+exports.onCreateNode = async ({ cache, node, getNode, actions, reporter }) => {
   if (
     node.internal.type === "Directory" &&
     node.sourceInstanceName === "documentation" &&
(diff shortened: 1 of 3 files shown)


docs/gatsby-node.js
```diff
@@ -30,7 +24,7 @@ exports.onCreateNode = ({ node, getNode, actions, reporter }) => {
             : node.absolutePath
         }", every directory in "src/documentation" should have one`,
       );
-      metaFileDataMap.set(url, {});
+      await cache.set(url, {});
```
CollierCZ, line 27: Would it make sense to make the cache key more specific? Like about it being the meta file? I don't know if there are other things that use similar keys or might be in the future. If so, it might help, but not sure it's a problem or not.

CollierCZ (changes requested):
It seems to solve the problem of breaking the build, which was the main problem. Builds passing are always a good thing. 💪 

There's still the problem that changing the data in a `meta.yml` file will be cached. So in situations where the cache is persisted, those changes won't apply unless the cache is cleared. Right now that's not the case with deploys, but is the case for development and the test workflow in the CI.

If there's not an easy fix for it, it's probably something that we can live with. But I'd at least like to see it documented somewhere so the issue doesn't come up in a few months and we don't remember why the changes aren't appearing.

mainframev (dismissed):

docs/gatsby-node.js
silvenon: I was thinking about this as well, but docs don't mention which cache keys I should avoid, so for now I'll pretend as if each plugin has namespaced cache, including the root `gatsby-node.js`, and that I don't have to worry about overriding anything.

CollierCZ: The cache is stored under `default-site-plugin`, which means all data from `gatsby-*` (node, config, etc.). The only cache I see there now is from this action, so it doesn't seem like a current problem. Just didn't know if it wouldn't be in the future, but probably not.

silvenon:
@CollierCZ added limited support for updating `meta.yml` files, and described the problem in the comments. I'm sad that I didn't find an easy way to fix this, but we'll get there eventually.

CollierCZ (approved):
The situation isn't ideal, but it's better than it was and if we can't do anything with it, then this will do.

PR #2982, opened 2021-06-09
improve: remove need for Streamline package

silvenon:
Aside from the fact that having this package in our dependencies required contributors to add Streamline environment variables, which we can't share with them, pipeline was also sometimes failing because it couldn't establish the connection when executing the package's postinstall script.

 Storybook: https://orbit-chore-avoid-streamline-package.surge.sh


CollierCZ:
I like the change, but do we have the license for it? Including the code directly makes it part of our MIT license, no?

silvenon:
Awaiting response from the Streamline team, but we'll probably have to relicense orbit.kiwi code regardless of whether we're using the npm package or not.

silvenon:
We aren't allowed to do this, so I'll try to contribute to the Streamline package and [make the environment variables _optional_](https://jira.kiwi.com/browse/ORBIT-1819), they are open to it.

Thanks @CollierCZ for motivating me to research this.

PR #2987, opened 2021-06-14
fix(sandbox): prettify the outputed code

mainframev:
Storybook: https://orbit-fix-sandbox-prettify.surge.sh

### DIFF
--- a/docs/plugins/sandbox/gatsby-node.js
+++ b/docs/plugins/sandbox/gatsby-node.js
@@ -1,3 +1,6 @@
+const { format } = require("prettier");
+const parserTypeScript = require("prettier/parser-typescript");
+
 const { getScope, omitTypes } = require("./helpers");
 
 exports.onCreateNode = async ({ node, actions, loadNodeContent }) => {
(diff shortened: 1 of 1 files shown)


silvenon:
Damnit, JS doesn't work 😫  I don't think we can keep avoiding this issue.

silvenon (commented):
Damnit, JS doesn't work again 😫  We can't keep avoiding this issue, could you see if you can decrease the bundle size? (Not in this PR, of course.)

docs/src/components/ReactExample/index.tsx
```diff
@@ -85,8 +87,13 @@ const ReactExample = ({ exampleId, minHeight }: Props) => {
     })
     .join("\n");
 
+  const code = format(fields.example, {
+    parser: "typescript",
+    plugins: [parserTypeScript],
+  });
```
silvenon, line 93: This achieves the desired effect, but is it possible to do it in build time, rather than client-side? That way we can avoid adding prettier to the bundle.

silvenon (approved):

docs/src/components/ReactExample/index.tsx
mainframev: done

silvenon (approved):

PR #2994, opened 2021-06-22
fix(InputField): add aria-required

mainframev:
Noticed, that `VoiceOver` does not recognize the field as required. Added `aria-required` to fix that. 
 Storybook: https://orbit-fix-required-inputs.surge.sh


silvenon (changes requested):
[`aria-required` docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute) say that `aria-required` should only be used for backwards compatibility, and to use `required` instead. Have you considered this? I realize it's probably a breaking change because in that case browsers will natively prevent submitting the form, but I think that's expected functionality.

mainframev:
> [`aria-required` docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute) say that `aria-required` should only be used for backwards compatibility, and to use `required` instead. Have you considered this? I realize it's probably a breaking change because in that case browsers will natively prevent submitting the form, but I think that's expected functionality.

we do not use `required` for a reason, we do not want to show the default browser message. but for screen reader we have to have it as `requried` field.

silvenon (approved):
👍 Just add the comment explaining the situation.

PR #2995, opened 2021-06-22
feat(docs): open search by keyboard

mainframev:
Will suggested adding a feature to open the `SearchModal` by keyboard navigation by `cmd+k` or `cmd+f` 🤔  
 Storybook: https://orbit-feat-keyboard-search.surge.sh

### DIFF
--- a/docs/src/components/Navbar.tsx
+++ b/docs/src/components/Navbar.tsx
@@ -23,9 +23,7 @@ import Search from "./Search";
 import Modal from "./Modal";
 import { useBookmarks } from "../services/bookmarks";
 import { MAX_CONTENT_WIDTH, CONTENT_PADDING } from "../consts";
-
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const { useKeyboard } = require("../services/KeyboardProvider");
+import { useKeyboard } from "../services/KeyboardProvider";
 
 const StyledWrapper = styled.header`
   position: relative;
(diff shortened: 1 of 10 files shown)


docs/src/components/Navbar.tsx
```diff
@@ -96,7 +97,7 @@ interface Props {
 }
 
 const Navbar = ({ location, docNavigation }: Props) => {
-  const [searchOpen, setSearchOpen] = React.useState<boolean>(false);
+  const { isSearchOpened, setSearchOpened } = useKeyboard();
```
silvenon, line 100: I recommend making `useKeyboard` behave similar to `useState` and returning array rather than an object, at least that's what I've been doing so far. Also, keep the same naming convention, i.e. `searchOpen` and `setSearchOpen` instead of `isSearchOpened` and `setSearchOpened`, just like `menuOpen` and `setMenuOpen` below.

docs/src/services/KeyboardProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }, [setSearchOpened]);
+
+  return (
+    <KeyboardContext.Provider value={{ isSearchOpened, setSearchOpened }}>
+      {children}
+    </KeyboardContext.Provider>
+  );
+};
+
+export const useKeyboard = () => React.useContext(KeyboardContext);
```
silvenon, line 32: Considering that logic in here is strictly intended for search, I'd move it to `components/Search/service.tsx` (or something similar), rename `useKeyboard` and `KeyboardContextProvider` to `useSearch` and `SearchProvider` (`SearchProvider` instead of `SearchContextProvider` is more consistent with how we name other providers) and export both `useSearch` and `SearchProvider` from `components/Search/index.tsx` so that you can easily import them elsewhere.

Unrelated, if you want feel free to also rename `hooks/useDevMode.tsx` to `services/dev-mode.tsx` to maintain the naming convention of when we export both the hook and the context provider. Also, use named export for the `useDevMode` hook instead of default export.

silvenon (changes requested):

silvenon:
Also, it would be nice to show to people some kind of a placeholder about this shortcut. This is what Docusaurus has, you can ask Will what to do. 🤷

Screen Shot 2021-07-07 at 10 35 37

silvenon (approved):

PR #2998, opened 2021-06-28
feat: add descriptions for overview pages

CollierCZ:
Storybook: https://orbit-docs-overview-descriptions.surge.sh

### DIFF
--- a/docs/src/documentation/03-components/06-accessibility/meta.yml
+++ b/docs/src/documentation/03-components/06-accessibility/meta.yml
@@ -1,2 +1,3 @@
 title: Accessibility
+description: Components to increase the accessibility of your designs.
 type: folder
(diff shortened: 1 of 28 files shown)


docs/src/documentation/03-components/06-accessibility/meta.yml
```diff
@@ -1,2 +1,3 @@
 title: Accessibility
+description: Components to increase the accessibility of your designs.
```
silvenon, line 2:
```suggestion
description: Components to improve the accessibility of your designs.
```

docs/src/documentation/03-components/12-layout/meta.yml
```diff
@@ -1,2 +1,3 @@
 title: Layout
+description: Components to help you lay out your content.
```
silvenon, line 2: Maybe this would be slightly more informative.
```suggestion
description: Components to help you lay out your content on the page.
```

silvenon (changes requested):

docs/src/documentation/03-components/12-layout/meta.yml
CollierCZ: I've tried to stay away from the term "page" because it's less universal, but given the components here, it's probably fine.

silvenon (approved):

PR #3004, opened 2021-06-28
docs: fix development mode toggle in footer

silvenon:
The development mode toggle didn't work in pages that use `Layout` instead of `DocLayout` because `Layout` didn't provide the context.

 Storybook: https://orbit-docs-dev-mode-toggle-footer.surge.sh


mainframev:
@silvenon oh, sorry, did not notice you already created the PR. Which solution is better, wdy?

mainframev (approved):

PR #3005, opened 2021-06-28
docs: add some missing features to overview pages

silvenon:
We want to always have breadcrumbs, so we're adding a reasonable default when they aren't explicitly provided.

 Storybook: https://orbit-docs-default-breadcrumbs.surge.sh


CollierCZ (dismissed):
Would it be possible to quickly implement something for multiple levels (like https://e51deade.orbit.pages.dev/components/overlay)? If not quickly, I don't think it's worth changing, so this is fine. Just if it's not too long, it'd be nice to have.

silvenon:
Should be good now. 👍 As a bonus I added trailing slashes to overview pages and fixed that the necessary navigation items open. I refactored the code for overview pages and now it's being tested better.

docs/services/overview-pages.js
```diff
@@ -51,39 +55,55 @@ function createPages(obj, slug) {
...
+        const { data } = matter(await fsx.readFile(file, "utf-8"));
+        _.set(structure, relativePath.split(path.sep), data);
+      }
+    }),
+  );
 
-  files.forEach(file => {
-    const wholePath = omitNumbers(file)
-      .replace(/\.[^/.]+$/, "")
-      .split("/");
```
silvenon, line 71: This was unclear, so I tried to clear it up with `path.parse`.

docs/services/overview-pages.js
```diff
@@ -51,39 +55,55 @@ function createPages(obj, slug) {
...
 }
 
-const renderOverviewPages = (allPages, callback) => {
-  allPages.forEach(page => {
-    const { pages } = page;
-    if (!pages) return undefined;
-    callback(page);
-    return renderOverviewPages(pages, callback);
-  });
+const forEachNestedPage = (pages, callback) => {
```
silvenon, line 58: This function is very similar, but I renamed it for clarity.

docs/services/overview-pages.js
```diff
@@ -51,39 +55,55 @@ function createPages(obj, slug) {
...
+        ...page,
+        // keep pages nested only one level deep
+        pages: page.pages.map(p => _.omit(p, ["pages"])),
+      });
+      forEachNestedPage(page.pages, callback);
+    });
 };
 
-async function createOverviewPages(callback) {
+async function getOverviewPages() {
```
silvenon, line 71: I renamed this function so that it better describes that it now returns an array of overview pages.

docs/utils/document.js
```diff
@@ -1,6 +1,6 @@
 const path = require("path");
 
-const rePathNumbers = new RegExp(`(${path.sep})\\d+-`, "g");
+const rePathNumbers = new RegExp(`(^|${path.sep})\\d+-`, "g");
```
silvenon, line 3: This was needed so that `omitNumbers` works for `01-*` as well, although I'm not sure if it's needed now, but it doesn't hurt.

docs/src/components/DocNavigation/index.tsx
```diff
@@ -39,7 +39,7 @@ export function groupTrails(trails: Trail[]): Navigation {
       if (!acc[name]) {
         acc[name] = { result: [] };
         if (index < trail.length - 1) {
-          acc.result.push({ type: "branch", name, items: acc[name].result });
+          acc.result.push({ type: "branch", name, url, items: acc[name].result });
```
silvenon, line 42: This was needed to determine which navigation item to expand.

silvenon (commented):
A little help with reviewing.

docs/services/overview-pages.js
```diff
@@ -51,39 +55,55 @@ function createPages(obj, slug) {
...
-      _.set(structure, ommitedPath, { title, type, description });
-    } else {
-      const { data } = matter(fs.readFileSync(file, "utf-8"));
-      _.set(structure, ommitedPath, data);
-    }
+  // derive overview pages by recursively looping through pages that have nested pages
+  // this way we're essentially flattening the pages to be only one level deep
+  forEachNestedPage(allPages, page => {
+    overviewPages.push(page);
   });
```
CollierCZ, line 102: As above, I'm not sure this logic is necessary before the page is created, but if it's here, wouldn't it be a good use for reduce?

docs/utils/document.js
CollierCZ: I don't see any cases where omitNumbers is passed a relative path where it might be an issue, but I don't suppose it hurts.

While looking it up, i did notice we now have two `omitNumbers` functions. I'd get rid of the one in `docs/src/utils` since it doesn't seem to be used anywhere. Nor is `capitalize` in that file, though it doesn't have another function it might get confused with.

CollierCZ (changes requested):
A couple of minor things probably not fully related to this PR. It'd be nice to at least clean up the `omitNumbers` function, though. Otherwise, it looks good to me.

docs/utils/document.js
silvenon: In the test environment we use relative paths, see in [`gatsby-node.js`](https://github.com/kiwicom/orbit/pull/3005/files#diff-97ce157ffa078c33fbccfb2e69ae8be278c4074fbd0ec7c750ce35cceaa454f1R193).

docs/services/overview-pages.js
silvenon: I made this logic slightly more clear by inlining the recursive function at the place of invocation. I don't think this would be a good use case for `reduce` because `reduce` is best for deriving a single non-array value out of an array of elements. It's being overused in JavaScript because you can chain it to other array methods, but it often results in very unclear logic.

docs/gatsby-node.js
```diff
@@ -180,13 +182,19 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+  const overviewPages = await getOverviewPages();
+  await Promise.all(
+    overviewPages.map(async page => {
+      const trail = await getDocumentTrail(cache, page.slug);
+      const overviewPath = path.join(__dirname, "src/templates/Overview.tsx");
+      createPage({
+        path: page.slug,
+        component:
+          process.env.NODE_ENV === "test" ? path.relative(ROOT, overviewPath) : overviewPath,
+        context: { ...page, trail },
```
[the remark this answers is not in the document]
silvenon: In order to programmatically create pages in Gatsby a lot of this stuff had to be computed in advance. We can probably fetch some of this data via GraphQL, which I think would split the logic somewhat, and we will because we need it for other stuff, I'd postpone this as a separate effort.

CollierCZ (approved):

PR #3012, opened 2021-07-04
fix(docs): developer mode, move provider to gatsby-browser

mainframev:
The developer mode did not work on the main page, I guess it could not receive the values from provider. So I moved that to `gatsby-browser`


mainframev:
close. Already fixed in #3004

PR #3015, opened 2021-07-06
fix: generate table of contents dynamically

silvenon:
Generating table of contents by statically analyzing MDX is incomplete in places where headings are rendered via JSX, so now heading componetns are registering themselves in the context, which is then used to generate the complete table of contents.

 Storybook: https://orbit-docs-complete-toc.surge.sh


docs/config/jest-setup.js
```diff
@@ -1,6 +1,5 @@
 import { configure } from "@testing-library/react";
 import "@testing-library/jest-dom";
-import "jest-styled-components";
```
silvenon, line 3: We don't seem to need this at all, I'm removing it because of an error while generating a snapshot of table of contents, I assume because of some invalid CSS that was most probably not caused by these changes.

docs/src/components/ComponentStatus/index.tsx
```diff
@@ -63,11 +63,7 @@ export const ComponentStatusList = () => (
   <>
     {Object.values(Groups).map(group => (
       <React.Fragment key={group}>
-        <HeadingWithLink>
-          <Heading as="h2" type="title2">
-            {group}
-          </Heading>
-        </HeadingWithLink>
+        <H2>{group}</H2>
```
silvenon, line 66: There are a bunch of these changes because these headings need to end up in the table of contents.

docs/src/components/DocLayout/index.tsx
```diff
@@ -90,157 +77,151 @@ export default function DocLayout({
         description={description}
         path={path}
       />
-      <ThemeProvider theme={theme}>
-        <DevModeProvider>
```
silvenon, line 94: There's a bunch of indentation changes in this file, but there's not a lot going on aside from removing these two context providers and updating the table of contents, so I recommend telling GitHub to hide whitespace changes while reviewing.

docs/src/components/TableOfContents.tsx
```diff
@@ -1,11 +1,56 @@
 import React from "react";
 import Scrollspy from "react-scrollspy";
 import styled, { css } from "styled-components";
+import { v4 as uuidv4 } from "uuid";
 
-export interface TocItemObject {
+import { slugify } from "../utils/common";
+
+const DEPTH = 3;
```
silvenon, line 8: This excludes `h4` and `h5` from the table of contents, i. e. it only goes 3 levels deep.

docs/src/components/TableOfContents.tsx
```diff
@@ -1,11 +1,56 @@
...
+    setTableOfContents(prevToc => {
+      if (level >= DEPTH) return prevToc;
+      if (prevToc.find(item => item.id === id)) {
+        return prevToc.map(item =>
+          item.id === id ? { ...item, title, slug: slugify(title), level } : item,
+        );
+      }
+      return [...prevToc, { id, title, slug: slugify(title), level }];
+    });
+  }, [title, level, id, setTableOfContents]);
```
silvenon, line 53: This part is probably slightly overengineered because it relies on the assumption that headings might be dynamically updated. It attempts to update the heading if it already exists by checking its `id`, rather than always appending registered headings to the bottom.

docs/src/components/TableOfContents.tsx
```diff
@@ -69,48 +114,27 @@ const StyledTocListItem = styled.li<{ level: number }>`
...
-    </StyledTocListItem>
-  ));
+  return tree.map(item => {
+    const nestedContent = getTocContent(item.items, nextLevel, activeScrollSpy);
+    return (
+      <StyledTocListItem key={item.id} level={level}>
+        <StyledAnchor level={level} href={`#${item.slug}`} active={activeScrollSpy === item.slug}>
+          {item.title}
+        </StyledAnchor>
+        {item.items && React.Children.count(nestedContent) > 0 && <ul>{nestedContent}</ul>}
```
silvenon, line 130: Prevents adding empty lists if there are no items.

docs/wrap-with-providers.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { TableOfContentsProvider } from "./src/components/TableOfContents";
+import theme from "./src/theme";
+
+export default function wrapWithProviders({ element }: { element: React.ReactNode }) {
+  return (
+    <ThemeProvider theme={theme}>
+      <DevModeProvider>
+        <TableOfContentsProvider>{element}</TableOfContentsProvider>
+      </DevModeProvider>
+    </ThemeProvider>
```
silvenon, line 14: The main reason why I moved these providers here is because I needed the context for table of contents to already be available in `DocLayout` in order to check whether there are any items, so I decided to do some extra optimization and moved other repeating providers here as well, as suggested by Viktor. Now we won't get into problems with missing context and having to repetitively provide it.

mainframev (commented):
could you split the logic inside the `TableOfContents.tsx` into separate files? It looks a bit messy 🤔  Maybe move the context logic to a separate file.

docs/src/components/TableOfContents.tsx
```diff
@@ -1,11 +1,56 @@
...
+}
+
+export function useTableOfContents() {
+  const [tableOfContents] = React.useContext(TableOfContentsContext);
+  return tableOfContents;
+}
+
+export function useTableOfContentsRegister({ title, level }: { title: string; level: number }) {
+  const [, setTableOfContents] = React.useContext(TableOfContentsContext);
+  const id = React.useMemo(() => uuidv4(), []);
```
mainframev, line 41: i think you can use our `randomID` function, if you will pass there title of current item? 🤔

mainframev (approved):

PR #3016, opened 2021-07-08
fix: reduce bundle size

mainframev:
so as i wrote before, the problem is in icons, seems like a forgot to change one import and that's why webpack analyzer does not show reduce of bundle size. Now it's reduced and is about 800kb compare to 2.2mb in gzip. 
 Storybook: https://orbit-fix-bundle-size.surge.sh


silvenon (approved):
Looks good! If pipeline passes go ahead and merge. Great job! 💯

PR #3018, opened 2021-07-09
fix(sandbox): viewport preview

mainframev:
fixed the viewport previewer, so it applies the width on `Wrapper` before `iframe`, so `iframe` takes that width and `window` inside `iframe` receives proper width
 Storybook: https://orbit-fix-window-size.surge.sh


CollierCZ:
It seems like it's fixed nicely in some places, but other problems exist elsewhere. In [some examples](https://006a0321.orbit.pages.dev/examples/text-types) only the mobile widths (large and small) are available. They also don't have the [right automatic height](https://006a0321.orbit.pages.dev/foundation/typography/#text-styles).

Also, the full-page sandbox seems to have only mobile large, mobile small, desktop and full width. Is that intentional to not have the others?

mainframev:
> It seems like it's fixed nicely in some places, but other problems exist elsewhere. In [some examples](https://006a0321.orbit.pages.dev/examples/text-types) only the mobile widths (large and small) are available. They also don't have the [right automatic height](https://006a0321.orbit.pages.dev/foundation/typography/#text-styles).
> 
> Also, the full-page sandbox seems to have only mobile large, mobile small, desktop and full width. Is that intentional to not have the others?

fixed

docs/src/components/ReactExample/helpers.tsx
```diff
@@ -1,38 +1,33 @@
...
-  if (isFull) return "FullWidth";
-  if (size <= QUERIES.mediumMobile) return `Mobile small`;
-  if (size <= QUERIES.largeMobile) return `Mobile medium`;
-  if (size <= QUERIES.tablet) return `Mobile large`;
-  if (size <= QUERIES.desktop) return `Tablet`;
-
-  return `Desktop`;
-};
+import * as Components from "@kiwicom/orbit-components";
+import * as Icons from "@kiwicom/orbit-components/lib/icons";
```
silvenon, line 2:
```suggestion
import * as Icons from "@kiwicom/orbit-components/icons";
```

docs/src/templates/sandbox/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import Board from "../../components/ReactExample/Board";
+import ViewportsRuler from "../../components/ReactExample/ViewportsRuler";
+import Frame from "../../components/ReactExample/Frame";
+import useSandbox from "./useSandbox";
+
+const StyledWrapper = styled.div`
+  display: grid;
+  min-height: 100%;
+  grid-template-rows: auto 1fr auto;
+  grid-template-columns: 100%;
```
silvenon, line 17:
```suggestion
  grid-template-columns: 1fr;
```

docs/src/templates/sandbox/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+`;
+
+const Sandbox = ({ pathContext }) => {
+  const { example, id, example_id, scope } = pathContext;
+  const [isEditorOpened, setOpenEditor] = React.useState(true);
+  const [width, setPreviewWidth] = React.useState(0);
+  const { code, origin } = useSandbox(example_id);
+
+  const getCurrentWidth = React.useCallback(size => setPreviewWidth(size), []);
```
silvenon, line 46: `getCurrentWidth` doesn't get a width, it sets it, I recommend a name like `handleChangeRulerSize`.

silvenon (changes requested):

silvenon (approved):

PR #3022, opened 2021-07-13
docs: fix overview inner spacing

vepor:
Storybook: https://orbit-docs-fixes-overview.surge.sh

### DIFF
--- a/docs/src/components/Tile.tsx
+++ b/docs/src/components/Tile.tsx
@@ -209,7 +209,7 @@ export default function Tile({
             <TileTitle>{title}</TileTitle>
             <div
               css={css`
-                margin: 0.5rem 0 ${href ? "1.5rem" : "0"};
+                margin: 0.5rem 0 ${href ? (linkContent && "1.5rem") || "0.5rem" : "0"};
               `}
             >
               {/* wrap plain strings in a p tag and otherwise render children */}
--- a/docs/src/components/Tile.tsx
+++ b/docs/src/components/Tile.tsx
@@ -209,7 +209,7 @@ export default function Tile({
             <TileTitle>{title}</TileTitle>
             <div
               css={css`
-                margin: 0.5rem 0 ${href ? (linkContent && "1.5rem") || "0.5rem" : "0"};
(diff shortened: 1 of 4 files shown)


docs/src/components/Tile.tsx
```diff
@@ -209,7 +209,7 @@ export default function Tile({
             <TileTitle>{title}</TileTitle>
             <div
               css={css`
-                margin: 0.5rem 0 ${href ? "1.5rem" : "0"};
+                margin: 0.5rem 0 ${href ? (linkContent && "1.5rem") || "0.5rem" : "0"};
```
silvenon, line 212: No reason to add padding just because it's a link. Also, at the bottom add `StyledEndLinkWrapper` if both `href` and `linkContent` are provided, not only `href`, otherwise there will be empty elements.
```suggestion
                margin: 0.5rem 0 ${href && linkContent ? "1.5rem" : "0"};
```

silvenon (changes requested):

docs/src/templates/Overview.tsx
```diff
@@ -30,7 +30,8 @@ const Overview = ({ location, pageContext }: Props) => {
         columns="1fr"
         gap="2rem"
         largeMobile={{ columns: "repeat(2, 1fr)" }}
-        tablet={{ columns: "repeat(3, 1fr)" }}
+        desktop={{ columns: "repeat(3, 1fr)" }}
+        largeDesktop={{ columns: "repeat(2, 1fr)" }}
```
silvenon, line 34: I forgot about this. Merge master in `master` to get latest table of contents changes, then use the `useTableOfContents` hook to check whether there are any items (no items = no table of contents), then use three columns if there aren't any, otherwise two columns:

```tsx
const [tableOfContents] = useTableOfContents()

// inside Grid props
{...(tableOfContents.length > 0 ? {
  desktop: { columns: "repeat(3, 1fr)" },
  largeDesktop: { columns: "repeat(2, 1fr)" },
} : {
  tablet: { columns: "repeat(3, 1fr)" },
})}
```

silvenon (changes requested):

silvenon (approved):

PR #3023, opened 2021-07-13
docs: replace Spectrum chat for Figma community link

vepor:
Storybook: https://orbit-docs-figma.surge.sh


docs/src/pages/index.tsx
```diff
@@ -36,19 +36,17 @@ interface Props {
...
-
 export default function Home({ location, path }: Props) {
   const [searchOpen, setSearchOpen] = React.useState<boolean>(false);
   return (
     <Layout
       location={location}
       title="Orbit design system"
       description="An open source design system for your next travel project."
-      path="/"
+      path={path}
```
silvenon, line 49: We're deliberately hardcoding the path. ATM I forgot why, but using `path` didn't work out well in some cases, so delete it from the destructuring statement instead.

docs/src/pages/index.tsx
```diff
@@ -95,7 +91,7 @@ export default function Home({ location, path }: Props) {
                 iconRight={<ArrowRight />}
                 // @ts-expect-error asComponent has wrong type declaration
                 asComponent={GatsbyLinkToButton}
-                href="/getting-started/"
+                href="/getting-started/for-designers/"
```
silvenon, line 94: Many of these Tile changes appear to be reverting the changes we did for overview pages, not sure what happened on your end.

silvenon (changes requested):
The contrast is not great.

Screen Shot 2021-07-13 at 14 06 53

silvenon (approved):
Nailed it 👍

PR #3026, opened 2021-07-13
fix: always add footer gradient in document layout

silvenon:
Storybook: https://orbit-fix-gradient-overview-pages.surge.sh

### DIFF
--- a/docs/src/components/DocLayout/index.tsx
+++ b/docs/src/components/DocLayout/index.tsx
@@ -220,7 +220,7 @@ export default function DocLayout({
               )}
             </StyledMain>
           </StyledMiddle>
-          <Footer hasGradient={!noElevation} />
+          <Footer hasGradient />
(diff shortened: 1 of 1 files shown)


docs/src/components/Footer.tsx
```diff
@@ -84,11 +80,7 @@ function Dot() {
   return <Desktop>·</Desktop>;
 }
 
-interface Props {
-  hasGradient?: boolean;
-}
-
-export default function Footer({ hasGradient }: Props) {
```
vepor, line 91: Do we want to get rid of passing `hasGradient` property to the two or three usages of Footer? 🙂

silvenon: I decided to go a different way and simply always add a gradient in DocLayout, which is the goal of this PR anyway.

vepor (approved):

PR #3029, opened 2021-07-14
docs: fix Safari related bugs

vepor:
[Slack thread](https://skypicker.slack.com/archives/CAMS40F7B/p1626212264189900)
 Storybook: https://orbit-docs-fix-shadow.surge.sh

### DIFF
--- a/docs/src/mdx-components.tsx
+++ b/docs/src/mdx-components.tsx
@@ -122,9 +122,7 @@ export const th = ({ children, align }: React.ThHTMLAttributes<HTMLTableHeaderCe
 export const ul = ({ children }: React.HTMLAttributes<HTMLUListElement>) => (
   <ul
     css={css`
-      list-style-position: outside;
-      list-style-image: none;
-      list-style-type: disc;
+      list-style: disc none outside;
       margin-left: ${({ theme }) => theme.orbit.spaceMedium};
       ol,
       ul {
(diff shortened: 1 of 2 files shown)


docs/src/mdx-components.tsx
```diff
@@ -65,7 +65,9 @@ export const hr = () => <Separator spaceAfter="largest" />;
 export const ol = ({ children }: React.OlHTMLAttributes<HTMLOListElement>) => (
   <ol
     css={css`
-      list-style: outside none decimal;
+      list-style-position: outside;
+      list-style-image: none;
+      list-style-type: decimal;
```
silvenon, line 70: We ordered the values in the shorthand wrong, according to [`list-style`'s MDN documentation](http://developer.mozilla.org/en-US/docs/Web/CSS/list-style) the correct order is `type | image | position`, which works in Safari.
```suggestion
      list-style: decimal none outside;
```

docs/src/mdx-components.tsx
```diff
@@ -122,7 +124,9 @@ export const th = ({ children, align }: React.ThHTMLAttributes<HTMLTableHeaderCe
 export const ul = ({ children }: React.HTMLAttributes<HTMLUListElement>) => (
   <ul
     css={css`
-      list-style: outside none disc;
+      list-style-position: outside;
+      list-style-image: none;
+      list-style-type: disc;
```
silvenon, line 129:
```suggestion
      list-style: disc none outside;
```

silvenon (changes requested):

silvenon (approved):

PR #3030, opened 2021-07-14
feat(docs): don't hide search results when out of focus

vepor:
It's a bit annoying if you lose focus out of the search input and the results will hide.
 Storybook: https://orbit-docs-search-results.surge.sh


silvenon (approved):
I agree, this hiding behavior is intended for popovers I think.

PR #3033, opened 2021-07-14
sandbox: knobs feature 

mainframev:
Storybook: https://orbit-silvenon-docs-knobs-feature.surge.sh


mainframev:
> I've also briefly explored whether there are some obvious alternatives to traversing the AST, but I didn't find any because everything is very connected and needs to work in different environments. Your solution is really clever, I just want to find a simpler way in the future when we have time.

not sure if we will find, only if the completely different implementation of the whole sandbox, without `react-live`. It operates only with strings, so that's why a lot of string transformations via ast here.

docs/plugins/sandbox/gatsby/on-source-nodes.js
```diff
@@ -0,0 +1,61 @@
...
+
+    files.forEach(file => {
+      const { name } = path.parse(file);
+
+      const exampleFolder = path.dirname(file).split("/").slice(-1).join("").toLowerCase();
+      const content = fs.readFileSync(file, "utf-8");
+      const id = [exampleFolder, "-", name.toLowerCase()].join("");
+      const scope = getScope(content);
+      const example = getByName(content, "Example");
+      const knobCode = getByName(content, "exampleKnobs");
```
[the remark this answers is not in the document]
mainframev: Hmm, it's almost the same as a pick with regexp, just more comfortable. I can rewrite in a single function to pick all values, not sure if it will boost the speed.

docs/plugins/sandbox/gatsby/create-schema-customization.js
```diff
@@ -0,0 +1,42 @@
...
+        exampleKnobs: `[ExampleKnobs]`,
+      },
+    });
+
+    createTypes([
+      `
+    type Knob {
+      name: String!
+      type: String!
+      defaultValue: String!
```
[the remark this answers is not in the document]
mainframev: i can use union type with `buildUnionType`, i guess

docs/plugins/sandbox/gatsby/on-source-nodes.js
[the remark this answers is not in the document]
silvenon: Parsing code string to AST is much slower than couple of regexes, I think it would be really good to minimize doing it because it works on such a large scale. `getByName` can take AST instead of code string, and you can export `getCode` so that you can compute AST only once and pass it around.

docs/src/components/ReactExample/helpers.tsx
```diff
@@ -29,3 +35,97 @@ export const copyImports = (scope: Scope[]) =>
...
+          }
+        }
+      });
+    },
+  });
+
+  const { code } = generate(ast);
+
+  return code;
+};
```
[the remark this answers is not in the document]
silvenon: Huh, `transform` fails to parse the string for some reason, I wasn't able to figure out why yet, but considering that this works it's not worth the additional effort at this point.

docs/src/components/ReactExample/components/knobs/Boolean.tsx
```diff
@@ -6,18 +6,16 @@ import { Knob } from "../../Example";
 
 interface Props extends Pick<Knob, "name"> {
   checked: boolean;
-  onChange: (ev: React.SyntheticEvent<HTMLInputElement, Event>) => void;
+  onChange: (ev: React.ChangeEvent<HTMLInputElement>) => void;
```
silvenon, line 9: Sorry, I meant `*EventHandler`, not `*Event` 🤦‍♂️ 🙏
```suggestion
  onChange: React.ChangeEventHandler<HTMLInputElement>;
```

docs/src/__examples__/List/DEFAULT.tsx
```diff
@@ -9,9 +9,25 @@ export default {
...
-    title: "Default list",
-    description:
-      "Default lists display their items in primary color with small circles for each item.",
-  },
+  exampleKnobs: [
+    {
+      component: "List",
+      knobs: [
+        { name: "label", type: "text", defaultValue: "" },
+        { name: "icon", type: "icon", defaultValue: "CircleSmall" },
```
silvenon, line 17: These should be for `ListItem`, not `List`. We should find a way to automatically prevent these mistakes in the future.

```suggestion
        { name: "label", type: "text", defaultValue: "" },
        { name: "icon", type: "icon", defaultValue: "CircleSmall" },
```

docs/src/__examples__/Button/DEFAULT.tsx
```diff
@@ -1,10 +1,45 @@
...
+        {
+          name: "type",
+          type: "select",
+          defaultValue: "primary",
+          options: ["primary", "secondary", "critical", "primarySubtle", "criticalSubtle", "white"],
+        },
+      ],
+    },
+  ],
 };
```
silvenon, line 45: Playground's layout for Button is all over the place. Same for ButtonLink, SocialButton, TextLink and probably some others.

Screen Shot 2021-08-31 at 20 19 11

docs/src/__examples__/ButtonGroup/DEFAULT.tsx
```diff
@@ -18,9 +18,4 @@ export default {
       </Button>
     </ButtonGroup>
   ),
-  info: {
-    title: "Default button group",
-    description:
-      "Button groups visually join their children to show related actions as a single unit.",
-  },
 };
```
silvenon, line 21: Sandbox showing examples without knobs should not display the button "Playground".

docs/src/components/ReactExample/components/knobs/Icon.tsx
```diff
@@ -0,0 +1,35 @@
...
+            label: key,
+            value: key,
+            key,
+          }))}
+      />
+    </>
+  );
+};
+
+export default Icon;
```
silvenon, line 35: Icon knob works now, but always shows the same value, `Accommodation`.

docs/src/components/ReactExample/components/knobs/Text.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <Text weight="bold" size="small" type="secondary">
+        {name.split("-")[1]}
+      </Text>
+      {/* @ts-expect-error: fix in orbit */}
+      <InputField name={name} value={value} onChange={onChange} />
+    </>
+  );
+};
+
+export default TextKnob;
```
silvenon, line 23: Textual knobs like Card's `title` and `description` don't have spaces and don't work.

silvenon (changes requested):
Found some more stuff.

docs/src/__examples__/Button/DEFAULT.tsx
silvenon: My bad, now it looks ok to me, so I guess I forgot to build `@kiwicom/orbit-components` or something.

silvenon (approved):

PR #3036, opened 2021-07-15
feat: lock scrolling in components with overlay

silvenon:
Modal, Dialog, and on mobile Popover and Tooltip, have an overlay, but scrolling the content of these components also scrolls the content behind it. This locks the scrolling only to the scrolling container of those components.

To achieve this we created a new `useLockScrolling` hook which we're exposing separately in case anyone wants.

This is the first iteration of solving this problem, let me know what you think.

 Storybook: https://orbit-lock-scrolling.surge.sh


packages/orbit-components/src/hooks/useLockScrolling/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```jsx
+import useLockScrolling from "@kiwicom/orbit-components/lib/hooks/useLockScrolling";
+```
+
+Then you can use it:
+
+```jsx
+const Component = (props: Props) => {
+  const scrollableElementRef = (useRef < HTMLELement) | (null > null);
```
mainframev, line 15: this line is not correct :)

mainframev (approved):

mainframev (approved):

PR #3039, opened 2021-07-15
feat(docs): adopt new design tokens

vepor:


mainframev:
already solved in #3333

PR #3045, opened 2021-07-21
feat: add lockScrolling prop

silvenon:
This prop allows people to enable scrolling in components where it's otherwise locked for better user experience. While locking scrolling was a highly requested feature, we still don't want to assume that people never wanted to be able to scroll, so we're adding this prop as a way to return to the previous behavior. Later we will see whether anyone actually used it and potentially remove it.

I decided to go with the prop where people need to do `={false}` because I like the name the best and becuase not being able to scroll is desired, so turning this behavior off should look very strong.

 Storybook: https://orbit-silvenon-enable-scroll.surge.sh


packages/orbit-components/src/Dialog/README.md
```diff
@@ -19,13 +19,14 @@ After adding import into your project you can use it simply like:
...
+| renderInPortal    | `boolean`               |         | Optional prop, set it to `false` if you're rendering Dialog inside a custom portal, defaults to `true`                              |
+| description       | `React.Node`            |         | Optional description of the main action that Dialog performs.                                                                       |
+| illustration      | `React.Node`            |         | Optional illustration of the Dialog.                                                                                                |
+| **primaryAction** | `React.Node`            |         | Primary and required action that user can do with the Dialog.                                                                       |
+| secondaryAction   | `React.Node`            |         | Optional, secondary action that user can perform - possibility to close the Dialog most of the time.                                |
+| lockScrolling     | `boolean`               | `true`  | Whether to prevent scrolling of the rest of the page while Dialog is open. This is on by default to provide better user experience. |
```
mainframev, line 30:
```suggestion
| lockScrolling     | `boolean`               | `true`  | Whether to prevent scrolling of the rest of the page while Dialog is open. This is on by default to provide a better user experience. |
```

packages/orbit-components/src/Modal/README.md
```diff
@@ -25,6 +25,7 @@ Table below contains all types of the props available in the Modal component.
 | Name                | Type                       | Default    | Description                                                                                                                                                                                          |
 | :------------------ | :------------------------- | :--------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | children            | `React.Node`               |            | The content of the Modal. [See Subcomponents](#subcomponents)                                                                                                                                        |
+| lockScrolling       | `boolean`                  | `true`     | Whether to prevent scrolling of the rest of the page while Modal is open. This is on by default to provide better user experience.                                                                   |
```
mainframev, line 28:
```suggestion
| lockScrolling       | `boolean`                  | `true`     | Whether to prevent scrolling of the rest of the page while Modal is open. This is on by default to provide a better user experience.                                                                   |
```

packages/orbit-components/src/Popover/README.md
```diff
@@ -18,23 +18,24 @@ After adding import into your project you can use it simply like:
...
+| opened            | `boolean`               |                     | Prop for programmatically controlling Popover inner state. If `opened` is present open triggers are ignored.                                   |
+| preferredAlign    | [`enum`](#enum)         | `"start"`           | The preferred position to choose [See Functional specs](#functional-specs).                                                                    |
+| preferredPosition | [`enum`](#enum)         | `"bottom"`          | The preferred align to choose [See Functional specs](#functional-specs).                                                                       |
+| overlapped        | `boolean`               | `false`             | If `true`, the content of Popover will overlap the trigger children.                                                                           |
+| lockScrolling     | `boolean`               | `true`              | Whether to prevent scrolling of the rest of the page while Popover is open on mobile. This is on by default to provide better user experience. |
```
mainframev, line 34:
```suggestion
| lockScrolling     | `boolean`               | `true`              | Whether to prevent scrolling of the rest of the page while Popover is open on mobile. This is on by default to provide a better user experience. |
```

packages/orbit-components/src/Tooltip/README.md
```diff
@@ -18,19 +18,20 @@ After adding import into your project you can use it simply like:
...
+| preferredAlign       | [`enum`](#enum) |         | The preferred align to choose [See Functional specs](#functional-specs)                                                                        |
+| preferredPosition    | [`enum`](#enum) |         | The preferred position to choose [See Functional specs](#functional-specs)                                                                     |
+| removeUnderlinedText | `boolean`       |         | Removes underline on child component, when underline is not desired.                                                                           |
+| size                 | [`enum`](#enum) |         | The maximum possible size of the Tooltip.                                                                                                      |
+| stopPropagation      | `boolean`       |         | If `true` the click event on children won't bubble. Useful when you use Tooltip inside another clickable element.                              |
+| lockScrolling        | `boolean`       | `true`  | Whether to prevent scrolling of the rest of the page while Tooltip is open on mobile. This is on by default to provide better user experience. |
```
mainframev, line 33:
```suggestion
| lockScrolling        | `boolean`       | `true`  | Whether to prevent scrolling of the rest of the page while Tooltip is open on mobile. This is on by default to provide a better user experience. |
```

packages/orbit-components/src/Tooltip/index.js
```diff
@@ -16,6 +16,7 @@ const Tooltip = ({
   size = SIZE_OPTIONS.SMALL,
   content,
   preferredPosition,
+  lockScrolling,
```
mainframev, line 19:
```suggestion
  lockScrolling = true,
```

packages/orbit-components/src/primitives/MobileDialogPrimitive/index.js
```diff
@@ -19,6 +19,7 @@ const MobileDialog = ({
   stopPropagation = false,
   removeUnderlinedText,
   block = false,
+  lockScrolling,
```
mainframev, line 22:
```suggestion
  lockScrolling = true,
```

mainframev (changes requested):
`MobileDialogPrimitive` readme file is missing the `lockScrolling` prop name and description :)

packages/orbit-components/src/Tooltip/index.js
silvenon: This seems unnecessary because it's ultimately being passed to `DialogContent`, which is the one that actually does something with it.

mainframev (approved):

packages/orbit-components/src/Tooltip/index.js
mainframev: but in the README file you have `default true`

PR #3046, opened 2021-07-21
refactor(Popover): replace calculate logic with popper.js

mainframev:
As we discussed before, it would be nice to refactor `Popover` with `Popper`, because of the complex calculations and constant bugs with positioning. It will help us to save time. 
 Storybook: https://orbit-silvenon-feat-popover-replace-for-popper.surge.sh


packages/orbit-components/src/Popover/components/ContentWrapper.js
```diff
@@ -131,86 +145,112 @@ StyledPopoverPadding.defaultProps = {
...
+  const [actionsDimensions, setActionsDimensions] = React.useState(0);
   const { isInsideModal } = React.useContext(ModalContext);
   const { isLargeMobile, isTablet } = useMediaQuery();
-  const popover: {| current: React.ElementRef<*> |} = React.useRef(null);
+
   const content: {| current: React.ElementRef<*> |} = React.useRef(null);
   const scrollingElementRef = React.useRef<HTMLElement | null>(null);
   useLockScrolling(scrollingElementRef, lockScrolling && !isLargeMobile);
+
+  const popoverRef: {| current: React.ElementRef<*> |} = React.useRef(null);
```
silvenon, line 208: While you're changing refs you can also update how we type them:
```suggestion
  const popoverRef = React.useRef<HTMLElement | null>(null);
```

packages/orbit-components/src/Popover/index.js
```diff
@@ -1,52 +1,77 @@
...
+  ${mq.largeMobile(css`
+    border-radius: 3px;
+  `)}
+`;
+
+// $FlowFixMe: https://github.com/flow-typed/flow-typed/issues/3653#issuecomment-568539198
+StyledPopoverContent.defaultProps = {
+  theme: defaultTheme,
+};
+
```
silvenon, line 38: This component is unused. Too bad ESLint doesn't report it, I guess because of setting `defaultProps`.

packages/orbit-components/src/Popover/index.js
```diff
@@ -1,52 +1,77 @@
...
   fixed,
   lockScrolling,
+  noPadding,
+  width,
   actions,
-  renderInPortal = true,
+  overlapped,
+  dataTest,
 }: Props): React.Node => {
+  const ref: {| current: React.ElementRef<*> |} = React.useRef(null);
```
silvenon, line 56:
```suggestion
  const ref = React.useRef<HTMLElement | null>(null);
```

packages/orbit-components/src/Popover/index.js
```diff
@@ -59,7 +84,7 @@ const Popover = ({
   const handleOut = React.useCallback(
     ev => {
       // If open prop is present ignore custom handler
-      if (container.current && !container.current.contains(ev.target)) {
+      if (ref && !ref.current.contains(ev.target)) {
```
silvenon, line 87: This happened because of wrong type above. `ref` will always exist, but `ref.current` is the one that needs to be checked because it might be `null`.
```suggestion
      if (ref.current && !ref.current.contains(ev.target)) {
```

silvenon (changes requested):
Changes were a bit hard to follow because lots of stuff have been moved around, but seems to work great! However, I think breaking the API should be a separate decision, let's talk about it tomorrow. 🤙

silvenon (dismissed):

silvenon:
However, let's hold the merge until tomorrow so we can discuss the API change.

packages/orbit-components/src/Popover/index.js
```diff
@@ -2,51 +2,56 @@
...
-  noPadding,
-  overlapped,
+  content,
   onClose,
   onOpen,
+  offset,
+  placement = PLACEMENTS.BOTTOM_START,
   fixed,
   lockScrolling,
+  noFlip,
```
silvenon, line 29: `allowOverflow` should also be forwarded to `PopoverContent`.

silvenon (changes requested):

silvenon (approved):
Works well now 👍

PR #3047, opened 2021-07-21
improve: set up writing components in TypeScript

silvenon:
This setup allows us to gradually start converting our source code from Flow to TypeScript, and any new code can now be written in TypeScript. As a reward for doing this, both TypeScript and Flow declaration files will be automatically generated from the source code, so we can finally keep our types in a single file and prevent TypeScript and Flow types from falling out of sync, yay!

These changes appear to support a local test case I was practicing with, but it's hard to tell how well generating Flow declaration files will work for the rest of the code, we might need to tweak the process as we go, for example in this PR I needed to fix inconsistently named type exports in `defaultTheme.js`. But the only way to find out what else we need to do is to start converting. Off we go! 🔥

## Tips for converting files to TypeScript

- while writing your TypeScript code, keep in mind the conversion to Flow, e.g. you must always import types with `import type { MyType } from ...` instead of just regular `import { MyType } from ...` so that this can properly convert
  - keep `flowgen`'s limitations in mind while converting, they are listed in the readme, and double-check the generated Flow declaration file
- TypeScript files can import JavaScript files, but not the other way around, so be careful not to convert too much at once
- for maximum type coverage add type declaration for files you're not converting
  - e.g. if you converted file A which is importing file B, but you don't want to also convert file B because other files depend on it, create `B.d.ts` (if it's missing)
- delete declaration files (`.d.ts` and `.js.flow`) of files you're converting, they will be generated in the build step
- remove `// @flow` annotation
- remove Flow-related comment above `defaultProps` of styled components
- use `as const` for objects like the ones in `const.ts` files, see [Objects vs Enums](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums)
- use `keyof obj` to get keys of an object, and `typeof obj[keyof typeof obj]` to get values
  - these pair well with `as const`, you can use them to avoid manually listing keys and values in a type
- use [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) like [`Readonly`](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) over the `readonly` keyword to avoid repetition, this is the opportunity to tidy up our types
- use `const MyComp: React.FC<Readonly<Props>> = ...` style of defining components, or we can agree on an alternative approach, as long as we do it consistenly

 Storybook: https://orbit-silvenon-improve-typescript-first-components.surge.sh


silvenon:
This PR should wait for Yarn v3 update. Also, it should be broken down to multiple PRs, at least:

- remove `declare module` from TypeScript definition template
- rename types in `defaultTheme.d.ts` and export it from `src/index.d.ts`

PR #3048, opened 2021-07-21
chore: disallow useless `index` in import path

silvenon:
No more.

 Storybook: https://orbit-silvenon-index-import.surge.sh


mainframev:
looks like it unnecessarily removed imports inside examples 🤔

mainframev (approved):

PR #3052, opened 2021-07-26
improve: build tokens on postinstall

silvenon:
This script is important for development, so running it automatically after every install helps when switching between branches where tokens need to be rebuilt.

 Storybook: https://orbit-silvenon-improve-postinstall-build-tokens.surge.sh

### DIFF
--- a/packages/orbit-design-tokens/package.json
+++ b/packages/orbit-design-tokens/package.json
@@ -26,11 +26,10 @@
     "build:xml": "NODE_RUN=true babel-node ./scripts/generateXMLDesignTokens.js",
     "prebuild": "rimraf lib src/js",
     "build": "yarn clean && yarn run build:lib && yarn run generate-json && yarn run build-html && yarn run build-web && yarn run build-ios && yarn run build-android && yarn build:xml",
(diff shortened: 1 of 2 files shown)


packages/orbit-design-tokens/package.json
```diff
@@ -26,10 +26,13 @@
     "build:xml": "NODE_RUN=true babel-node ./scripts/generateXMLDesignTokens.js",
     "prebuild": "rimraf lib src/js",
     "build": "yarn clean && yarn run build:lib && yarn run generate-json && yarn run build-html && yarn run build-web && yarn run build-ios && yarn run build-android && yarn build:xml",
+    "quickbuild": "yarn build:lib",
```
mainframev, line 29: why you need that `quickbuild` script? can't you just use the `yarn build:lib`?

silvenon: I named it so that the intention is a bit more obvious. Doesn't help much, but I thought it's at least something.

mainframev: i think it's not necessary 🤔

mainframev (approved):

Issue #3054, opened 2021-07-29
Broken SSR


oreqizer:
https://github.com/kiwicom/orbit/blob/a8ef089ef53cb0c53f88a7b3e385249f8988903f/packages/orbit-components/src/Card/CardSection/index.js#L87

this breaks SSR, yo

use `react-uid`

silvenon:
It seems like we should replace all usage of `randomID` with `react-uid`, correct?

mainframev:
#3055 resolved this

PR #3055, opened 2021-07-30
fix(randomID): support SSR

mainframev:
As suggested in #3054, refactored internal `randomID` function with `react-uid` seed id hook.


packages/orbit-components/src/hooks/useRandomId/index.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import { useUIDSeed } from "react-uid";
+
+import type { useRandomID } from "./index.js.flow";
+
+const useRandomId: useRandomID = (name: string) => {
+  return useUIDSeed()(name);
```
silvenon, line 7: What's the benefit of `useUIDSeed` in this case over `useUID`? Considering that you're immediately calling the seed instead of saving it for using it multiple times.

packages/orbit-components/src/InputField/index.js
```diff
@@ -344,17 +344,10 @@ const InputField: React.AbstractComponent<Props, HTMLInputElement> = React.forwa
     dataAttrs,
   } = props;
 
-  const forID = React.useMemo(() => id || (label ? randomID("inputFieldID") : undefined), [
```
silvenon, line 347: What was the purpose of this additional `label ? randomID("inputFieldID") : undefined` logic? Probably nothing important, but did you leave it out of the new implementation on purpose?

silvenon (changes requested):
I think you can delete `utils/randomID` entirely, we're not using it and it's internal API so nobody else should be using it either.

packages/orbit-components/src/SmartPassIllustrations/illustrations/SmartPassV2.js
```diff
@@ -1,14 +1,14 @@
...
 import IllustrationWrapper from "./IllustrationWrapper";
 import type { Props } from "..";
 
 const SmartPassV2 = ({ primary, secondary, ...props }: Props): React.Node => {
-  const randomIDLinear = React.useMemo(() => randomID("linear1"), []);
-  const randomIDLinear2 = React.useMemo(() => randomID("linear2"), []);
-  const randomIDLinear3 = React.useMemo(() => randomID("linear3"), []);
+  const RandomIDLinear = useRandomId("linear1");
+  const RandomIDLinear2 = useRandomId("linear2");
+  const RandomIDLinear3 = useRandomId("linear3");
```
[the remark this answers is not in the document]
mainframev: it was replaced fast, so it was not my intention to rename it with a capital case.

packages/orbit-components/src/hooks/useRandomId/index.js
mainframev: it's similar to previous behavior with argument. Not sure what you mean by `multiple times`, because for several id's the hook should be defined several times. Like: 

```
 const id1 = useUID();
 const id2 = useUID();
 const id3 = useUID();

```

packages/orbit-components/src/InputField/index.js
mainframev: [it's just about labels.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) If the label is present, it should have `for` property and `Input` should have that `id`.

packages/orbit-components/src/PictureCard/index.js
```diff
@@ -229,7 +229,7 @@ const PictureCard = ({
     }
   }, []);
 
-  const cardID = React.useMemo(() => randomID("pictureCardID"), []);
+  const cardID = useRandomId()("pictureCardID");
```
silvenon, line 232: In order to give `useRandomId` more weight and make it clear what's going on I recommend always splitting it into two lines instead of immediately invoking.

```js
const randomId = useRandomId();
const cardID = randomId("pictureCardID");
```

packages/orbit-components/src/Accordion/AccordionSection/index.js
```diff
@@ -43,7 +43,12 @@ const AccordionSection = ({
           </SectionHeader>
         )}
 
-        <Slide maxHeight={height} expanded={isExpanded} id={slideID} ariaLabelledBy={slideID}>
+        <Slide
+          maxHeight={height}
+          expanded={isExpanded}
+          id={slideID("slideID")}
+          ariaLabelledBy={slideID("slideID")}
```
silvenon, line 50: We can decide about the convention in case we need to reuse the same ID, whether we're going to invoke seed inline like this multiple times or save it in variable, like we do in other places. I vote for variable because there's no chance of typos in that case.

silvenon (changes requested):
I was thinking about implementing `useUID` instead of `useUIDSeed`, in that case you would call `useRandomId` for each random ID you need, that would provide a simpler API:

```js
const cardID = useRandomId();
const forID = useRandomId();
```

Then we can introduce seed later in some other way if it's necessary, maybe as a new hook. I think the only benefit of seed is generating a `key` when looping in cases where you don't have a good alternative, which is nice, but maybe not such a frequent situation.

packages/orbit-components/src/PictureCard/index.js
mainframev: in my opinion it makes sense only if are multiple ids are needed :)

silvenon: Economically speaking, yes, but it's really uncommon to use hooks like this, they should be on their own, not be like regular functions. This usage makes me lean more towards removing the seed implementation and just using `useUID` instead.

packages/orbit-components/src/hooks/useRandomId/index.js
```diff
@@ -0,0 +1,12 @@
...
+
+import type { useRandomIdSeedType, useRandomIdType } from "./index.js.flow";
+
+export const useRandomIdSeed: useRandomIdSeedType = () => {
+  return useUIDSeed();
+};
+
+export const useRandomId: useRandomIdType = () => {
+  return useUID();
+};
```
silvenon, line 12: I recommend exporting `useRandomId` as default instead of named because the path is named `hooks/useRandomId` so people will expect it (I would). That would also send the correct message that `useRandomId` is the primary hook, and `useRandomIdSeed` is secondary, for more specific situations.

packages/orbit-components/src/primitives/TooltipPrimitive/__tests__/index.test.js
```diff
@@ -5,7 +5,9 @@ import { render } from "@testing-library/react";
 import Tooltip from "..";
 import Airplane from "../../../icons/Airplane";
 
-jest.mock("../../../utils/randomID", () => name => `${name}50000-id-50000`);
+jest.mock("../../../hooks/useRandomId", () => ({
+  useRandomId: () => 1,
+}));
```
silvenon, line 10: Mocking `useRandomId` is unnecessary, I tested locally and works without.

silvenon (changes requested):
**Much** better 👌

packages/orbit-components/src/primitives/TooltipPrimitive/__tests__/index.test.js
mainframev: it was failing on CI with screenshots, so i thought the reason is in the changing id.

silvenon (approved):

docs/src/documentation/05-development/03-hooks/usrandomid.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: useRandomId
+description: Generates unique random id
+redirect_from:
+  - /utilities/userandomid/
+---
+
+import UseRandomId from "@kiwicom/orbit-components/src/hooks/useRandomId/README.md";
```
vepor, line 8: @mainframev the documentation should mention real use case for the hook, currently it looks like a component, but it's not.

vepor: Oh, true. Sorry, I didn't realize it's imported README 🤦 

However, there's missing whitespace for the heading and thus resulting in wrong render:
https://github.com/kiwicom/orbit/pull/3055/files#diff-3ed8cb5d7c9c8fdbc5da47586e5a7f83decff66e90e3017399d4a2782a27fd5cR1

https://beta.orbit.kiwi/development/hooks/usrandomid/

PR #3071, opened 2021-08-10
feat(Skeleton): introduce Skeleton component

mainframev:
Storybook: https://orbit-silvenon-feat-skeleton.surge.sh


packages/orbit-components/src/Skeleton/Svg.js
```diff
@@ -0,0 +1,169 @@
...
+  const foreground = getColor(foregroundColor);
+  const keyTimes = `0; ${animationInterval}; 1`;
+  const duration = `${animationSpeed}s`;
+
+  React.useEffect(() => {
+    if (!children && rows && rowOffset) setCalculatedHeight(rows * rowOffset);
+  }, [rowOffset, rows, setCalculatedHeight]);
+
+  return (
+    <UID>
```
silvenon, line 85: Why not `useRandomId`? I assumed the only use case for `UID` are class components.

packages/orbit-components/src/Skeleton/__tests__/index.test.js
```diff
@@ -0,0 +1,36 @@
...
+    const children = "kek";
+    const ariaLabelledby = "bur";
+
+    render(
+      <Skeleton dataTest={dataTest} title="Loading" ariaLabelledby={ariaLabelledby}>
+        <rect>{children}</rect>
+      </Skeleton>,
+    );
+
+    expect(screen.getByTestId(dataTest)).toBeInTheDocument();
```
silvenon, line 25: In the future we can switch to simply calling `getBy*` without `expect` because `getBy*` itself fails when it doesn't exist, so it never reaches the `expect` functionality.

packages/orbit-components/src/Skeleton/__tests__/index.test.js
```diff
@@ -0,0 +1,36 @@
...
+    expect(screen.getByText(children)).toBeInTheDocument();
+    expect(screen.getByRole("img")).toHaveAttribute("aria-labelledby", ariaLabelledby);
+    expect(screen.getByText("Loading")).toBeInTheDocument();
+  });
+
+  it("should have rows", () => {
+    render(<Skeleton dataTest="test" rows={6} />);
+    const clipPath = document.getElementById("154-clip");
+    expect(clipPath?.childNodes).toHaveLength(6);
```
silvenon, line 34:
```suggestion
    expect(document.querySelectorAll("clipPath > *")).toHaveLength(6);
```

packages/orbit-components/src/Skeleton/helpers.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import defaultTheme from "../defaultTheme";
+
+const firstToUpper = string => string.charAt(0).toUpperCase() + string.slice(1);
```
silvenon, line 4: `Box` also has this, I suggest moving it to common utils.

packages/orbit-components/src/Skeleton/helpers.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const getColor = (token?: string): string | null => {
+  if (token) return defaultTheme.orbit[`palette${firstToUpper(token)}`];
+  return null;
```
silvenon, line 8:
```suggestion
  return token ? defaultTheme.orbit[`palette${firstToUpper(token)}`] : null;
```

packages/orbit-components/src/Skeleton/index.js
```diff
@@ -0,0 +1,23 @@
...
+import Svg from "./Svg";
+
+import type { Props } from ".";
+
+const Skeleton = ({ preset, ...props }: Props): React.Node => {
+  if (preset === "List") return <List {...props} />;
+  if (preset === "Image") return <Image {...props} />;
+  if (preset === "Card") return <Card {...props} />;
+  if (preset === "Button") return <Button {...props} />;
+  if (preset === "Text") return <Text {...props} />;
```
silvenon, line 18: Add a `presets/index.js` file which exports all presets, then:

```jsx
import * as Preset from "./presets"

// ...

const Component = Preset[preset] || Svg;

return <Component {...props} />
```

silvenon (changes requested):
1. What about adding Skeleton to orbit.kiwi with some basic description?
2. There's something off with the animation, it feels janky and like it has a delayed ending.

Otherwise very nice!

mainframev:
> 1. What about adding Skeleton to orbit.kiwi with some basic description?
It's already planned after release here :)

packages/orbit-components/src/Skeleton/index.d.ts
```diff
@@ -0,0 +1,98 @@
...
+  /** Set offset between rect elements  */
+  /** default: `0` */
+  readonly rowOffset?: number;
+  /** Number of rect elements */
+  /** default: `1` */
+  readonly rows?: number;
+  /** Add text for svg accessible name element  */
+  readonly title?: string;
+  /** You can pick one of predefined presets  */
+  readonly variant?: Preset;
```
silvenon, line 89:
```suggestion
  readonly preset?: Preset;
```

silvenon (changes requested):

docs/src/documentation/03-components/10-progress-indicators/skeleton/01-guidelines.mdx
```diff
@@ -0,0 +1,36 @@
+---
+title: Guidelines
+redirect_from:
+  - /components/skeleton/
+---
+
+<ReactExample exampleId="Skeleton-default" />
```
silvenon, line 7: None of the examples are loading for some reason.

silvenon (changes requested):
The animation is now better on one side, but choppy on the other. See the left side here, not sure if that was the intention:

https://user-images.githubusercontent.com/471278/131477145-30f9c8f4-e8d0-482a-a663-236b3e3fdcaa.mov

Is there any official Kiwi guidance and what this animation should look like?

mainframev:
> The animation is now better on one side, but choppy on the other. See the left side in this RTL example, not sure if that was the intention:
> 
>  Screen.Recording.2021-08-31.at.11.17.09.mov 
> Is there any official Kiwi guidance and what this animation should look like?

i suppose there is no guidance on such kind of animations 🤔

docs/src/components/DocNavigation/Category.tsx
```diff
@@ -15,7 +11,7 @@ export default function Category({ name, children }: Props) {
   return (
     <StyledCategoryContainer>
       <StyledCategoryName>{name}</StyledCategoryName>
-      <StyledCategoryItems>{children}</StyledCategoryItems>
+      {children}
     </StyledCategoryContainer>
   );
 }
```
silvenon, line 17: These orbit.kiwi changes seem related to #3098, not this PR. I guess it doesn't matter if they are going to cancel themselves out after merging #3098, but I'm curious why they're here.

packages/orbit-components/src/Skeleton/README.md
```diff
@@ -0,0 +1,60 @@
...
+| height          | `number`                                                                          |          |         | Set height for Svg element                                                                                                                                     |
+| rowBorderRadius | `number`                                                                          |          | `3`     | Border-radius for row rect elements                                                                                                                            |
+| rowHeight       | `number`                                                                          |          | `21px`  | Set height for rect elements                                                                                                                                   |
+| rowOffset       | `number`                                                                          |          | `0`     | Set offset between rect elements                                                                                                                               |
```
silvenon, line 43:
```suggestion
| rowOffset       | `number`                                                                          |          | `20`    | Set offset between rect elements                                                                                                                               |
```

packages/orbit-components/src/Skeleton/index.d.ts
```diff
@@ -0,0 +1,74 @@
...
+  /** Set height for Svg element */
+  readonly height?: number;
+  /** Border-radius for row rect elements */
+  /** default: `3` */
+  readonly rowBorderRadius?: number;
+  /** Set height for rect elements  */
+  /** default: `21` */
+  readonly rowHeight?: number;
+  /** Set offset between rect elements  */
+  /** default: `0` */
```
silvenon, line 57:
```suggestion
  /** default: `20` */
```

silvenon (changes requested):

silvenon (approved):

PR #3078, opened 2021-08-19
fix(useLockScrolling): fix body positioning on iOS

silvenon:
This fixes a bug on iOS Safari where locking scroll was causing body not to take up full width of the screen because body-scroll-lock sets only top and left position.

 Storybook: https://orbit-silvenon-fix-scroll-lock-positioning.surge.sh


mainframev (approved):
LGTM. Hope it will not cause bugs somewhere else on the product. 🤞🏻

PR #3088, opened 2021-08-27
fix(CountryFlag): set "undefined" as default code

silvenon:
The default code for CountryFlag is now `"undefined"` rather than `"anywhere"`, a gray flag with a question mark. This code is also used if the given code isn't supported.

 Storybook: https://orbit-silvenon-fix-default-country-flag-code.surge.sh


packages/orbit-components/src/CountryFlag/index.js
```diff
@@ -76,12 +76,12 @@ StyledShadow.defaultProps = {
 };
 
 function getCountryProps(code: ?string, name: ?string): {| code: string, name: ?string |} {
```
mainframev, line 78: the line was not changed, but  I guess it should be: `getCountryProps(code?: string, name?: string): {| code: string, name: string |}`, right?

mainframev (approved):

packages/orbit-components/src/CountryFlag/index.js
```diff
@@ -76,12 +76,12 @@ StyledShadow.defaultProps = {
...
-  const codeNormalized = code ? code.toUpperCase().replace("-", "_") : "ANYWHERE";
+  const codeNormalized = code ? code.toUpperCase().replace("-", "_") : "UNDEFINED";
   const countryCodeExists = codeNormalized in CODES;
 
   warning(countryCodeExists, "Country code not supported: %s", code);
-  const countryCode = countryCodeExists ? CODES[codeNormalized] : CODES.ANYWHERE;
-  const countryName = countryCode === CODES.ANYWHERE && !name ? "Anywhere" : name;
+  const countryCode = countryCodeExists ? CODES[codeNormalized] : CODES.UNDEFINED;
+  const countryName = countryCode === CODES.UNDEFINED && !name ? "Undefined" : name;
```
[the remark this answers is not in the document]
mainframev: should be ok 🤷🏻

packages/orbit-components/src/CountryFlag/index.js
silvenon: Returned `name` can still be `null`, but I changed the rest.

silvenon:
Disallowing `null` is another breaking change, so make sure to check it.

PR #3089, opened 2021-08-27
test: implement E2E testing with Cypress

silvenon:
We need this for testing #2350.

For bundler and server I chose the awesome [Vite](https://vitejs.dev/), I was honestly just waiting for an opportunity to sneak it into our project 😆 but it was also a good way to ensure that Orbit works with modern tools. In order to use it I needed to make some adjustments, namely add the ESM build of tokens. I think esbuild had a problem with the fact that the ESM build of components uses the CJS build of tokens, implementing the ESM build fixed the issue.

I discovered that we don't include TypeScript definitions in ESM build of components, oops! Probably #3047 already fixed that, but until that's merged let's fix it right away.

I also shortened our shortcut scripts to `yarn tokens | components | docs` to be easier to type, they were unnecessarily long. I removed `dev:docs` in favor of running `yarn docs dev`, I should probably do the same with `components` later to declutter the scripts a little.

Regarding Cypress implementation itself, the tricky part was starting the server, running the tests and closing the server when that's done. For this I chose `start-server-and-test` module, which Cypress docs recommend for this problem, and it works great. I didn't know what to add as an example test, so I added a tiny test for locking scrolling. I wanted a more agnostic check for scrollability than checking CSS, but it was surprisingly tricky to write a passing test, I don't understand Cypress well enough yet, I guess. The two E2E scripts are:

- `e2e:run` — headlessly runs the tests, for when we want to quickly run everything
- `e2e:open` — opens the browser, this is what we'll use in development to debug certain tests

 Storybook: https://orbit-silvenon-test-cypress.surge.sh


mainframev (approved):
LGTM :) I'd use `cypress:dev` instead of `e2e`. Having a number in the name seems not comfortable for me

PR #3093, opened 2021-09-01
fix(Modal): fix bug with detached footer

silvenon:
When Modal footer contains transitioning elements which affect footer's height, for example a Button switching size, it can cause Modal to incorrectly calculate footer's height, so now we're recomputing this layout stuff every time any element inside footer finishes transitioning.

 Storybook: https://orbit-silvenon-fix-modal-footer-detach.surge.sh


mainframev:
I think we can test it in cypress. Wdy? :)

PR #3094, opened 2021-09-02
ci: limit library size with size-limit

silvenon:
Replaces the action we use for measuring compressed size with size-action.

- [x] #3158.

 Storybook: https://orbit-silvenon-ci-size-limit.surge.sh


silvenon:
Closing this because the existing GitHub action isn't good enough for us, it prints out every entry instead of just the ones whose size changed, so it's useless because nobody is going to read each row to find those few ones where size difference is not `0%`.

PR #3095, opened 2021-09-05
feat: update colour palette

mainframev:
Added new colors to the cloud palette according to the [task](https://jira.kiwi.com/browse/ORBIT-1820). Updated the colors of the ink palette. Changed property of icon from `secondary` to `tertiary` so it matches `inkLighter`. 

Too many visual changes, mostly because of the change in ink palette. In our codebase changed tokens from `paletteInkLighter` to `paletteCloudDarker`, because now this token has that value. Read more in the task decription.

Storybook: https://orbit-silvenon-feat-update-color-palette.surge.sh


silvenon (approved):
Don't forget to add `BREAKING CHANGE` to commit message when squashing.

PR #3099, opened 2021-09-09
Add new features to locking scrolling

silvenon:
- Nano Stores + Router plugin is tiny in comparison to React Router, so I thought I could add it to Cypress testing. Not that it matters much considering that we're not deploying it, I just wanted to try it out.

- Add `lockScrolling` theme flag, which can disable the functionality for locking scrolling, so that people can still use Orbit's latest without having to worry without running into bugs with it, as we weren't able to figure what the problem was yet. The flag is `true` by default.

- Allow specifying additional dependencies for `useLockScrolling` because sometimes the given scrollable ref changes value.

 Storybook: https://orbit-silvenon-feat-lock-scrolling-theme-flag.surge.sh


packages/orbit-components/src/hooks/useLockScrolling/index.js.flow
```diff
@@ -3,4 +3,5 @@
 declare export default function useLockScrolling(
   ref: {| current: HTMLElement | null |},
   lock?: boolean,
+  dependencies?: any[],
```
mainframev, line 6: i guess the alternative to `unknown` type in flow is `mixed` 🤔

mainframev (approved):

PR #3100, opened 2021-09-10
fix(CountryFlag): set "undefined" as default code

silvenon:
BREAKING CHANGE: The default code for CountryFlag is now `"undefined"` rather than `"anywhere"`, a gray flag with a question mark. This code is also used if the given code isn't supported.

This was meant to go in v1, but I forgot to repeat the change after reverting it, so it's scheduled for v2.

 Storybook: https://orbit-silvenon-fix-country-flag-code-undefined.surge.sh


docs/src/__examples__/CountryFlag/DEFAULT.tsx
```diff
@@ -3,6 +3,11 @@ import { CountryFlag } from "@kiwicom/orbit-components";
 
 export default {
   Example: () => <CountryFlag />,
+  info: {
```
mainframev, line 6: I removed all `info` in examples, since we do not need them anymore, could you remove?

mainframev (approved):

PR #3102, opened 2021-09-12
refactor(TooltipErrorForm): replace Tooltip with Popper

mainframev:
Replaced previous implementation of tooltip inside `TooltipErrorForm` with `Popper`
 Storybook: https://orbit-silvenon-fix-error-forms.surge.sh


packages/orbit-components/src/InputField/index.js
```diff
@@ -329,10 +331,10 @@ const StyledIconWrapper = styled.span`
   display: flex;
 `;
 
-const InputField: React.AbstractComponent<Props, HTMLInputElement> = React.forwardRef<
+const InputField: React.AbstractComponent<
   Props,
-  HTMLInputElement,
->((props, ref) => {
+  React.AbstractComponent<Props, HTMLInputElement>,
+> = React.forwardRef((props, ref) => {
```
silvenon, line 337: I'm not sure if this is right, it should probably be just `React.AbstractComponent<Props, HTMLInputElement>`, Flow doesn't seem to complain. You can simply import the type exported by `InputField/index.js.flow`:

```js
import typeof InputFieldType from ".";

// ...

const InputField: InputFieldType = // ...
```

That way you also check whether the library definition is good.

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -176,26 +222,27 @@ const ErrorFormTooltip = ({
...
-      shown={shown && dimensions.set}
+      shown={shown}
       isHelp={isHelp}
       data-test={dataTest}
-      inlineLabel={inlineLabel}
       aria-live="polite"
-      id={id}
-      {...dimensions}
+      placement={attrs.popper && attrs.popper["data-popper-placement"]}
+      {...popper}
```
silvenon, line 242: Be explicit instead so that you can remove `$FlowFixMe` because you already know what you need.
```suggestion
      top={popper.top}
      left={popper.left}
      position={popper.position}
      bottom={popper.bottom}
      right={popper.right}
      transform={popper.transform}
```

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -176,26 +222,27 @@ const ErrorFormTooltip = ({
...
     >
+      {/* $FlowFixMe: https://github.com/popperjs/react-popper/issues/318 */}
+      <StyledArrow isHelp={isHelp} ref={setArrowRef} {...arrow} />
```
silvenon, line 245: Same here. You might also be able to use some lodash function for this if you don't want to be repetitive.

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -12,20 +13,48 @@ import { Item } from "../../List/ListItem";
...
-import { POSITIONS, SIDE_NUDGE } from "./consts";
+import resolvePlacement from "./helpers/resolvePlacement";
+import { SIDE_NUDGE } from "./consts";
+import useTheme from "../../hooks/useTheme";
 
 import type { Props } from ".";
 
+const StyledArrow = styled(
+  React.forwardRef(({ className }, ref) => <div className={className} ref={ref} />),
+)`
```
silvenon, line 24: We should probably start using `$*` props instead of this. (Not now.)

silvenon (changes requested):
Positioning is broken now.

Screen Shot 2021-09-13 at 11 39 21

Ignore the broken UI for viewports, it's a [different issue](https://jira.kiwi.com/browse/ORBIT-1899) caused by hard refresh.

packages/orbit-components/src/InputField/index.js
mainframev: it fails for, types are correct, its actually two times forwarding refs, so the target element in the first type is not `HTMLInputElement` but another `AbstractComponent`

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
mainframev: lodash is not the part of the orbit, it should be used only in orbit.kiwi.

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
mainframev: yep, agree

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -131,40 +157,61 @@ StyledCloseButton.defaultProps = {
...
+    return [rtl ? 10 : -10, 7];
+  };
+
+  const { styles, attributes: attrs, update } = usePopper(referenceElement?.current, tooltipRef, {
+    placement: rtl ? "top-end" : "top-start",
+    modifiers: [
+      {
+        name: "offset",
+        options: {
+          offset: resolveOffset(),
```
silvenon, line 188: It would probably be a good idea to call `update` when the result of this function changes, in that case you'd have to memoize `resolveOffset`:

```js
const resolveOffset = React.useCallback(() => {
  if (inlineLabel) {
    if (inputSize === "small") return [rtl ? 10 : -14, 7];
    return [rtl ? 6 : -6, 6];
  }
  return [rtl ? 10 : -10, 7];
}, [inlineLabel, inputSize, rtl]);

// `usePopper()` call

React.useEffect(() => {
  if (update) update();
}, [update, resolveOffset);
```

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -176,26 +223,40 @@ const ErrorFormTooltip = ({
         onShown(false);
       }
     };
-
     window.addEventListener("keydown", handleTab);
     return () => {
       window.removeEventListener("keydown", handleTab);
     };
-  }, [onShown, isHelp]);
+  }, [onShown, isHelp, shown, tooltipRef, update]);
```
silvenon, line 230: If you move the call to `update()` in the new `React.useEffect` I suggested above, adding `shown` to dependencies there, you can remove `shown` from here:

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
```diff
@@ -131,40 +157,61 @@ StyledCloseButton.defaultProps = {
...
-  const tooltipRef = React.useRef<HTMLElement | null>(null);
-
-  const dimensions = useDimensions({
-    labelRef,
-    iconRef,
-    contentRef: tooltipRef,
-    children,
-    inlineLabel,
+  const { rtl } = useTheme();
+  const [tooltipRef, setTooltipRef] = React.useState(null);
```
silvenon, line 171: Why did you go for `useState` instead of `useRef` here? That's only needed for stuff like measuring.

silvenon (changes requested):
Ok, that's about it. Just for curiosity I tried enabling scroll listener to see what overflow modifier would look like, but it's off. Considering that we're not offering this functionality yet, it's fine, but it tells me that arrow might not be positioned/marked correctly.

https://user-images.githubusercontent.com/471278/133266887-a894c701-ac9d-45d3-8f56-d24c28d676e3.mov

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.js
mainframev: official docs are recommend to pass reference elements like that

mainframev: and it does not work properly with regular refs

PR #3103, opened 2021-09-13
fix(InputField): aria-labelledby should be null if label is provided

RobinCsl:
When a label is provided, the current implementation provides `aria-labelledby="false"` to the underlying `<input />` field, making it non-labellable by the provided label.

I think it would be worth it to add a relevant test to prevent future regressions.
 Storybook: https://orbit-silvenon-rcsl-inputfield-fix.surge.sh


silvenon (changes requested):
Thanks for spotting this! I would expect `screen.getByRole("textbox", { name: /Label/ })` to detect this this, maybe it's a bug with `@testing-library/dom` so I raised an issue: testing-library/dom-testing-library#1030. Would you mind adding a simple `ByLabelText` in the `should have expected DOM output` test? 🙏 

```js
// https://github.com/testing-library/dom-testing-library/issues/1030
screen.getByLabelText(/Label/);
```

silvenon (approved):

PR #3107, opened 2021-09-15
refactor(docs): do/donts guidelines

mainframev:


docs/src/components/HeadingWithLink.tsx
```diff
@@ -27,17 +46,14 @@ interface Props extends SpaceAfter {
...
-        <ButtonLink
-          iconLeft={<LinkIcon />}
-          title={`Link to heading: ${headingText}`}
-          href={`#${slugifiedText}`}
-          type="secondary"
-          compact
-        />
+        <StyledAnchor href={`#${slugifiedText}`} title={`Link to heading: ${headingText}`}>
+          <LinkIcon />
+        </StyledAnchor>
```
silvenon, line 56: It's better to do it right considering that you're refactoring this. I think `StyledAnchorWrapper` is the one that needs to be an anchor, not `StyledAnchor` (obviously you'd name them differently then), because elements that aren't focusable should not be hoverable. You can see that tabbing through these anchors is not good because they are invisible.

docs/src/components/HeadingWithLink.tsx
```diff
@@ -1,24 +1,43 @@
...
+    display: inline-block;
+    padding: 10px;
+    margin: -10px;
+    margin-left: 1px;
+    :hover,
+    :active,
+    :focus {
+      outline: none;
+      svg {
+        transition: fill ${theme.orbit.durationFast} ease-in;
```
silvenon, line 34: I'd prefer transitioning both ways. Transitioning only on hover makes the fill change instantaneously when mouse is moved away.

docs/src/components/HeadingWithLink.tsx
```diff
@@ -1,24 +1,43 @@
...
-  margin-top: -10px;
-
   svg {
     opacity: 0;
-    transition: opacity ${({ theme }) => theme.orbit.durationNormal} ease-in;
   }
 
   &:hover svg {
     opacity: 1;
   }
```
silvenon, line 16: This can be moved in `StyledAnchor`, so styles there would look something like:

```
svg {
  opacity: 0;
}

${StyledAnchorWrapper} {
  &:hover, &:focus {
    svg {
      opacity: 1;
    }
  }
}
```

docs/src/components/Guidelines/index.tsx
```diff
@@ -1,161 +1,146 @@
...
+    padding: ${theme.orbit.spaceXSmall} ${theme.orbit.spaceLarge} ${theme.orbit.spaceXSmall}
+      ${theme.orbit.spaceMedium};
   `}
 `;
 
-const ImageBorder = styled.div<GuidelineType>`
+const StyledImageContainer = styled.div<ImageContainerProps>`
   ${({ theme, type }) => css`
-    border-left: 4px solid
+    padding: 10%;
```
silvenon, line 58: This padding is a little much vertically on smaller resolutions:

Screen Shot 2021-09-15 at 18 17 20

I think it should stay `spaceMedium` and you could add a maximum width of the image if you can, and center it.

docs/src/components/Guidelines/index.tsx
```diff
@@ -1,161 +1,146 @@
...
-import { imageWrapperClass } from "gatsby-remark-images/constants";
+import {
+  Stack,
+  Text,
+  Heading,
+  Grid,
+  useMediaQuery,
+  mediaQueries as mq,
+} from "@kiwicom/orbit-components";
+import useTheme from "@kiwicom/orbit-components/lib/hooks/useTheme";
```
silvenon, line 12: Every time you see something that isn't being exported from the `@kiwicom/orbit-components` root please instead of importing from `lib` add that export (in a separate commit) and use it 🙏

docs/src/components/Guidelines/helpers.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const resolveBorders = () => {
+  return css`
+    ${setBorders(true)};
+    ${mq.desktop(css`
+      border-top: none;
+      ${setBorders(false)};
+    `)}
+  `;
+};
```
silvenon, line 51: `setBorders` introduces complexity, you don't need it.
```suggestion
export const resolveBorders = ({ theme, type, coloredBorder }) => {
  let borderColor: string;
  if (coloredBorder) {
    if (type === "do") {
      borderColor = theme.orbit.paletteGreenNormal;
    }
    if (type === "dont") {
      borderColor = theme.orbit.paletteRedNormal;
    }
  } else {
    borderColor = theme.orbit.paletteInkLight;
  }
  return css`
    border-top: 3px solid ${borderColor};
    ${mq.desktop(css`
      border-top: none;
      border-left: 3px solid ${borderColor};
    `)}
  `;
};
```

silvenon (changes requested):

docs/src/components/Guidelines/helpers.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const resolveColumns = (images, isDesktop) => {
+  if (isDesktop) {
+    if (images === 1) return `1fr 1fr`;
+    if (images === 2) return `1fr 1fr 1fr`;
+
+    return `1fr`;
+  }
+
+  return `1fr`;
```
silvenon, line 61: I think this also works:
```suggestion
 return isDesktop ? `repeat(${images + 1}, 1fr)` : "1fr";
```

docs/src/components/Guidelines/helpers.tsx
mainframev: not agree. I prefer a declarative way, rather than imperative with let. In my opinion, your suggestion looks more complex (edited)

docs/src/components/HeadingWithLink.tsx
mainframev: nope, the anchor should be only for that hidden anchor, the tile should not be inside `a` tag. I doubt that they should be accessible via tab too. Check for instance [here](https://leerob.io/blog/style-guides-component-libraries-design-systems#what-is-a-style-guide). It should appear on heading hover, not the link, i will fix that.

mainframev: nope, only icon should be wrapped in anchor, if i got u right.

mainframev: I think the best would be to have in the way like Lee R has in his [blog](https://leerob.io/blog/monorepo-lerna-yarn-workspaces). I know that we don't use the rehype or remark plugin for autolink-headings, but our solution is quite similar

silvenon (changes requested):
All good except the images are still huge on single column breakpoint. Is there no way to limit image width to their natural width?

silvenon (approved):

PR #3109, opened 2021-09-19
feat: add component structure component

silvenon:
The logic might be a little messy, I'm open to suggestions, but I added thorough tests so it should be good.

 Storybook: https://orbit-silvenon-feat-component-structure.surge.sh


silvenon:
Now tabs don't show up if there is only one.

docs/src/components/ComponentStructure/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+        &::before {
+          flex-shrink: 0;
+          content: counter(component-parts);
+          display: grid;
+          place-content: center;
+          width: 24px;
+          height: 24px;
+          background: #ec5192;
+          border-radius: 100%;
+          color: white;
```
mainframev, line 85:
```suggestion
          color: ${({theme}) => theme.orbit.paletteWhite};
```

docs/src/components/ComponentStructure/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+    display: grid;
+    place-content: center;
+    padding: ${theme.orbit.spaceXXLarge};
+  `};
+`;
+StyledImageContainer.defaultProps = { theme: docsTheme };
+
+const StyledComponentParts = styled.div`
+  ${({ theme }) => css`
+    display: grid;
```
mainframev, line 54: I think you don't need that

mainframev (changes requested):

docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
```diff
@@ -27,9 +29,29 @@ check out our [interactive guide on action components](/guides/using-button-type
 
 <ComponentStatus component="Button" />
 
-## Content structure
-
-![IconLeft: visually supports button visibility; label: describes the main action and works best when short; icon right: shows additional options for the button (expandable, link, etc.).](fileId:4QJZqvBvRrLu6t9mwObCkA;nodeId:289%3A14)
+## Component structure
+
+<ComponentStructure
```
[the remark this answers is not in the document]
silvenon: I was thinking about this as well, but:

1. this kind of API should only be if WebStructure renders at exactly that place, which is not true, I want to always render them in the same order for consistency

2. I’d need to do some React.Children kung-fu in order to extract info from these children, and generally the containing component shouldn’t have to know anything about its children

docs/src/components/ComponentStructure/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+  component: string;
+  web?: Structure;
+  ios?: Structure;
+  android?: Structure;
+}
+
+export default function ComponentStructure({ component, web, ios, android }: Props) {
+  const seed = useRandomIdSeed();
+
+  const tabsCount = (web ? 1 : 0) + (ios ? 1 : 0) + (android ? 1 : 0);
```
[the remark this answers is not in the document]
silvenon: I really don't like JavaScript's type coercion. In your suggestion I'd be coercing it _twice_, an object to a boolean to a number. I prefer not to do that just so I can save a little space, I hope you don't mind.

docs/src/components/ComponentStructure/index.tsx
silvenon: It's useful in vertically centering the list of parts when the example is taller than the list of parts.

mainframev (approved):

PR #3111, opened 2021-09-21
docs: little fixes

mainframev:
- fixed search button on the main page
- fixed developer mode in overview pages (worked only via navigation items)
- fixed bookmarks (did not work on the main page)
- fixed heading size of `when to use` / `when not to use` (it should have other changes, but we will implement it later, since it should be as component, now it just a mdx)
- adjusted search button in the navbar to the latest design 

 Storybook: https://orbit-silvenon-docs-lil-fixes.surge.sh

### DIFF
--- a/docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
+++ b/docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
@@ -16,7 +16,7 @@ check out our [interactive guide on action components](/guides/using-button-type
 
 - To offer the most important actions users can take.
 
-### When not to use
+## When not to use
 
 - When you have many actions at once---use a [button link](/components/action/buttonlink/).
 - You want to offer a less important action---use a [button link](/components/action/buttonlink/).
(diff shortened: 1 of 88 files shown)


docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
```diff
@@ -16,7 +16,7 @@ check out our [interactive guide on action components](/guides/using-button-type
 
 - To offer the most important actions users can take.
 
-### When not to use
+## When not to use
```
silvenon, line 19: I'm pretty sure all of these were intentional. 🤷  Because this can be treated as related to "When to use". Unless you're sure about this change, I wouldn't do it.

silvenon (changes requested):
Great changes otherwise!

docs/src/documentation/03-components/01-action/button/01-guidelines.mdx
silvenon: Confirmed that this changed was requested.

CollierCZ: :crying_cat_face: A loss of semantics.

silvenon (approved):

PR #3108, opened 2021-09-17
docs: add more knobs content

mainframev:

PR #3113, opened 2021-09-22
docs(sandbox): variants feature

mainframev:
Currently depends on #3108. Had to merge that branch here. I'll rebase it later after your review. Added variants only for button now. 
 Storybook: https://orbit-silvenon-docs-variants.surge.sh

### DIFF
--- a/docs/src/components/ReactExample/components/Board.tsx
+++ b/docs/src/components/ReactExample/components/Board.tsx
@@ -10,11 +10,13 @@ import {
 import { ChevronUp, ChevronDown } from "@kiwicom/orbit-components/icons";
 import styled, { css } from "styled-components";
 
+import Variants from "./Variants";
 import DarkMode from "../../../images/darkmode.svg";
 import NewWindow from "../../../images/new-window.svg";
 import Copy from "../../../images/copy.svg";
 import useCopyToClipboard from "../../../hooks/useCopyToClipboard";
 import { BgType } from "..";
+import { Variant } from "../Example";
 
 const StyledBoard = styled.div<{ isOpened: boolean }>`
   ${({ theme, isOpened }) => css`
@@ -33,10 +35,13 @@ interface Props {
   isFullPage?: boolean;
   isPlaygroundOpened: boolean;
   isEditorOpened: boolean;
+  isVariantsOpened: boolean;
   knobs: boolean;
+  variants: Variant[];
   background: BgType;
   onSelectBackground: (value: BgType) => void;
   onOpenPlayground: () => void;
+  onChangeVariant: (code: string) => void;
   onOpenEditor: () => void;
 }
 
@@ -45,9 +50,12 @@ const Board = ({
   exampleId,
   background,
   code,
+  variants,
+  isVariantsOpened,
   isEditorOpened,
   isPlaygroundOpened,
   onOpenPlayground,
+  onChangeVariant,
   onOpenEditor,
   onSelectBackground,
   knobs,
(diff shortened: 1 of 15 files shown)


docs/src/components/ReactExample/components/Board.tsx
```diff
@@ -96,6 +104,18 @@ const Board = ({
...
+          {variants.length > 0 && (
+            <Popover content={<Variants exampleVariants={variants} onChange={onChangeVariant} />}>
+              <ButtonLink
+                type="secondary"
+                asComponent="button"
+                ariaExpanded={isVariantsOpened}
+                iconRight={isVariantsOpened ? <ChevronUp /> : <ChevronDown />}
+              >
+                Variants
+              </ButtonLink>
```
silvenon, line 116: Clicking on this button throws off the scrolling position. Probably should have been a Button, not ButtonLink, no?

docs/src/components/ReactExample/components/Board.tsx
```diff
@@ -96,6 +104,18 @@ const Board = ({
               Playground
             </ButtonLink>
           )}
+          {variants.length > 0 && (
+            <Popover content={<Variants exampleVariants={variants} onChange={onChangeVariant} />}>
```
silvenon, line 108: It would be better if the current variant was already selected when opening this, would this be hard to do?

silvenon (changes requested):
Cool feature!

I noticed that it doesn't look quite like in the design, borders are missing:

Screen Shot 2021-09-24 at 11 58 48

docs/src/components/ReactExample/components/Board.tsx
mainframev: nope, fixed that already, should be a buttonlink visually, but `asComponent="button"`

docs/src/components/ReactExample/components/Board.tsx
```diff
@@ -96,6 +104,18 @@ const Board = ({
...
+              <ButtonLink
+                type="secondary"
+                asComponent="button"
```
silvenon, line 111: For me the scroll position still changes on click, not sure why because now it's just a button.

silvenon (changes requested):
1. Now the first radio in Popover is selected, and when I select another one, and reopen Popover, again the first one is selected, which is not the current one. That's why I was afraid that it might be too difficult to implement now.

2. Ideally this particular Popover shouldn't flip because it covers up the example. Did we add the feature to Popover to turn this off?

docs/src/components/ReactExample/components/Board.tsx
mainframev: weird, that it does not happen locally

silvenon: But it happens for you on staging, right? For me it happens locally as well, maybe you need to rebuild components. 🤷  Something in Popover is causing this, not sure what, because I'm getting the same behavior for dark mode button, and regardless of whether the component is ButtonLink, Button, regular `button` or anything else.

It doesn't throw my scrolling off all the way, just part of the way, and the amount seems inconsistent. There's also a very uncomfortable delay between clicking and actually opening the popover, maybe this is related, and it's something that was always bothering me about Popover.

Also, probably unrelated, but see if you can do something about these Poppper warnings, I created a [ticket](https://jira.kiwi.com/browse/ORBIT-1918) for this.

Screen Shot 2021-09-26 at 14 52 09

mainframev: yep, started to have that after the rebuild. Looks like refactor to Popper caused that

silvenon (approved):
I pushed a fix for that one issue with not retaining the selected variant after menu closes.

PR #3114, opened 2021-09-23
feat: add component status component

silvenon:


mainframev:
I guess on mobile it does not look good. I have to scroll a lot, plus it's too narrow :/ 

Can we change a layout on mobile? The row with the name will be on the top and component statuses will be below. 🤔

mainframev (approved):

PR #3115, opened 2021-09-24
feat: update Button page

silvenon:
I needed to temporarily stop remark from failing when checking local links because some headings are inside components, and there doesn't appear to be a way to disable remark-validate-links with `<!-- lint disable validate-links -->` comments because `remark-validate-links` is a remark plugin, not remark-lint. remarkjs/remark-validate-links#18

 Storybook: https://orbit-silvenon-feat-button-page-updates.surge.sh


docs/src/components/ReactExample/Example.tsx
```diff
@@ -18,14 +18,21 @@ const StyledWrapper = styled.div<{ isFullPage?: boolean }>`
...
   `};
 `;
 
-const StyledWrapperFrame = styled.div<{ width: number }>`
+const StyledWrapperFrame = styled.div<{ width: number | null }>`
   ${({ width }) => css`
     margin: 0 auto;
-    max-width: ${width}px;
     width: 100%;
+    ${width === null
```
mainframev, line 29: looks like it did not help, it's still broken after the hard refresh

docs/src/components/ReactExample/index.tsx
```diff
@@ -69,6 +76,7 @@ const ReactExample = ({ exampleId, background = "white", minHeight, maxHeight }:
 
   return (
     <Example
+      responsive={responsive}
```
mainframev, line 79: do not forget also to pass responsive prop in `docs/src/template/Sandbox` ;)

mainframev (changes requested):

mainframev:
Also, spacing between elements on the page should be bigger. For instance, currently betwee structure and playground spacing is 24px, sholuld be 32px ;)

docs/src/components/ReactExample/index.tsx
silvenon: It shouldn't be necessary because on full app viewport ruler should always be there, and in `Example` `responsive` is `true` by default.

docs/src/components/ReactExample/Example.tsx
silvenon: Fixed in another PR.

PR #3116, opened 2021-09-25
fix(InputField): label before tags and spacing

mainframev:
- [Fixes slack request. ](https://skypicker.slack.com/archives/CAMS40F7B/p1632485533109900). Plus there was a bug with too small spacing in normal inputs (not in error forms). 
- Fixed type definition in the `Popover`
 Storybook: https://orbit-silvenon-fix-small-fixes.surge.sh

### DIFF
--- a/packages/orbit-components/src/InputField/InputTags/index.js
+++ b/packages/orbit-components/src/InputField/InputTags/index.js
@@ -18,7 +18,6 @@ const StyledInputTags = styled.div`
   z-index: 2;
   min-width: 50px;
   overflow: hidden;
-  order: -1;
 `;
 
 // $FlowFixMe: https://github.com/flow-typed/flow-typed/issues/3653#issuecomment-568539198
--- a/packages/orbit-components/src/InputField/index.js
+++ b/packages/orbit-components/src/InputField/index.js
@@ -148,7 +148,10 @@ const StyledInlineLabel = styled.div`
   align-items: center;
   pointer-events: none;
   justify-content: center;
-  padding: ${({ theme }) => rtlSpacing(`0 0 0 ${theme.orbit.spaceXXSmall}`)};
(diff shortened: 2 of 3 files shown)


packages/orbit-components/src/InputField/InputTags/index.js
```diff
@@ -18,7 +18,6 @@ const StyledInputTags = styled.div`
   z-index: 2;
   min-width: 50px;
   overflow: hidden;
-  order: -1;
```
silvenon, line 21: Do you remember why you put this there in the first place?

packages/orbit-components/src/InputField/index.js
```diff
@@ -434,7 +435,7 @@ const InputField: InputFieldType = React.forwardRef((props, ref) => {
             prefix && <Prefix size={size}>{prefix}</Prefix>
           )}
           {label && inlineLabel && (
-            <StyledInlineLabel ref={labelRef} size={size}>
+            <StyledInlineLabel isErrorForms={!tags && (error || help)} ref={labelRef} size={size}>
```
silvenon, line 438: "Error forms" is an internal name of the feature that doesn't really describe what it is, instead I recommend adding two boolean props that have a clear meaning and derive styles from those.
```suggestion
            <StyledInlineLabel hasTags={!!tags} hasFeedback={!!(error || help)} ref={labelRef} size={size}>
```

silvenon (changes requested):

packages/orbit-components/src/InputField/InputTags/index.js
mainframev: not sure, it's too old. Looks like legacy thing from the previous PRs

silvenon (approved):

PR #3117, opened 2021-09-26
refactor: rewrite viewport ruler

silvenon:
There were multiple problems with the viewport ruler, the biggest one being that it was computing container width too soon, so the initial width ended up being 0, making the ruler unusable. This refactor uses ResizeObserver instead of using the `resize` event of the entire window, and fixes some other bugs along the way.

 Storybook: https://orbit-silvenon-fix-sandbox-viewport-ruler.surge.sh

### DIFF
--- a/docs/src/components/ReactExample/components/ViewportsRuler.tsx
+++ b/docs/src/components/ReactExample/components/ViewportsRuler.tsx
@@ -1,11 +1,12 @@
 import React from "react";
 import styled, { css } from "styled-components";
-import useMediaQuery from "@kiwicom/orbit-components/lib/hooks/useMediaQuery";
+import { useMediaQuery, Text } from "@kiwicom/orbit-components";
+import useResizeObserver from "use-resize-observer";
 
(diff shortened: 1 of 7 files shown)


docs/src/components/ReactExample/components/ViewportsRuler.tsx
```diff
@@ -51,62 +50,70 @@ const StyledCell = styled(({ className, children, onClick, onMouseEnter }) => (
...
+              onMouseEnter={() => setTargetViewport(viewport)}
+              onMouseLeave={() => setTargetViewport(null)}
+              onClick={() => setActiveViewport(viewport)}
+            />
+          ))}
+          <StyledLabel>
+            <Text as="div">
+              {targetViewport
+                ? `${targetViewport} (${QUERIES[targetViewport]})`
+                : `${activeViewport} (${QUERIES[activeViewport]})`}
```
mainframev, line 110: in one of the latests PRs i added pixels. Could you add that? so it's not just 320, but `320px` and etc.

mainframev (approved):

PR #3122, opened 2021-09-29
docs: more variants to sandbox examples and small fixes

mainframev:


docs/src/documentation/03-components/05-input/inputfield/01-guidelines.mdx
```diff
@@ -52,15 +52,11 @@ making it easier for users to complete the field.
 
 <!-- eslint-disable -->
 
-#### Help message
+#### Error and Help messages
```
silvenon, line 55: In InputFile the order is opposite, "Help and Error messages", so I recommend being consistent.

docs/src/__examples__/InputGroup/STATES.tsx
```diff
@@ -0,0 +1,229 @@
...
+          error="This text is not displayed"
+          placeholder="YYYY"
+          value="2020
+          type="number"
+        />
+      </InputGroup>
+    );`,
+    },
+    {
+      name: "Help",
```
silvenon, line 155: In this variant the error message remains, I assume it's because it's being merged with the main example, which has `error`, so error wins.

docs/src/__examples__/Loading/TYPES.tsx
```diff
@@ -1,63 +1,28 @@
...
+      code: `() => <Loading type="boxLoader">Useful when content in boxes like cards and card sections is being loaded.</Loading>`,
+    },
+    {
+      name: "pageLoader",
+      code: `() => <Loading>Useful when an entire page is loading.</Loading>`,
+    },
+    {
+      name: "InlineLoader",
+      code: `() => <Loading type="inlineLoader">Useful when content in paragraphs and other inline elements is being loaded.</Loading>`,
+    },
```
silvenon, line 26: The prop `loading` is false in these variants, so only the text appears, which I assume isn't what you wanted.

silvenon (changes requested):

docs/src/__examples__/Loading/TYPES.tsx
mainframev: Yep, it's because I've added variants, checked that, and added text after. It does not change on hot reload, so i did not check it after.

docs/src/__examples__/InputGroup/STATES.tsx
mainframev: nope, it's because i forgot to add `() =>`, it's strings, so it's easy to miss such thing :/

silvenon:
Assuming that the fact you resolved my comments means that you fixed them, InputGroup still has the same problem, and in Loader selecting the ButtonLoader variant doesn't appear to show anything.

mainframev:
> Assuming that the fact you resolved my comments means that you fixed them, InputGroup still has the same problem, and in Loader selecting the ButtonLoader variant doesn't appear to show anything.

button does not have to show something, i decided to remove that, others have

PR #3126, opened 2021-09-30
improve: split dependencies across workspaces

silvenon:
So far we've been relying on hoisting to avoid having to specify some dependencies multiple times because we use them in multiple workspaces.  In Yarn v3 this will no longer work, so we're strictly assigning dependencies to every workspace that uses them, and we're enforcing this by modifying the `import/no-extraneous-dependencies` ESLint rule.


docs/src/components/MasonryLayout.tsx
```diff
@@ -29,7 +29,7 @@ const MasonryLayout = ({ columns = 2, gap = 24, children }: Props) => {
   const masonryColumns = isTablet ? columns : 1;
 
   const columnWrapper: Record<string, Array<React.ReactNode>> = {};
-  React.Children.forEach(children, (child, index) => {
+  React.Children.toArray(children).forEach((child, index) => {
```
silvenon, line 32: TypeScript was failing here because `child` was `unknown`.

package.json
```diff
@@ -39,67 +39,30 @@
   },
   "author": "kiwi.com",
   "license": "MIT",
+  "resolutions": {
+    "@types/react": "^17.0.26",
+    "@types/react-dom": "^17.0.9"
+  },
```
silvenon, line 45: Multiple versions of `@types/react` and `@types/react-dom` caused problems similar to the ones we had with `@types/react-native`, i.e. a bunch of errors inside `@types/react` and `@types/react-dom` complaining that modules were redeclared with different types.

packages/orbit-components/package.json
```diff
@@ -83,7 +83,85 @@
...
+    "@babel/plugin-transform-runtime": "^7.15.0",
+    "@babel/preset-env": "^7.14.0",
+    "@babel/preset-flow": "^7.12.13",
+    "@babel/preset-react": "^7.12.13",
+    "@babel/preset-typescript": "^7.13.0",
+    "@babel/runtime": "^7.11.2",
+    "@babel/types": "^7.13.0",
+    "@emotion/core": "^11.0.0",
+    "@kiwicom/browserslist-config": "^1.0.0",
+    "@kiwicom/orbit-components": "file:.",
```
silvenon, line 104: I'm a bit nervous about self-referencing `@kiwicom/orbit-components` like this, the goal was to make ESLint accept `@kiwicom/orbit-components` imports in `__examples__` and `__testfixtures__` work. I checked and this doesn't cause an infinite nesting nesting within `packages/orbit-components/node_modules/@kiwicom/orbit-components`, it only copies necessary files.


This is `devDependencies` so I doubt something can go wrong, and it's only temporary anyway, we'll eventually move examples to `docs`, and we can move codemods to a separate workspace.

docs/plugins/contributors/gatsby-node.js
```diff
@@ -93,22 +116,10 @@ exports.sourceNodes = async ({ actions, createContentDigest, createNodeId }, { r
...
-          content: JSON.stringify(user),
-          contentDigest: createContentDigest(user),
-        },
-      });
-    });
     console.error(error);
-    warnMissingAccessToken(error);
+    console.info(
+      "You may have forgotten to include the repo and read:org scopes for your GitHub access token.",
+    );
```
silvenon, line 122: Clumsily adapting this logic to `dotenv-safe` instead of `dotenv`. We'll come up with better dotenv utilities later.

PR #3128, opened 2021-10-01
sandbox(knobs): object type knob

mainframev:


docs/src/components/DocLayout/index.tsx
```diff
@@ -158,7 +168,18 @@ export default function DocLayout({
                         {tabs && <Tabs activeTab={location.pathname} tabs={tabs} />}
                         {headerLink && (
                           <Hide on={["smallMobile", "mediumMobile"]}>
-                            <HeaderButtonLink href={headerLink} />
+                            <Stack flex spacing="small">
```
silvenon, line 171: The spacing is too big, I'd remove it completely because `HeaderButtonLink` already provides a lot of spacing. Also the label "Storybook" might be useful next to the Storybook icon. I'd ask @williamkolmacka for guidance here.

docs/src/components/ReactExample/__tests__/transform.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        },
+      ),
+    ).toMatchInlineSnapshot(`
+      "() => <Component iconRight={<Visibility />} options={[{
+        \\"kek\\": \\"bur\\",
+        \\"blin\\": \\"cheburek\\"
+      }]} disabled fullWidth title=\\"kek\\">children</Component>;"
+    `);
+  });
+});
```
silvenon, line 80: Awesome that you added tests for this!

docs/src/components/ReactExample/components/Playground.tsx
```diff
@@ -71,90 +71,86 @@ const Playground = ({ exampleKnobs, onChange }: Props) => {
             </Text>
             <StyledKnobsWrapper>
               {sortBy(knobs, ["type"]).map(({ type, name, options }) => {
-                const fieldName = `${component}-${name}`;
+                const baseKnobProps = {
+                  key: name,
```
silvenon, line 75: I don't think `key` can be spread, see the [`jsx-key`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) rule, they must be used directly. Although I'm a little confused that there apparently aren't any warnings, but maybe React removes these warnings in production.

silvenon (changes requested):

docs/src/components/DocLayout/index.tsx
mainframev: do we really need that padding inside `HeaderLink`?  it looks better without.

PR #3129, opened 2021-10-04
refactor(TooltipPrimitive): use Popper instead of custom logic

mainframev:
BREAKING CHANGE:
`preferredAlign` and `preferredPosition` were replaced by Placement prop
new props: `noFlip`, `offset`

 Storybook: https://orbit-silvenon-refactor-tooltip-to-popper.surge.sh


packages/orbit-components/src/primitives/TooltipPrimitive/README.md
```diff
@@ -18,29 +18,42 @@ After adding import into your project you can use it simply like:
...
+| `"top-end"`      |
+| `"right-start"`  |
+| `"right-end"`    |
+| `"bottom-start"` |
+| `"bottom-end"`   |
+| `"left-start"`   |
+| `"left-end"`     |
+| `"auto"`         |
+| `"auto-start"`   |
+| `"auto-end"`     |
```
silvenon, line 55: Tooltip readme should also be updated.

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.js
```diff
@@ -27,50 +20,56 @@ const StyledTooltip = styled.div`
...
-    width: 0;
-    height: 0;
-    border-style: solid;
-    content: " ";
-    display: block;
-    position: absolute;
+const StyledTooltipArrow = styled(
+  React.forwardRef(({ className }, ref) => <div className={className} ref={ref} />),
+)`
+  ${({ transform, position }) => css`
```
silvenon, line 26: It would be better to prefix custom props with `$` and avoid having to forward ref manually.
```suggestion
const StyledTooltipArrow = styled.div`
  ${({ $transform, $position }) => css`
```

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.js
```diff
@@ -113,69 +114,91 @@ const TooltipContent = ({
...
         onMouseEnter={onEnter}
         onMouseLeave={onClose}
+        contentHeight={contentHeight}
         onClick={handleInnerClick}
+        top={popper.top}
+        left={popper.left}
+        right={popper.right}
+        bottom={popper.bottom}
+        position={popper.position}
+        transform={popper.transform}
```
silvenon, line 188: Would this be ok? Not sure if it would have some performance regressions. 🤔  Same with arrow styles.
```suggestion
        $popper={popper}
```

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.js.flow
```diff
@@ -15,9 +16,9 @@ export type Props = {|
   onClose: () => void,
   onCloseMobile: () => void,
   onEnter: () => void,
-  preferredPosition: ?Positions,
-  preferredAlign: ?Aligns,
-  containerRef: React.ElementRef<*>,
+  placement?: Placement,
```
silvenon, line 19: I think `noFlip` is missing, but I'm not sure how Flow doesn't complain. 🤔

packages/orbit-components/src/Tooltip/index.js.flow
```diff
@@ -45,8 +46,7 @@ export type Props = {|
   +size?: Sizes,
   +renderInPortal?: boolean,
   +stopPropagation?: boolean,
-  +preferredPosition?: Positions,
-  +preferredAlign?: Aligns,
+  +placement?: Placement,
```
silvenon, line 49: I think `noFlip` is missing.

packages/orbit-components/src/primitives/TooltipPrimitive/index.js.flow
```diff
@@ -47,14 +22,14 @@ export type Props = {|
...
-  +preferredPosition?: Positions,
-  +preferredAlign?: Aligns,
   +renderInPortal?: boolean,
   +tooltipShown?: boolean,
   +enabled?: boolean,
   +tabIndex?: string | number,
   +removeUnderlinedText?: boolean,
   +block?: boolean,
+  +offset?: [number, number],
+  +placement?: Placement,
```
silvenon, line 32: Also, probably missing `noFlip`.

silvenon (changes requested):

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.js.flow
mainframev: it does not complain because of the default value 😞

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.js
mainframev: not sure why with `$`, it can be just `popper={popper} than

silvenon (approved):
> not sure why with `$`, it can be just `popper={popper}` than 

I assume that `styled-components` would interpret `popper` as trying to set an HTML attribute, but if doesn't then you can omit the `$`.

PR #3143, opened 2021-10-14
fix(Modal): improve accessibility and testability

silvenon:
- Correctly label the Modal with its title instead of all of the content. This makes the Modal possible to find with the `ByLabelText` test matcher.

- Use the heading HTML element to mark up the Modal title, making it possible to find with singular `ByRole` matchers, e.g.  `getByRole("heading")`.

- Remove the fixed mobile title from the [accessibility tree](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion) to prevent screen readers from reading the title twice.

https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html

 Storybook: https://orbit-silvenon-fix-modal-title-a11y.surge.sh


mainframev:
@silvenon i think this change is not necessary at all. It makes Modal even more complex than it’s already is :smile:

PR #3161, opened 2021-10-26
improve/size limit esm

silvenon:
Use ESM to measure library size, the size ends up being significantly smaller, which indicates right away how much we can gain from deprecating the Babel plugin and moving all Orbit imports to ESM. But to make sure there are no sudden size changes like the Babel runtime one, I configured measuring Orbit's CommonJS build as well.

Other changes:

- export `useFocusTrap` from root
- export RTL utilities from root as `rlt`
- fix type definitions for InputStepper
- add library size to readme

 Storybook: https://orbit-silvenon-improve-size-limit-esm.surge.sh


silvenon:
Ok, this is ready. I updated the PR description.

packages/orbit-components/package.json
```diff
@@ -124,12 +124,12 @@
     "dotenv-safe": "^8.2.0",
     "fs-extra": "^9.0.1",
     "glob": "^7.1.2",
-    "globby": "^11.0.4",
     "isomorphic-unfetch": "^3.1.0",
     "jest-matchmedia-mock": "^1.1.0",
     "jest-styled-components": "^7.0.5",
     "jscodeshift": "^0.11.0",
     "jsdom": "^16.5.3",
+    "lodash": "^4.17.21",
```
mainframev, line 132: i think we should add also `babel-plugin-lodash` then

silvenon: Why? 🤔 `lodash` is used only for sorting in `.size-limit.js`.

mainframev: ah, right, so in this case i think it will be better to just have own functions and do not add dependency. Like take it from [here](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_sortby-and-_orderby), for instance

mainframev (approved):

PR #3170, opened 2021-11-02
refactor(Grid): drop IE support

mainframev:
Dropped the support of IE 11, since we stopped supporting it
 Storybook: https://orbit-silvenon-fix-grid-valeus.surge.sh


silvenon (approved):
👌

PR #3175, opened 2021-11-03
fix(useMediaQuery): optimize, support custom breakpoints, and make it work without context

silvenon:
In #3138 we moved the entire `matchMedia` logic to QueryContext, but this caused `useMediaQuery` to depend on the context, where before it was possible to use it independently. That breaking change was unnecessary, so this change restores this feature.

QueryContext was introduced in #1659 apparently to avoid FOUC, but at that time queries were being set after an unnecessary delay which was removed in #1982 (which also [deletes query object initialization](https://github.com/kiwicom/orbit/pull/1982/files#diff-838641df8a4076c2924fe7a116b270b31b765f584a6cc096b82a6de1a52e741dL64), perhaps accidentally), so the initial reaction to that was to remove QueryContext altogether, but there is actually still room for it as a way to optimize `useMediaqQuery` by reducing the number of renders because newly mounted components can use the context value instead of having to initialize the values again. Additional optimization was to keep referential equality in case breakpoints haven't changed instead of always creating a new object.

Lastly, now that QueryContext has been moved inside styled-component's ThemeProvider, `useMediaQuery` will actually respect the given `theme`. It used to always use the default theme because the value was being set outside StyledThemeProvider (this was a problem before #3138), so if people using Orbit changed the default breakpoints in ThemeProvider, they might run into very unexpected and hard to debug behavior while using `useMediaQuery`.

 Storybook: https://orbit-silvenon-fix-use-media-query-backwards-compat.surge.sh


packages/orbit-components/src/ThemeProvider/index.jsx
```diff
@@ -4,25 +4,25 @@ import { ThemeProvider as StyledThemeProvider } from "styled-components";
...
 const ThemeProvider = ({ theme, dictionary, children }: Props): React.Node => {
   return (
-    <QueryContextProvider>
-      <UIDReset>
-        <UIDFork>
-          <StyledThemeProvider theme={theme}>
+    <UIDReset>
+      <UIDFork>
+        <StyledThemeProvider theme={theme}>
+          <QueryContextProvider>
```
silvenon, line 16: I moved the query provider inside theme provider so that it has access to `theme`, before it was always using the default theme, oops!

packages/orbit-components/src/hooks/useMediaQuery/index.js
```diff
@@ -1,10 +1,89 @@
 // @flow
```
silvenon, line 1: I moved the logic from `QueryContextProvider` here because `useMediaQuery` needs to work both with and without it. And this is a better place for it anyway.

mainframev (approved):
LGTM

PR #3182, opened 2021-11-08
improve: upgrade to Gatsby v4

silvenon:
There's nothing in here that's actually taking advantage of new Gatsby yet, but no harm in upgrading it anyway, we can start using the features later.

Gatsby 3+ was throwing because we were using Babel packages on client side that didn't support running in the browser environment. We probably could have done some more webpack configuration magic to get around that, but instead we switched to `@babel/standalone`, which is meant to exactly for these uses cases and is simpler to use.

 Storybook: https://orbit-silvenon-improve-gatsby4.surge.sh


mainframev:
The problem with updating the babel is still persist, bundle size increased on 164kb

mainframev (commented):
update of adeira did not help :/

silvenon:
> update of adeira did not help :/

Yeah, that was not meant to fix the Babel problem, it was another problem. I'll let you know when I find the fix.

silvenon:
The change that is needed here is to get rid of `@babel/runtime`, because people using this library might be using a different version, which will render the size optimization useless.

PR #3187, opened 2021-11-10
feat(Toast): introduce Toast component

mainframev:
[JIRA ticket](https://jira.kiwi.com/browse/ORBIT-1567)

 Storybook: https://orbit-silvenon-feat-toast-component.surge.sh


packages/orbit-components/src/Toast/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+  100% {
+    transform: translateX(0%);
+  }
+}`;
+
+export const resolveFadeAnimations: resolveAnimation = ({ animate, isHovered }) => {
+  if (animate) {
+    return css`
+      animation: ${fadeIn} 1s linear, ${fadeOut} 1s linear 5s forwards;
+      animation-play-state: ${isHovered ? "paused" : "running"};
```
silvenon, line 62: When hovered I think it's best to bring it back completely. Currently it pauses in whatever state it is, even when it's partially faded out.

packages/orbit-components/src/Toast/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+// $FlowFixMe: https://github.com/flow-typed/flow-typed/issues/3653#issuecomment-568539198
+StyledWrapper.defaultProps = {
+  theme: defaultTheme,
+};
+
+const StyledInnerWrapper = styled.div`
+  ${({ theme, animate }) => css`
+    position: relative;
+    cursor: pointer;
```
silvenon, line 37: Why `cursor: pointer`? Nothing happens on click.

silvenon (changes requested):
- I think the "countdown" should happen _after_ Toast faded in, not during.
- I think when hovering the Toast can still be able to expire, but not go away until the mouse is removed, if you understand what I mean.

packages/orbit-components/src/Toast/index.jsx
mainframev: it's because in the first version it's supposed to be clickable. Plus

mainframev:
I decided to change implementation of toast, because currently it looks as a draft. I will reopen it or create a new PR.

PR #3190, opened 2021-11-12
fix(Box): flex-grow and flex-shrink, allow any number

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1636723584134200)
 Storybook: https://orbit-silvenon-fix-box-grow.surge.sh


silvenon (approved):
Maybe also update stories and orbit.kiwi examples to support custom number input rather than just 0 and 1.

PR #3191, opened 2021-11-13
feat: add new experimental hook `useTransition`

silvenon:
The Toast implementation in #3187 uses CSS Animations, but I think this is unnecessary, you can get by with just CSS Transitions. The problem with the current Toast implementation is that it stays in the DOM after fading out, so you won't be able to click on something below it if you need to, which will be very confusing because Toast is invisible, so I recommend unmounting Toast after fading out.

The problem with animating the presence of an element in the DOM is a well-known one, and has already been solved by libraries like react-transition-group (more specifically, its CSSTransition component), which I used as an inspiration for this hook. Currently it's undocumented because it's intended primarily for internal use, but people are welcome to try it.

I considered using [Headless UI's Transition component](https://headlessui.dev/react/transition), but it seems significantly bigger, so by having our own minimal hook we're saving about 3 kB.

 Storybook: https://orbit-silvenon-feat-use-transition.surge.sh

### DIFF
--- a/packages/orbit-components/src/hooks/useTransition/index.js
+++ b/packages/orbit-components/src/hooks/useTransition/index.js
@@ -26,7 +26,7 @@ const useTransition: UseTransitionType = ({ show, appear = false }) => {
 
   React.useEffect(() => {
     if (firstRender.current) {
-      if (appear && show && firstRender.current) {
+      if (appear && show) {
         setMounted(true);
       }
     }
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/hooks/useTransition/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+  // if (module.hot) {
+  //   module.hot.dispose(() => {
+  //     firstRender.current = true
+  //     setMounted(!appear && show)
+  //     setEnter(!appear && show)
+  //     setDone(!appear || !show)
+  //     forceUpdate()
+  //   })
+  // }
```
silvenon, line 25: I would like this to stay here for now because it's a way to keep this hook working through hot reloads, but I'm not sure how to include it so it doesn't end up in the build.

packages/orbit-components/src/hooks/useTransition/index.js
```diff
@@ -0,0 +1,83 @@
...
+      if (show) {
+        setMounted(true);
+      } else {
+        setEnter(false);
+        setDone(false);
+        listener = () => {
+          setMounted(false);
+          setDone(true);
+        };
+        el?.addEventListener("transitionend", listener);
```
silvenon, line 48: By listening for `transitionend` instead of using `setTimeout` slowing down animations via Chrome's developer tools will apply to JavaScript as well, not just CSS.

mainframev (approved):

mainframev:
could you also document this hook? 🙏🏻

silvenon:
I was going to, a little later. I will send a documentation PR separately then.

PR #3195, opened 2021-11-16
docs(Gudelines): fix not visible images

mainframev:
A quick fix for [slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1636970158141300). 
 Storybook: https://orbit-silvenon-fix-voice-and-tone.surge.sh


silvenon (changes requested):
By doing this we would be fighting against the tool that's supposed to help us. The reason why this happens is because multiple the containers are set to be as wide as their content, which is `0` in this case because of `position: absolute`, so the content has nowhere to go. Those containers should stretch as wide as they can, so set `align` of relevant Stack components to `stretch`, and do whatever else is necessary.

Let me know if you need help 😉

mainframev:
> By doing this we would be fighting against the tool that's supposed to help us. The reason why this happens is because multiple the containers are set to be as wide as their content, which is `0` in this case because of `position: absolute`, so the content has nowhere to go. Those containers should stretch as wide as they can, so set `align` of relevant Stack components to `stretch`, and do whatever else is necessary.
> 
> Let me know if you need help 😉

weird, i tried `stretch` and it did not work, maybe i missed something 🤔 I agree that the fix is dirty

docs/src/data/CHANGELOG.md
```diff
@@ -3,32 +3,58 @@
...
+-   **Wizard:** close after click on WizardStep in compact ([#3176](https://github.com/kiwicom/orbit/issues/3176)) ([247e7ae](https://github.com/kiwicom/orbit/commit/247e7ae43a8e6983c183bd100806d64c3636b0a7))
+
 ### [2.0.1](https://github.com/kiwicom/orbit/compare/@kiwicom/orbit-components@2.0.0...@kiwicom/orbit-components@2.0.1) (2021-11-03)
 
 #### Bug Fixes
 
-*   missing storybook provider ([#3173](https://github.com/kiwicom/orbit/issues/3173)) ([4967fb7](https://github.com/kiwicom/orbit/commit/4967fb7dc2f34a716db7fda30c48952992979562))
-*   **useToggle:** make typings reflect initial param is optional ([95821c5](https://github.com/kiwicom/orbit/commit/95821c514e8ccf970ec606de219eff301309f914))
+-   missing storybook provider ([#3173](https://github.com/kiwicom/orbit/issues/3173)) ([4967fb7](https://github.com/kiwicom/orbit/commit/4967fb7dc2f34a716db7fda30c48952992979562))
+-   **useToggle:** make typings reflect initial param is optional ([95821c5](https://github.com/kiwicom/orbit/commit/95821c514e8ccf970ec606de219eff301309f914))
```
silvenon, line 37: Again these changes, we shouldn't commit this. In fact, I don't see a reason for this generated file to be in the version control at all. Could you remove it and add it to `.gitignore`?

silvenon (changes requested):
Great, works! Just one thing.

silvenon (approved):

PR #3196, opened 2021-11-17
test: remove test warnings

mainframev:


packages/orbit-components/src/InputField/__tests__/index.test.jsx
```diff
@@ -64,7 +64,7 @@ describe("InputField", () => {
     expect(screen.getByTestId("prefix")).toBeInTheDocument();
     expect(screen.getByTestId("suffix")).toBeInTheDocument();
     fireEvent.focus(input);
-    expect(screen.getByTestId("help")).toBeInTheDocument();
+    await waitFor(() => expect(screen.getByTestId("help")).toBeInTheDocument());
```
silvenon, line 67: What do you think about using `FindBy` queries instead? It's shorter.
```suggestion
    expect(await screen.findByTestId("help")).toBeInTheDocument();
```
It's not symmetrical to waiting for disappearance, if that bothers you you can keep using `waitFor`.

silvenon (changes requested):
🎉 FINALLY 🎉 Thanks!

I pushed the last one, in `gatsby-node.js`, because of missing changelog paths.

silvenon (changes requested):
I thought about this some more, and this is only a workaround for us, it will continue happening to others. These tests worked synchronously (the error forms ones), so making them async just to eat the `act` warning is a hack, we should fix the source of the problem, which is setting refs in ErrorFormTooltip via state. Merely using ref objects unfortunately won't work, which is why you used state, so could you find another way?

silvenon (approved):

PR #3197, opened 2021-11-18
fix: prevent tab shadow clipping

silvenon:
The tab shadow was very noticeably clipping on the left, especially stacked tabs on hover.

 Storybook: https://orbit-silvenon-docs-tabs-shadow-clip.surge.sh

### DIFF
--- a/docs/src/components/Tabs/consts.ts
+++ b/docs/src/components/Tabs/consts.ts
@@ -2,7 +2,9 @@ import { mediaQueries } from "@kiwicom/orbit-components";
 
 export const TAB_HEIGHT = "64px";
 export const BORDER_RADIUS = "12px";
-// to prevent tab shadow from clipping
-export const SHADOW_PADDING_TOP = "16px";
-export const SHADOW_PADDING_RIGHT = "24px";
+// taken from the stronger shadow that appears when hovering stacked tabs
+// we use these values for padding to prevent that shadow from clipping
+export const SHADOW_TOP = "16px";
+export const SHADOW_LEFT = "24px";
+export const SHADOW_RIGHT = "24px";
 export const FULL_WIDTH_BREAKPOINT: keyof typeof mediaQueries = "largeMobile";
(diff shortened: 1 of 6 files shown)


docs/src/components/Tabs/consts.ts
```diff
@@ -2,7 +2,9 @@ import { mediaQueries } from "@kiwicom/orbit-components";
...
 export const TAB_HEIGHT = "64px";
 export const BORDER_RADIUS = "12px";
-// to prevent tab shadow from clipping
-export const SHADOW_PADDING_TOP = "16px";
-export const SHADOW_PADDING_RIGHT = "24px";
+// taken from the stronger shadow that appears when hovering stacked tabs
+// we use these values for padding to prevent that shadow from clipping
+export const SHADOW_TOP = "16px";
+export const SHADOW_LEFT = "24px";
+export const SHADOW_RIGHT = "24px";
```
mainframev, line 9: i guess instead of consts you can use tokens. `spaceMedium` and `spaceLarge`

silvenon: `boxShadowRaised`, which is used for tabs, doesn't use the spacing tokens, the value is instead hardcoded, so using spacing tokens wouldn't be appropriate because changing the spacing tokens would ruin my fix. Instead I'll parse the shadow token and retrieve these values from there.

docs/src/components/Tabs/index.tsx
```diff
@@ -27,8 +23,9 @@ const StyledContainer = styled.div<{ collapsed: boolean }>`
     z-index: 1; /* place tabs above the main container */
     display: flex;
     flex-wrap: wrap; /* so that we can detect whether tabs wrap, and in that case collapse them */
-    padding-top: ${SHADOW_PADDING_TOP};
-    padding-right: ${SHADOW_PADDING_RIGHT};
+    padding-top: ${getTabShadowReachTop(theme)};
```
mainframev, line 26: you don't need to pass the theme as a parameter explicitly, change the function, so you can just pass the function as a reference, as we usually do with SC.

mainframev (changes requested):

mainframev (approved):

PR #3198, opened 2021-11-19
fix: restore examples for error forms

CollierCZ:
#3122 by @mainframev removed examples for InputField, which made sense for the InputField component page. But it left the [Forms and errors page](https://orbit.kiwi/design-patterns/form-errors/) with 3 instances of `Could not find example with the id:`. This PR restores those examples.

### DIFF
--- a/.github/actions/node/action.yml
+++ b/.github/actions/node/action.yml
@@ -21,5 +21,3 @@ runs:
     - name: Install dependencies
       run: yarn install --frozen-lockfile
       shell: bash
-      env:
-        STREAMLINE_FAMILIES: ${{ secrets.STREAMLINE_FAMILIES }}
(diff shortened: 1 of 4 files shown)


silvenon (approved):
Indeed, thanks!

CollierCZ:
Is that a custom action that's making all the test fail, @silvenon ? Interesting. What's it supposed to do?

silvenon:
Just a way to share common steps across workflows.

CollierCZ:
Is it cause it's from a forked repo? :shrug:

silvenon:
Might be! But let's see if this change fixes it. Although if it does I have no idea how it ever worked before.

silvenon:
I'll assume these failures are because it's a fork, and deal with this problem separately.

PR #3199, opened 2021-11-19
feat(icons): added new icons

mainframev:
Added new icons `no-guarantee` and `alert-octagon`. Pngs are generated and updated in `images.kiwi`.
 Storybook: https://orbit-silvenon-feat-add-icons.surge.sh


silvenon (changes requested):
It's a negated form of `kiwicom-guarantee`, so rename it to `no-kiwicom-guarantee`. Make sure to do it in images.kiwi.com as well.

silvenon (approved):

PR #3204, opened 2021-11-30
docs: flatten and alphabetize list of components in navigation

silvenon:
Tests are here, but implementation is missing.

 Storybook: https://orbit-silvenon-docs-nav-components-alphabetical.surge.sh


silvenon (commented):
Works great! 💯  I can't approve PRs I started, so consider it approved!

mainframev (approved):

PR #3212, opened 2021-12-07
fix(Heading): temporary fix for `viewports` being `undefined`

silvenon:
For some reason `viewports` ends up being `undefined` in some cases, even though it should always be at least an empty object considering that it's a rest prop. Until we figure out why this happens we're adding a temporary fix so that this doesn't cause an error in production.

 Storybook: https://orbit-silvenon-fix-heading-viewports-undefined.surge.sh


mainframev:
i would also change it there from `...viewports`, to:
```
const {mediumMobile,
    largeMobile,
    tablet,
    desktop,
    largeDesktop} = props;
   
 const viewports = {   mediumMobile,
    largeMobile,
    tablet,
    desktop,
    largeDesktop};
```

check `Box` component

PR #3231, opened 2022-01-05
docs(tracking): add allRepositories page and instance props search

mainframev:


docs/plugins/dashboard/gatsby/on-create-pages.js
```diff
@@ -37,9 +38,19 @@ module.exports = async ({ graphql, actions, reporter, cache }) => {
       };
     });
 
+    pages.push({
```
silvenon, line 41: Best make it the first item.
```suggestion
    pages.unshift({
```

silvenon (commented):
Some suggestions, not required, except maybe the one where the page stops responding which is kind of critical:

- [x] I suggest adding "all repositories" to the search so that it's possible to quickly search for something in that area
- [ ] In "all repositories" when opening the modal to see the instances of a certain prop, the link name should also contain the repository name. It's fine to leave it out when viewing instances of specific projects because it's implied, but on this page it isn't.
- [ ] Something in the search is sometimes causing the page to stop responging, for example when I try t search for something on this page: http://localhost:8000/dashboard/tracking/booking/accordion/#spaceAfter
- [ ] It would be nice to list the projects where a certain component is used. Maybe as another column, to say "Used in:" and list the projects.
- [ ] One small request, unrelated to these changes, that you can add to this PR: instead of fetching `allSitePage` via GraphQL, could you assign `restPages` to an array of `path` + `name`? That way it will look nicer in the search results, instead of listing just the path.

PR #3232, opened 2022-01-05
feat(Tag): add new colors and props

mainframev:
[Figma](https://www.figma.com/file/Y79I18H9w8HJZVxPdAkAHb/%5BOrbit%5D-Desktop-Components?node-id=1324%3A0)
 Storybook: https://orbit-silvenon-feat-tag-adjustments.surge.sh


silvenon (approved):
It would be good to also add it to the playground in orbit.kiwi.

silvenon (commented):
I actually meant add to the playground in **orbit.kiwi**, not Storybook. 😄

PR #3241, opened 2022-01-18
error forms small changes

mainframev:
- added new prop for error forms, if someone would like to disable closable help

small visual changes:
- label icons for help changed colors from secondary to info.
- close icon moved on top
 Storybook: https://orbit-silvenon-fix-error-forms.surge.sh

### DIFF
--- a/packages/orbit-components/src/InputField/README.md
+++ b/packages/orbit-components/src/InputField/README.md
@@ -55,7 +55,7 @@ Table below contains all types of the props available in InputField component.
 | tabIndex     | `string \| number`          |            | Specifies the tab order of an element                                                                                                                                            |
 | **type**     | [`enum`](#enum)             | `"text"`   | The type of the InputField.                                                                                                                                                      |
 | value        | `string`                    |            | Specifies the value of the InputField.                                                                                                                                           |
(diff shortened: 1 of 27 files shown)


packages/orbit-components/src/InputField/README.md
```diff
@@ -55,6 +55,7 @@ Table below contains all types of the props available in InputField component.
 | tabIndex     | `string \| number`          |            | Specifies the tab order of an element                                                                                                                                            |
 | **type**     | [`enum`](#enum)             | `"text"`   | The type of the InputField.                                                                                                                                                      |
 | value        | `string`                    |            | Specifies the value of the InputField.                                                                                                                                           |
+| helpClosable | `boolean`                   | `true`     | Responsible for hiding tooltip with help message by clicking on close button or automatically onBlur                                                                             |
```
silvenon, line 58: This would be better I think, if you agree apply it to other readmes as well 🙏

```suggestion
| helpClosable | `boolean`                   | `true`     | Whether to display help as a closable tooltip, or have it open only while the field is focused, like error.                                                                             |
```

silvenon (changes requested):
Looks much better with the info color!

PR #3242, opened 2022-01-18
UI fixes for Collapse, SocialButton, Popover and Separator

silvenon:
UI fixes for:

- Collapse ([ORBIT-1979](https://jira.kiwi.com/browse/ORBIT-1979))
  - the problem was that `:focus` styles in ButtonPrimitve were overriding `:hover` styles, so I did some editing there, also now `:active` styles should work better
- SocialButton ([ORBIT-1992](https://jira.kiwi.com/browse/ORBIT-1992))
- Popover ([ORBIT-1984](https://jira.kiwi.com/browse/ORBIT-1984))
  - bottom corners weren't rounded when actions were present because of the actions' white background, one solution was to set `overflow: hidden` on the entire Popover, but I didn't want some unintended consequences so I went for the safer option of rounding corners of the actions container
- Separator ([ORBIT-1987](https://jira.kiwi.com/browse/ORBIT-1987))

The two remaining ones are ButtonMobileStore and Card, but those aren't cleared up yet so we'll do them in another PR.

I also refactored CardWrapper as a preparation for UI fixes in Card, so I'm including that in this PR.

 Storybook: https://orbit-silvenon-fix-ui-fixes.surge.sh

### DIFF
--- a/packages/orbit-components/src/Popover/components/ContentWrapper.jsx
+++ b/packages/orbit-components/src/Popover/components/ContentWrapper.jsx
@@ -35,7 +35,7 @@ const StyledContentWrapper = styled.div`
 
     ${mq.largeMobile(css`
       max-height: 100%;
-      border-radius: 3px;
+      border-radius: ${theme.orbit.borderRadiusNormal};
       bottom: auto;
       left: auto;
       position: relative;
@@ -66,6 +66,8 @@ const StyledActions = styled.div`
       position: relative;
       bottom: auto;
       left: auto;
+      border-bottom-left-radius: ${theme.orbit.borderRadiusNormal};
+      border-bottom-right-radius: 3px;
       ${StyledButtonPrimitive} {
         width: auto;
         flex-grow: 0;
--- a/packages/orbit-components/src/Separator/Separator.stories.jsx
+++ b/packages/orbit-components/src/Separator/Separator.stories.jsx
@@ -3,8 +3,9 @@ import * as React from "react";
 import { select } from "@storybook/addon-knobs";
 
 import SPACINGS_AFTER from "../common/getSpacingToken/consts";
+import RenderInRtl from "../utils/rtl/RenderInRtl";
 
-import Separator from ".";
+import Separator, { type Indent } from ".";
 
 export default {
   title: "Separator",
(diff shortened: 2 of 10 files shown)


packages/orbit-components/src/Separator/index.jsx
```diff
@@ -1,29 +1,69 @@
...
+      padding-${right}: ${getIndentAmount};
+    `};
+    ${align === "right" &&
+    css`
+      padding-${left}: ${getIndentAmount};
+    `};
+    ${align === "center" &&
+    css`
+      padding-${left}: ${getIndentAmount};
+      padding-${right}: ${getIndentAmount};
```
mainframev, line 42:
```suggestion
      padding: 0 ${getIndentAmount};
```

no need for rtl utility here for both sides :)

packages/orbit-components/src/Separator/Separator.stories.jsx
```diff
@@ -25,7 +26,10 @@ export const Playground = (): React.Node => {
     [null, ...Object.values(SPACINGS_AFTER)],
     SPACINGS_AFTER.LARGEST,
   );
-  return <Separator spaceAfter={spaceAfter} />;
+  const align = select("align", ["left", "right", "center"], "left");
```
mainframev, line 29: could you also add that to docs? 🙏🏻

packages/orbit-components/src/Popover/components/ContentWrapper.jsx
```diff
@@ -66,6 +66,8 @@ const StyledActions = styled.div`
       position: relative;
       bottom: auto;
       left: auto;
+      border-bottom-left-radius: 3px;
```
mainframev, line 69: use `borderRadiusNormal` token here :)

mainframev (changes requested):
The card ticket in jira has poor description, just says remove borders in responsive. You removed only borders from sides, but left on the top and bottom. Is it correct? Because there is no borders in figma, only for section. 🤔

packages/orbit-components/src/Separator/Separator.stories.jsx
silvenon: What do you mean? I added `align` prop to `Separator/README.md` already.

silvenon:
> You removed only borders from sides

I only refactored Card, without any design change, because I wasn't sure what I'm supposed to do. The refactor was a preparation step, but I decided to do the actual UI fix in another PR, once it becomes clear what to do exactly.

packages/orbit-components/src/Separator/Separator.stories.jsx
mainframev: I meant to add them to our sandbox

mainframev (approved):

PR #3243, opened 2022-01-20
fix(Toast): fix infinite loop in unit tests

silvenon:
https://github.com/timolins/react-hot-toast/issues/107#issuecomment-1017213020

I solved the problem by memoizing the ref, so it doesn't create a new one on every render, avoiding continuously updating state in react-hot-toast.

 Storybook: https://orbit-silvenon-fix-toast-infinite-loop-jsdom.surge.sh

### DIFF
--- a/packages/orbit-components/src/Toast/__tests__/index.test.jsx
+++ b/packages/orbit-components/src/Toast/__tests__/index.test.jsx
@@ -7,11 +7,15 @@ import { ToastRoot, createToast } from "..";
 import { Airplane } from "../../icons";
 import Toast from "../ToastMessage";
 import Button from "../../Button";
-import { SWIPE_DISMISS_DELAY } from "../hooks/useSwipe";
+import { EXPIRE_DISMISS_DELAY, SWIPE_DISMISS_DELAY } from "../consts";
 
 beforeEach(() => {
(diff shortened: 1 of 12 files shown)


packages/orbit-components/src/Toast/__tests__/index.test.jsx
```diff
@@ -1,12 +1,22 @@
...
 import Button from "../../Button";
+import { EXPIRE_DISMISS_DELAY, SWIPE_DISMISS_DELAY } from "../consts";
+
+beforeEach(() => {
+  // reset mocks before each test
+  jest.useFakeTimers();
+});
+
+afterAll(() => {
+  jest.useRealTimers();
```
mainframev, line 18: Official docs recommend to use `usePendingTimers` before `useRealTimers`. [Here](https://testing-library.com/docs/using-fake-timers/).
> It's important to also call runOnlyPendingTimers before switching to real timers. This will ensure you flush all the pending timers before you switch to real timers.

mainframev (changes requested):
Great work on solving this tricky issue! :) It was quite fast and fuuuuurious! :D

mainframev (approved):

PR #3247, opened 2022-01-21
fix(Itinerary): truncate possible long texts

mainframev:
Truncate long texts in `ItinerarySegmentDetail`. Let's wait until the discussion and design will be updated for badges, we can use there `HorizontalScroll`
 Storybook: https://orbit-silvenon-fix-itinerary.surge.sh


silvenon (approved):
Works great!

PR #3249, opened 2022-01-24
docs(ios): update iOS components availability

PavelHolec:


silvenon (approved):
Nice, thanks! We should really find a way to automate this.

PR #3255, opened 2022-01-27
Better usage of `@testing-library/user-event` and clear up remaining `act` warnings

silvenon:
We still had `act` warnings, so I fixed those and changed the existing fixes with what I think is the most straightforward solution. One way to make it self-documenting is to create a utility that does the empty `act` call named `flushQueue` or something, and add comments to document it, let me know what you think.

This solution should work both for Popper and Floating UI. floating-ui/floating-ui#1521

 Storybook: https://orbit-silvenon-test-popper-act-warnings.surge.sh


mainframev (approved):

mainframev:
LGTM :)

PR #3264, opened 2022-02-01
Document useTransition hook

silvenon:
This also adds a tweak to `useTransition` API. Instead of the `enter` property I decided it would be better to pass the `state` property.

 Storybook: https://orbit-silvenon-docs-use-transition.surge.sh


mainframev (approved):
Thanks! <3

PR #3330, opened 2022-04-07
fix(docs): remove Spectrum

CollierCZ:
Spectrum is now read-only and shouldn't be promoted as a communications channel. This removes it from the docs.


CollierCZ:
Seems like the checks aren't set up to work on forks :see_no_evil:

PR #3365, opened 2022-05-20
docs: component status is N/A for all Android components

hrach:
Currently, there is a mistake – component that we do not have and is not defined in our repo is defined here with released state, so rather making everything here a default N/A.
 Storybook: https://orbit-mainframev-hrach-patch-1.surge.sh

### DIFF
--- a/docs/src/data/component-status.yaml
+++ b/docs/src/data/component-status.yaml
@@ -16,28 +16,28 @@
   figma: Released
   react: Released
   ios: Released
-  android: Released
+  android: N/A
   docs: Released
   group: Information
 - component: Badge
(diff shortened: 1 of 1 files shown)


mainframev:
Hi @hrach, 

thanks, i will check, it's automated, i will need to update the script, it seems like a mistake there 🤔

hrach:
IMO we're quite ok just to merge this and no other change is required 😁 🤞

mainframev:
> IMO we're quite ok just to merge this and no other change is required 😁 🤞
yep, i will merge it, but still will need to check the script, why it does not match actual status :)

PR #3582, opened 2022-09-29
feat(SegmentedSwitch): introduce SegmentedSwitch component

mainframev:
New SegmentedSwitch 
[ORBIT-2200](https://jira.kiwi.com/browse/ORBIT-2200)
 Storybook: https://orbit-mainframev-feat-segmented-switch.surge.sh


packages/orbit-components/src/SegmentedSwitch/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+  label,
+}: Props): React.Node => {
+  const {
+    tooltipShown,
+    tooltipShownHover,
+    setTooltipShown,
+    setTooltipShownHover,
+    labelRef,
+    handleFocus,
+  } = useErrorTooltip({ onFocus });
```
DSil, line 50: I believe that doing this is causing the tooltip to be triggered when we select an option. Is that expected? I had the feeling we would only want the tooltip to be manually triggered **or** on hover. How can we know for sure when is the tooltip expected to be shown?

DSil, line 50: I believe that doing this is causing the tooltip to be triggered when we select an option. Is that expected? I had the feeling we would only want the tooltip to be manually triggered **or** on hover. How can we know for sure when is the tooltip expected to be shown?

DSil, line 50: I believe that doing this is causing the tooltip to be triggered when we select an option. Is that expected? I had the feeling we would only want the tooltip to be manually triggered **or** on hover. How can we know for sure when is the tooltip expected to be shown?

mainframev: Yes, it's expected. The behavior is similar to our ErrorForms. It should be triggered onFocus, if the help or error message is provided. We can adjust the behavior to have it controlled via property, like `showTooltip`.

packages/orbit-components/src/SegmentedSwitch/SwitchSegment/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+      ? `0 0 0 2px ${theme.orbit.paletteBlueNormal}`
+      : `0 0 0 1px ${theme.orbit.paletteCloudDark}`};
+
+    &:nth-child(odd) {
+      border-radius: 5px 0 0 5px;
+    }
+
+    &:nth-child(even) {
+      border-radius: 0 5px 5px 0;
+    }
```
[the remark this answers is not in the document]
mainframev: I started to use indexes for that, it's a simple approach, but then I remembered about the styling document I'm working on. So, there is a problem in other libraries, which allows building styles behind the runtime, that you cannot have dynamic styling based on indexes. 

I will fix that with interpolation in StyledWrapper

DSil (approved):

PR #3588, opened 2022-10-05
feat: add onShow properties for Tooltip and TooltipPrimitive 

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1664966991843809)
 Storybook: https://orbit-mainframev-feat-add-onshow-callback.surge.sh

### DIFF
--- a/packages/orbit-components/src/Tooltip/__tests__/index.test.jsx
+++ b/packages/orbit-components/src/Tooltip/__tests__/index.test.jsx
@@ -1,6 +1,7 @@
 // @flow
 import * as React from "react";
 import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import Tooltip from "..";
 
@@ -15,9 +16,16 @@ jest.mock("../../hooks/useMediaQuery", () => {
 describe("Tooltip", () => {
   it("it should render Tooltip", () => {
     const content = "Write some message to the user";
+    const onShow = jest.fn();
 
-    render(<Tooltip content={content}>kek</Tooltip>);
+    render(
+      <Tooltip content={content} onShow={onShow}>
+        kek
+      </Tooltip>,
+    );
 
     expect(screen.getByText("kek")).toBeInTheDocument();
+    userEvent.hover(screen.getByText("kek"));
+    expect(onShow).toHaveBeenCalled();
   });
 });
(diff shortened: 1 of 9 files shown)


packages/orbit-components/src/primitives/TooltipPrimitive/README.md
```diff
@@ -18,22 +18,23 @@ After adding import into your project you can use it simply like:
...
+| stopPropagation      | `boolean`                     |          | If `true` the click event on children won't bubble. Useful when you use TooltipPrimitive inside another clickable element. |
+| placement            | [`placement`](#placement)     | `auto`   | 12 different placements to choose from                                                                                     |
+| noFlip               | `boolean`                     | `false`  | Turns off automatic flipping of the Tooltip when there is not enough space                                                 |
+| offset               | `[number, number]`            | `[0, 5]` | Set offset `[vertical, horizontal]` for Tooltip                                                                            |
+| tabIndex             | `string \| number`            | `"0"`    | Specifies the tab order of an element                                                                                      |
+| onShow               | `() => void \| Promise<void>` |          | Callback for Tooltip, which will be fired when it will be shown                                                            |
```
DSil, line 37:
```suggestion
| onShow               | `() => void \| Promise<void>` |          | Callback for Tooltip, which will be fired when it is shown                                                            |
```
💅

packages/orbit-components/src/Tooltip/__tests__/index.test.jsx
```diff
@@ -15,9 +16,16 @@ jest.mock("../../hooks/useMediaQuery", () => {
...
-    render(<Tooltip content={content}>kek</Tooltip>);
+    render(
+      <Tooltip content={content} onShow={onShow}>
+        kek
+      </Tooltip>,
+    );
 
     expect(screen.getByText("kek")).toBeInTheDocument();
+    userEvent.hover(screen.getByText("kek"));
+    expect(onShow).toHaveBeenCalled();
```
DSil, line 29: Don't you think this deserves its own test? Separate the testing responsibilities so that when one fails it is for (hopefully) just one reason.

mainframev: Good question. It says "it should render Tooltip", so I guess it's ok to have it like that :)

DSil: Yes, when I suggested a separate test I was thinking of something like "it fires onShow". I'm ok with leaving them together, I was just considering more granular tests, but I think both work.

DSil (approved):

PR #3591, opened 2022-10-07
fix(Card): Adjust icon size on expandable card

DSil:
[ORBIT-1977](https://jira.kiwi.com/browse/ORBIT-1977)
Icon size was inconsistent with the design. I has been adjusted.

[Slack thread](https://skypicker.slack.com/archives/GSGN9BN6Q/p1664975130259849)
 Storybook: https://orbit-mainframev-fix-card-icon-size.surge.sh


mainframev:
@DSil 

It's a bit complex problem here. Designers changed all normal sized icons from 24px to 20px. It has to be updated in our repo automatically, but because of other design changes (breaking in the palette), it's not updating. It can cause a lot of visual inconsistencies, so we will have to merge palette colors first, then update icon sizes and run Loki to check where are the problems.

mainframev:
Closing in favor of #3613

PR #3597, opened 2022-10-11
fix(Stack): keep old behavior, enable gap only if flex or spacing is …

mainframev:


packages/orbit-components/src/Stack/helpers/getGap.js
```diff
@@ -5,18 +5,60 @@ import getSpacing from "./getSpacing";
...
       `;
     }
+
+    return css`
+      & > * {
+        margin: ${margin && rtlSpacing(margin)}!important;
+        ${isMobile &&
+        css`
+          &:last-child {
+            margin: 0 !important;
```
DSil, line 57: What is this for? Do we always want to add margin: 0 to the last child on mobile?

mainframev: yes, it's because last child does not need to have margin

packages/orbit-components/src/Stack/helpers/getGap.js
```diff
@@ -5,18 +5,60 @@ import getSpacing from "./getSpacing";
...
 import { QUERIES } from "../../utils/mediaQuery/consts";
 import type { GetGap } from "./getGap";
+import { rtlSpacing } from "../../utils/rtl";
+import { DIRECTIONS } from "../../utils/layout/consts";
+import type { Direction, Spacing } from "..";
+
+const getDirectionSpacingTemplate = (direction: Direction | Spacing): string => {
+  switch (direction) {
+    case DIRECTIONS.COLUMN:
+      return "0 0 __spacing__ 0";
```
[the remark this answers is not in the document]
mainframev: I just copied that from old implementation. Sure

packages/orbit-components/src/Stack/helpers/getGap.js
mainframev: In general to the last child, but it was like that from the beginning, so not sure if we should change or keep that. We can change since the gap has the same behaviour. It's probably it was like that, for mobile, where usually is colomn direction and last element does not need to have a margin.

PR #3599, opened 2022-10-13
docs(improvements): add new rocket image and add navbar links

DSil:
There is a new rocket image and the navbar now has a set of links for quick access, featuring mobile responsibility.
 Storybook: https://orbit-mainframev-docs-improvements.surge.sh


mainframev:
for some reason cloudflare does not build that with the image, it's not loaded, locally it's fine

PR #3600, opened 2022-10-14
chore: split cron jobs and add contributors update on action

mainframev:
- It's better to split the cron jobs, so they are independent, and merge requests will be more descriptive
- add an update of contributors on the action, so it does not run with every building process

for now, used the `.mjs` extensions, later we can rewrite that to `ts-node` and `.ts`
 Storybook: https://orbit-mainframev-chore-split-cron-jobs.surge.sh


.github/workflows/cron-update-contributors.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v4
+        with:
+          base: master
+          branch: docs/update-contributors
+          token: ${{ secrets.PAT }}
+          delete-branch: true
+          title: "chore: update contributors"
+          labels: Icons
```
DSil, line 38: Do we want this label? Or is this a copy-paste leftover? 👀

docs/plugins/contributors/users.json
```diff
@@ -10,6 +10,7 @@
     "github": "https://github.com/CollierCZ",
     "twitter": "",
     "website": "https://collier.cz/",
+    "core": true,
```
DSil, line 13: Is this property new? Why do we need it? 🤔

docs/src/components/Github/OrbitTeam.tsx
```diff
@@ -8,16 +8,12 @@ import Member from "./TeamMember";
 
 import { Contributor } from ".";
 
-interface AllMembers {
+interface Props {
   current: Contributor[];
   previous: Contributor[];
```
DSil, line 13: I know these are old, but at least the interface would tell us it was about members. Now I think the readability would improve a lot if we could rename these to `currentMembers` and `previousMembers`, for example. WDYT?

.github/workflows/cron-update-contributors.yml
mainframev: good catch, not, for contributors we don't ;)

docs/plugins/contributors/users.json
mainframev: yeah, to distinguish core team from other contributors :)

docs/src/components/Github/OrbitTeam.tsx
mainframev: sure, makes sense, will rename ;)

DSil (approved):
👍

PR #3604, opened 2022-10-17
fix(Table): responsive shadow was not correct

DSil:
While working on the new Table for Docs, I noticed this small bug on the Table. The shadow on the mobile version was not disappearing even if the user scrolled all the way horizontally.

Before:
before

Now:
now

 Storybook: https://orbit-mainframev-fix-table-shadow-responsiveness.surge.sh


mainframev:
LGTM :)

Good catch and the videos you attached, that's saving time for reviewing 👍🏻

PR #3606, opened 2022-10-18
fix(Accordion): not expanding

### LINKED ISSUE #3561: Accordion Component Broken in Latest Version
## Expected Behavior

The accordion component should open on click when set up correctly (e.g. as outlined in [this post](https://github.com/kiwicom/orbit/issues/2691)). This was working correctly for us previously but has become broken in one of the recent versions (over past couple of months).

I have tried implementing the simple example from that old issue to confirm it's not something specific to how we are using the component (despite our implementation having worked correctly previously).

## Current Behavior

The accordion section does not open on click.

## Steps to Reproduce

Try implementing accordion as outlined in this [old issue](https://github.com/kiwicom/orbit/issues/2691). The accordion will not open.

Nadhum:

- [x] New Components has both `.flow` and `d.ts` files and are exported in `index.js.flow` and `index.d.ts`

Fixes #3561 

### Description
An id update was missed during some id work, which resulted in the accordion checking against the wrong id to determine whether it should be expanded or not.

Tested in storybook

### DIFF
--- a/.github/contribution/README.md
+++ b/.github/contribution/README.md
@@ -40,7 +40,7 @@
 First, you need to build the project:
 
 ```
-yarn build
+yarn build-ci
 ```
 
 Next, run Storybook:
--- a/.github/contribution/README.md
+++ b/.github/contribution/README.md
@@ -40,7 +40,7 @@
 First, you need to build the project:
 
 ```
-yarn build-ci
+yarn build
 ```
 
 Next, run Storybook:
--- a/packages/orbit-components/src/Accordion/index.jsx
+++ b/packages/orbit-components/src/Accordion/index.jsx
@@ -37,7 +37,7 @@ const Accordion = ({
 
           const { id: innerId } = item.props;
           // Determine if section is expanded
-          const isExpanded = expandedSection === id;
+          const isExpanded = expandedSection === innerId;
           // Callback with section id
           // onExpand is not required prop to have easier loading use case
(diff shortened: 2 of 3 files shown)


packages/orbit-components/src/Accordion/index.jsx
```diff
@@ -37,7 +37,7 @@ const Accordion = ({
 
           const { id: innerId } = item.props;
           // Determine if section is expanded
-          const isExpanded = expandedSection === id;
+          const isExpanded = expandedSection === innerId;
```
Nadhum, line 40: This is the actual fix.

.github/contribution/README.md
```diff
@@ -40,7 +40,7 @@
 First, you need to build the project:
 
 ```
-yarn build
+yarn build-ci
```
mainframev, line 43: could you remove that? we already building the tokens during the build, so no need in that ;)

Nadhum: This is the README in the contribution guide, there is no step yarn build, I thought it's out of date? Am I missing something here?

Nadhum: The steps to start with development here https://github.com/kiwicom/orbit/blob/master/.github/contribution/README.md#starting-with-development
should be 
```
yarn build-ci
yarn dev
```
right? or at least that's what I had to do, as there was no target for yarn build

PR #3607, opened 2022-10-18
docs(layout): update props table to a new design

DSil:
New approach for the table that renders the props from the components.

- [x] Initial styling and responsiveness
- [x] Data extraction to apply sorting feature
- [x] Sorting logic
- [x] Final touches
  - [x] Update the icon when sorting

 Storybook: https://orbit-mainframev-table-docs-component.surge.sh


docs/src/docs-table.tsx
```diff
@@ -0,0 +1,238 @@
...
+    &:hover {
+      background-color: inherit;
+
+      ${mq.desktop(css`
+        border-radius: 6px;
+        transform: scale(1.025);
+        box-shadow: 0px 4px 8px 0px #252a311f, 0px 1px 4px 0px #252a3129;
+
+        ${StyledTableCell} {
+          border-top-style: solid;
```
mainframev, line 55: I do not think you need that, it's causing a layout shift and the whole table jumps a bit 🤔 

maybe you can also consider using box-shadow instead of borders, so it will be more smooth for such cases when it will be hovering.

 like `box-shadow: 1px 0 0 0 token` and on hover just to use normal box-shadow elevation token

docs/src/docs-table.tsx
```diff
@@ -0,0 +1,238 @@
...
+      border-left-style: solid;
+    }
+
+    &:last-child {
+      border-top-right-radius: 6px;
+      border-right-style: solid;
+    }
+  }
+  ${StyledTableBody} > ${StyledTableRow} {
+    transition: transform ease-in-out 0.2s;
```
mainframev, line 44: use the token `theme.orbit.durationFast` fort transition

docs/src/docs-table.tsx
```diff
@@ -0,0 +1,238 @@
...
+  showLeft: boolean;
+}>`
+  max-width: 100%;
+  width: 100%;
+  position: relative;
+
+  &::after,
+  &::before {
+    content: " ";
+    display: ${({ showShadows }) => (showShadows ? "block" : "none")};
```
mainframev, line 132: just a small detail, as we discussed before, to look it cleaner - use single destructure on the top ;)

mainframev (changes requested):
it looks good and you are in the right direction. So let's resolve that visual small issues and second round review after sorting ;)

PR #3608, opened 2022-10-18
revert(Stack): remove gap change

mainframev:
We forgot about margin collapsing and that created unnecessary complications on FE repositories, let's revert for now and left it as a feature change. When everybody will be ready for that. 

For LinkList just reverted as it was before
 Storybook: https://orbit-mainframev-revert-spacing-changes.surge.sh

### DIFF
--- a/packages/orbit-components/src/Stack/helpers/getGap.js
+++ b/packages/orbit-components/src/Stack/helpers/getGap.js
@@ -4,10 +4,10 @@ import { css } from "styled-components";
 import getSpacing from "./getSpacing";
 import getProperty from "./getProperty";
 import { QUERIES } from "../../utils/mediaQuery/consts";
-import type { GetGap } from "./getGap";
 import { rtlSpacing } from "../../utils/rtl";
 import { DIRECTIONS } from "../../utils/layout/consts";
 import type { Direction, Spacing } from "..";
+import type { GetGap } from "./getGap";
 
 const getDirectionSpacingTemplate = (direction: Direction | Spacing): string => {
   switch (direction) {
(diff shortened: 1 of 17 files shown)


yarn.lock
```diff
@@ -8263,7 +8263,7 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react-dom@>=16.9.0":
+"@types/react-dom@>=16.9.0", "@types/react-dom@^17.0.9":
```
DSil, line 8266: Non-blocking, but why did this happen? 🤔 The resolved version is the same and all.

packages/orbit-components/src/Stack/helpers/getGap.js
```diff
@@ -4,10 +4,10 @@ import { css } from "styled-components";
 import getSpacing from "./getSpacing";
 import getProperty from "./getProperty";
 import { QUERIES } from "../../utils/mediaQuery/consts";
-import type { GetGap } from "./getGap";
 import { rtlSpacing } from "../../utils/rtl";
 import { DIRECTIONS } from "../../utils/layout/consts";
 import type { Direction, Spacing } from "..";
+import type { GetGap } from "./getGap";
```
DSil, line 10: Unnecessary diff 🤔

DSil (commented):
LGTM, I would just rethink the commit message so that it is also clearer to the changelog. Something like:
`feat(Stack): add legacy prop to apply old spacing properties`. The same could be done for LinkList

yarn.lock
mainframev: looks like it's changed before and now it was fixed 🤔

PR #3611, opened 2022-10-20
docs: visual adjustments to the latest design

DSil:
WIP. Adding gray background to the top of documentation pages
 Storybook: https://orbit-mainframev-docs-improvements.surge.sh


mainframev:
Adjusted the background gradient and TableOfContent. As we thought initially, that gradient should be full-width - I do not think it has to be actually, it looks weird on big screens.

mainframev:
The problem with the image is not being loaded: 

- When you switch between pages there is no additional request happens because the image has been cached already, and kinda there is no need for loading it. 

- there is another event - onStartLoad. That event fires every time, even on page change, and has a wasCached parameter, that indicates whether the image is cached.

docs/src/components/DocLayout/primitives/StyledMain.tsx
```diff
@@ -1,11 +1,13 @@
...
 
+import { MAX_CONTENT_WIDTH } from "../../../consts";
+
 const StyledMain = styled.main`
   flex: 1;
   position: relative;
   display: flex;
-  padding-top: 6px; /* align with DocNavigation */
   flex-direction: column;
+  max-width: calc(${MAX_CONTENT_WIDTH} - 22ch);
```
DSil, line 10: Why 22ch? Could this be some meaningful named const?

DSil (commented):
There's something odd about this. If you look at the Foundation/Colors page, it has no gray background, but if you navigate to other tabs inside the Colors page, the background appears…

docs/src/components/DocLayout/primitives/StyledMain.tsx
mainframev: because the navigation bar has width `22ch`, decided not to change that. Initially i was more for using `rems` everywhere, just fyi ;)

DSil (commented):
LGTM (I can't approve it, as I'm the owner of the PR 😅)

mainframev (approved):

PR #3612, opened 2022-10-20
feat(Timeline): new component design

mainframev:
Updated component to new visual [design](https://www.figma.com/file/wziyIIJCCQi2Ote2PnFfm6/Timeline?node-id=1511%3A2192). 

[ORBIT-2293](https://jira.kiwi.com/browse/ORBIT-2293)

Had to remove prop `asText`, since there are no badges now in the component. It's kinda breaking, but according to tracking data, there is no currently `Timeline` component on production, so maybe we can release it as a minor.


packages/orbit-components/src/Timeline/TimelineStep/components/TypeIcon.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+import CheckCircle from "../../../icons/CheckCircle";
+import Circle from "../../../icons/CircleEmpty";
+import AlertCircle from "../../../icons/AlertCircle";
+import CloseCircle from "../../../icons/CloseCircle";
+import StyledIconWrapper from "../primitives/StyledIconWrapper";
+import type { Type } from "..";
+
+const getTypeIcon = ({ type, active }) => {
+  if (type === "success" && active)
```
DSil, line 12: If it is success and last is it also success and active? If so, it will be animated, but it shouldn't 🤔 what determines if it is active? I'm looking at the _inside modal_ story and the last step has this animation but I feel it should only have the green check, as the process is apparently finished. Is it controllable and it's just a problem on the story? Or is there some logic missing here?

packages/orbit-components/src/Timeline/TimelineStep/index.d.ts
```diff
@@ -5,11 +5,10 @@ import * as React from "react";
 
 import * as Common from "../../common/common";
 
-export type Type = "success" | "warning" | "critical";
+export type Type = "success" | "warning" | "critical" | "info";
```
DSil, line 8: Should we have a story showing this new info type?

packages/orbit-components/src/Timeline/TimelineStep/primitives/StyledProgressLine.js
```diff
@@ -3,23 +3,48 @@ import styled, { css } from "styled-components";
...
-const renderStatus = (type, theme) => {
-  if (type === "success") return theme.orbit.colorTextSuccess;
-  if (type === "warning") return theme.orbit.colorTextWarning;
-  if (type === "critical") return theme.orbit.colorTextCritical;
+const getBorderStyle = ({ desktop, status, nextStatus, theme, last, prevStatus }) => {
+  if (desktop) {
+    if (status && !nextStatus && !last) {
+      return css`
+        border-image-slice: 1;
+        border-image-source: linear-gradient(
```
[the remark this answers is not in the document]
mainframev: here it [has](https://www.figma.com/file/wziyIIJCCQi2Ote2PnFfm6/Timeline?node-id=1511%3A574) ;)

packages/orbit-components/src/Timeline/TimelineStep/components/TypeIcon.jsx
mainframev: yes, it should be controlled, you are right ;) thanks

packages/orbit-components/src/Timeline/TimelineStep/primitives/StyledProgressLine.js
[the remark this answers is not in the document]
DSil: Hmm, but not [here](https://www.figma.com/file/wziyIIJCCQi2Ote2PnFfm6/Timeline). How do we know where to look for? 🤔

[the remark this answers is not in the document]
mainframev: I do not know why WIll put it there, but in PM messages he sent me that link 😕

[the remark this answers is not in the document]
DSil: Alright, so I'll consider that he sent you the right one 😄

packages/orbit-components/src/Timeline/README.md
```diff
@@ -40,13 +40,13 @@ import TimelineStep from "@kiwicom/orbit-components/lib/Timeline/TimelineStep";
...
-| type         | [`enum`](#enum) |         | Type of current process step                                              |
-| subLabel     | `string`        |         | SubLabel text of current process step                                     |
-| asText       | `boolean`       |         | Label is rendered as `Text` instead of `Badge`                            |
+| Name         | Type            | Default | Description                          |
+| :----------- | :-------------- | :------ | :----------------------------------- |
+| **children** | `React.Node`    |         | The content of the component         |
+| label        | `React.Node`    |         | Text for `label` component inside    |
+| subLabel     | `React.Node`    |         | Text for `subLabel` component inside |
+| type         | [`enum`](#enum) |         | Type of current process step         |
+| active       | `boolean`       |         | Controlled state of the step ds      |
```
DSil, line 49:
```suggestion
| active       | `boolean`       |         | Controlled state of the step      |
```
ds?

packages/orbit-components/src/Timeline/TimelineStep/index.jsx
```diff
@@ -48,25 +45,24 @@ const TimelineStep = ({ children, label, subLabel, type, asText }: Props): React
...
         {children}
       </TimelineStepDesktop>
     );
 
   return (
     <TimelineStepMobile
-      typeIcon={<TypeIcon type={type} />}
       nextType={nextType}
       label={label}
-      asText={asText}
```
DSil, line 68: If we're removing this prop this is a breaking change, no?

mainframev: yes, mentioned that in the PR description

DSil: Oh, I missed that. I don't know for sure if we can/should trust the tracking, because as it is open source we never know for sure who else might be using it, and so we should respect the breaking change. But I'll leave the decision to you.

mainframev: why? the purpose of the tracking exact for that reason, to make decisions about breaking changes

mainframev: kinda agree about the open source, but for us the product is prio, so we should more care about that. I do not want to block that changes, because in case if we would decide to release it with other breaking changes, it would have to wait a couple of weeks 🤔

DSil (approved):

PR #3613, opened 2022-10-20
feat(icons): change icon sizes according to figma

mainframev:
- [x] change icon size tokens
- [x] run loki

 Storybook: https://orbit-mainframev-feat-icon-size-changes.surge.sh


packages/orbit-components/src/CarrierLogo/index.tsx
```diff
@@ -98,11 +98,9 @@ export const StyledCarrierLogo = styled.div<{
   ${({ theme, carriers, size, inlineStacked }) => css`
     background-color: ${theme.orbit.backgroundCarrierLogo};
     height: ${carriers.length > 1 && !inlineStacked
-      ? theme.orbit.heightIconLarge
-      : `${getRenderSize({ theme, size })}px`};
-    width: ${carriers.length > 1
-      ? theme.orbit.widthIconLarge
+      ? theme.orbit.spaceXLarge
```
DSil, line 101: Should we be using spacing tokens to define the height of elements? Sounds odd to me

DSil (commented):
Looks ok except for the small question raised. I am just a bit unsure about how we should communicate this change. It can be a visual breaking change with icons (unexpectedly) changing its size. How will the commits be arranged to give a proper message on the changelog?

packages/orbit-components/src/CarrierLogo/index.tsx
mainframev: right, created the token for CarrierLogo

DSil (commented):
LGTM, but tests are failing, so please fix it. And please also adjust the commits for a proper changelog message 🤞

DSil (approved):

PR #3614, opened 2022-10-21
fix: focus state multiple components

DSil:
This PR changes the focus in multiple components to look like the native browser ones
 Storybook: https://orbit-mainframev-fix-focus-state-multiple-components.surge.sh

### DIFF
--- a/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.jsx
+++ b/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.jsx
@@ -28,16 +28,10 @@ const StyledBreadcrumbsItemAnchor = styled(
       transition: color ${theme.orbit.durationFast} ease-in-out;
       cursor: pointer;
 
-      &:hover,
-      &:focus {
+      &:hover {
         outline: none;
         text-decoration: none;
-      }
-      &:hover {
         color: ${theme.orbit.paletteProductNormalHover};
-      }
-      :focus {
-        color: ${theme.orbit.paletteProductNormalActive};
       }
     `};
   `}
--- a/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.jsx
+++ b/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.jsx
@@ -27,9 +27,14 @@ const StyledBreadcrumbsItemAnchor = styled(
     css`
       transition: color ${theme.orbit.durationFast} ease-in-out;
       cursor: pointer;
+      border: none;
+      background-color: inherit;
+      padding: 0;
+      outline-offset: 1px;
+      font-size: inherit;
+      font-weight: inherit;
 
       &:hover {
-        outline: none;
         text-decoration: none;
         color: ${theme.orbit.paletteProductNormalHover};
       }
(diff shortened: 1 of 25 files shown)


packages/orbit-components/src/utils/common/index.js
```diff
@@ -1,2 +1,9 @@
 // @flow
+import { css } from "styled-components";
+
 export const firstToUpper = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
+
+export const defaultFocus = (): any => css`
+  outline: 1px solid Highlight;
```
mainframev, line 7: wdyt? maybe default focus should have 2px, just check the difference between default browser outline and that utility ;)

packages/orbit-components/src/Tag/index.jsx
```diff
@@ -120,16 +115,7 @@ export const StyledTag: any = styled.div`
         box-shadow: none;
       }
 
-      &:focus {
-        background: ${getBackgroundColor(STATES.DEFAULT)};
```
mainframev, line 124: it should have focus-ring and background. So you will have to get background back ;)

mainframev (changes requested):

packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.jsx
```diff
@@ -28,17 +28,11 @@ const StyledBreadcrumbsItemAnchor = styled(
...
-      &:hover,
-      &:focus {
+      &:hover {
         outline: none;
         text-decoration: none;
-      }
-      &:hover {
         color: ${theme.orbit.paletteProductNormalHover};
       }
-      :focus {
```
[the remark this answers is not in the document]
DSil: I just noticed that problem was actually already happening, fixed it to render a button and added some styles reset so they all look alike.

packages/orbit-components/src/InputField/helpers/formElementFocus.js
```diff
@@ -1,19 +1,6 @@
 // @flow
-import { css } from "styled-components";
-import { convertHexToRgba } from "@kiwicom/orbit-design-tokens";
+import { defaultFocus } from "../../utils/common";
 
-const formElementFocus = (): any => css`
```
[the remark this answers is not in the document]
DSil: Yes, it was agreed with design to look like this :)

packages/orbit-components/src/utils/common/index.js
DSil: Right. I don't know where I got the 1px from 🤯

packages/orbit-components/src/SegmentedSwitch/SwitchSegment/index.jsx
```diff
@@ -64,7 +64,8 @@ const StyledInput = styled.input`
       z-index: 10;
       border-radius: 5px !important;
       color: ${theme.orbit.paletteInkNormal};
-      box-shadow: 0 0 0 2px ${theme.orbit.paletteBlueNormal};
+      outline: 1px solid Highlight;
```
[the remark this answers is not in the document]
DSil: It should actually use the defaultFocus 😝

packages/orbit-components/src/Tag/index.jsx
DSil: I am ok with that, but does it really? I thought that was because hover and focus styles were kinda mixed.
So, if I do that, when we have a tag focused and we hover it, we have no visual feedback that we are hovering it. Should it be like that?
For now, I'll add the background effect back as you requested, but I think we should confirm with designers, if it hasn't been confirmed already.

mainframev: this background color applies after you click on it. So I guess we should not care about hover, because it was already hovered and clicked. If it was focused by Tab, then you have a cursor pointer and it should be enough to let you know that it's hovered ;) Ask designers to be sure on 100% percents, but that how it was designed initially ;)

DSil: Clarified [here](https://skypicker.slack.com/archives/GSGN9BN6Q/p1666858014346839).

mainframev (approved):

PR #3620, opened 2022-10-26
fix: missing onShow prop for MobileDialog inside Tooltip 

mainframev:
[Slack request #1](https://skypicker.slack.com/archives/CAMS40F7B/p1666695579504919)
[Slack request #2](https://skypicker.slack.com/archives/CAMS40F7B/p1666696607559829)
 Storybook: https://orbit-mainframev-fix-missing-on-shown.surge.sh

### DIFF
--- a/packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.jsx
+++ b/packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.jsx
@@ -18,15 +18,17 @@ afterAll(() => {
 });
 describe("MobileDialogPrimitive", () => {
   it("should have expected DOM output", () => {
+    const onShown = jest.fn();
     render(
       // eslint-disable-next-line jsx-a11y/tabindex-no-positive
-      <MobileDialog tabIndex="1" dataTest="test" content="My text link">
+      <MobileDialog tabIndex="1" dataTest="test" onShow={onShown} content="My text link">
         children
       </MobileDialog>,
     );
     const children = screen.getByText("children");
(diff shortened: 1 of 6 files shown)


packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.jsx
```diff
@@ -18,15 +18,17 @@ afterAll(() => {
...
       // eslint-disable-next-line jsx-a11y/tabindex-no-positive
-      <MobileDialog tabIndex="1" dataTest="test" content="My text link">
+      <MobileDialog tabIndex="1" dataTest="test" onShow={onShown} content="My text link">
         children
       </MobileDialog>,
     );
     const children = screen.getByText("children");
     expect(children).toHaveAttribute("tabindex", "1");
     userEvent.click(children);
+    expect(onShown).toHaveBeenCalled();
```
DSil, line 31: I think we can be more precise here and assert that it was called only once. By doing this, we are asserting that the function is called, but it can be called for whatever reason, or even multiple times. If we assure it is only called once, at least we can know it is not called on every state update. WDYT?

DSil (commented):
Just a non-blocking question about the test. Everything else LGTM 👍

packages/orbit-components/src/primitives/MobileDialogPrimitive/__tests__/index.test.jsx
mainframev: makes sense, fixed

DSil (approved):

PR #3621, opened 2022-10-27
test: remove test warnings

mainframev:
Removed the test warnings. One issue is connected with `react-popper` and there is nothing that we can do currently, only hide that test warnings. 
 Storybook: https://orbit-mainframev-fix-test-warnings.surge.sh

### DIFF
--- a/packages/orbit-components/src/Box/index.jsx
+++ b/packages/orbit-components/src/Box/index.jsx
@@ -10,8 +10,8 @@ import normalize from "./normalize";
 import type { Props } from ".";
 
 const StyledBox: any = styled(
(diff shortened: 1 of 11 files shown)


DSil (commented):
LGTM, but I noticed yet another warning. So on LinkList it is passing the `legacy?{true}` prop to the `<li>` element, but it should not.
```
Warning: Received `true` for a non-boolean attribute `legacy`.
      
If you want to write it to the DOM, pass a string instead: legacy="true" or legacy={value.toString()}.
```
So maybe we should just extract it on the `StyledNavigationLinkListChild` definition and avoid it being passed to the html `li` element.

DSil (approved):

PR #3631, opened 2022-10-28
docs(Navbar): remove useMediaQuery, appears only after hydration

mainframev:
Just a small refactoring, supposed to remove that delay with hydration, only after which the links inside navbar appear. 
 Storybook: https://orbit-mainframev-fix-navbar-hydartion.surge.sh

### DIFF
--- a/docs/babel.config.js
+++ b/docs/babel.config.js
@@ -1,6 +1,7 @@
 module.exports = {
   presets: [require.resolve("babel-preset-gatsby"), require.resolve("@babel/preset-typescript")],
   plugins: [
+    "@kiwicom/orbit-components",
     require.resolve("babel-plugin-styled-components"),
     [
       require.resolve("babel-plugin-inline-react-svg"),
--- a/docs/src/components/Navbar/Links.tsx
+++ b/docs/src/components/Navbar/Links.tsx
@@ -1,0 +1,97 @@
+import React from "react";
+import { Hide, TextLink, Stack, Button, Popover } from "@kiwicom/orbit-components";
+import { MenuHamburger } from "@kiwicom/orbit-components/lib/icons";
+
+import PuzzleIcon from "../icons/Puzzle";
+import ColorBrushIcon from "../../images/streamline-light/color-brush.svg";
+import ReadArt from "../../images/streamline-light/read-art.svg";
+import IrisScanIcon from "../../images/streamline-light/iris-scan.svg";
+import ScriptIcon from "../../images/streamline-light/script.svg";
+
+type Link = "Components" | "Guides" | "Tokens" | "Accessibility" | "Foundation";
+
+interface Props {
+  links?: Link[];
+}
+
+const LinkIcon = ({ name }: { name?: string }) => {
+  const iconStyles = {
+    stroke: "#252A31",
+    width: "20px",
+    height: "20px",
+  };
+
+  switch (name) {
+    case "Components":
+      return <PuzzleIcon {...iconStyles} />;
+    case "Guides":
+      return <ReadArt {...iconStyles} />;
+    case "Tokens":
+      return <ColorBrushIcon {...iconStyles} />;
+    case "Accessibility":
+      return <IrisScanIcon {...iconStyles} />;
+    default:
+      return <ScriptIcon {...iconStyles} />;
+  }
+};
+
+const Icon = ({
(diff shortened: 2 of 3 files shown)


docs/src/components/Navbar/Links.tsx
```diff
@@ -0,0 +1,84 @@
...
+      case "Tokens":
+        return <ColorBrushIcon stroke="#252A31" width="20px" height="20px" />;
+      case "Accessibility":
+        return <IrisScanIcon stroke="#252A31" width="20px" height="20px" />;
+      default:
+        return <ScriptIcon stroke="#252A31" width="20px" height="20px" />;
+    }
+  }
+
+  return <>{isFirst ? null : <span>&sdot;</span>}</>;
```
DSil, line 41: 💅 Fragment not needed here

docs/src/components/Navbar/Links.tsx
```diff
@@ -0,0 +1,84 @@
...
+  isDesktop,
+}: {
+  name?: string;
+  isFirst?: boolean;
+  isDesktop?: boolean;
+}) => {
+  if (!isDesktop) {
+    switch (name) {
+      case "Components":
+        return <PuzzleIcon stroke="#252A31" width="20px" height="20px" />;
```
DSil, line 29: Those styles (stroke, width, height) are common to all icons, so I'd rather have them stored in a const object and spread here, so that if one changes, they all change at the same time.

docs/src/components/Navbar/Links.tsx
```diff
@@ -0,0 +1,84 @@
...
+import IrisScanIcon from "../../images/streamline-light/iris-scan.svg";
+import ScriptIcon from "../../images/streamline-light/script.svg";
+
+type Link = "Components" | "Guides" | "Tokens" | "Accessibility" | "Foundation";
+
+interface Props {
+  links?: Link[];
+}
+
+const LinkIcon = ({
```
DSil, line 17: Honestly I'm not sure if this is the best approach, in terms of readability/maintainability. Because it conditionally renders either an icon or the dot separating the links. Shouldn't it be just for the icon and then have the condition of either rendering the Icon (this component) or the dot on the `LinkList` component? (Suggestion below)

docs/src/components/Navbar/Links.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return <>{isFirst ? null : <span>&sdot;</span>}</>;
+};
+
+const LinkList = ({ isDesktop, links }: { isDesktop?: boolean; links: Props["links"] }) => (
+  <>
+    {links &&
+      links.map((link, idx) => (
+        <Stack inline align="center">
+          <LinkIcon name={link} isFirst={idx === 0} isDesktop={isDesktop} />
```
DSil, line 49:
```suggestion
         {isDesktop ? idx === 0 ? null : <span>&sdot;</span> : <LinkIcon name={link} />}
```
And then remove the need to have `isDesktop` and `isFirst` props on `LinkIcon`. The problem here is that apparently we have an EsLint rule to disallow nested ternary. I don't necessarily disagree with this rule, but I also don't necessarily agree with it 😅
Simply because, in this situation, I think the code is more readable even with the nested ternary than having the icon/dot logic on the LinkIcon component.
WDYT?

mainframev: think it's ok, except that you suggest to use double ternary

mainframev: ```{isDesktop && idx >= 0 ? <span>&sdot;</span> : <LinkIcon name={link} />}```

should be better ;)

docs/babel.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   presets: [require.resolve("babel-preset-gatsby"), require.resolve("@babel/preset-typescript")],
   plugins: [
+    "@kiwicom/orbit-components",
```
DSil, line 4: What's this and why is it needed? 🤔

docs/src/components/Navbar/Links.tsx
DSil: Good approach you took 👍

docs/babel.config.js
mainframev: added our babel plugin to config

PR #3636, opened 2022-11-02
docs: change homepage image

DSil:
Rocket image is now using the new gatsby-plugin-image and no longer has an enter transition.
Grey background on the homepage was added on desktop version
 Storybook: https://orbit-mainframev-docs-change-homepage-image.surge.sh

### DIFF
--- a/docs/src/components/HeaderBox.tsx
+++ b/docs/src/components/HeaderBox.tsx
@@ -1,0 +1,24 @@
+import styled, { css } from "styled-components";
+import { mediaQueries as mq } from "@kiwicom/orbit-components";
+
+import { MAX_CONTENT_WIDTH } from "../consts";
+
+const HeaderBox = styled.div`
+  display: block;
+  border-radius: 24px;
+  width: 100%;
+  max-width: ${MAX_CONTENT_WIDTH};
+  max-height: 530px;
+  margin: 0 auto;
+
+  ${mq.desktop(css`
+    position: relative;
+    height: 100vh;
+    padding: 64px;
+    background: linear-gradient(80.68deg, #f1f4f7 7.05%, #f5f7f9 57.05%);
+    z-index: 0;
+    margin-bottom: -200px;
+  `)}
+`;
+
+export default HeaderBox;
--- a/docs/src/consts.ts
+++ b/docs/src/consts.ts
@@ -1,2 +1,2 @@
-export const MAX_CONTENT_WIDTH = "80rem";
+export const MAX_CONTENT_WIDTH = "90rem";
 export const CONTENT_PADDING = "2rem";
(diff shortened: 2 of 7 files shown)


docs/src/components/HeaderBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { MAX_CONTENT_WIDTH } from "../consts";
+
+const HeaderBox = styled.div`
+  ${mq.desktop(css`
+    position: relative;
+    height: 100vh;
+    padding: 64px;
+    background: linear-gradient(80.68deg, #f1f4f7 7.05%, #f5f7f9 57.05%);
+    z-index: 0;
```
mainframev, line 12: could you place mq query below the normal CSS code? With mobile-first approach it's better to read it top-down ;)

docs/src/components/HeaderBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+  `)}
+
+  display: block;
+  border-radius: 24px;
+  border-bottom-right-radius: 50%;
```
mainframev, line 17: could you explain that? I just don't get why it has to be 50%, it's visible only while the image is not loaded

docs/src/components/HeaderBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+  width: 100%;
+  max-width: ${MAX_CONTENT_WIDTH};
```
mainframev, line 19: I would not use this constant here. Just set it as 90rem. It will not be probably used anywhere else.

docs/src/consts.ts
```diff
@@ -1,2 +1,2 @@
-export const MAX_CONTENT_WIDTH = "80rem";
+export const MAX_CONTENT_WIDTH = "90rem";
```
mainframev, line 1: Keep it as it was before. You will not need to do calculations then ;)

mainframev (changes requested):

mainframev:
LGTM. Just a few small things.

docs/src/components/HeaderBox.tsx
DSil: Yeah, I didn't think of that. It was set to 50% because of the image positioning not matching the border radius, but now I just noticed that it actually matches perfectly, so I'll remove it.

docs/src/components/HeaderBox.tsx
DSil: Hmm I don't think I can. Because the TopBar and the footer are supposed to have the same width as the HeaderBox. The content is the one that is actually smaller, hence the change. The content stays at 80rem while everything else grows to 90 rem. WDYT?

docs/src/consts.ts
DSil: Justified above ☝️

mainframev (approved):

PR #3640, opened 2022-11-03
chore: add stylelint and fix stylelint errors

mainframev:
I guess we should add `stylelint`, to prevent such issues as in #3638 

It's can be added for now like that. The [processor](https://github.com/styled-components/stylelint-processor-styled-components) mentioned in the SC documentation seems to be deprecated, so I had to turn off some of the rules, but it's catching the issues mentioned above with `TextLink`
 Storybook: https://orbit-mainframev-chore-add-stylelint.surge.sh


packages/orbit-components/src/Toast/ToastMessage.jsx
```diff
@@ -56,7 +56,6 @@ const StyledInnerWrapper = styled.div`
       width: initial;
       padding: ${theme.orbit.spaceSmall};
     `)}
-    }
```
DSil, line 59: I wonder how this was not a problem before 🤯

DSil (approved):

packages/orbit-components/src/Toast/ToastMessage.jsx
mainframev: SC parses that later and fixing :)

PR #3642, opened 2022-11-08
Docs: add custom topbar links + fix tabs documentation page

DSil:
It is now possible to add custom storybook and headerlinks (usually for Github) on the meta.yml file.
This also allows to hide the button when it is defined to `false`.

The Tabs documentation page was broken and is now fixed.
 Storybook: https://orbit-mainframev-docs-add-custom-topbar-links.surge.sh

### DIFF
--- a/docs/gatsby-node.ts
+++ b/docs/gatsby-node.ts
@@ -148,10 +148,19 @@ export const onCreateNode = async ({ cache, node, getNode, actions, reporter }) 
       return "";
     };
 
+    const headerLink =
+      metaFileData.headerLink || getLinkForComponents() || node.frontmatter.headerLink;
+
     createNodeField({
       node,
       name: "headerLink",
-      value: getLinkForComponents() || node.frontmatter.headerLink || metaFileData.headerLink,
+      value: headerLink,
+    });
+
+    createNodeField({
+      node,
+      name: "hasHeaderLink",
+      value: metaFileData.hasHeaderLink ?? Boolean(headerLink),
     });
 
     if (node.fields.collection === "documentation") {
@@ -260,6 +269,7 @@ export const createSchemaCustomization: GatsbyNode["createSchemaCustomization"] 
       title: String!
       breadcrumbs: [BreadcrumbsPart]
       storybookLink: String
+      headerLink: String
     }
 
     type BreadcrumbsPart {
--- a/docs/gatsby-node.ts
+++ b/docs/gatsby-node.ts
@@ -93,6 +93,12 @@ export const onCreateNode = async ({ cache, node, getNode, actions, reporter }) 
 
     createNodeField({
       node,
+      name: "hasStorybook",
+      value: metaFileData.hasStorybook ?? dir.startsWith("/03-components"),
+    });
+
+    createNodeField({
+      node,
       name: "storybookLink",
       value: metaFileData.storybook,
     });
@@ -245,8 +251,11 @@ export const onPreBuild = async () => {
   await parseChangelog();
 };
 
-export const createSchemaCustomization: GatsbyNode["createSchemaCustomization"] = ({ actions }) => {
-  actions.createTypes(
+export const createSchemaCustomization: GatsbyNode["createSchemaCustomization"] = ({
+  actions,
+  schema,
+}) => {
+  const typeDefs = [
     `
     type Mdx implements Node {
       frontmatter: MdxFrontmatter!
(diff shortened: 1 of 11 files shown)


docs/gatsby-node.ts
```diff
@@ -253,6 +259,8 @@ export const createSchemaCustomization: GatsbyNode["createSchemaCustomization"]
       tabCollection: String
       title: String!
       breadcrumbs: [BreadcrumbsPart]
+      storybookLink: String
```
mainframev, line 262: instead of extending current types like that with string, do something like that: 

```suggestion
  const typeDefs = [
    `
  type Mdx implements Node {
    frontmatter: MdxFrontmatter!
    fields: MdxFields
  }

  type MdxFrontmatter {
    date: Date @dateformat
    description: String
    redirect_from: [String]
    title: String!
    type: String
  }

  type BreadcrumbsPart {
    name: String!
    url: String!
  }
`,
    schema.buildObjectType({
      name: "MdxFields",
      fields: {
        breadcrumbs: {
          type: "[BreadcrumbsPart]",
        },
        collection: "String!",
        description: "String",
        slug: "String!",
        tabCollection: "String",
        title: "String!",
        hasHeaderLink: {
          type: "Boolean",
          resolve: source => !!source.headerLink,
        },
        hasStorybook: {
          type: "Boolean",
          resolve: source => !source.storybookLink,
        },
      },
    }),
  ];

  actions.createTypes(typeDefs);
```

mainframev (changes requested):

docs/src/components/DocLayout/TopBar.tsx
```diff
@@ -103,14 +104,16 @@ const TopBar = ({
               {headerLink && (
                 <Hide on={["smallMobile", "mediumMobile"]}>
                   <Stack flex spacing="XXSmall">
-                    <HeaderButtonLink href={headerLink} />
-                    {title && (
+                    {headerLink !== "false" && <HeaderButtonLink href={headerLink} />}
```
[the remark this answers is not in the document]
DSil: I agree. My initial approach included this, but then I thought I could optimize and avoid it. Truth is I think I made it more confusing, so I followed your advice and added the two fields. I think it looks better now 👍

mainframev (approved):

docs/gatsby-node.ts
```diff
@@ -91,6 +91,22 @@ export const onCreateNode = async ({ cache, node, getNode, actions, reporter })
     const metaFileData = (await cache.get(getParentUrl(omitNumbers(fileUrl)))) || {};
     const hasTabs = metaFileData.type === "tabs";
 
+    const getDefaultHasStorybook = () => {
```
mainframev, line 94: no need for that function, use the condition directly below

docs/src/components/DocLayout/index.tsx
```diff
@@ -52,11 +53,15 @@ interface Props {
     url: string;
   }>;
   custom?: boolean;
+  noTopBar?: boolean;
+  hasStorybook?: boolean;
+  storybookLink?: false | string;
```
mainframev, line 58: set it to a boolean and when it's pass down to this component add `!!storybookLink or Boolean(storybookLink)`

mainframev (changes requested):

docs/src/components/DocLayout/TopBar.tsx
```diff
@@ -49,10 +48,16 @@ const TopBar = ({
...
   location,
   tocHasItems,
+  hasHeaderLink,
   headerLink,
   breadcrumbs,
   description,
+  hasStorybook,
+  storybookLink,
 }) => {
+  const hasTabs = tabs && tabs.length > 1;
```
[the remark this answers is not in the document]
DSil: If the page only has one tab we don't need to display that as a tab (it was even displaying just an empty blank tab, not the actual title of the tab), because if there is no navigation to other tabs, why display tabs at all?

docs/src/components/DocLayout/index.tsx
DSil: Which type are you referring to? `storybookLink`? I think it's better to have it like this so we can never pass `true` to the prop

docs/src/components/DocLayout/TopBar.tsx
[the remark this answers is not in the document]
mainframev: there is might be a case, when it released only in Figma and have guidelines tab, but not react

docs/src/components/DocLayout/index.tsx
mainframev: but storybookLInk can be only a string or null, no? hasStorybook is supposed to be a boolean. In schema it's type string, so if you will pas false there in yaml file it should throw error

DSil: Oh, yes, it should be string. It's a leftover from previous implementation. Fixed

docs/src/components/DocLayout/TopBar.tsx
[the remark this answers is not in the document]
DSil: Right, missed that case. Added

mainframev (approved):

PR #3643, opened 2022-11-09
fix(Itinerary): render text as div

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1667995401356349)
 Storybook: https://orbit-mainframev-fix-itinerary-stop-render-text-as-div.surge.sh

### DIFF
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
@@ -98,6 +98,7 @@ const ItinerarySegmentStop = ({
             {time && (
               <Text
                 weight="medium"
+                as="div"
                 type={cancelledTime ? textType : "primary"}
                 withBackground={!!cancelledTime}
               >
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.jsx
```diff
@@ -102,6 +102,7 @@ const ItinerarySegmentStop = ({
             {time && (
               <Text
                 weight="medium"
+                as="div"
```
DSil, line 105: I am not a fan of this. From the request, I've seen they are trying to pass pretty much anything to these props, so should we do anything at all? Rendering a `Text` as `div` will allow them to pass direct strings and have them rendered as `div` when they should be `p`. Based on that, who should be responsible for that styling? What is the point of styling this with a "Text as div" wrapper if then they are putting other components with their own (overridden) styles inside?

mainframev: It's an edge case when they need to pass CountryFlag and Text together in the case with THC (See the slack thread). 
When they need to pass a station, they will not use the Text, just plain string. I guess I answered your question

DSil: > When they need to pass a station for a classic SegmentStop, they will not use the Text, just plain string.

That was my concern. Is that correct? To have just plain text inside a div styled as a paragraph? It feels weird to me.
I understand the decision, and I agree that we should avoid complicating the API even more. But then we should probably consider how the implementation is and what are the real use cases that are possible

mainframev: As far as I know, it's still correct according to a11y, ofc it's preferable to have paragraphs, but currently, it seems a quick solution. Let's bare it in my mind and we can discover later, how we can improve that. Especially since we are allowed to do the same in our frontends via `as` prop.

DSil (approved):

PR #3645, opened 2022-11-11
chore: small type fixes

mainframev:
- removed redundant type exports
- adjusted ts template files


packages/orbit-components/src/hooks/useTheme/index.ts
```diff
@@ -1,8 +1,8 @@
 import { useContext } from "react";
 import { ThemeContext } from "styled-components";
 
-import defaultTheme from "../../defaultTheme";
+import defaultTheme, { Theme } from "../../defaultTheme";
 
-const useTheme = () => useContext(ThemeContext) || defaultTheme;
+const useTheme = (): Theme => useContext(ThemeContext) || defaultTheme;
```
DSil, line 6: Non-blocking, but I think the Theme type could have a different name. Conventions suggest something like `ITheme` but I'm also not a big fan of it. I'd just like to not have a type called `Theme` because it can mislead with an actual theme. If anything, it can stay like this, but I'd like you to think about it. WDYT?

mainframev: as far as I know, naming interfaces with `I` is not recommended. I do not think it can mislead with an actual theme, because of lowercase and uppercase. Also, we do not much use `theme` except styled-components, usually, it's `defaultTheme`.

DSil: Hmm alright, nothing to do here then 👍

PR #3651, opened 2022-11-15
feat(InputField): set autoCorrect and autoCapitalize explicitly to off

mainframev:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#autocorrect

[why](https://baymard.com/labs/touch-keyboard-types)

### DIFF
--- a/packages/orbit-components/src/InputField/index.tsx
+++ b/packages/orbit-components/src/InputField/index.tsx
@@ -251,6 +251,8 @@ Suffix.defaultProps = {
 interface StyledInputProps extends Partial<Props> {
   min: number;
   max: number;
+  autoCorrect: string;
+  autoCapitalize: string;
   ariaLabelledby?: string;
 }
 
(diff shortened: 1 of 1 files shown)


DSil:
Can't we have it with the default value as off but still allow to customize it, if needed?

DSil (approved):

DSil:
However, for the autocapitalize, I find it weird that it is recommended to be always off. I'm thinking about a booking form where I have to input my name, I expect it to be capitalized by default… 🤔

mainframev:
> However, for the autocapitalize, I find it weird that it is recommended to be always off. I'm thinking about a booking form where I have to input my name, I expect it to be capitalized by default… 🤔

No. It does not recommend it to be off always. It's just recommended not to use the attributes, that only partially supported. 

> The autocapitalize attribute doesn't affect behavior when typing on a physical keyboard. Instead, it affects the behavior of other input mechanisms, such as virtual keyboards on mobile devices and voice input.

PR #3653, opened 2022-11-16
docs(DocsLayout): fix max-width

mainframev:
fixed the container width for DocLayout, it has to be the same as navigation bar

### DIFF
--- a/docs/src/components/DocLayout/index.tsx
+++ b/docs/src/components/DocLayout/index.tsx
@@ -37,6 +37,7 @@ import StyledProse from "./primitives/StyledProse";
 import StyledMobileTocWrapper from "./primitives/StyledMobileTocWrapper";
 import { getDocumentPageTitle } from "../../utils/document";
 import TopBar from "./TopBar";
+import { MAX_CONTENT_WIDTH } from "../../consts";
 
 interface Props {
   children: React.ReactNode;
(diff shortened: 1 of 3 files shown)


docs/src/components/DocLayout/index.tsx
```diff
@@ -113,7 +113,9 @@ export default function DocLayout({
               display: flex;
               margin: 0 auto;
               width: 100%;
-              max-width: 80rem;
+              box-sizing: border-box;
+              padding: 0 2rem;
+              max-width: 90rem;
```
DSil, line 118: We can use the `MAX_CONTENT_WIDTH` global const, no?

docs/src/components/Navbar/index.tsx
```diff
@@ -44,7 +44,6 @@ const StyledWrapper = styled.header`
 const StyledInner = styled.div`
   max-width: ${MAX_CONTENT_WIDTH};
   padding: 0 ${CONTENT_PADDING};
-  box-sizing: content-box;
```
DSil, line 47: If we remove this, I also don't think it gets according to the designs 🤔 it gets smaller than the content in the homepage

docs/src/components/DocLayout/index.tsx
mainframev: yes

docs/src/components/Navbar/index.tsx
mainframev: now it's correct, in Figma it's 1440px including paddings, before it was bigger.

mainframev: changed the size of HeaderBox, so it's visually ok

DSil (approved):

PR #3656, opened 2022-11-17
Docs: add Orbit Themer to docs

DSil:
Orbit Themer now lives inside our documentation 🎉


packages/orbit-themer/package.json
```diff
@@ -1,13 +1,15 @@
...
+  "version": "1.0.0",
   "private": "true",
   "description": "Design your own orbit-components theme with custom colors",
-  "main": "index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "dev": "webpack-dev-server --mode development --open --hot --host 0.0.0.0",
-    "flow": "flow",
```
mainframev, line 9: could you rewrite it to a typescript? 

simply rewrite files from `jsx` to `tsx` and add tsconfig, build it with babel.

docs/src/pages/themer/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import Themer from "@kiwicom/orbit-themer";
+
+import DocLayout from "../../components/DocLayout";
+
+export default ({ location }) => {
+  return (
+    <DocLayout
```
mainframev, line 8: Instead of `DocLayout`, I suggest rendering it inside a different Layout, without SideNavigation. It will visually look better in my opinion, links in the navbar and logo should be enough to navigate, plus breadcrumbs are there.

packages/orbit-themer/package.json
mainframev: or you can also check how we use `tsup` inside `orbit-tracking` or `orbit-design-tokens` and also use it ;)

mainframev (changes requested):
LGTM. Few suggestions how to improve.

packages/orbit-themer/package.json
DSil: Done

docs/src/pages/themer/index.tsx
DSil: Done. Although the navbar links are not there, because they're set to only be displayed on the homepage. Should we change that rule...?

mainframev: yeah, i think for themer it makes sense :)

mainframev (changes requested):

mainframev (approved):

PR #3660, opened 2022-11-21
docs: Present token values in hex

DSil:
Token values are now presented in hex, instead of rgba

[ORBIT-1922](https://kiwicom.atlassian.net/browse/ORBIT-1922)

### DIFF
--- a/docs/src/components/DesignTokensList/index.tsx
+++ b/docs/src/components/DesignTokensList/index.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from "react";
 import tokensList from "@kiwicom/orbit-design-tokens/output/theo-spec.json";
+import { convertRgbaToHex } from "@kiwicom/orbit-design-tokens";
 import {
   InputField,
   Stack,
@@ -27,6 +28,9 @@ const allTokens: Array<{ name: TokenName; value: TokenValue }> = Object.entries(
 
 const DesignToken = ({ value }: { value: TokenValueType }) => {
   const [isCopied, copy] = useCopyToClipboard(2000);
+  const formattedValue =
+    typeof value !== "string" || !value.startsWith("rgb") ? value : convertRgbaToHex(value);
+
   return isCopied ? (
     <Stack inline spacing="XSmall" shrink align="center">
       <StyledDesignTokenOther />
@@ -36,8 +40,12 @@ const DesignToken = ({ value }: { value: TokenValueType }) => {
     <Tooltip content="Click to copy">
       <Stack inline spacing="XSmall" shrink align="center">
         <DesignTokenIcon value={value} />
-        <button type="button" onClick={() => copy(String(value))} title={String(value)}>
-          <Truncate maxWidth="200px">{value}</Truncate>
+        <button
+          type="button"
+          onClick={() => copy(String(formattedValue))}
+          title={String(formattedValue)}
+        >
+          <Truncate maxWidth="200px">{formattedValue}</Truncate>
         </button>
       </Stack>
     </Tooltip>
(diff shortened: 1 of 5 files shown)


packages/orbit-design-tokens/src/convertRgbaToHex.ts
```diff
@@ -0,0 +1,12 @@
+export type RgbaToHex = (red: number, green: number, blue: number, alpha: number) => string;
+
+const convertRgbaToHex: RgbaToHex = (r, g, b, a) => {
```
mainframev, line 3: I would rewrite that function to accept only one parameter (rgba). 

and do something like that: 
```suggestion
const convertRgbaToHex = (rgba: string) => {
  const [r, g, b, a] = rgba
    .replace("rgba(", "")
    .replace(")", "")
    .split(",")
    .map((v) => parseInt(v, 10));

  return `#${((1 << 24) + (r << 16) + (g << 8) + b).toString(16).slice(1)}`;
};
```

docs/src/components/DesignTokensList/index.tsx
```diff
@@ -25,8 +26,18 @@ const allTokens: Array<{ name: TokenName; value: TokenValue }> = Object.entries(
   .map(([name, value]) => ({ name, value }))
   .sort((a, b) => (a.name < b.name ? -1 : 1));
 
+const getFormattedValue = (value: TokenValueType): TokenValueType => {
```
mainframev, line 29: in case with my suggestion below you will not need that function

mainframev (changes requested):

packages/orbit-design-tokens/src/convertRgbaToHex.ts
DSil: I was considering something like that at first, but I think the readability of the code is worse. I personally prefer to have lines of code that I can read and understand, if possible. Unless there is a major gain in performance or effectiveness, I don't see a big advantage in changing the code… WDYT?

DSil: It's a bit of all. Also, it's not bulletproof, as it is not considering when it is just `rgb` (and not `rgba`). And it is missing the alpha conversion for the hex. It can be added, but the function won't be this cleaner, and readability will be (even more) worse, imo

DSil: It's a bit of all. Also, it's not bulletproof, as it is not considering when it is just `rgb` (and not `rgba`). And it is missing the alpha conversion for the hex. It can be added, but the function won't be this cleaner, and readability will be (even more) worse, imo

mainframev: anyway, my point is, that the function can look whatever you would like it to look, but should accept the string instead of several parameters. So it can be reused next time somewhere without implementing additional logic :)

DSil: Ah, got it, that can be done, for sure 😄

DSil: Eh… I tried to, but the `value` in DesignTokenList, which will be passed to this function, is of type `TokenValueType`, which is string or number. So I either do some verification beforehand, or I convert to string, but in both scenarios I thinks that is overcomplicating.
Also, if I call this function `convertRgbaToHex` and I pass a random string, I am not returning an Hex (should I return the same string or throw some error?), so the name is misleading, for it to be in such abstracted level. I think it should be the caller's responsibility to pass the right parameters, that is why I deconstructed it into multiple arguments at first. WDYT?

mainframev (approved):
LGTM 🚀

PR #3663, opened 2022-11-23
chore: remove deprecated components

mainframev:


DSil (changes requested):
I've found the following:
- There are examples on docs using `PictureCard` that should be removed
- `PictureCard`, `Sticky` and `RatingStars` are still being exported on `index.ts` inside orbit-components
- `PictureCard`, `Sticky` and `RatingStars` are still considered on orbit-tracking (should they?)
- `PictureCard`, `Sticky` and `RatingStars` are still on the babel-plugin

DSil (commented):
The Popover story is also using Sticky

DSil (approved):

PR #3664, opened 2022-11-23
fix(Box): wrap property was calculated incorrectly

DSil:
The wrap property in Box component was evaluating its value wrongly.
One test was also changed so it actually failed before and passes now.

[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1669201347964819)


mainframev (approved):
🚀

PR #3670, opened 2022-11-28
fix(Modal): don't render overlay if is full page

DSil:
When a Modal is full page on Modal, it does not render the dark overlay, avoiding a flash happening occasionally.
[Slack report](https://skypicker.slack.com/archives/CAMS40F7B/p1669303701390719)


mainframev (approved):

DSil:
@mainframev I tested it without the `animationDisabled` prop and I noticed no flicker. What is the scenario you've found a problem?

mainframev (approved):

PR #3675, opened 2022-12-01
fix(ItinerarySegment): fix access of hidden prop in the context

mainframev:


packages/orbit-components/src/Itinerary/ItinerarySegment/index.tsx
```diff
@@ -44,29 +44,34 @@ const ItinerarySegment = ({
...
               toggleOpened: () => setOpened(prev => !prev),
               last: i === React.Children.count(children) - 1,
-              // @ts-expect-error expected
-              isNextHidden: content[i + 1] && content[i + 1].props.hidden,
-              // @ts-expect-error expected
-              isPrevHidden: content[i - 1] && content[i - 1].props.hidden,
+              isNextHidden: !!(
+                content[i + 1] &&
+                content[i + 1].props &&
+                content[i + 1].props.hidden
```
DSil, line 65: I'd personally prefer if it was just
```suggestion
              isNextHidden: Boolean(content[i + 1]?.props?.hidden ?? false)
```

and the same for the one below

mainframev: suggestion is good, except nullish operator, not needed. Still will be evaluated to false

DSil: True 😃

DSil (approved):

PR #3678, opened 2022-12-01
feat: add margin utility properties

mainframev:
[ORBIT-2408](https://kiwicom.atlassian.net/browse/ORBIT-2408)
[ORBIT-2410](https://kiwicom.atlassian.net/browse/ORBIT-2410)
[ORBIT-2411](https://kiwicom.atlassian.net/browse/ORBIT-2411)

Margin utility prop for Text, Card and Illustration

### DIFF
--- a/packages/orbit-components/src/Card/__tests__/index.test.tsx
+++ b/packages/orbit-components/src/Card/__tests__/index.test.tsx
@@ -25,9 +25,9 @@ describe("Card", () => {
     expect(screen.getByRole("button"));
     expect(screen.getByTestId("test")).toHaveStyle({
       marginTop: "12px",
-      marginBottom: "0",
-      marginRight: "0",
-      marginLeft: "0",
+      marginBottom: undefined,
+      marginRight: undefined,
+      marginLeft: undefined,
     });
   });
 
--- a/packages/orbit-components/src/Card/__tests__/index.test.tsx
+++ b/packages/orbit-components/src/Card/__tests__/index.test.tsx
@@ -14,6 +14,7 @@ describe("Card", () => {
         title="kek"
         description="description"
         icon={<Airplane dataTest="airplane" />}
+        margin={{ top: "12px" }}
         actions={<Button>button</Button>}
       />,
     );
@@ -22,6 +23,12 @@ describe("Card", () => {
     expect(screen.getByText("description")).toBeInTheDocument();
     expect(screen.getByTestId("airplane"));
     expect(screen.getByRole("button"));
+    expect(screen.getByTestId("test")).toHaveStyle({
+      marginTop: "12px",
+      marginBottom: "0",
+      marginRight: "0",
+      marginLeft: "0",
+    });
   });
 
   it("section", () => {
--- a/packages/orbit-components/src/Card/index.tsx
+++ b/packages/orbit-components/src/Card/index.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 
 import Loading from "../Loading";
 import CardWrapper from "./components/CardWrapper";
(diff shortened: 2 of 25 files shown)


packages/orbit-components/src/Card/types.d.ts
```diff
@@ -6,10 +6,13 @@ import type * as React from "react";
 import type * as Common from "../common/types";
 import type { As } from "../Heading/types";
 
+// TODO: remove spaceAfter in the next major version
+/** spaceAfter is deprecated, use margin instead */
 export interface Props extends Common.Globals, Common.SpaceAfter {
   readonly children?: React.ReactNode;
   readonly title?: React.ReactNode;
   readonly titleAs?: As;
+  readonly margin?: string | Common.ObjectProperty;
```
DSil, line 15: Isn't it safer if we define the type as
```suggestion
  readonly margin?: React.CSSProperties["margin"] | Common.ObjectProperty;
```

This applies for all, ofc. WDYT?

mainframev: Hmm, always thought it was mostly for style prop, but it works  ok, except that the only difference is that it will allow string and number.

DSil: Yeah, it doesn't offer that much validation in the end (eg: it could validate the value passed is valid to define a margin). But it at least complies with the system, if they ever decide to apply that kind of validation 😄

packages/orbit-components/src/Card/index.tsx
```diff
@@ -68,7 +77,12 @@ const Card = ({
   };
 
   return (
-    <StyledCard spaceAfter={spaceAfter} data-test={dataTest} id={id}>
+    <StyledCard
+      // TODO: remove spaceAfter in the next major version
```
DSil, line 81: Should we to document that `spaceAfter` is deprecated?

packages/orbit-components/src/Card/__tests__/index.test.tsx
```diff
@@ -22,6 +23,12 @@ describe("Card", () => {
     expect(screen.getByText("description")).toBeInTheDocument();
     expect(screen.getByTestId("airplane"));
     expect(screen.getByRole("button"));
+    expect(screen.getByTestId("test")).toHaveStyle({
+      marginTop: "12px",
+      marginBottom: "0",
+      marginRight: "0",
+      marginLeft: "0",
+    });
```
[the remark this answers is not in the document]
mainframev: nope. It should be tested here too because it's also based on space after property. 
Isolated testing of the utility function is not guaranteed, that property is passed down to a styled component and will be applied. So it's ok in that case to have that test. We were testing spaceAfter before and do not think we should change it here.

packages/orbit-components/src/Card/index.tsx
mainframev: better will be to remove that from READMEs

DSil: Isn't it possible to add some linter warning when using the prop? Just removing from the docs seems a bit too silent

packages/orbit-components/src/Card/__tests__/index.test.tsx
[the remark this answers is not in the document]
DSil: I am not saying we should not test. We should. But the assertion shouldn't be the result of the function. Even with the spaceAfter prop we should not break the abstraction barrier. The components should trust that the function returns what is expected, because the function should be tested in its place. And therefore the test should be that the function is called when it should be called.

Also, what should be the expected behavior when one uses `margin` and `spaceAfter`? From reading the code, it seems that the `margin` prop is ignored. I don't think it should be like that. Also, there is no test for such scenario, and there should be, once it is correctly defined what should happen.

packages/orbit-components/src/utils/common/index.ts
```diff
@@ -1,9 +1,31 @@
...
   outline: 2px solid -webkit-focus-ring-color;
 `;
+
+export const marginUtility = (
+  margin?: React.CSSProperties["margin"] | ObjectProperty,
+): FlattenSimpleInterpolation | null => {
+  if (!margin) return null;
+  if (typeof margin === "string" || typeof margin === "number") {
+    return css`
+      margin: ${margin};
```
DSil, line 19: I noticed a problem here. If we use it like (on Card, for example):
`<Card margin="2" … />`
or
`<Card margin={2} … />`
it will produce this
image

The same happens with the object. So we have to treat the input, at least, when it is a number and add the "px".
If it is a string I think we can disregard any treatment.

DSil (changes requested):

packages/orbit-components/src/utils/common/index.ts
```diff
@@ -1,9 +1,34 @@
...
+
 export const firstToUpper = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
 
 export const defaultFocus = (): FlattenSimpleInterpolation => css`
   outline: 2px solid Highlight;
   outline: 2px solid -webkit-focus-ring-color;
 `;
+
+const setValue = (value: string | number): string =>
+  typeof value === "number" ? `${value}px` : value;
```
DSil, line 14:
```suggestion
  typeof value === "number" && !== 0 ? `${value}px` : value;
```

Because if it is 0, it should just be 0 💅

packages/orbit-components/src/utils/common/index.ts
```diff
@@ -1,9 +1,34 @@
...
+    `;
+  }
+
+  const { top, right, bottom, left } = margin;
+
+  return css`
+    margin-top: ${top && setValue(top)};
+    margin-right: ${right && setValue(right)};
+    margin-bottom: ${bottom && setValue(bottom)};
+    margin-left: ${left && setValue(left)};
```
DSil, line 32: If these values are 0, they will evaluate to false, no?

PR #3679, opened 2022-12-02
fix(Card): inconsistent expanded state is now impossible

DSil:
It is now impossible to define `expanded` or `initialExpanded` when `expandable` is not defined

### DIFF
--- a/packages/orbit-components/src/Card/CardSection/types.d.ts
+++ b/packages/orbit-components/src/Card/CardSection/types.d.ts
@@ -6,19 +6,30 @@ import type * as React from "react";
 import type * as Common from "../../common/types";
 import type { As } from "../../Heading/types";
 
-export interface Props extends Common.Globals {
+type ExpandableConditionalProps =
+  | {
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/Card/CardSection/types.d.ts
```diff
@@ -6,19 +6,30 @@ import type * as React from "react";
 import type * as Common from "../../common/types";
 import type { As } from "../../Heading/types";
 
-export interface Props extends Common.Globals {
+type ExpandableConditionalProps =
```
mainframev, line 9: Could you try to write a type abstraction for this? It's kinda becoming a pattern for our components, so it would be nice to have that

mainframev (changes requested):

packages/orbit-components/src/Card/CardSection/types.d.ts
DSil: I am not sure I understand your request. You want a generic conditional type?

PR #3622, opened 2022-10-27
docs: sync code and playground

DSil:
- [x] Adapt webpack configuration
- [ ] Have code and playground in sync
 Storybook: https://orbit-mainframev-docs-sync-code-and-playground.surge.sh

PR #3690, opened 2022-12-12
docs: make code persistent between tabs in playground

mainframev:
This PR is related to #3622


docs/src/components/ReactExample/transform.ts
```diff
@@ -1,95 +1,140 @@
...
+                true,
+              ),
             );
           }
 
-          return t.jsxAttribute(t.jsxIdentifier(name), t.stringLiteral(stringified));
+          if (typeof value === "boolean") return t.jsxExpressionContainer(t.booleanLiteral(value));
+          if (typeof value === "number") return t.jsxExpressionContainer(t.numericLiteral(value));
+          if (typeof value === "string") return t.stringLiteral(value);
+
```
DSil, line 62: I remember we hardcoded 🔨  this. Is there no turning back?

mainframev: We will solve this in another PR, that's outside of the scope of this, currently, controlled/uncontrolled components also do not work correctly :)

DSil:
Took some time but here's a slightly more extensive review. Some problems are unrelated but are improvements to be done that should generate tasks in our board.
- On `Accordion` playground, the `expandable` knob should be on by default, because the default example has this prop set to true.
- On `Alert`, the playground does not work.
- On `Breadcrumbs` the playground is odd. It asks for href… but for which item? Maybe we don't even need it for this component. But this one is non-blocking on this PR
- On `Button` and `ButtonLink` playgrounds, the size prop should be defined to `normal` by default, because that is what the components have by default
- On `ButtonMobileStore` playground, changing the `variant` prop has no effect. Probably unrelated to this PR.
- On `CountryFlag` playground, the default selected `code` is _ad_, but that is not what is rendered by default.
- The `HorizontalScroll` should have at least one more `Tile` so it really needs to scroll, when viewing in full screen. Unrelated to this PR.
- On `Icon`, `Illustration` and `IllustrationPrimitive` playgrounds, the size prop should be defined to `medium` by default, because that is what the components have by default
- On `InputFile` playground the `help` prop is empty even though there is an help tooltip being displayed.
- On `InputGroup`, when changing to Playground, the label changes to "This text is not displayed"
- On `LinkList` playground the `direction` prop is on `row` by default even though the component renders with `column` by default.
- On `Pagination`, the default rendered example changes completely when we change to Playground for the first time.
- On `Seat`, every time I switch from Editor to Playground, the `label` prop gets bigger (adds chars)
- On `ServiceLogo` playground the size prop should be defined to `medium` by default.
- On `Slider` playground I didn't see any effect on `histogramLoading` and `histogramLoadingText` props 🤔 
- On `SocialButton` playground the size prop should be defined to `normal` by default.
- On `Stepper`, the size prop does not seem to have any effect 🤔 
- On `Switch` there's that problem with switching from Playground to Editor disallows the visual effect of the checked switch when one interacts with it
- On `Table`, when selecting Playground for the first time, the rendered component changes
- On `Tag` playground the size prop should be defined to `normal` by default.
- On `Tile` playground the icon prop is defined as `null` by default but there is an icon rendered (`<Visibility />`)
- On `Timeline`, when selecting Playground for the first time, every step changes to the warning state 🤯 
- Should the `Wizard` component have Playground? It changes what's rendered on the first time it's opened. Probably unrelated to this PR.

PR #3696, opened 2022-12-20
fix(Card): remove additional padding in CardSection

mainframev:
Looks like redundant padding, check how currently it looks with "onlySection" in storybook


DSil:
This changes the padding on the Expandable scenario, probably for the correct ones. They used to be 24px, and they are now only 16px. If this change is expected, then it LGTGM

DSil (approved):

PR #3698, opened 2022-12-20
fix(icons): add prefixes to icon ids

mainframev:
[slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1671462588755959)


DSil (commented):
Not sure about how to test it, but there is a snapshot failing (in need of an update? 🤔

packages/orbit-components/src/Illustration/types.d.ts
```diff
@@ -88,12 +88,8 @@ export type Name =
...
   | "WomanWithPhone";
 
-// TODO: remove spaceAfter in the next major version
-
-/** spaceAfter is deprecated, use margin instead */
 export interface Props extends Common.Globals, Common.SpaceAfter {
   readonly size?: "extraSmall" | "small" | "medium" | "large" | "display";
   readonly name: Name;
-  readonly margin?: React.CSSProperties["margin"] | Common.ObjectProperty;
   readonly alt?: string;
```
DSil, line 94: Why is this happening here? We are saying spaceAfter is deprecated and to use margin, but margin is being removed. And not related to this PR (?)

mainframev: it's because it's generated from the template

PR #3699, opened 2022-12-21
fix(Breadcrumbs): make last element always bold

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1671450724375909)


DSil:
The change is bigger than the request. The font-weight is now different on every link, from 500 to 400 on the links. Is this expected?

image

image

mainframev:
> https://ccf8b507.orbit.pages.dev

yes, this is expected, figma has that like that :)

PR #3706, opened 2023-01-04
Add aria-invalid and aria-describedby to form components

DSil:
Added `aria-invalid` and `aria-describedby` to InputField, Select and Textarea components.

[ORBIT-1960](https://kiwicom.atlassian.net/browse/ORBIT-1960)


DSil:
I didn't try voiceover either, but I checked on the DOM elements and they had everything, so I don't know if anything is missing 🤔

PR #3709, opened 2023-01-05
feat(icons): update icons from figma

mainframev:


packages/orbit-components/src/icons/svg/calendar-duration.svg
```diff
@@ -0,0 +1,6 @@
+<!--character:4038-->
+  <svg width="24" height="24" viewBox="0 0 24 24" >
+<path d="M16.4253 14.6384V16.7984L18.0453 18.4184" stroke="#252A31" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
```
hrach, line 3: Hi @mainframev.
We would need this icon to be color-less, the same as others: https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/icons/svg/calendar-anytime.svg?short_path=1782d7c

E.g. the stroke color should be here. Do you know why this export is different? Thanks.

mainframev: thanks @hrach, I will check why this happened, our script should remove it but it did not.

PR #3710, opened 2023-01-05
docs: fix knobs inconsistencies 

mainframev:
[ORBIT-2465](https://kiwicom.atlassian.net/browse/ORBIT-2465)

### DIFF
--- a/docs/src/__examples__/Stepper/DEFAULT.tsx
+++ b/docs/src/__examples__/Stepper/DEFAULT.tsx
@@ -25,12 +25,6 @@ export default {
         { name: "maxValue", type: "number", defaultValue: 10 },
         { name: "step", type: "number", defaultValue: 1 },
         { name: "disabled", type: "boolean", defaultValue: false },
-        {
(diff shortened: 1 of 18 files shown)


docs/src/__examples__/Stepper/DEFAULT.tsx
```diff
@@ -25,12 +25,6 @@ export default {
         { name: "maxValue", type: "number", defaultValue: 10 },
         { name: "step", type: "number", defaultValue: 1 },
         { name: "disabled", type: "boolean", defaultValue: false },
-        {
-          name: "size",
-          type: "select",
-          options: ["small", "normal"],
-          defaultValue: "small",
-        },
```
DSil, line 33: I've noticed that this is not really a problem of the docs, but of the component itself. The prop is documented on the README but it actually does not exist on the component

DSil (approved):
Just one small adjustment. It can be done on this PR or separately

DSil (approved):

PR #3713, opened 2023-01-11
Remove state update from useErrorTooltip hook

DSil:
The `useErrorTooltip` hook now receives a new parameter that indicates if it has a tooltip to handle or not. If not, it avoids state update calls that were unnecessary.
Tests to the hook were also added.

All components using the hook were also updated to the correct usage.

[ORBIT-2464](https://kiwicom.atlassian.net/browse/ORBIT-2464)
After this is merged, an update on #3705 is advised, so it avoids further conflicts in the future.


mainframev:
LGTM

mainframev (approved):

PR #3716, opened 2023-01-12
fix(icons): fix calendar icon after fix in Figma

mainframev:


DSil:
Is this the same as #3717 ?

mainframev:
> Is this the same as #3717 ?

nope, that PR solves mainly by removing the stroke attributes from icons, plus icon update. But it had to be also updated in Figma after.

mainframev:
> Is this the same as #3717 ?

ah, yes, forgot about the cron job :D

PR #3730, opened 2023-01-25
fix: Itinerary status color fixes

mainframev:
As we agreed to have temporary fix, I made it in a such way, so later we can quickly remove that


DSil (approved):
As mentioned in parallel conversation, I feel this approach could be better, but would require a lot more time and effort, so we are going with this for now, and refactoring it later.

Just reorder the commits, please (or squash them), because if Github squashes it automatically it might not be correct in the changelog 🙏

PR #3731, opened 2023-01-25
chore(actions): fix branch_url vulnerability

mainframev:
[Understanding the risk of script injections
](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable)


DSil (commented):
This is the script to deploy storybook both when we open a PR and for production, right? How do we know it is working? This did not deploy for this branch, did it?

mainframev:
It did :) you can check url it's created on Storybook action

DSil (approved):

PR #3734, opened 2023-02-06
feat: introduce Tabs component

mainframev:
[ORBIT-2572](https://kiwicom.atlassian.net/jira/software/c/projects/ORBIT/boards/117?modal=detail&selectedIssue=ORBIT-2572)


packages/orbit-components/src/Tabs/components/Tab/types.d.ts
```diff
@@ -0,0 +1,10 @@
+import type { Globals } from "../../../common/types";
+
+export type Type = "default" | "basic" | "medium" | "top";
```
DSil, line 3: Since we are defining the enum, we should take advantage of it and create the type from the enum. By doing that, we avoid possible inconsistencies in the future, as we only need to change new types in one place.

docs/src/__examples__/Tabs/DEFAULT.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
DSil, line 1: This example is not being used.

packages/orbit-components/src/Tabs/components/Tab/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+import defaultTheme from "../../../defaultTheme";
+
+const getColorBackground = ({ type, theme }) => {
+  if (type === TYPE_OPTIONS.MEDIUM) return theme.orbit.backgroundColorTabBundleMedium;
+  if (type === TYPE_OPTIONS.BASIC) return theme.orbit.backgroundColorTabBundleBasic;
+
+  return theme.orbit.backgroundColorTabBundleTop;
+};
+
+const getBackgroundType = (state?: "hover" | "active") => ({
```
DSil, line 17: I don't think we need these functions. Can't we just use the correct values where they are needed? This is extra runtime cost

packages/orbit-components/src/Tabs/components/TabPanel/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
DSil, line 1: It took me a while to understand what `TabPanel` would be. I honestly think this is overcomplicating the component. To me, the Tab component is just the clickable Tab. The content container is something else, and it can be anything. I don't think we need a component for that.

packages/orbit-components/src/Tabs/components/TabPanels/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import type { Props } from "./types";
+import { useTabs, TabPanelProvider } from "../../TabContext";
+
+const TabPanels = ({ children, dataTest }: Props) => {
```
DSil, line 6: Same as `TabPanel`

packages/orbit-components/src/Tabs/TabContext.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from "react";
```
DSil, line 1: I would rather have the apps control the state of the tabs individually. I think having a Context and trying to control the tabs behavior not only is more expensive but is also less flexible. It would also remove the need of iterating over children.
The basic Tabs usage is rather simple to control on the app side, and it simplifies the component's logic and implementation by a lot. I honestly don't think having a Context is necessary here. I would let the state management to the user side.

DSil (changes requested):
Noticed some problems and implementation details that are avoidable, imo. More details in the comments.

I also noticed that MMB is expecting tabs in colors different from the ones we are making available. Maybe a sync with designers is needed and more variations should be implemented.

The breaking change must also be documented.

packages/orbit-components/src/utils/common/index.ts
```diff
@@ -15,22 +15,23 @@ const setValue = (value?: string | number): string | null => {
   return typeof value === "number" && value !== 0 ? `${value}px` : String(value);
 };
 
-export const marginUtility = (
-  margin?: React.CSSProperties["margin"] | ObjectProperty,
+export const spacingUtility = (
```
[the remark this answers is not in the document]
mainframev: It's ok to have it in that PR, it's still will be in the changelog, but yes, I see your point

packages/orbit-components/src/Tabs/components/Tab/index.tsx
mainframev: It's then not readable and it has to be functions, otherwise, we would need to use nested ternary conditions.

mainframev: I doubt it's heavily affects runtime cost

packages/orbit-components/src/Tabs/components/Tab/types.d.ts
```diff
@@ -0,0 +1,11 @@
+import type { Globals } from "../../../common/types";
+import type { TYPE_OPTIONS } from "./consts";
+
+export type Type = Lowercase<keyof typeof TYPE_OPTIONS>;
```
DSil, line 4: I believe this can be simplified into
```suggestion
export type Type = `${TYPE_OPTIONS}`;
```

mainframev: yeah, forgot about that, thanks ;)

DSil (changes requested):
There is a bug with the controlled and uncontrolled version. If I try to use a completely uncontrolled version of the component, it will always select the first tab by default. It has to do with defining the context always with a `defaultSelected` value. Please fix that and the other commented things.

packages/orbit-components/src/Tabs/__tests__/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    userEvent.click(screen.getByText("Tab 2"));
+    userEvent.click(screen.getByText("Tab 1"));
+
+    expect(onClick).toHaveBeenCalled();
+    expect(onChange).toHaveBeenCalled();
+    expect(screen.getByTestId("first-tab")).toHaveAttribute("aria-selected", "false");
+    expect(screen.getByTestId("second-tab")).toHaveAttribute("aria-selected", "true");
+    expect(screen.getByRole("tablist")).toBeInTheDocument();
+    expect(screen.getAllByRole("tab")).toHaveLength(2);
```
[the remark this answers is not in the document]
mainframev: it does not make sense to test onChange, because it supposed to be passed only in uncontrolled usecase

packages/orbit-components/src/Tabs/components/TabPanels/types.d.ts
```diff
@@ -0,0 +1,7 @@
+import type { Globals } from "../../../common/types";
+import type { Props as TabPanelProps } from "../TabPanel";
+
+export interface Props extends Globals {
+  children: React.ReactElement<TabPanelProps> | React.ReactElement<TabPanelProps>[];
+  selectedTab?: number;
```
DSil, line 6: I saw this was added but it is not documented, and it should.

mainframev:
> Should there be Loki tests for this?

Not for now, only in future, if visual regression will detect any unrelated change

packages/orbit-components/src/Tabs/components/TabPanels/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+const TabPanels = ({ children, selectedTab, dataTest }: Props) => {
+  const { selected } = useTabs();
+  const isSelected = Boolean(selectedTab || selected);
+
+  return (
+    <div data-test={dataTest}>
+      {React.Children.map(children, (child, idx) => {
+        if (!React.isValidElement(child)) return null;
+        return (
+          (idx === selectedTab || idx === selected) && (
```
DSil, line 15: This relies on the order of the tabs and panels being the same, and unchanged. But imagine one has a disabled Tab and doesn't even has content for the corresponding TabPanel (if it's disabled, why would it even have something there…?). We could rely on id or something like that.

It also mixes the controlled and uncontrolled behaviors. I would expect that in a scenario where I am in full control of the state of the Tab I am also in full control of the TabPanel. I wouldn't expect to be TabPanels handling that, and relying on a Context. Can't we have a possible `active` prop on `TabPanel` that controls whether it is rendered or not? We would remove that responsibility from this `TabPanels` component.

packages/orbit-components/src/Tabs/__tests__/index.test.tsx
```diff
@@ -0,0 +1,103 @@
+import * as React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import Tabs, { TabList, Tab, TabPanels, TabPanel } from "..";
+
+describe("Tabs", () => {
```
DSil, line 7: We should assert that the correct TabPanel is in the document on the different interactions. Since the component is responsible for handling that.

DSil (changes requested):
Besides my comments, there is still an incompatibility between uncontrolled and controlled version. Try having the controlled version with a tab selected on the initial state that is not the first one.

> Not for now, only in future, if visual regression will detect any unrelated change

For that to happen don't we need to have the base test for comparison? 🤔 How do we perform the visual regression test if we don't have the original screenshots from when the component is created?

mainframev:
> Besides my comments, there is still an incompatibility between uncontrolled and controlled version. Try having the controlled version with a tab selected on the initial state that is not the first one.
> 
> > Not for now, only in future, if visual regression will detect any unrelated change
> 
> For that to happen don't we need to have the base test for comparison? 🤔 How do we perform the visual regression test if we don't have the original screenshots from when the component is created?

It's a part of the release process, not this component

packages/orbit-components/src/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <TabPanel>Tab 2 content</TabPanel>
+        <TabPanel>Tab 3 content</TabPanel>
+        <TabPanel>Tab 4 content</TabPanel>
+      </TabPanels>
+    </Tabs>
+  );
+};
+
+export const Controlled = () => {
+  const [selected, setSelected] = React.useState(0);
```
DSil, line 30: Do this an you will see the bug.

```suggestion
  const [selected, setSelected] = React.useState(2);
```

DSil (changes requested):
I think we can't have a default value on `defaultSelected`, the bug is still there. I have pointed a step to reproduce on the story.

packages/orbit-components/src/Tabs/Tabs.stories.tsx
mainframev: It's not a bug. defaultSelected is not supposed to be used with controlled. README md is mentioning it.

DSil (commented):
Behavior looks ok, but a few visual tweaks are needed. Sorry for not pointing them out earlier, I was focused on behavior and didn't extend the review to the visual part.

As for the re-render thing, I would not delay this because of that for now.

packages/orbit-components/src/Tabs/components/Tab/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    position: relative;
+    appearance: none;
+    flex-direction: row;
+    justify-content: space-between;
+    box-sizing: border-box;
+    align-items: center;
+    font-family: ${theme.orbit.fontFamily};
+    cursor: ${!disabled && "pointer"};
+    opacity: ${disabled && "0.5"};
+    padding: ${compact ? "5px 16px" : "9px 16px"};
```
DSil, line 57: I would have done it like this. With this, the content is actually centrally aligned with the full height of the Tab, like in Figma. But I am not sure if it is what is intended, or if it should be centrally aligned with the white part. Therefore, I'll leave the suggestion but it is up to you, non-blocking.
```suggestion
    padding: ${compact ? "6px 16px 4px" : "10px 16px 8px"};
```

DSil (approved):
Just a non-blocking suggestion, otherwise, LGTM 👍 
Keep in mind the commits, as a squash to a single commit might lead to a non-accurate changelog. We need the commit regarding the Tabs component and the commit about the rename of the utility.

PR #3738, opened 2023-02-08
feat(ChoiceTile): add ChoiceTile component

DSil:
This introduces the `ChoiceTile` component.

~The PR is opened against a different branch because it depends on the changes made there. It will be targeted to `master` once #3735 is merged.~

[ORBIT-2510](https://kiwicom.atlassian.net/browse/ORBIT-2510)


packages/orbit-components/src/ChoiceTile/ChoiceTile.stories.tsx
```diff
@@ -0,0 +1,138 @@
...
+DefaultRadio.story = {
+  parameters: {
+    info: `ChoiceTile component can have any custom content as children. Title, icon/flag and description can be defined via props. 
+    If clickable elements are used in custom content, be sure to stop the propagation of the click event in order to prevent effects on the ChoiceTile.`,
+  },
+};
+
+export const DefaultCheckbox = () => {
+  return (
+    <div style={{ display: "flex", flexDirection: "column", gap: "12px" }}>
```
mainframev, line 46: why not stack? 😄

docs/src/__examples__/ChoiceTile/DEFAULT.tsx
```diff
@@ -0,0 +1,30 @@
...
+  },
+  exampleKnobs: [
+    {
+      component: "ChoiceTile",
+      knobs: [
+        { name: "title", type: "text", defaultValue: "Title" },
+        { name: "description", type: "text", defaultValue: "Some additional description" },
+        { name: "label", type: "text", defaultValue: "Label" },
+        { name: "type", type: "select", options: ["radio", "checkbox"], defaultValue: "radio" },
+        { name: "inline", type: "boolean", defaultValue: "false" },
```
mainframev, line 26: if it's boolean, you can use boolean, not string

packages/orbit-components/src/ChoiceTile/styled.ts
```diff
@@ -0,0 +1,105 @@
...
+    font-weight: ${theme.orbit.fontWeightHeadingTitle3};
+    line-height: ${theme.orbit.lineHeightHeadingTitle3};
+  `}
+`;
+
+StyledChoiceTileHeading.defaultProps = {
+  theme: defaultTheme,
+};
+
+export const StyledChoiceTileDescription = styled(StyledText)<{ hasIconSpacing?: boolean }>`
```
mainframev, line 48: description font-size is not right, in figma it's 14px, in storybook 16px 🤔

packages/orbit-components/src/ChoiceTile/styled.ts
```diff
@@ -0,0 +1,105 @@
...
+  `}
+`;
+
+StyledChoiceTile.defaultProps = {
+  theme: defaultTheme,
+};
+
+export const StyledChoiceTileHeading = styled(StyledText)`
+  ${({ theme }) => css`
+    font-size: ${theme.orbit.fontSizeHeadingTitle3};
```
mainframev, line 38: In Figma, it has two variations, with font-size 18px and line-height 24px and font size 16px and line-height 20px. You have a font-size 16px and a line height 24px. Not sure why it's like that, but there is something wrong, I think

packages/orbit-components/src/ChoiceTile/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <StyledChoiceTile
+      selected={selected}
+      onClick={onClick}
+      tabIndex={0}
+      data-test={dataTest}
+      id={id}
+    >
+      <Stack inline spacing="medium" align="center">
+        {icon}
+        {title && (
```
mainframev, line 57: could you try to use a slightly different markup here? 

something like: 

```
 <Stack inline shrink spacing="medium">
        {icon}
        <Stack direction="column" spacing="XXSmall">
          {title && (
            <Stack inline spacing="XSmall" align="center">
              <StyledChoiceTileHeading>{title}</StyledChoiceTileHeading>
              {inline && isTablet && description && <Text type="secondary">{description}</Text>}
            </Stack>
          )}
          {(!isTablet || !inline) && description && <Text type="secondary">{description}</Text>}
        </Stack>
        {isTablet && <InputContent selected={selected} label={label} type={type} />}
      </Stack>
```

feel free to resolve it, if won't working because of some use case, but I will definitely try to place description under title in the same Stack, in order to prevent having custom text wrapper with margin offset

packages/orbit-components/src/ChoiceTile/README.md
```diff
@@ -0,0 +1,34 @@
...
+| inline      | `boolean`                  | Determines if the description is presented inline with the title in larger viewports. |
+| title       | `React.Node`               | Defines the title of the ChoiceTile.                                                  |
+| icon        | `React.Node`               | Defines the icon or flag of the ChoiceTile.                                           |
+| description | `React.Node`               | Defines the description of the ChoiceTile.                                            |
+| label       | `React.Node`               | Defines the label to be presented next to the Checkbox or Radio of the ChoiceTile.    |
+| selected    | `boolean`                  | Determines if the Checkbox or Radio of the ChoiceTile are selected.                   |
+| onClick     | `event => void \| Promise` | Executed when clicking on the ChoiceTile. Can be used for managing its state.         |
+| children    | `any`                      | Allows for custom content to be added to the ChoiceTile.                              |
```
mainframev, line 29: children is also a `React.ReactNode`

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    render(<ChoiceTile type="checkbox" />);
+    expect(screen.getByRole("checkbox")).toBeInTheDocument();
+  });
+
+  it("can be selected", () => {
+    render(<ChoiceTile type="radio" selected />);
+    expect(screen.getByRole("radio")).toBeChecked();
+  });
+
+  it("calls onClick", () => {
```
mainframev, line 33: add that to the first section `renders with correct properties`, no need to have a separate assertion for this

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    render(<ChoiceTile dataTest="choice-tile" type="radio" onClick={onClick} />);
+    userEvent.click(screen.getByTestId("choice-tile"));
+
+    expect(onClick).toHaveBeenCalled();
+  });
+
+  it("can be inline", () => {
+    const onClick = jest.fn();
+    const mqMock = jest.spyOn(mq, "default").mockImplementationOnce(() => {
```
mainframev, line 44: instead of spy, use the mock of `useMediaQuery`, something like [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Pagination/__tests__/index.test.tsx)

docs/src/documentation/03-components/05-input/choicetile/01-guidelines.mdx
```diff
@@ -0,0 +1,15 @@
...
+title: Guidelines
+redirect_from:
+  - /components/choicetile/
+---
+
+<ReactExample exampleId="ChoiceTile-default" />
+
+## When to use
+
+- To present a set of similar options to chose from. Either only one option or multiple.
```
mainframev, line 11: I think we need more info here, something like we have it with other components. When to use / When not to use. Ask designers for more information about usage.

mainframev (changes requested):

docs/src/documentation/03-components/05-input/choicetile/01-guidelines.mdx
DSil: I asked already. I just added this here so we don't have a blank page

packages/orbit-components/src/ChoiceTile/styled.ts
DSil: On the sync, Will talked about the possibility of only having one size variation, and not two. Until the decision is made, I have implemented only one, as agreed

mainframev: Okay, but if the font-size 16px, should not it have 20px line-height?

DSil: Hmm, yes, but then our tokens are a bit misaligned. The `fontSizeHeadingTitle3` is 16px and `lineHeightHeadingTitle3` is 24px. But in this design the conjugation is different. I'll sync with designers.

packages/orbit-components/src/ChoiceTile/index.tsx
DSil: It seems to work for title, description and input if `shrink` is applied to that new Stack and `InputContent` changes it `align-self` property to `start`. However, the Flag would need a wrapper, so that it is aligned to the top, with correct line-height and all. I think in the end it would be the same, removing one wrapper to add another 🤔

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
DSil: There are a few things I don't particularly like on that example. We have to mock and restore mock for every test (even if viewport is irrelevant for what's being tested) and by defining the type to `any` we can mock it to be anything really, not following types. Is there any major advantage in that approach that I am not seeing?

packages/orbit-components/src/ChoiceTile/index.tsx
mainframev: For Flag, the wrapper will be smaller. Also for that case you can use only a div with `css` property ;) should be small and no need to have external styled wrapper for that

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
mainframev: you are already restoring it for almost all tests. afterEach can help to restore it in one function, instead of calling mockRestore at the end of each assertion. 

It just looks less verbose, there is a default exported hook and by mocking the module you do not need to spy on default. It's up to you

packages/orbit-components/src/ChoiceTile/styled.ts
DSil: Agreed with Will to keep it as is (with Title3 tokens) as the font-size for Title3 is expected to grow to 18px. The line-height is already correct.

packages/orbit-components/src/ChoiceTile/index.tsx
DSil: I managed to be able to do it without any wrapper 👍 so I could remove the `StyledChoiceTileDescription` and adjust the icon positioning so it is inline with the title. Thanks for the suggestion!

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
DSil: With the new markup implemented (following the suggestion in the other thread), I have removed theses tests, as they were relying in the Description wrapper.

Therefore, I had to rethink the tests and came up with an approach that I don't particularly like. I implemented that on a separate commit, so that it is easier to review, and eventually revert, if we decide we don't like it. I would still like to have the logic on the `inline` prop tested. But maybe only Loki tests are enough for this, no?

packages/orbit-components/src/ChoiceTile/index.tsx
mainframev: Great ;)

packages/orbit-components/src/ChoiceTile/__tests__/index.test.tsx
mainframev: Yes, Loki testing might be enough ;)

mainframev (changes requested):
could you also update the component status yaml file?

DSil:
As agreed with Design, this will not move forward, as specs are still being refined on the Design side. Closing this but keeping the branch, for the future.

PR #3742, opened 2023-02-13
feat(tokens): change fontSizeSm to 13px and fontSizeMd to 15px

DSil:
As per design decision, the font size is changing for small and medium text sizes. They both went up by 1px. The line height is not expected to change. However, some components use line-height with a relative value (multiplier unit) instead of a fixed value in px. For that reason, some visual changes were identified:
- `BadgePrimitive` is using `line-height: 1`. With the increase of font-size, the height of the badge also increases by 1px. Possible solutions:
  - use a fixed `line-height` of the same value (we currently don't have a token for that value)
  - limit the height of the badge
  - Change the value of the `line-height` property to a relative value that ends up with the same calculated value as before
- `ButtonPrimitive` is also using `line-height: 1`. However, it has `height` defined. Therefore, the height of buttons do not change, but their width does.
- There is a `lineHeight` token defined as "1.4" on the base palette. This token is not used directly in Orbit (we can't know in other repos) but it is used to define the `lineHeightText` token, which is used in some other components:
  - It is used in the internal `Feedback` component in `ChoiceGroup`, so the height changes
  - It is used in `ErrorFormTooltip`, used by many input components. Its height also changes with the change of the font-size.
  - It is used in the internal `IconWrapper` component in `Timeline`. It doesn't seem to have any effect on the component alone. But on the example inside a Modal there are some text overflows that cause the Modal to grow in size. This hints a possible problem in custom modals built throughout the product.


mainframev:
@DSil could you also run loki? 🙏🏻

mainframev (approved):

PR #3746, opened 2023-02-16
fix(ItinerarySegmentStop): fix CircleSmall icon

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1676542344349339)

We need to support both use cases. It's still by default rendering the icon based on type, but if `null` is explicitly set to `icon` prop - it will render default circle empty icon


packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
```diff
@@ -395,8 +395,8 @@ export const Stop = () => {
         station={station}
         hidden={hidden}
         date={date}
-        time={time}
         type={type}
+        time={time}
```
DSil, line 399: Please avoid unnecessary file changes 🙏

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
```diff
@@ -1,16 +1,13 @@
...
-import CircleEmpty from "../../../icons/CircleSmallEmpty";
-import Circle from "../../../icons/CircleSmall";
 import { useWidth } from "../../context";
 import defaultTheme from "../../../defaultTheme";
 import Stack from "../../../Stack";
 // TODO: remove after designers will resolve status colors
 // https://skypicker.slack.com/archives/GSGN9BN6Q/p1674568716519889
 import Text from "../../ItineraryTemporaryText";
-import ItineraryIcon from "../ItineraryIcon";
+import SegmentStopIcon from "../ItineraryIcon";
```
DSil, line 10: Can't we call it `ItineraryIcon`?

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
```diff
@@ -49,41 +46,23 @@ StyledHiddenCity.defaultProps = {
...
 const ItinerarySegmentStop = ({
   date,
-  icon,
   time,
   cancelledDate,
   cancelledCity,
   cancelledStation,
   cancelledTime,
   city,
+  icon,
```
DSil, line 57: Same as above, unnecessary diff

DSil (commented):
LGTM, just minor changes. I feel this should also be documented somehow

PR #3749, opened 2023-02-21
fix: remove type and theme on DOM elements

mainframev:


DSil (commented):
LGTM. Pipeline is broken tho

mainframev:
> LGTM. Pipeline is broken tho

Fixed in Tab component PR ;)

PR #3761, opened 2023-03-10
Fix Itinerary expand issues

DSil:
This addresses two issues:

- Text can now be selected inside the expanded part of ItinerarySegmentDetails without triggering the close action
- When a segment was wrapped in ItineraryStatus, it would always have the actionable style. It can now be controlled. (A story was updated to display this use case)

Docs were also updated regarding new and required props.

[Slack thread](https://skypicker.slack.com/archives/CAMS40F7B/p1678437139773419)

### DIFF
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/index.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/index.tsx
@@ -57,7 +57,9 @@ const ItinerarySegment = ({
             value={{
               index: i,
               opened,
-              toggleOpened: () => setOpened(prev => !prev),
+              toggleOpened: () => {
+                if (document && document.getSelection()?.type !== "Range") setOpened(prev => !prev);
+              },
               last: i === content.length - 1,
               isNextHidden: Boolean(content[i + 1]?.props?.hidden),
               isPrevHidden: Boolean(content[i - 1]?.props?.hidden),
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/Itinerary/ItinerarySegment/index.tsx
```diff
@@ -57,7 +57,9 @@ const ItinerarySegment = ({
             value={{
               index: i,
               opened,
-              toggleOpened: () => setOpened(prev => !prev),
+              toggleOpened: () => {
+                if (document.getSelection()?.type !== "Range") setOpened(prev => !prev);
```
mainframev, line 61: maybe add a check here for the document, so on SSR it won't have a type error 

```suggestion
                if (document && document.getSelection()?.type !== "Range") setOpened(prev => !prev);
```

packages/orbit-components/src/Itinerary/README.md
```diff
@@ -70,8 +70,8 @@ ItinerarySegmentStop is an atomic unit of the Itinerary component, shows two loc
 
 | Name             | Type                | Required           | Default       | Description                                              |
 | ---------------- | ------------------- | ------------------ | ------------- | -------------------------------------------------------- |
-| date             | `React.Node`        | :heavy_check_mark: |               | The date of `ItinerarySegmentStop`                       |
-| time             | `React.Node`        | :heavy_check_mark: |               | The time of `ItinerarySegmentStop`                       |
+| date             | `React.Node`        |                    |               | The date of `ItinerarySegmentStop`                       |
```
mainframev, line 73: Can you also remove the empty row in this table? It was not caused by your changes, but still can be done here ;)

DSil: What empty row? 🤔

PR #3763, opened 2023-03-13
fix(Stack): fix gap issue for iOS Safar < 14.1

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1678714906521569)

Based on [`-webkit-touch-callout`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout) and `supports` query. It's being used only on iOS Safari, that allow us to trigger only it. 

tested on iOS Safari 13

Screenshot 2023-03-13 at 17 12 59

### DIFF
--- a/packages/orbit-components/src/Stack/helpers/getGap.ts
+++ b/packages/orbit-components/src/Stack/helpers/getGap.ts
@@ -41,9 +41,16 @@ const getGap = ({ index, devices }: { index: number; devices: Devices[] }) => (
       getSpacing(props.theme)[spacing],
     );
 
+  // workaround to make it work on Safari iOS < 14.1
+  // TODO: remove that after dropping support for iOS < 14.1
   if (props.flex) {
     return css`
       gap: ${gap};
+      @supports (-webkit-touch-callout: none) and (not (translate: none)) {
+        & > *:not(:last-child) {
+          margin: ${margin && rtlSpacing(margin)} !important;
+        }
+      }
     `;
   }
 


packages/orbit-components/src/Stack/helpers/getGap.ts
```diff
@@ -41,9 +41,16 @@ const getGap = ({ index, devices }: { index: number; devices: Devices[] }) => (
       getSpacing(props.theme)[spacing],
     );
 
+  // workaround to make it work on Safari iOS < 14.1
+  // TODO: remove that after dropping support for iOS < 14.1
   if (props.flex) {
     return css`
       gap: ${gap};
+      @supports (-webkit-touch-callout: none) and (not (translate: none)) {
```
DSil, line 49: I understand this is the "hacky" way and if it works I'm ok with it. I have two questions, though:
- Couldn't we do it using javascript to check the browser version?
- Can you please attach the source (as a comment) as to why those css properties in `@supports` correspond to that Safari version? I think it's better documented hacky code 🙏

mainframev: using js will be an even hackier solution and less neat than this

technically it's not even hacky code, it's standard code that only not being used quite ofter. Support query is quite known, the options that it received a targeting only Safari mobile browsers, because `-webkit-touch-callout` is being used only there

DSil: I call it hacky for the readability. I mean, how does one know that it refers to that version of safari? I trust your approach, I'd just like to have it better understandable when reading the code

> are targeting only Safari mobile browsers

Don't we want to target all Safari ≤14.1?

DSil: Alright, got it 👍 thanks

DSil (approved):

PR #3765, opened 2023-03-14
fix: segmented switch outline and role

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1678724132897789)

As we decided to use the default browser outline instead of box-shadow before, which in my opinion is totally ok, however, there are situations like this one, when we rely on default input focus behavior and the color of the outline may be different depending on the browser, operating systems and user's preferences, that's why I think we should implicitly force it to be blue only, so requests like this will not appear anymore.

### DIFF
--- a/packages/orbit-components/src/utils/common/index.ts
+++ b/packages/orbit-components/src/utils/common/index.ts
@@ -2,12 +2,16 @@ import type { FlattenSimpleInterpolation } from "styled-components";
 import { css } from "styled-components";
 
 import type { ObjectProperty } from "../../common/types";
+import type defaultTheme from "../../defaultTheme";
 
 export const firstToUpper = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
 
-export const defaultFocus = (): FlattenSimpleInterpolation => css`
-  outline: 2px solid Highlight;
-  outline: 2px solid -webkit-focus-ring-color;
+export const defaultFocus = ({
+  theme,
+}: {
+  theme: typeof defaultTheme;
+}): FlattenSimpleInterpolation => css`
+  outline: 2px solid ${theme.orbit.paletteBlueNormal};
 `;
 
 const setValue = (value?: string | number): string | null => {
--- a/packages/orbit-components/src/Checkbox/index.tsx
+++ b/packages/orbit-components/src/Checkbox/index.tsx
@@ -157,7 +157,7 @@ const StyledInput = styled.input<StyledInputProps>`
   }
 
   &:focus + ${StyledIconContainer} {
-    ${defaultFocus}
+    ${defaultFocus};
   }
 `;
 
--- a/packages/orbit-components/src/InputField/helpers/formElementFocus.ts
+++ b/packages/orbit-components/src/InputField/helpers/formElementFocus.ts
@@ -1,5 +1,5 @@
 import { defaultFocus } from "../../utils/common";
 
-const formElementFocus = () => defaultFocus;
+const formElementFocus = (): typeof defaultFocus => defaultFocus;
 
 export default formElementFocus;
--- a/packages/orbit-components/src/Radio/index.tsx
+++ b/packages/orbit-components/src/Radio/index.tsx
@@ -134,7 +134,7 @@ const StyledInput = styled.input`
   }
 
   &:focus + ${StyledIconContainer} {
-    ${defaultFocus}
+    ${defaultFocus};
 
     ${media.largeMobile(css`
       border-width: 1px;
--- a/packages/orbit-components/src/SegmentedSwitch/SegmentedSwitch.stories.tsx
+++ b/packages/orbit-components/src/SegmentedSwitch/SegmentedSwitch.stories.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { text, number, boolean } from "@storybook/addon-knobs";
+import { text, boolean } from "@storybook/addon-knobs";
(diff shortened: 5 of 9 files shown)


packages/orbit-components/src/utils/common/index.ts
```diff
@@ -1,15 +1,25 @@
...
+import defaultTheme from "../../defaultTheme";
 
 export const firstToUpper = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
 
-export const defaultFocus = (): FlattenSimpleInterpolation => css`
-  outline: 2px solid Highlight;
-  outline: 2px solid -webkit-focus-ring-color;
+export const defaultFocus = (): FlattenInterpolation<ThemeProps<typeof defaultTheme>> => css`
+  ${({ theme }) => css`
+    outline: 2px solid ${theme.orbit.paletteBlueNormal};
```
DSil, line 15: This changes all the focus, which removes the native color entirely. I think our mistake is using this for the selected state (as we do with Segmented switch). I don't think we should change this, the default focus is to keep. What we want to change is the color of the selected state. Therefore, the selected state can't use this. It must use its own outline that explicitly states blue

mainframev: It will lead us to do it in the future in the same way. I can expect, that we will receive the same request about InputField and other components. I do not see a problem to force it to be a blue color, do you? What's your argument for keeping the default color? The main reason why we did that before was, that box-shadow had not sufficient contrast as blue outline, but I personally do not mind to implicitly have that as blue

DSil: I would keep the default focus color precisely for accessibility purposes. If it's something that can differ on the browser side for whatever reason (settings, default, etc.), I think we should keep it.
I'm all in favor of switching the selected/active state. But back in time designers agreed on going with the native colors, and I don't think that's meant to be changed. And if native colors can be so different… that's a consequence we should not worry about.
So the SegmentedSwitch option that is selected has our blue outline. But its focus color should be the native from whatever browser is being used. That was my understanding from the thread and that's what I think we should do, honestly

mainframev: Ok, how it's different from the native browser/OS/user's preferences colors? For A11y is important only contrast here, so I do not think it would lead to a11y issues. 

I prefer to add those styles based on the checked/unchecked state, it's closer to the default input behavior. As I mentioned above, it would lead us to change that probably in the future for every component, so I do not think it's good. 

I would probably change my opinion if you will convince me with some real arguments :) 

Designers agreed that before, because for them it was the same, I am not even sure that they got what we did, because reading the thread in slack, I can say, that Milan is not aware of how the outline works. If we told them before, that somewhere it will be black - I think they could change their opinion.

DSil: My understanding is that that `-webkit-focus-ring-color` can change based on accessibility preferences configured by the user, that's why I raised accessibility questions. I am not certain it happens, but it would make sense, otherwise why would browsers define it differently amongst them? 💭 

The black outline was clearly undesired, but I understood they wanted native colors for focus, but not for active/selected, so that's what I was expected to see applied here.

Why can't we have a color for focus (the native one) and another for active/selected? 🤔

mainframev: Plus, it's weird to me, to have InputField with a black outline and SegmentedSwitch with blue. We either should do the same for all other Inputs, as you suggested in SegmentedSwitch or do it as I suggested and just update it inside defaultFocus.

DSil: Maybe I wasn't clear, but in case we do it differently (focus vs active), we should do it for all, yes.
The black outline would occur only on browsers with it defined, but that's something we can't (and probably shouldn't) control. The active state is "ours" so we can put it blue as we want. Does it make sense?

But let's confirm with designers if focus is always equal to selected (and therefore different from the native) or if we should have native for focus and custom for selected 🙏

mainframev: I do not think that there are our/not our. Even the textarea you are currently writing it here is overwritten :D 
So I still prefer to implicitly set the color for the outline

DSil: That's what GitHub's design system owners decide for themselves 😄 I am only raising the question because the change is quite recent. But it can be as you said: unintentional

packages/orbit-components/src/utils/common/index.ts
```diff
@@ -1,15 +1,25 @@
...
+  `};
 `;
 
+defaultFocus.defaultProps = {
+  theme: defaultTheme,
+};
```
DSil, line 21: Do we need this? `defaultFocus` is not a component 🤔 weird calling it defaultProps…

DSil: If anything it should the component's responsibility to have theme as default prop (like you added in StyledStepperButton

mainframev: yes, we need, defaultFocus need access to theme object

mainframev: right, forgot to remove that, when pipeline was failing and i was not sure where exactly it can't access theme object, turned out that it's in Stepper

DSil: But it's not a component and we're defining default props 🤯 can't it be default value for the argument instead? It's non-blocking though

DSil (approved):

PR #3768, opened 2023-03-15
fix(Switch): background color was not correct

DSil:
On the tokens migration introduced in 5.0.0, the migration from `cloudDarker` to `cloudDark` was incorrect in Switch (migrated to `cloudNormal` instead). This PR fixes that.


packages/orbit-components/src/Switch/index.tsx
```diff
@@ -19,7 +19,7 @@ const StyledSwitchBase = styled.div<{ checked?: boolean; disabled?: boolean }>`
     cursor: ${!disabled && "pointer"};
     width: 40px;
     height: 20px;
-    background-color: ${theme.orbit.paletteCloudNormal};
+    background-color: ${theme.orbit.paletteCloudDark};
```
mainframev, line 22: maybe it's good to actually create token for this inside tokens

DSil: Agreed. Will do

mainframev (approved):

PR #3769, opened 2023-03-16
chore(tests): fix Dialog loki tests

DSil:
Portal rendered stories are not considered by Loki. This changes Dialog stories so they are not rendered in a Portal.

It also removes an unnecessary icon from a test that was causing the snapshot testing locally to be different from on server. The icon is not needed on the test, so it was replaced by a much simpler text

[ORBIT-1462](https://kiwicom.atlassian.net/browse/ORBIT-1462)


packages/orbit-components/src/primitives/MobileDialogPrimitive/MobileDialogPrimitive.stories.tsx
```diff
@@ -33,6 +33,7 @@ export const Playground = () => {
 
 Playground.story = {
   parameters: {
+    loki: { skip: true },
```
mainframev, line 36: 👍🏻

mainframev (approved):

PR #3770, opened 2023-03-16
feat: remove translations from Orbit 

mainframev:
[Slack request ](https://skypicker.slack.com/archives/CAMS40F7B/p1678381722546969)

Removed dictionaries, translation components, and hooks. Replaced translation values with props.


packages/orbit-components/src/OrbitProvider/ThemeProvider.stories.tsx
```diff
@@ -1,63 +1,12 @@
...
 import OrbitProvider from ".";
 
-const ButtonWithTranslation = () => {
-  const translate = useTranslate();
-  return <Button>{translate("button_close")}</Button>;
-};
-
-export default {
-  title: "OrbitProvider",
-};
```
DSil, line 21: Please don't delete these lines, otherwise the whole component is gone from Storybook 😄

packages/orbit-components/src/Pagination/index.tsx
```diff
@@ -21,6 +19,9 @@ const handlePageChange = (onPageChange, pageCount) => nextPageIndex => {
 
 const Pagination = ({
   pageCount,
+  labelPrev,
+  labelNext,
+  labelProgress,
```
DSil, line 24: Should we have default values for these? Like in the other components

packages/orbit-components/src/Pagination/Pagination.stories.tsx
```diff
@@ -12,16 +12,33 @@ export default {
...
   const pageCount = number("pageCount", 100);
+  const labelPrev = text("labelPrev", "prev");
+  const labelNext = text("labelNext", "next");
+  const labelProgress = text("labelProgress", "of");
 
   return (
     <Pagination
+      labelPrev={labelPrev}
+      labelNext={labelNext}
+      labelProgress={labelProgress}
```
DSil, line 41: This should be something like
```suggestion
      labelProgress={`${selectedPage} ${labelProgress} ${pageCount}`}
```
And the same for all other stories

packages/orbit-components/src/Popover/README.md
```diff
@@ -38,6 +38,7 @@ Table below contains all types of the props available in the Popover component.
 | lockScrolling  | `boolean`                | `true`              | Whether to prevent scrolling of the rest of the page while Popover is open on mobile. This is on by default to provide a better user experience. |
 | noFlip         | `boolean`                | `false`             | Turns off automatic flipping of the Popover when there is not enough space                                                                       |
 | allowOverflow  | `boolean`                | `false`             | Allows the Popover to be cut off instead of moving it while scrolling to keep it visible.                                                        |
+| labelClose     | `React.Node`             |                     | Label for close button.                                                                                                                          |
```
DSil, line 41: It has the default value of "Close"

packages/orbit-components/src/Popover/Popover.stories.tsx
```diff
@@ -254,12 +254,14 @@ OpenedByProp.story = {
...
   const overlapped = boolean("overlapped", true);
   const opened = boolean("opened", false);
+  const labelClose = text("Label close", "Close");
 
   return (
     <Popover
       overlapped={overlapped}
       content={content}
       actions={actions}
+      labelClose={labelClose}
```
DSil, line 264: Popovers that have `actions` don't use this `labelClose` prop. This knob should be added to other stories, like:
- Default
- With Tooltip
- Placement
- With ListChoice
- Opened by prop
- Scrolling Content
- RTL
- RTL reverse
- LazyContentSimlulated (this story name has a typo 😄 should be _simulated_)

packages/orbit-components/src/SkipNavigation/index.tsx
```diff
@@ -41,7 +40,7 @@ const StyledSelectWrapper = styled.div`
   max-width: 800px;
 `;
 
-const SkipNavigation = ({ actions, feedbackUrl }: Props) => {
+const SkipNavigation = ({ actions, feedbackUrl, feedbackLabel }: Props) => {
```
DSil, line 43: We should have the default value

packages/orbit-components/src/SkipNavigation/index.tsx
DSil, line 1: Wondering if this component is still relevant 😅 our tracking tool marks as unused… should we deprecate it?

Raising this question because the component has a couple of TODOs to add dictionary. Those TODOs are now obsolete, but the hardcoded strings they contain should also be handled by props.
But, again, if no one is using this… we might just deprecate it on the next version 🫥

packages/orbit-components/src/Wizard/index.tsx
```diff
@@ -100,12 +99,7 @@ const Wizard = ({
           }}
         >
           <Stack as="span" inline>
-            <b>
-              {translate("wizard_progress", {
-                number: activeStep + 1,
-                total: stepsCount,
-              })}
-            </b>
+            <b>{labelProgress}</b>
```
DSil, line 102: Stories need to be updated, as this was being applied automatically. They need to get this prop with the current step, the string and the total steps

packages/orbit-components/src/primitives/MobileDialogPrimitive/index.tsx
```diff
@@ -9,6 +9,7 @@ import type { Props } from "./types";
 
 const MobileDialog = ({
   children,
+  labelClose,
```
DSil, line 12: It should have default value

packages/orbit-components/src/deprecated/InputStepper/index.tsx
DSil, line 1: Any reason for not getting rid of the other deprecated components in this release? 😇

DSil (commented):
Great work 👍 left some comments of some things, nothing major 👌

packages/orbit-components/src/deprecated/InputStepper/index.tsx
mainframev: it was removed, didn't you notice? 🤔

packages/orbit-components/src/Pagination/index.tsx
mainframev: here is probably makes sense force them to pass it, they are not optional, plus it needs the translation key for progress, we can't have a default value for that

packages/orbit-components/src/deprecated/InputStepper/index.tsx
DSil: I still see some. I was referring to the other components inside the `deprecated` folder

packages/orbit-components/src/Pagination/index.tsx
DSil: Yes, I was only referring to Next and Prev. But if they are required we don't need default value 👍

packages/orbit-components/src/SkipNavigation/index.tsx
mainframev: if it was a regular component - i would deprecate it, but this one is important for a11y, maybe we should keep it just for that purpose, for future, if someone will decide to use it or for open source

DSil: Alright, but then shouldn't we also add props for the other labels?

packages/orbit-components/src/OrbitProvider/ThemeProvider.stories.tsx
mainframev: accident :D

packages/orbit-components/src/Wizard/Wizard.stories.tsx
```diff
@@ -12,13 +12,17 @@ export default {
...
+  const labelProgress = text("labelProgress", "of");
 
   return (
     <Wizard
       id="wizard"
       direction={direction}
       completedSteps={3}
       activeStep={3}
+      labelClose={labelClose}
+      labelProgress={`3 ${labelProgress} 3`}
```
DSil, line 25: Missing the RTL story

DSil (commented):
Just missed the question about SkipNavigation and also the adjustments made from auto-commit to my suggestion are breaking, so please fix them 🙏

docs/src/__examples__/Pagination/DEFAULT.tsx
```diff
@@ -16,6 +21,21 @@ export default {
...
+        },
+        {
+          name: "labelNext",
+          type: "text",
+          defaultValue: "Next",
+        },
+        {
+          name: "labelProgress",
+          type: "text",
+          defaultValue: "Page 2 of 7",
```
DSil, line 37: This is actually not being used. I would just remove it, because we can't be flexible to allow the construction of the whole string with the `currentPage` and `pageCount` variables…

And we can't make it be just for the `"of"` word, because that is not how the prop actually works, so it would be misleading. That's why I think we should not have this knob here.

DSil:
Besides the above comments I also noticed one thing more. The commit messages state that translations were removed, but they should also include the added props, no? Otherwise, the changelog will only mention the new props on the breaking change section (and some have typos, saying "transaltion" 😄)

PR #3772, opened 2023-03-19
refactor: remove useMemo

mainframev:
I decided to remove `useMemo`, where it's not necessary

### DIFF
--- a/docs/src/services/table-of-contents.tsx
+++ b/docs/src/services/table-of-contents.tsx
@@ -35,17 +35,19 @@ export function useTableOfContents() {
 
 export function useTableOfContentsRegister({ title, level }: { title: string; level: number }) {
   const [, setTableOfContents] = React.useContext(TableOfContentsContext);
-  const id = React.useMemo(() => uuidv4(), []);
+  const idRef = React.useRef<string>("");
+  const id = uuidv4();
+  idRef.current = id;
 
   React.useEffect(() => {
     setTableOfContents(prevToc => {
       if (level >= DEPTH) return prevToc;
-      if (prevToc.find(item => item.id === id)) {
+      if (prevToc.find(item => item.id === idRef.current)) {
         return prevToc.map(item =>
-          item.id === id ? { ...item, title, slug: slugify(title), level } : item,
+          item.id === idRef.current ? { ...item, title, slug: slugify(title), level } : item,
         );
       }
-      return [...prevToc, { id, title, slug: slugify(title), level }];
+      return [...prevToc, { id: idRef.current, title, slug: slugify(title), level }];
     });
-  }, [title, level, id, setTableOfContents]);
+  }, [title, level, idRef, setTableOfContents]);
 }
--- a/packages/eslint-plugin-orbit-components/docs/rules/unique-id.md
+++ b/packages/eslint-plugin-orbit-components/docs/rules/unique-id.md
@@ -16,11 +16,11 @@ const App = () => (
 ```
 
 ```jsx
-import React, { useMemo } from "react";
+import React from "react";
 import randomID from "../utils/randomID";
 
-const firstID = useMemo(() => randomID("firstID"), []);
-const secondID = useMemo(() => randomID("secondID"), []);
+const firstID = randomID("firstID");
+const secondID = randomID("secondID");
 
 const App = () => (
   <>
--- a/packages/orbit-components/src/Dialog/index.tsx
+++ b/packages/orbit-components/src/Dialog/index.tsx
@@ -127,9 +127,7 @@ const Dialog = ({
   useLockScrolling(wrapperRef, lockScrolling);
   const ref = React.useRef<HTMLDivElement | null>(null);
   const theme = useTheme();
-  const transitionLength = React.useMemo(() => parseFloat(theme.orbit.durationFast) * 1000, [
-    theme.orbit.durationFast,
-  ]);
+  const transitionLength = parseFloat(theme.orbit.durationFast) * 1000;
   const [shown, setShown] = React.useState(false);
 
   useFocusTrap(ref);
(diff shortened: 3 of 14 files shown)


docs/src/services/table-of-contents.tsx
```diff
@@ -35,17 +35,19 @@ export function useTableOfContents() {
 
 export function useTableOfContentsRegister({ title, level }: { title: string; level: number }) {
   const [, setTableOfContents] = React.useContext(TableOfContentsContext);
-  const id = React.useMemo(() => uuidv4(), []);
+  const idRef = React.useRef<string>("");
+  const id = uuidv4();
 
   React.useEffect(() => {
+    idRef.current = id;
```
DSil, line 42: This current code is causing infinite loop, because on every re-render there is a new id. And if there's a new id, the useEffect is executed. And if the useEffect is executed, there's a re-render. And if there's a re-render, there's a new id. And it goes on a loop ➿ 

IMO, this line should be outside the useEffect, and the dependency should change from `id` to `idRef`. This way the useEffect is only executed when the ref changes (which is exactly what we don't expect to happen).

packages/eslint-plugin-orbit-components/docs/rules/unique-id.md
```diff
@@ -16,11 +16,11 @@ const App = () => (
 ```
 
 ```jsx
-import React, { useMemo } from "react";
+import React from "react";
 import randomID from "../utils/randomID";
 
-const firstID = useMemo(() => randomID("firstID"), []);
-const secondID = useMemo(() => randomID("secondID"), []);
```
DSil, line 23: How can we test this? In this situation, this `useMemo` is actually preventing a new ID from being generated everytime. By removing `useMemo`, a new ID will be generated every time. I don't know the consequences of it, because I don't know how to test this plugin…

packages/orbit-components/src/Dialog/index.tsx
```diff
@@ -127,9 +127,7 @@ const Dialog = ({
   useLockScrolling(wrapperRef, lockScrolling);
   const ref = React.useRef<HTMLDivElement | null>(null);
   const theme = useTheme();
-  const transitionLength = React.useMemo(() => parseFloat(theme.orbit.durationFast) * 1000, [
-    theme.orbit.durationFast,
-  ]);
+  const transitionLength = parseFloat(theme.orbit.durationFast) * 1000;
```
DSil, line 130: This const is actually only used inside another hook call (useEffect) so what we should do is to define it inside the hook and define `theme.orbit.durationFast` as the dependency for that hook. This way we avoid the `parseFloat` call on every render.

Something like:
```
  React.useEffect(() => {
    const transitionLength = parseFloat(theme.orbit.durationFast) * 1000;

    const timer = setTimeout(() => {
      setShown(true);
      if (ref.current) {
        ref.current.focus();
      }
    }, transitionLength);
    return () => clearTimeout(timer);
  }, [theme.orbit.durationFast]);
```

On line 135, and remove this line 130

packages/orbit-components/src/InputGroup/index.tsx
```diff
@@ -222,8 +222,8 @@ const InputGroup = React.forwardRef<HTMLDivElement, Props>(
     const labelRef = React.useRef(null);
     const iconRef = React.useRef(null);
 
-    const foundErrors = React.useMemo(() => findPropInChild("error", children), [children]);
-    const foundHelp = React.useMemo(() => findPropInChild("help", children), [children]);
+    const foundErrors = findPropInChild("error", children);
+    const foundHelp = findPropInChild("help", children);
```
DSil, line 226: This is one of the cases it can probably make sense to keep the useMemo, because we are iterating over children twice per call. However, the number of children should not be too big on each iteration, so probably it's still worth it…

Did you try any metrics to see which perform better?

packages/orbit-components/src/Popover/index.tsx
```diff
@@ -37,10 +37,7 @@ const Popover = ({
 
   const theme = useTheme();
 
-  const transitionLength = React.useMemo(() => parseFloat(theme.orbit.durationFast) * 1000, [
-    theme.orbit.durationFast,
-  ]);
-
+  const transitionLength = parseFloat(theme.orbit.durationFast) * 1000;
```
DSil, line 40: Same as Drawer comment.

packages/eslint-plugin-orbit-components/docs/rules/unique-id.md
mainframev: Actually, this is something that we might want to remove entirely because we have a hook `useRandomId` and `useRandomIdSeed` that is working with SSR. 

But that would be breaking and outside of the scope of this PR 

I do not see a problem with having a new id on rerender, it won't affect component behavior and it does not have any performance cost, if needed to keep it the same, I guess we can use the same approach with useRef.

packages/orbit-components/src/InputGroup/index.tsx
mainframev: I run locally profiler, no difference 🤔

docs/src/services/table-of-contents.tsx
mainframev: Yeah, did not pay much attention to this one

packages/orbit-components/src/Drawer/index.tsx
```diff
@@ -170,9 +170,7 @@ const Drawer = ({
   const overlayRef = React.useRef(null);
   const closeButtonRef = React.useRef<HTMLButtonElement | null>(null);
   const scrollableRef = React.useRef<HTMLElement | null>(null);
-  const timeoutLength = React.useMemo(() => parseFloat(theme.orbit.durationNormal) * 1000, [
-    theme.orbit.durationNormal,
-  ]);
+  const timeoutLength = parseFloat(theme.orbit.durationNormal) * 1000;
```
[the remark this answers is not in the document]
DSil: You did for all others but missed this one 😄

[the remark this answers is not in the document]
mainframev: Updated, but still do not see much benefit in that. It still will be called every time on render

DSil (approved):

PR #3773, opened 2023-03-21
Add download prop to Button, TextLink and ButtonLink

DSil:
Originally requested [here](https://skypicker.slack.com/archives/CAMS40F7B/p1679396428317989).
Added to all major components that can render `<a>`.


packages/orbit-components/src/primitives/ButtonPrimitive/README.md
```diff
@@ -31,11 +31,13 @@ Table below contains all types of the props available in ButtonPrimitive compone
...
 | contentAlign     | `string`                   | `center`   | The justify-content of inner container that wraps icons and the children.                                                                                      |
 | contentWidth     | `string`                   |            | The with of inner container that wraps the children.                                                                                                           |
 | dataTest         | `string`                   |            | Optional prop for testing purposes.                                                                                                                            |
 | id               | `string`                   |            | Set `id` for `ButtonPrimitive`                                                                                                                                 |
+| download         | `boolean \| string`        |            | Can only be used when `href` is defined. Adds the `download` attribute to the link.                                                                            |
```
mainframev, line 40: not really an issue, but might be better to be precise and change to `to the anchor element`, technically `link` is a completely different thing in HTML

packages/orbit-components/src/TextLink/types.d.ts
```diff
@@ -4,10 +4,11 @@
 import type * as React from "react";
 
 import type * as Common from "../common/types";
+import type { DownloadWithHrefConditionalProps } from "../primitives/ButtonPrimitive/types";
 
 export type Type = "primary" | "secondary" | "info" | "success" | "warning" | "critical" | "white";
 
-export interface Props extends Common.Globals {
+interface IProps extends Common.Globals {
```
mainframev, line 11: do not use this convention with interfaces, rewrite it into something like `CommonProps`, `BaseProps`, `Properties` etc.

mainframev: Just FYI. The Typescript Handbook tells the developer not to prefix interfaces with I.

mainframev (changes requested):

mainframev:
can you also add a prop to those components to `__examples__` folder?

PR #3774, opened 2023-03-22
chore: add translations codemod

mainframev:
Skipped Pagination and SkipNavigation :D, according to our tracking data components are not being used anywhere

### DIFF
--- a/packages/orbit-components/transforms/__testfixtures__/translations.input.tsx
+++ b/packages/orbit-components/transforms/__testfixtures__/translations.input.tsx
@@ -1,0 +1,42 @@
+import React from "react";
+import Popover from "@kiwicom/orbit-components/lib/Popover";
+import Modal, { ModalSection } from "@kiwicom/orbit-components/lib/Modal";
+import Card, { CardSection } from "@kiwicom/orbit-components/lib/Card";
+import Alert from "@kiwicom/orbit-components/lib/Alert";
+import NavigationBar from "@kiwicom/orbit-components/lib/NavigationBar";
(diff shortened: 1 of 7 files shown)


packages/orbit-components/transforms/__testfixtures__/translations.input.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </Card>
+      <Alert>kek</Alert>
+      <NavigationBar>bar</NavigationBar>
+      <Drawer shown>drawer shown</Drawer>
+      <Tooltip content="bur">
+        <button>kek</button>
+      </Tooltip>
+      <MobileDialogPrimitive>dialog</MobileDialogPrimitive>
+      <Wizard id="kek" completedSteps={1} activeStep={1}>
+        <WizardStep title="Step 1">step</WizardStep>
```
DSil, line 36: I know these are just tests, but can we please not have TS errors in here? 🙏

DSil (commented):
LGTM, just one minor request.

Also, this branch should be started from the translations branch, and the PR against that branch, and not master. Since it handles implementations from that branch. Not critical, but advisable

mainframev:
> Also, this branch should be started from the translations branch, and the PR against that branch, and not master. Since it handles implementations from that branch. Not critical, but advisable, since it also would allow us to merge this into the other branch, that can get put on hold before the next major release

Nope, it should not. It's two separated PRs which are related only by context, not technically

DSil (approved):

PR #3780, opened 2023-03-24
refactor: icon fetching script

mainframev:
Slightly improved icon fetching script, changed request URL to get components, so no need to parse recursively frames. 
Added types, and removed axios as dependency in orbit-components, also updated prettier, because form version `2.5.0` it supports `mts` extension


yarn.lock
```diff
@@ -12496,20 +12496,15 @@ caniuse-api@^3.0.0:
...
-caniuse-lite@^1.0.30001400:
-  version "1.0.30001400"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001400.tgz#3038bee70d8b875604cd8833cb0e5e254ee0281a"
-  integrity sha512-Mv659Hn65Z4LgZdJ7ge5JTVbE3rqbJaaXgW5LEI9/tOaXclfIZ8DW7D7FCWWWmWiiPS7AC48S8kf3DApSxQdgA==
+caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001125, caniuse-lite@^1.0.30001297, caniuse-lite@^1.0.30001400:
+  version "1.0.30001473"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001473.tgz"
+  integrity sha512-ewDad7+D2vlyy+E4UJuVfiBsU69IL+8oVmTuZnH5Q6CIUbxNfI50uVpRHbUPDD6SUaN2o0Lh4DhTrvLG/Tn1yg==
+
+caniuse-lite@^1.0.30001473:
```
DSil: This should not appear repeated. When I run `yarn` on my machine it successfully deletes one of the entries

yarn.lock
```diff
@@ -27022,6 +27022,11 @@ prettier@^2.5.1, prettier@^2.7.1:
   resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.7.1.tgz#e235806850d057f97bb08368a4f7d899f7760c64"
   integrity sha512-ujppO+MkdPqoVINuDFDRLClm7D78qbDt0/NR+wp5FqEZOoTNAjPHWj17QRhu7geIHJfcNhRk1XVQmF8Bp3ye+g==
 
+prettier@^2.8.7:
```
DSil, line 27025: With this one, we have 4 different prettier versions installed. One of them is restricted by storybook (maybe we should update it too soon), but the others are all in the same range, so they can be unified into one.

Manually deleting the entries from yarn.lock and running `yarn` again will solve this.

DSil (commented):
Overall LGTM. Just pointed two things that can optimized in yarn.lock

yarn.lock
mainframev: We should not manually edit yarn.lock, but i will run it again to regenerate

mainframev: they are still might be installed because they specified as pear dependencies somewhere, not only storybook needs it, flowgen needs 2.7.1. I won't manually remove them, cause running it again will add them again, i removed it inside our babel-plugin as dev dep instead, to remove 2.2.1

DSil: > cause running it again will add them again

No, it won't.
Flowgen needs `^2.5.1` and is resolving to 2.7.1 but could resolve to 2.8.7 like all others and avoid 2.7.1 to be installed entirely.
Even if they are peer deps, version 2.8.7 satisfies all the requirements. Only version 2.3.0 will be installed, because it's the maximum limit version required by a storybook package.
VSCode disallowing you should be some setting on your side, because I can delete entries on yarn.lock. But if you don't want to do it, we can actually use [a tool](https://www.npmjs.com/package/yarn-deduplicate) for it.

mainframev: It's not a problem, I will just use nvim, let's see

DSil (approved):

PR #3783, opened 2023-03-27
chore: update supported browsers

mainframev:


DSil:
Why does this cause an increase in the size of some components? 🤯

mainframev:
> Why does this cause an increase in the size of some components? 🤯

because of the babel, it also checks the browser we support

yarn.lock
```diff
@@ -4420,6 +4420,11 @@
   resolved "https://registry.yarnpkg.com/@kiwicom/browserslist-config/-/browserslist-config-3.0.0.tgz#5710f4c94c4165ad9b9e7c0f43feee3dfb0d989b"
   integrity sha512-g0TZoOEEiFmtQncFDiabku44p/AFRP3ZeRjhl+z15Cx+1OcNFqI0O84Alj7YKfcjF7SFwWuDkZJ7ERt+rlE1Hg==
 
+"@kiwicom/browserslist-config@^4.0.0":
```
DSil, line 4423: I find it weird that the other two versions are still here (2.1.0 and 3.0.0). I tried to manually delete them and run `yarn` and they did not come back, which means they're not requested by any package/dependency. Question is why weren't they deleted during the common `yarn` 🤔

DSil (approved):
Besides the question, LGTM

yarn.lock
mainframev: It's probably a yarn issue, we have it with other packages too. There are not many options for what we can do, deleting it and regenerating again - not a good one, the second one - update yarn or as we planned before - migrate to PNPm.

PR #3786, opened 2023-03-27
feat(Tabs): add fullWidth prop

mainframev:
[Design request](https://skypicker.slack.com/archives/GSGN9BN6Q/p1679931228652949)


DSil (commented):
Feature looks good to me, but I have two questions:
1 - Could we have a story for the full width variation?
2 - On Orbit.kiwi, there seems to be a problem with the example. On the current implementation it seems to be also happening (the color is lost when one opens the playground) but now in this PR it's arguable unusable.
Mar-28-2023 10-44-52

Regarding the first one, is non-blocking. The second one can be addressed here or in a different PR (I'd prefer here, but it's up to you).

DSil (approved):

PR #3790, opened 2023-03-28
fix(ModalFooter): count children properly

mainframev:
The issue is quite the same as we had before inside `Itinerary` when conditions were falsy and it was still counting it with `React.Children.count`, there are several ways how to fix it, filter the nodes with `React.isValidElement` and left `React.Children.count` or use `React.Children.toArray(children).length`

### DIFF
--- a/packages/orbit-components/src/Modal/Modal.stories.tsx
+++ b/packages/orbit-components/src/Modal/Modal.stories.tsx
@@ -89,6 +89,23 @@ Sizes.story = {
   },
 };
 
+export const ModalFooterOnly = () => {
+  const isBackButton = boolean("isBackButton", false);
+  const isNextButton = boolean("isNextButton", true);
+
+  return (
+    <Modal>
+      <ModalSection>
+        The dinosaurs looked at Chuck Norris the wrong way once. You know what happened to them.
+      </ModalSection>
+      <ModalFooter flex={["0 0 auto", "1 1 100%"]}>
+        {isBackButton ? <Button>Back</Button> : null}
+        {isNextButton ? <Button>Next</Button> : null}
+      </ModalFooter>
+    </Modal>
+  );
+};
+
 export const ShortModal = () => {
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/Modal/Modal.stories.tsx
```diff
@@ -89,6 +89,23 @@ Sizes.story = {
   },
 };
 
+export const ModalFooterOnly = () => {
+  const isBackButton = boolean("isBackButton", false);
+  const isNextButton = boolean("isNextButton", true);
```
DSil, line 94: I understand this story in the context of this problem and this PR. But having these knobs makes it seem they're props from the component. Moreover, if anything, they should be called `hasBackButton`.

I do agree to have some kind of test to the problem fixed by this PR, but in the form of a test, not of a story.

DSil (approved):
Non blocking, but I'd add a test over a story to test the approach. Can be done separately, though

packages/orbit-components/src/Modal/ModalFooter/index.tsx
```diff
@@ -68,35 +68,34 @@ StyledModalFooter.defaultProps = {
...
+        })}
+      </StyledChild>
+    );
   });
 };
 
 const ModalFooter = ({ dataTest, children, flex = "0 1 auto" }: Props) => {
   const { isMobileFullPage, setFooterHeight } = React.useContext(ModalContext);
   const containerRef = React.useRef<HTMLDivElement | null>(null);
+  const filteredChildren = React.Children.toArray(children).filter(React.isValidElement);
```
DSil, line 98: Couldn't we avoid two iterations over the children? Like, do the map executed in `wrappedChildren` here but filtering out the null ones? Basically what we were doing but not counting the original children, but the wrapped one? 💭

DSil (approved):
Nice 👍

PR #3791, opened 2023-03-28
docs: fix changelog

mainframev:
[Slack request](https://skypicker.slack.com/archives/CAMS40F7B/p1680008830557049)

Instead of having it `onPreBuild` hook inside Gatsby, it makes sense to update it every time we are making a release and that's why moved it to our publish script.

### DIFF
--- a/scripts/publish.mjs
+++ b/scripts/publish.mjs
@@ -164,6 +164,7 @@ async function previewChangelog() {
 
     await $`yarn install`;
     await $`yarn lerna publish --no-private --conventional-commits --create-release github`;
+    await $`yarn docs changelog`;
 
     const timestamp = await getLatestReleaseTime();
     const names = await getUserNames(timestamp);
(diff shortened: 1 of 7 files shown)


scripts/publish.mjs
```diff
@@ -164,6 +164,7 @@ async function previewChangelog() {
 
     await $`yarn install`;
     await $`yarn lerna publish --no-private --conventional-commits --create-release github`;
+    await $`yarn docs changelog`;
```
DSil, line 167: If this is executed after the publish, when/where will this generated file (docs/src/data/log-md) be pushed to the repository, so that the docs are built with it?

DSil: Ok so the release process will include an extra commit and push, made manually, right?

DSil (approved):

scripts/publish.mjs
mainframev: Right, maybe later will add it also to a part of the script, as we are still gonna change it in Q2, the release process.

PR #3792, opened 2023-03-28
chore: improve ts

mainframev:
As we did recently #3782, it wasn't good, because stories were added to the build, good thing is, that no releases happened meanwhile, so it did not get to npm. 

Separated configs on build and regular, one for checking types only, the second one is for build (includes only necessary files and creates declarations)  

Also, because of these changes, a few exciting missings founded, so fix them too (including new types package)


packages/orbit-components/src/SegmentedSwitch/__tests__/index.test.tsx
```diff
@@ -28,7 +27,7 @@ describe("SegmentedSwitch", () => {
     expect(screen.getByText(label)).toBeInTheDocument();
     expect(screen.getByTestId(dataTest)).toBeInTheDocument();
 
-    // $FlowFixMe
+    // @ts-expect-error TODO
     userEvent.tab(screen.getByLabelText("Male"));
```
DSil, line 31: I don't know when this changed, but the `tab` function no longer receives the element to be tabbed. So we can remove the argument (and the TS annotation) and the test will still pass 🙂

DSil (approved):
Non-blocking comment. Good catch 👍 🙏

packages/orbit-components/src/SegmentedSwitch/__tests__/index.test.tsx
mainframev: ah, right, thanks ;)

PR #3797, opened 2023-04-03
feat(BadgeListItem): add size prop

DSil:
BadgeListItem now accepts size prop for changing the text size. Size can be small (default) or normal.

The icon wrapper size was also updated to match Figma's size. The Icon size is also off regarding Figma, but it seemed too small if figma size is applied, so no changes were made.

**Update:** as confirmed with designers, icon size is always the same and the `size` prop should be on the item, rather than on the parent wrapper component.

### DIFF
--- a/packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
+++ b/packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
@@ -52,8 +52,8 @@ export const StyledVerticalBadge = styled.div<{ $type?: Props["type"] }>`
     justify-content: center;
     margin-${right}: ${theme.orbit.spaceXSmall};
     flex-shrink: 0;
-    height: ${theme.orbit.heightIconMedium};
-    width: ${theme.orbit.widthIconMedium};
+    height: ${theme.orbit.heightIconLarge};
+    width: ${theme.orbit.widthIconLarge};
     border-radius: ${theme.orbit.borderRadiusCircle};
     svg {
       height: ${theme.orbit.heightIconSmall};
@@ -70,11 +70,6 @@ export const StyledBadgeContent = styled.div`
   ${({ theme }) => css`
     display: inline-flex;
     align-items: center;
-    &,
-    ${StyledText} {
-      font-size: ${theme.orbit.fontSizeTextSmall};
-      line-height: ${theme.orbit.lineHeightTextSmall};
-    }
 
     ${StyledTooltipChildren} ${StyledText} {
       font-weight: ${theme.orbit.fontWeightMedium};
@@ -90,6 +85,7 @@ const BadgeListItem = ({
   icon,
   strikeThrough,
   type = TYPE_OPTIONS.NEUTRAL,
+  size = "small",
   dataTest,
   children,
 }: Props) => {
@@ -102,7 +98,7 @@ const BadgeListItem = ({
           })}
       </StyledVerticalBadge>
       <StyledBadgeContent>
-        <Text type="secondary" size="small" as="span" strikeThrough={strikeThrough}>
+        <Text type="secondary" size={size} as="span" strikeThrough={strikeThrough}>
           {children}
         </Text>
       </StyledBadgeContent>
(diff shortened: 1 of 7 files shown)


mainframev:
@DSil I find it a bit strange, that we need to pass size prop for each item explicitly, is there a usecase, where only a list item should be bigger or smaller than another? because, in my opinion, the size prop would be better to add to the `BadgeList` parent, not to item 🤔

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
```diff
@@ -90,6 +85,7 @@ const BadgeListItem = ({
   icon,
   strikeThrough,
   type = TYPE_OPTIONS.NEUTRAL,
+  size = "small",
```
mainframev, line 88: could you also have it as const as above with type options? 🙏🏻

packages/orbit-components/src/BadgeList/BadgeListItem/types.d.ts
```diff
@@ -5,9 +5,12 @@ import type * as React from "react";
 import type * as Common from "../../common/types";
 
 export type Type = "neutral" | "info" | "success" | "warning" | "critical";
+export type Size = "small" | "normal";
+
 export interface Props extends Common.Globals {
   readonly children: React.ReactNode;
   readonly type?: Type;
+  readonly size?: Size;
```
mainframev, line 13: add the prop also to `index.js.flow` 🙏🏻

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
```diff
@@ -52,8 +52,8 @@ export const StyledVerticalBadge = styled.div<{ $type?: Props["type"] }>`
     justify-content: center;
     margin-${right}: ${theme.orbit.spaceXSmall};
     flex-shrink: 0;
-    height: ${theme.orbit.heightIconMedium};
-    width: ${theme.orbit.widthIconMedium};
+    height: ${theme.orbit.heightIconLarge};
```
mainframev, line 55: so if I got it right, only the text should be changed and the icon should be always large (24px)? right?

mainframev (commented):
Could you also add size prop to doc examples? 🙏🏻

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
DSil: I am not sure, because of what I reported that icon sizes in Figma (or the icon placeholders used there) are too small to be perceived. I'd wait on designers' feedback on this one

packages/orbit-components/src/BadgeList/BadgeListItem/types.d.ts
DSil: For some reason I thought the flow files were being generated automatically 🤔

DSil:
@mainframev I had the same thought as you. I decided not to approach the addition of the prop on the parent component for three reasons:
1 - to apply Encapsulation, delegating the responsibility of handling sizes to the component that actually needs it  
2 - to give devs (and designers) more flexibility
3 - to avoid children iteration and manipulation

packages/orbit-components/src/BadgeList/BadgeListItem/types.d.ts
mainframev: we left that idea because of slow flowgen, it takes him a lot to generate files. We use it only on some root package files, like `index.ts` and others

mainframev:
> @mainframev I had the same thought as you. I decided not to approach the addition of the prop on the parent component for three reasons: 1 - to apply Encapsulation, delegating the responsibility of handling sizes to the component that actually needs it 2 - to give devs (and designers) more flexibility 3 - to avoid children iteration and manipulation

> 3 - to avoid children iteration and manipulation

children manipulation here can be avoided, by targeting the style element 

> 1 - to apply Encapsulation, delegating the responsibility of handling sizes to the component that actually needs it 

Fair point, has an advantage, but I can imagine having a list of this like 10 items and the need to add the size everywhere is kinda strange :) 

> 2 - to give devs (and designers) more flexibility

I would ask designers about if it's possible to mix different sizes inside BadgeList, if not, I would change it to parent ;)

mainframev (approved):

PR #3805, opened 2023-04-11
fix(Toast): icon on Promisse Toast was not rendered + infinite re-render fixed

DSil:
The icon was not rendered when the Toast was triggered by a promise.
There was also an infinite re-render problem, solved in this commit. The infinite re-render problem was caused by a workaround to a known issue on react-hot-toast that was fixed in the latest version. By upgrading, the workaround can be removed, removing the infinite re-render as well.

BREAKING CHANGE: the createToastPromise function now receives three arguments. The first two are required and similar to what was accepted before, except the icon. The third one is for handling the icon on different toast types. Check documentation for further details.

### DIFF
--- a/packages/orbit-components/src/Toast/ToastMessage.tsx
+++ b/packages/orbit-components/src/Toast/ToastMessage.tsx
@@ -5,10 +5,9 @@ import mq from "../utils/mediaQuery";
 import Stack from "../Stack";
 import defaultTheme from "../defaultTheme";
 import Text from "../Text";
-import { fadeIn, fadeOut, lightAnimation, getPositionStyle, createRectRef } from "./helpers";
+import { fadeIn, fadeOut, lightAnimation, getPositionStyle } from "./helpers";
 import useTheme from "../hooks/useTheme";
 import useSwipe from "./hooks/useSwipe";
-import mergeRefs from "../utils/mergeRefs";
 import type { Toast as Props } from "./types";
 
 const StyledWrapper = styled(({ className, children, ariaLive }) => (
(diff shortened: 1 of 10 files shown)


packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -115,7 +110,7 @@ const ToastMessage = ({
     >
       <StyledInnerWrapper
         visible={visible}
-        ref={mergedRef}
+        ref={innerRef}
```
mainframev, line 113:
```suggestion
        ref={ref}
```

packages/orbit-components/src/Toast/types.d.ts
```diff
@@ -44,15 +44,19 @@ export interface Toast {
...
-) => void;
+export type IconType = Pick<DefaultToastOptions, "icon">;
 
-export type createToastPromise = <T>(promise: Promise<T>, options: Options<T>) => Promise<T>;
+export type createToast = (message: ValueOrFunction<Renderable, Toast>, options?: IconType) => void;
+
+export type createToastPromise = <T>(
+  promise: Promise<T>,
+  messages: Options<T>,
+  options?: IconType & { [key in ToastType]?: IconType },
```
mainframev, line 59:
```suggestion
  options?: IconType & Record<ToastType, IconType>
```

mainframev (changes requested):
LGTM, just small things

packages/orbit-components/src/Toast/types.d.ts
mainframev: Partial<Record<<ToastType>, IconType>> can work then

DSil: Right, they're the same. But I'll switch to your suggestion to make the code more Typescript conventional 👍

mainframev (approved):

PR #3809, opened 2023-04-13
Convert Slider components to hooks

DSil:
Slider component was still a Class component. This PR changes it into a functional component that uses hooks.

PureSlider was being exported but wasn't used nor documented. Therefore, it is no longer exported.


packages/orbit-components/src/Slider/index.tsx
```diff
@@ -69,403 +70,381 @@ const StyledSliderInput = styled.div`
...
+
+  React.useEffect(() => {
+    const newValue = Array.isArray(defaultValue)
+      ? defaultValue.map(item => Number(item))
+      : Number(defaultValue);
+
+    setValue(newValue);
+  }, [defaultValue]);
+
+  const calculateValue = React.useCallback(
```
mainframev, line 107: I think you are overusing the useCallbacks here, profile first before those optimizations, they seem as not needed to be here everywhere

packages/orbit-components/src/Slider/components/Handle/index.tsx
```diff
@@ -96,9 +96,9 @@ StyledHandle.defaultProps = {
 
 interface Props {
   tabIndex: string | number;
-  onMouseDown: React.MouseEventHandler<HTMLDivElement>;
```
mainframev, line 99: Remove also the `index.js.flow` files from child components, they are not needed, because we do not export them

mainframev (changes requested):
LGTM, just one

packages/orbit-components/src/Slider/index.tsx
DSil: Agreed. I ended up abstracting a lot of the functions to the utils file, so that they are just called and not redefined on every render. The only useCallback usage is for props that are not expected to change often and, therefore, can have a callback optimized.

mainframev (approved):

PR #3812, opened 2023-04-17
Button update

DSil:
Addressing the Figma review, this PR includes the following:

- Icon size on Large Button (and ButtonLink) is now 24px
- Space between Icon and Text inside Button (and ButtonLink) is now always 8px, regardless of Button size.
- Tokens for spacing between Icon and Text inside Button (and ButtonLink) were all merged into one single token - BREAKING CHANGE (can be detached from this PR to prevent breaking change)
- Subtle AlertButton now has custom colors, generated via opacity from token colors. Expected to be used only inside Alert.

https://kiwicom.atlassian.net/browse/ORBIT-2709


packages/orbit-components/src/primitives/ButtonPrimitive/common/getIconContainer.ts
```diff
@@ -45,9 +37,19 @@ const getIconContainer: getIconContainerType = ({
   size = SIZE_OPTIONS.NORMAL,
   iconForeground,
 }) => {
-  const onlyIcon = Boolean(iconLeft && !children);
-  const sizeIcon = size === ICON_SIZES.SMALL ? ICON_SIZES.SMALL : ICON_SIZES.MEDIUM;
+  let sizeIcon: ICON_SIZES;
```
mainframev, line 40: we usually try to avoid using let variables and switch/case, we prefer to have it declarative way, instead of imperative. Rewrite it to a function that will return the values you need 🙏🏻

mainframev (changes requested):

mainframev (approved):

PR #3815, opened 2023-04-19
chore(Illustration): add type to consts file

DSil:
Small change so that the array only has possible illustration names. It is also valid for other stories that are relying on this array for knobs.


mainframev:
Rebase it with master and drop that commit about ModalFooter ;)

mainframev:
> @mainframev no, the other branch was merged to this one, as described on the PR description. It's ok to merge this one as is, including both commits.

It's not ok, it's not related commit to PR, if we later will need to get back to that PR in slack, it will be better to contain only related stuff here

PR #3816, opened 2023-04-19
feat(Modal): remove flex-none from ModalFooter

DSil:
BREAKING CHANGE: visual breaking change on ModalFooter. Flex-none is no longer applied on largeMobile and bigger. A layout component wrapper might be needed.

https://kiwicom.atlassian.net/browse/ORBIT-2711

This PR is opened against #3815 to avoid pipeline errors. It will be redirected to `master` once the other branch is merged

### DIFF
--- a/packages/orbit-components/src/Modal/ModalFooter/index.tsx
+++ b/packages/orbit-components/src/Modal/ModalFooter/index.tsx
@@ -16,10 +16,6 @@ const StyledChild = styled.div<{ flex?: Props["flex"] }>`
     flex: ${flex};
     box-sizing: border-box;
     padding: ${rtlSpacing(`0 ${theme.orbit.spaceXSmall} 0 0`)};
-
-    ${media.largeMobile(css`
-      flex: none;
-    `)};
   `}
 `;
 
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/Modal/ModalFooter/index.tsx
```diff
@@ -16,10 +16,6 @@ const StyledChild = styled.div<{ flex?: Props["flex"] }>`
     flex: ${flex};
     box-sizing: border-box;
     padding: ${rtlSpacing(`0 ${theme.orbit.spaceXSmall} 0 0`)};
-
-    ${media.largeMobile(css`
```
mainframev, line 20: I guess it's better of just removal, add there `width: 100% and display:flex;`

mainframev:
LGTM. Could you also correct a bit of breaking the change message? It seems too long, I would prefer to have it shorter and no need to mention the wrapper layout component there ;)

packages/orbit-components/src/Modal/ModalFooter/index.tsx
DSil: I would refrain from adding display properties to inner wrapper elements that can't be controlled. If devs need a display flex, they'll add it with a wrapper component. But if they don't want it, it will always be there… What would be the advantage of adding a display and a width there?

mainframev: Ok, I forgot, that flex prop still be applied there, so no need for that then

mainframev (approved):

PR #3817, opened 2023-04-20
Add card changes

DSil:
`icon` prop is deprecated. An example on how to keep using `CarrierLogos` was added on the Modal example, as it is the only context expected to feature this.
`ButtonLink` is expected to be used for actions. Therefore, all docs now use it.
A spacing between actions and title was added.

[ORBIT-2712](https://kiwicom.atlassian.net/browse/ORBIT-2712)


packages/orbit-components/src/Card/components/Header/index.tsx
```diff
@@ -28,6 +28,25 @@ const StyledCloseContainer = styled.div`
   z-index: 1;
 `;
 
+const StyledActionsWrapper = styled.div`
+  ${({ theme }) => css`
+    display: inline-flex;
+    align-items: flex-start;
+    justify-content: flex-end;
+    flex-grow: 0;
+    flex-shrink: 0;
```
mainframev, line 37: use `flex: 0 0` instead

packages/orbit-components/src/Card/components/Header/index.tsx
```diff
@@ -81,11 +100,7 @@ const Header = ({
     {header && icon}
     {header}
     {expandable && !actions && <ChevronIcon color="secondary" expanded={expanded} />}
-    {actions && (
-      <Stack inline grow={false} justify="end">
-        {actions}
-      </Stack>
-    )}
+    {actions && <StyledActionsWrapper>{actions}</StyledActionsWrapper>}
```
mainframev, line 103: could not you achieve the same with the Stack as before?

mainframev (changes requested):

packages/orbit-components/src/Card/components/Header/index.tsx
DSil: I tried keeping the Stack but I was unable to achieve. I don't think stacking two Stacks is a better approach than having one wrapper component. How would you suggest it?

DSil: Managed to do it without the wrapper, thanks 👍

mainframev (approved):

PR #3820, opened 2023-04-21
docs: add missing babel-node

mainframev:


DSil (commented):
Running this locally causes changes to yarn.lock. Small ones, but worth updating. Can you please do it?

DSil (approved):

PR #3821, opened 2023-04-21
Update TextLink

DSil:
Fixed colors on active state and also the size of the icons

[ORBIT-2720](https://kiwicom.atlassian.net/browse/ORBIT-2720)

### DIFF
--- a/packages/orbit-components/src/TextLink/index.tsx
+++ b/packages/orbit-components/src/TextLink/index.tsx
@@ -78,16 +78,21 @@ const getSizeToken = ({ theme, size }: { theme: Theme; size: Common.Size }): str
   return sizeTokens[size];
 };
 
-const StyledIconContainer = styled(({ children, className }) => (
-  <span className={className}>{children}</span>
-))`
-  ${({ theme }: { theme: Theme }) => css`
+const getIconSize = ({ theme, size }: { theme: Theme; size: Props["size"] }) => {
+  if (size === SIZE_OPTIONS.LARGE)
+    return { width: theme.orbit.widthIconLarge, height: theme.orbit.heightIconLarge };
+  if (size === SIZE_OPTIONS.SMALL)
+    return { width: theme.orbit.widthIconSmall, height: theme.orbit.heightIconSmall };
+  return { width: theme.orbit.widthIconMedium, height: theme.orbit.heightIconMedium };
+};
+
+const StyledIconContainer = styled.span`
+  ${({ theme, size }: { theme: Theme; size: Props["size"] }) => css`
     display: flex;
     align-items: center;
 
     svg {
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/TextLink/index.tsx
```diff
@@ -78,16 +78,23 @@ const getSizeToken = ({ theme, size }: { theme: Theme; size: Common.Size }): str
...
 
+const getIconSize = (theme: Theme, size: Props["size"]) => {
+  if (size === SIZE_OPTIONS.LARGE)
+    return { width: theme.orbit.widthIconLarge, height: theme.orbit.heightIconLarge };
+  if (size === SIZE_OPTIONS.SMALL)
+    return { width: theme.orbit.widthIconSmall, height: theme.orbit.heightIconSmall };
+  return { width: theme.orbit.widthIconMedium, height: theme.orbit.heightIconMedium };
+};
+
 const StyledIconContainer = styled(({ children, className }) => (
```
mainframev, line 89: I know it was not touched, but I think we can remove here this arrow function with span and have only `styled.span` ;)

packages/orbit-components/src/TextLink/index.tsx
```diff
@@ -78,16 +78,23 @@ const getSizeToken = ({ theme, size }: { theme: Theme; size: Common.Size }): str
   return sizeTokens[size];
 };
 
+const getIconSize = (theme: Theme, size: Props["size"]) => {
```
mainframev, line 81:
```suggestion
const getIconSize = ({ theme, size }: { theme: Theme, size: Props["size"] }) => {
```

packages/orbit-components/src/TextLink/index.tsx
```diff
@@ -78,16 +78,23 @@ const getSizeToken = ({ theme, size }: { theme: Theme; size: Common.Size }): str
...
 ))`
-  ${({ theme }: { theme: Theme }) => css`
+  ${({ theme, size }: { theme: Theme; size: Props["size"] }) => css`
     display: flex;
     align-items: center;
 
     svg {
-      width: ${theme.orbit.widthIconSmall};
-      height: ${theme.orbit.heightIconSmall};
+      ${getIconSize(theme, size)};
```
mainframev, line 97:
```suggestion
      ${getIconSize};
```

mainframev (changes requested):
LGTM :) nothing serious, just some small suggestions

packages/orbit-components/src/TextLink/index.tsx
DSil: I like the previous suggestion but on this one I'd rather do `getIconSize({theme, size})`, like we do on `getSizeToken`

mainframev (approved):

PR #3823, opened 2023-04-21
feat(InputField): add changes after components audit

mainframev:
[ORBIT-2723](https://kiwicom.atlassian.net/browse/ORBIT-2723)


packages/orbit-components/src/FormLabel/index.tsx
```diff
@@ -80,6 +81,9 @@ const FormLabel = styled(
     color: ${!filled || disabled ? theme.orbit.colorFormLabel : theme.orbit.colorFormLabelFilled};
     line-height: ${theme.orbit.lineHeightTextNormal};
     margin-bottom: ${theme.orbit.spaceXXSmall};
+    ${mq.largeMobile(css`
+      margin-left: ${theme.orbit.spaceXXSmall};
+    `)}
```
DSil, line 86: Their request with this was to align the tooltip with the start of the input, but it was already happening… with this change, it actually became misaligned. Check the prints:
Before:
image

Now:
image

So we either don't move forward with this change (my personal preference, but just personal) or we have to adjust the tooltip positioning as well

DSil (changes requested):
Overall LGTM, just one detail about the new label margin that needs to be addressed. And also two possible improvements regarding commit messages:

- On the breaking change message on the first commit, I would replace `Normal size now is default.` to `Normal size is now the only option.` because saying default seems that there are other options. If anything, the correct phrasing would be `is now` instead of `now is`, but that is minor change.
- On commit 3 there is a typo on the name of the component 😄 It should be `ErrorFormTooltip`.
- On commit 5, if it is to keep 😅, the message is truncated. I am not sure how the changelog will be generated in that case. I would just rephrase it to something `add margin to ensure message is aligned with input`.

DSil:
I also noticed it is missing the request to change the border color on hover, so please add that 🙏

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -92,6 +76,7 @@ const StyledFormFeedbackTooltip = styled.div<{
 
     ${media.largeMobile(css`
       width: auto;
+      border-radius: ${theme.orbit.borderRadiusLarge};
```
DSil, line 79: This is actually not correct. It should be Large on default and on desktop (and bigger) should be Normal.

I also checked with Will and it should also be applied to the actual input, and not only the tooltip.

packages/orbit-components/src/FormLabel/index.tsx
mainframev: I will update the position ;)

DSil:
I still see the tooltip misaligned on Mobile
Screenshot 2023-04-26 at 12 00 40

And the typo on the commit message is still there.

And it is still missing the border-radius of 6px on the input responsive.

mainframev:
> And it is still missing the border-radius of 6px on the input responsive

Just incorrect breakpoint ;)

packages/orbit-components/src/InputField/index.tsx
```diff
@@ -85,7 +85,7 @@ export const FakeInput = styled(({ children, className }) => (
       : theme.orbit.backgroundInput};
     font-size: ${theme.orbit.fontSizeInputNormal};
     transition: all ${theme.orbit.durationFast} ease-in-out;
-    border-radius: ${theme.orbit.borderRadiusNormal};
+    border-radius: ${theme.orbit.borderRadiusLarge};
```
DSil, line 88: This change should be on its own commit

DSil (commented):
Just one final touch regarding commits and it's 👍

DSil (approved):

PR #3827, opened 2023-04-24
Add tag changes

DSil:
This fixes the border-radius on responsive and also adds the `iconLeft` prop

[ORBIT-2722](https://kiwicom.atlassian.net/browse/ORBIT-2722)

### DIFF
--- a/packages/orbit-components/src/Tag/index.tsx
+++ b/packages/orbit-components/src/Tag/index.tsx
@@ -10,6 +10,7 @@ import { SIZES, STATES, TYPES } from "./consts";
 import KEY_CODE_MAP from "../common/keyMaps";
 import resolveColor from "./helpers/resolveColor";
 import resolveCircleColor from "./helpers/resolveCircleColor";
+import mq from "../utils/mediaQuery";
 import type { Props, Type } from "./types";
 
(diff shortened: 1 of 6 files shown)


packages/orbit-components/src/Tag/index.tsx
```diff
@@ -168,6 +173,25 @@ StyledClose.defaultProps = {
   theme: defaultTheme,
 };
 
+const StyledIconContainer = styled.div`
```
mainframev, line 176: I think you can use Stack here. Just wrap `iconLeft`, `children` and `CloseContainer` inside Stach with props `inline`, `align="center"` and `spacing="XSmall` :)

mainframev: Yeah, I would personally prefer to have less code. Additional div won't make any difference in the markup, but the custom styled-component will be bundled, while Stack will be tree just reused, so less addition to bundle ;)

DSil: Fair point. I'll leave it like this for now but maybe we can reassess all this during component refactoring for new styling solution and have a consistent approach across all components in Orbit 🙂

PR #3333, opened 2022-04-08
feat: various token changes

mainframev:
This PR solves:

- update design-tokens to style-dictionary
- token changes (new token names)
- update of orbit and docs codebases to a new tokens

 Storybook: https://orbit-mainframev-dev-tokens.surge.sh

PR #3829, opened 2023-04-24
feat: new tailwind preset package

mainframev:
New tailwind preset package, that will use our design tokens. 

Refactoring of design tokens to styled-dictionary, which we wanted to do before and this seems to be a good reason to finally do that, although without breaking changes like token naming changes introduced in #3333 

Created two presets (with default foundation and component level)

Component level preset should be later used only as internal, as users are supposed to consume only foundation tokens and base preset. But for now, we also will export it, as on Search they will need it for travel pages. 

The PR is kinda big, mostly because of token package changes, so we can review that starting from the tokens and then tailwind presets, as it's heavily dependent on tokens. 

- [x] Refactor tokens package from theo to styled-dictionary
- [x] Deprecate old token structure
- [x] Create two presets for tailwind (foundation and components)
- [x] Add tests for configs
- [x] Run loki
- [x] Add README.md and page to Gatsby docs
- [x] Add more fixtures and tests to test applied CSS from the config for foundation
- [x] Release beta versions


mainframev:
@DSil you can start reviewing it. I suggest starting from the design tokens package, as I plan to finish the presets while you can review tokens :)

.eslintrc.js
```diff
@@ -172,39 +172,6 @@ module.exports = {
         "react/no-unstable-nested-components": "off",
       },
     },
-    {
-      files: "packages/eslint-plugin-orbit-components/**/*.ts",
-      rules: {
-        "@typescript-eslint/prefer-readonly-parameter-types": "off",
-        "@typescript-eslint/explicit-function-return-type": "off",
-        "@typescript-eslint/explicit-module-boundary-types": "off",
```
DSil, line 180: Why are we changing this on the scope of this task?

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
+# Schema of @kiwicom/orbit-design-tokens
+
+Orbit design tokens could be considered as a simple way to describe, document and use the visual language of Orbit design system, but there are multiple requirements that it needs to meet and support.
+
+Therefore, we use structured and standardized way Orbit design tokens are being created, stored and how we work with them.
```
DSil, line 5: I am not sure what this intends to say 😕 Maybe this?

```suggestion
Therefore, we use a structured and standardized way to create, store and work with Orbit design tokens.
```

DSil (commented):
I've done an overall initial review especially on documentation and tests. I am a bit confused about the whole structure of the dictionary.

For instance, in one place we have
badge > badge-types, where we define the different tokens for the different badge types
But then we also have
button > button-primary, where we define the tokens for the primary button only. But primary is a type, so why don't we follow the same structure for both components?

What is the process of deciding if a set of tokens has its own file or falls under the "type"?

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
...
+}
+```
+
+The above example is being translated with custom attribute transforms. For information how we work with `namespace.object.variant.property` hierarchical tree structure check transforms document.
+
+### Namespace
+
+In our design tokens an `namespace` is considered the top-level layer of design tokens that determines the intention of the token – where and how exactly it will be used.
+
+Possible values of the object level are:
```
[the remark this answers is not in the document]
mainframev: Yep

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
...
+
+- The `foundation` is being used for the internal design tokens that are used as aliases for `global` and `component` design tokens, and/or interpretation of values that can be overwritten by custom theming,
+- the `global` is being used for the most common design tokens that interprets the fundamental visual language of the Orbit design system,
+- the `component` is being used for grouped design tokens that can be used for similar components, e.g. shared design token for the height of the buttons and/or inputs.
+
+### Object
+
+The `object` is an attribute to specify either the name of some component (component group) or [foundation object](#foundation-objects).
```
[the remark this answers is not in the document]
mainframev: Object here is not an object in classing meaning, above it's described the structure like: 

`## Namespace / Object / Variant / Property`

so for instance, in case of the size foundation: 

```
   "foundation": {
    "size": {
      "small": {
        "type": "size",
        "value": 16,
        "internal": true
      },
    .... 
```

where the foundation is the namespace and "size" is the Object that contains normal object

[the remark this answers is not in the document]
mainframev: feel free to suggest the alternative :) I also just moved that part from the old MR

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
...
+
+For typography, we use combination of the [sizing scale](#sizing-scale) and scale that represents the visual hierarchy of heading levels:
+
+- display
+- displaySubtitle
+- title1 to title5 according to existing HTML heading tags
+
+### Elevation scale
+
+Based on our [elevation system](https://orbit.kiwi/foundation/elevation-hierarchy/), we use only following naming variants for the definition of them:
```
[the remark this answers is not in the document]
mainframev: elevations :)

packages/orbit-design-tokens/src/__tests__/index.test.ts
```diff
@@ -43,40 +43,32 @@ describe("getTokens should accept some palette and base foundation", () => {
...
-    expect(theme.heightIconMedium).toBe(brand.base.iconSizeMd);
-    expect(theme.widthIconLarge).toBe(brand.base.iconSizeLg);
-    expect(theme.heightIconLarge).toBe(brand.base.iconSizeLg);
-    expect(theme.heightInputNormal).toBe(brand.base.sizeXl);
-    expect(theme.heightInputLarge).toBe(brand.base.size2xl);
-    expect(theme.heightInputSmall).toBe(brand.base.sizeLg);
-    expect(theme.heightButtonNormal).toBe(brand.base.sizeXl);
-    expect(theme.heightButtonLarge).toBe(brand.base.size2xl);
-    expect(theme.heightButtonSmall).toBe(brand.base.sizeLg);
+    expect(theme.heightInputNormal).toBe(brand.size.extraLarge);
```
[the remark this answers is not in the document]
DSil: I meant if we change `heightInputNormal` to use any other token, this test will fail. But what's being tested here won't be broken. This test relies on `heightInputNormal` using one token, but that can change. My suggestion would be to have a custom token (let's call it `customToken`) that uses `foundation.size.extraLarge` and assert that when we customize `extraLarge` the `customToken` also gets the new value, which is exactly what this test is supposed to test. That way, if we ever change `heightInputNormal`, this test won't fail, as it shouldn't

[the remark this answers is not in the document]
mainframev: I will leave it as it is because otherwise, we need to use another function, not the `getTokens`. In case, if the default tokens that depended on the base will be changed - it's not a problem to adjust the test, as it's expected will be failed.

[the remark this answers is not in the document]
mainframev: I will leave it as it is, in case it will be changed - it's expected to adjust the test also in that case, not a lot of extra work. Currently, this function is typed and we will need to add suppression in order to add "fake" token

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
...
+- font-family
+- font-weight
+- z-index
+- modifier
+- spacing
+- opacity
+
+## Additional attributes
+
+For some cases we define and use additional attributes that help us maintain the whole structure of Orbit design tokens.
```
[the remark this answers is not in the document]
mainframev: I will remove that paragraph, because currently it's just useless

packages/orbit-design-tokens/README.md
```diff
@@ -28,15 +28,15 @@ Now you can use tokens by typing `defaultTokens` and just continue with the dot-
 
 ### Generating your theme
 
-**orbit-design-tokens** offers a possibility to generate different themes based on the inputted foundations.
+**orbit-design-tokens** offer the possibility to generate different themes based on the inputted foundations.
```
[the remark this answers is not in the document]
mainframev: Grammatically it is correct. So suggestion should be then`**orbit-design-tokens** package offers

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,276 @@
...
+- suppressed
+- flat
+- action
+- fixed
+- raised
+- overlay
+
+### Viewports
+
+smallMobile, mediumMobile, largeMobile, tablet, desktop, largeDesktop
```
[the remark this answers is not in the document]
mainframev: Do not think so, our docs are already covering that

mainframev:
> I've done an overall initial review especially on documentation and tests. I am a bit confused about the whole structure of the dictionary.
> 
> For instance, in one place we have badge > badge-types, where we define the different tokens for the different badge types But then we also have button > button-primary, where we define the tokens for the primary button only. But primary is a type, so why don't we follow the same structure for both components?
> 
> What is the process of deciding if a set of tokens has its own file or falls under the "type"?

ideally yes, but button just have more types, and having all of them in a single will end up as quite a big one, so better was here to split.

packages/orbit-design-tokens/docs/SCHEMA.md
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+
+The above example is being translated with custom attribute transforms. For information on how we work with the `namespace.object.variant.property` hierarchical tree structure, check the transforms document.
+
+### Namespace
+
+In our design tokens, a `namespace` is considered the top-level layer of a token. It determines the intention of the token – where and how exactly it will be used.
+
+Possible values of the `namespace":
```
DSil, line 34:
```suggestion
Possible values of the `namespace`:
```

packages/orbit-design-tokens/docs/SCHEMA.md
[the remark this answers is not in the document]
DSil: Ahh, understood. I got confused by `object` (the JS primitive) and the name `Object` (used in this tokens scope) 😅 I think it is acceptable as is, but I would personally prefer something like

```
The `object` attribute is used to specify either the name of some component (component group) or the [foundation object](#foundation-objects).
```

Just because I think it leaves no doubt between the two _objects_. But I will leave the decision for you.

packages/orbit-tailwind-preset/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "scripts": {
+    "build": "tsup",
+    "pretest": "yarn workspace @kiwicom/orbit-design-tokens build && yarn build && yarn tailwindcss -o style.css",
+    "test": "jest"
+  },
+  "peerDependencies": {
+    "tailwindcss": ">=3.3.2"
+  },
+  "dependencies": {
+    "@kiwicom/orbit-design-tokens": "*"
```
DSil, line 36: As agreed, this should actually start on the version of orbit-design-tokens that is compatible with tailwind

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -0,0 +1,158 @@
...
+
+const cfg = (options?: Options): Config => {
+  const { disablePreflight = true } = options || {};
+
+  return {
+    content: ["auto"],
+    corePlugins: {
+      preflight: disablePreflight ? false : undefined,
+    },
+    theme: {
```
DSil, line 40: I think the content of this object needs to be wrapped in `extends: { … }`. Otherwise it will replace the original tailwind classes, which is something we don't want to do

packages/orbit-tailwind-preset/src/presets/foundation/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import defaultFoundation from "./theme/defaultFoundation";
+import { spacing, screens, font, boxShadow, duration } from "./theme";
+
+const config: Config = {
+  content: ["auto"],
+  corePlugins: {
+    preflight: false,
+  },
+  theme: {
```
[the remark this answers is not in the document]
mainframev: I did, but not for all, for instance colors we do not want to extend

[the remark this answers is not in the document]
DSil: Actually, I gave it a second thought and maybe we should not extend spacing as well. I can imagine we only want devs to be able to use our spacing tokens. If anything, they can extend them (hopefully won't be needed).
The direct consequence of this is that the currently existing Tailwind pages will have their spaces broken, but that's a needed migration that they will need to perform upon adoption of Orbit's Tailwind preset. WDYT?

packages/orbit-design-tokens/package.json
```diff
@@ -17,39 +17,35 @@
...
-    "build-html": "theo --setup ./src/theo/setup.cjs ./src/theo/theo-spec.json --format kiwi.html --dest ./pages",
-    "build-web": "theo ./src/theo/theo-spec.json --transform web --format json,scss,sass,less,styl,cssmodules.css,common.js --dest ./output",
-    "build-ios": "theo ./src/theo/theo-spec.json --transform ios --format ios.json --dest ./output",
-    "build-android": "theo ./src/theo/theo-spec.json --transform android --format android.xml --dest ./output",
-    "build:xml": "NODE_RUN=true ts-node-esm ./scripts/generateXMLDesignTokens.ts",
-    "build": "yarn clean && tsup && yarn run generate:json && yarn run build-html && yarn run build-web && yarn run build-ios && yarn run build-android && yarn build:xml && yarn generate:flow",
-    "size:build": "yarn run build",
     "prepublishOnly": "yarn build",
-    "clean": "rimraf dist pages output"
+    "clean": "del dist output src/js/createTokens.* src/js/defaultFoundation.*"
```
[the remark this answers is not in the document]
mainframev: it has cli, so can be called without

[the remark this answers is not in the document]
mainframev: but I will change it to del-cli, since del is the command on windows :D just in case if somebody will run it on win

packages/orbit-tailwind-preset/package.json
mainframev: that will be possible only after publishing it to npm, otherwise on CI it will be broken here

packages/orbit-tailwind-preset/src/presets/foundation/index.ts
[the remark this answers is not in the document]
mainframev: I agree 👍🏻

packages/orbit-tailwind-preset/src/presets/__tests__/spacing.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  "spaceXXXSmall",
+  "spaceXXSmall",
+  "spaceXSmall",
+  "spaceSmall",
+  "spaceMedium",
+  "spaceLarge",
+  "spaceXLarge",
+  "spaceXXLarge",
+  "spaceXXXLarge",
+] as const;
```
DSil, line 17: Is this cast needed? :thinking:

packages/orbit-tailwind-preset/README.md
```diff
@@ -0,0 +1,52 @@
+# Orbit tailwind preset
+
+This package contains the Orbit Tailwind presets.
+
+Currently, there are two presets available:
+
+- `orbitPreset` - Orbit's foundation styles, including typography, colors, spacing, etc.
```
DSil, line 7: Should we mention this is the default export? Or maybe we can decide not to have default exports, so all exports are named and explicitly imported 💭

packages/orbit-tailwind-preset/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+```js
+const orbitPreset = require("@kiwicom/orbit-tailwind-preset");
+
+module.exports = {
+  presets: [orbitPreset],
+};
+```
+
+We also have preset with component tokens, which means to be used inside Orbit components. In case you need to use our component-level tokens, add the `orbitComponentsPreset` to your `tailwind.config.js` file:
```
DSil, line 35:
```suggestion
We also have a preset with component tokens, which is meant to be used inside Orbit components. In case you need to use our component-level tokens, add the `orbitComponentsPreset` to your `tailwind.config.js` file:
```

packages/orbit-tailwind-preset/README.md
```diff
@@ -0,0 +1,52 @@
...
+module.exports = {
+  presets: [
+    orbitComponentsPreset({
+      // Enable normalizing of the browser's default styles, which is disabled by default
+      disabledPreflight: false,
+    }),
+  ],
+};
+```
+
```
DSil, line 49: Wasn't there more documentation text here? I feel this is missing a connection between this and the snippet below... I feel it only tells how to import the preset but then not how to use it. It just has a snippet that seems a bit out of context and short. WDYT?

DSil (commented):
Overall looking good. Just some small comments / question mostly about documentation and maintenance, not implementation 👍

packages/orbit-tailwind-preset/src/presets/__tests__/spacing.test.tsx
mainframev: nope, removed

packages/orbit-tailwind-preset/README.md
mainframev: hm, inside usage there is an example that's showing it as default

DSil: Yes, I saw it. But it happened to me in the past 😄 when trying to use this on another repository I was trying to import this name just to later find out that I should import it as default. I think something as simple as this should be enough and clarifying
```suggestion
- `orbitPreset` **(default export)** - Orbit's foundation styles, including typography, colors, spacing, etc.
```

packages/orbit-tailwind-preset/README.md
mainframev: now :)

packages/orbit-tailwind-preset/README.md
mainframev: Im planning to do that in separate MR :) adding more docs

DSil (commented):
LGTM 👍 Good work! 🍾 
I'd say to hold the merge until the next release (With the open PRs) and I'd say commits can be reduced. WDYT?

mainframev:
> LGTM 👍 Good work! 🍾 I'd say to hold the merge until the next release (With the open PRs) and I'd say commits can be reduced. WDYT?

yes, it's planned like that, I will release your fix for Slider and type fixes today, and after we can release that. I will reduce commits

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      extend: {
+        colors: {
+          ...COLORS.reduce((acc, name) => {
+            // eslint-disable-next-line no-param-reassign
+            acc = {
+              ...acc,
+              ...getComponentLevelToken(name, "background"),
+              ...getComponentLevelToken(name, "backgroundHover"),
+              ...getComponentLevelToken(name, "backgroundActive"),
+              ...getComponentLevelToken(name, "foreground"),
```
DSil, line 52: While migrating the Heading component to Tailwind, I noticed that the foreground-inverted class was not being generated. Should it be added here? Or since only Heading (for now) has this rule we should just add it manually at the bottom of this object, as we do with other classes in other properties? 🤔

mainframev: I think for now we can use the paletteWhite color, as the token does not exists for that inside Orbit tokens package

DSil: I'm not sure I understand. The token is defined [here](https://github.com/kiwicom/orbit/blob/feat/tailwind-config/packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-foreground.json#L8). And if we have a token for it we should use it 💭

packages/orbit-tailwind-preset/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "keywords": [
+    "orbit",
+    "tailwind",
+    "preset",
+    "design-system",
+    "tokens"
+  ],
+  "scripts": {
+    "build": "yarn tsup",
+    "pretest": "yarn workspace @kiwicom/orbit-design-tokens build && yarn build && yarn tailwindcss -o style.css",
```
DSil, line 29: Unfortunately, this only works if the test command is executed directly on the package. I am used to running `yarn test` from the root folder, sometimes with the `--watch` flag. And if for some reason I don't have the `style.css` generated (or updated) it will fail because it can't access the file. I understand this only happens when one uses it for the first time, but is it ideal? As the styles.css is untracked I don't know if we should rely on it to run tests. I see that for the pipeline this is being built before executing the tests, but that does not happen locally, and one can develop and work on components without ever generating the `style.css`, that is why I would like to avoid the need to manually execute this script before the tests.
If one clones the repository for the first time, builds everything and runs tests… they fail. They won't know that this `pretest` needs to be executed. What would be the problem of having the `yarn tailwindcss -o style.css` as part of the `build` script? 🤔 

This is non-blocking, but I think the experience could be optimized here. WDYT?

mainframev: the problem is that it's been running with jest from the root folder, while it should be run with Lerna for all workspaces to execute test commands in each workspace. This problem will not happen with nx, but I can add it here with Lerna. pre commands executed automatically if a test command is run but just does not do it from the test command in the root. :)

packages/orbit-design-tokens/src/dictionary/definitions/component/form-element/form-element-font-size.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "normal": {
+          "type": "size",
+          "value": "{foundation.font-size.normal}",
+          "deprecated": true,
+          "deprecated-replace": "{component.form-element.normal.font-size}",
+          "deprecated-version": "1.0.0"
+        }
+      },
+      "form-feedback": {
```
DSil, line 34: On the context of this question, why is this declared [here](https://skypicker.slack.com/archives/C04HMMQ6VNE/p1688480433368959)? Does it make sense to deprecate a token that only existed in the context of another repo? :thinking:
What is the value we are trying to achieve with this? Ease their migration? If so, I can understand it, but I am afraid it can leave to more dangling classes through time... WDYT?

mainframev: wdym in the context of only another repo?  

what's [this](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Checkbox/index.tsx#L122) and [this](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Checkbox/index.tsx#L122) then? :) 

> What is the value we are trying to achieve with this? Ease their migration? If so, I can understand it, but I am afraid it can leave to more dangling classes through time

In general, having one repo using deprecated tokens -> yes, that is supposed to help them to migrate, instead of making breaking changes immediately, we can properly communicate that and deprecate them first and have it as a part of the documentation.

packages/orbit-design-tokens/src/__tests__/deprecatedTokens.test.ts
```diff
@@ -0,0 +1,592 @@
...
+    heightButtonLarge: theme.base.size2xl,
+    heightButtonSmall: theme.base.sizeLg,
+    heightCarrierLogo: theme.base.sizeLg,
+    widthCarrierLogo: theme.base.sizeLg,
+    heightRadioButton: "20px",
+    widthRadioButton: "20px",
+    heightCheckbox: "20px",
+    widthCheckbox: "20px",
+    heightCountryFlag: "auto",
+    widthCountryFlag: theme.base.sizeMd,
```
DSil, line 431: This is refering to sizeMd, which is 24px, but on the generated `createTokens` on the test snapshot and on the output folder it has a value of 20px 😕 can you further investigate why? This is deprecated and we won't need this token anymore anyway, but it would be interesting to understand why this is happening, because it can be happening somewhere else 💭

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-line-height.json
```diff
@@ -0,0 +1,118 @@
...
+      "display-subtitle": {
+        "line-height": {
+          "type": "size",
+          "value": 28
+        }
+      },
+      "title-1": {
+        "line-height": {
+          "type": "size",
+          "value": 36
```
DSil, line 19: This used to be 32

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-line-height.json
```diff
@@ -0,0 +1,118 @@
...
+      "title-5": {
+        "line-height": {
+          "type": "size",
+          "value": 20
+        }
+      },
+      "title-6": {
+        "line-height": {
+          "type": "size",
+          "value": 20
```
DSil, line 49: This used to be 16

DSil: I can change both on the Heading branch, but something is off here. I didn't manually check each value, but now we don't know if there are more values off

mainframev: You can check it in figma library in typography Orbit section. Fixed that, good catch

packages/orbit-tailwind-preset/src/presets/components/index.ts
mainframev: Ah, right, did not check that as I assumed we do not have it inside definition. Yeah, will make sense actually to spread in the same way (maybe in future we can use it for other components in dark mode, for instance)

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-line-height.json
DSil: Yeah, there are more inconsistent ones, according to Figma:
- title 3 should be 18px, not 16px (it is using foundation.fontSize.large but it should use XLarge)
- title 4 should be 16px, not 15px (it is using foundation.fontSize.normal but it should use Large)
- title 4 should have fontWeight 700, not 500 (it is using foundation.fontWeight.medium but it should use Bold) - this is actually wrong in the current Orbit implementation as well
- title 5 should be 15px, not 13px (it is using foundation.fontSize.small but it should use Normal)

Also we're missing the line height for xlarge for text, no? Something like `lineHeightTextLarge` with the value of 24px 💭

mainframev: Fixed that for headings. Line-height large is 24px, we do not have text extra large, or wdym?

DSil: We do have text extra large, it was added last month. What happens is that the line height is the same on both text large and extra large, and so in React we are reusing the same token (`lineHeightTextLarge`). But shouldn't we have a token for each size, in this case, with the same value?

mainframev: I guess it make sense if in future the value might change

package.json
```diff
@@ -117,9 +126,12 @@
...
     "stylelint-config-styled-components": "^0.1.1",
+    "tailwindcss": "^3.3.1",
     "ts-node": "^10.9.1",
     "tsc-files": "^1.1.3",
+    "tsup": "^6.7.0",
     "typescript": "^4.8.3",
     "zx": "^7.0.7"
-  }
+  },
+  "version": "1.0.0-beta.2"
```
DSil, line 136: Is this expected?

mainframev: it's not :D

docs/package.json
```diff
@@ -29,8 +29,9 @@
     "@jsdevtools/rehype-url-inspector": "^2.0.2",
     "@kiwicom/babel-plugin-orbit-components": "*",
     "@kiwicom/orbit-components": "*",
-    "@kiwicom/orbit-design-tokens": "*",
+    "@kiwicom/orbit-design-tokens": "5.0.0-beta.1",
     "@kiwicom/orbit-themer": "*",
+    "@kiwicom/orbit-tailwind-preset": "1.0.0-beta.0",
```
DSil, line 34: Should we merge this with the beta versions? How are we going to handle this? Same question for design tokens package

packages/orbit-design-tokens/src/dictionary/definitions/component/form-element/form-element-foreground.json
```diff
@@ -0,0 +1,141 @@
+{
+  "component": {
+    "form-element": {
+      "foreground": {
+        "type": "color",
+        "value": "{foundation.palette.cloud.dark}",
+        "comment": "Should be used for the text when they value is not filled."
```
DSil, line 7:
```suggestion
        "comment": "Should be used for the text when the value is not filled."
```

packages/orbit-design-tokens/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@kiwicom/orbit-design-tokens",
-  "version": "4.2.1",
+  "version": "5.0.0-beta.1",
```
DSil, line 3: How will lerna handle the versioning on the next release? :thinking:

DSil (commented):
It looks good to me! Just minor questions for consideration, but I don't see up ahead anything else that needs a change right now. Good work! :muscle: :+1:

docs/package.json
mainframev: for docs it's ok to use beta versions :)

packages/orbit-design-tokens/package.json
mainframev: I think I will exclude those packages from Lerna for a while since it's not generating the changelog properly, and better will be to generate those after we will get them out of beta :) It should be soon, after I will add docs, we can release the first major

DSil:
Please resolve conflicts with yarn.lock and build the tokens and commit the latest changes before merging 🙏

mainframev:
> 

no new changes except timestamp of the build, but ok :)

DSil:
And the comment that had a typo fixed 😄 that is what was causing diffs in other branches

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-font-size.json
```diff
@@ -28,19 +28,19 @@
...
           "type": "size",
-          "value": "{foundation.font-size.large}"
+          "value": "{foundation.font-size.extra-large}"
         }
       },
       "title-4": {
         "font-size": {
           "type": "size",
-          "value": "{foundation.font-size.normal}"
+          "value": "{foundation.font-size.large}"
```
DSil, line 37: I have just noticed a problem. This is being ignored because in deprecated (line 92 of this file) the value defined is different. Therefore, when we do `yarn tokens build`, the value for the tokens is actually the deprecated one. I think this is not expected, and blocks the correctness of this token definition. Can you take a look at it?

DSil (changes requested):
🚨 Problem found with the definition of deprecated tokens

DSil (approved):
Looks good to me 👍 Hoping for the best 🤞 Good work!

PR #3831, opened 2023-04-26
feat(Popover): add maxHeight prop

DSil:
maxHeight prop is added to Popover. It affects the content only. If actions are defined, they are over the maxHeight


packages/orbit-components/src/Popover/README.md
```diff
@@ -31,7 +31,8 @@ Table below contains all types of the props available in the Popover component.
...
 | overlapped     | `boolean`                | `false`             | If `true`, the content of Popover will overlap the trigger children.                                                                             |
 | renderInPortal | `boolean`                | `true`              | Optional prop, set it to `false` if you're rendering Popover inside a custom portal, defaults to `true`                                          |
-| width          | `string`                 |                     | Width of popover, if not set the with is set to `auto`.                                                                                          |
+| width          | `string`                 |                     | Width of popover, if undefined, it is set to `auto`.                                                                                             |
+| maxHeight      | `string`                 |                     | The maximum height of the content of the popover, if undefined, it is set to `auto`.                                                             |
```
mainframev, line 35:
```suggestion
| maxHeight      | `string`                 |                     | The maximum height of the content of the popover, if undefined, it is set to `100%`.                                                             |
```

mainframev (changes requested):
LGTM. Could you also add that prop to docs/__examples__? 🙏🏻 🙂

mainframev (approved):

PR #3834, opened 2023-04-26
docs: add new playground page

mainframev:
Added a new playground page. Currently for testing purposes, if all will be ok - we can try to release it for beta testing :D 

The page includes all examples by default, so users will be able to select them and edit - faster than using an actual page with current single examples, that do not have other components and icons in scope. 

Or they can just simply play with components as they want. 

I've some concerns about the bundle size increase and with Gatsby v5 it might be easier to achieve, but I used here loadable components, so it should not get to our main bundle and be in the separate chunk 🤔


DSil (commented):
Overall LGTM 👍 useful tool! Just minor tweaks needed 

`yarn.lock` is outdated. Please run `yarn` to resolve to the latest versions.

On the search modal, the results no longer include the path. It looks a bit odd like this
image

DSil (approved):

PR #3838, opened 2023-05-02
chore: add scope verification to commitlint

DSil:
The scope of a commit, when provided, must now be one of the following:
- The name of a component - directory inside `orbit-components/src/`
- The name of a hook - directory inside `orbit-components/src/hook`
- 'docs'
- 'tokens'

More valid scopes can be easily added by expanding the returned value from the function


commitlint.config.js
```diff
@@ -1,8 +1,24 @@
-// @noflow
+const fs = require("fs");
+
+function fetchComponentDirs() {
```
mainframev, line 3: instead of running only on parent folders, try to use globby, to include also children folders, for instance, if i made a change inside `ItinerarySegmentStop`, I want to have that scope allowed too :)

commitlint.config.js
```diff
@@ -1,8 +1,24 @@
...
+  const dirs = fs
+    .readdirSync("./packages/orbit-components/src/", { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
```
mainframev, line 6: currently `__typetests__` , data,  folders are also valid scopes, I think it shouldn't be

commitlint.config.js
```diff
@@ -1,8 +1,24 @@
...
+    .map(dirent => dirent.name);
+
+  const hookDirs = fs
```
mainframev, line 9: in case of using globby, you would not need that additional iteration, just to add excludes :)

mainframev (changes requested):

commitlint.config.js
DSil: Good point. I did not use `globby`, but its original form `fast-glob` instead. But the end result is achieved 🙂

mainframev: hm, it's ok, but we have globby already as a dependency from `zx` :) so that's why I suggested that, no need for additional dependency :)

DSil: We also have `fast-glob` 😄 you see that yarn.lock did not change when I added as a devDep, so I think that's pretty much the same.

Plus, I was not able to use `globby` with `require` and so, to avoid more changes to the file, I proceeded with `fast-glob`

mainframev: np :) LGTM

mainframev (approved):

PR #3841, opened 2023-05-04
feat(Popover): add renderTimeout prop

mainframev:
It was asked [here](https://skypicker.slack.com/archives/CAMS40F7B/p1683046391312509) and [here](https://skypicker.slack.com/archives/CAMS40F7B/p1677168767536869). 

In my opinion, we should not have that delay we had before, although, it looks like setTimeout here was also used for removing the flickering on the initial render. It works fine with 0, because of the nature of the event loop. So I decided to keep the hook, but reduce the timeout to 0, just to look it more visually faster. 

Added the prop to control that, if somebody still would need to have control, but I won't be useful much, just for case.

Also, do not think we should be 100% neat on that and mark it as breaking change. Can be as normal feature by mentioning that we just reduced the timeout.

### DIFF
--- a/docs/src/__examples__/Popover/DEFAULT.tsx
+++ b/docs/src/__examples__/Popover/DEFAULT.tsx
@@ -42,6 +42,7 @@ export default {
         { name: "fixed", type: "boolean", defaultValue: false },
         { name: "width", type: "text", defaultValue: "" },
         { name: "maxHeight", type: "text", defaultValue: "" },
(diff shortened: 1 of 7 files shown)


DSil (commented):
Agreed about the non-breaking change. The timeout was undocumented and there are no API breaking changes, so no need to further document it 👍 

I think this can just be slightly improved by:
- Adding the prop on the Popover Playground story
- Adding a test to the new prop (probably too tricky, so non-blocking for this)

DSil (approved):

PR #3849, opened 2023-05-18
Allow "start" and "end" align, deprecate rtl/textAlign

jozef-mikusinec-kiwicom:
The utility functions for text alignment `left` and `right` in `packages/orbit-components/src/utils/rtl/index.ts` have their native versions directly in css -- "start" and "end". I think it would be nice if we could use those directly. Since this looked simple, I decided to shoot a MR, but if I missed some nontrivial things, feel free to take this MR as just a feature request 😄

## This Pull Request meets the following criteria:

- [x] New Components has both `.flow` and `d.ts` files and are exported in `index.js.flow` and `index.d.ts`

### DIFF
--- a/docs/src/__examples__/Box/DEFAULT.tsx
+++ b/docs/src/__examples__/Box/DEFAULT.tsx
@@ -93,7 +93,7 @@ export default {
         {
           name: "textAlign",
           type: "select",
-          options: ["left", "center", "right"],
+          options: ["start", "end", "left", "center", "right"],
           defaultValue: "center",
         },
         {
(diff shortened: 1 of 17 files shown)


mainframev:
@jozef-mikusinec-kiwicom Yeah, I guess you missed a bit. It's ok to use it with text-align, but we can't deprecate it, because it's also used with the branch of other CSS properties like margins and paddings, and normal left/right positions. 

So remove the `deprecation` comment above `left/right` and can keep it above `textAlign`

jozef-mikusinec-kiwicom:
Alright, removed those two comments 👍

mainframev (approved):

PR #3854, opened 2023-05-23
fix(ItinerarySeparator): rename imported Props to avoid naming conflict

RobinCsl:
Otherwise, TypeScript would report the following error:
```
 Import declaration conflicts with local declaration of 'Props'.
```

### DIFF
--- a/packages/orbit-components/src/Itinerary/ItinerarySeparator/types.d.ts
+++ b/packages/orbit-components/src/Itinerary/ItinerarySeparator/types.d.ts
@@ -1,9 +1,9 @@
 import type * as React from "react";
 
-import type { Props } from "../../Separator/types";
+import type { Props as SeparatorProps } from "../../Separator/types";
 
 export interface Props {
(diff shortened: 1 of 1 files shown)


RobinCsl:
Could this somehow be detected earlier?

mainframev:
> Could this somehow be detected earlier?

Yes, it could be, I will check why it didn't. Thanks for contribution :)

DSil:
Having `skipLibCheck` set to `true` causes this error to be silent for us. Do we know why we have it set to true?

mainframev (approved):

mainframev (approved):

mainframev (approved):

RobinCsl:
> Having `skipLibCheck` set to `true` causes this error to be silent for us. Do we know why we have it set to true?

Not sure, probably because you would otherwise need to fix quite a few errors after toggling it off. But it's strange that `skipLibCheck: true` would affect your own library code. 🤔 

Thank you for merging!

PR #3858, opened 2023-05-24
chore: fix translations codemods to use translate functions instead of components

mainframev:
Replacing component with translations from `useIntl` (react-intl) and old `useIntl` (nitro service).


packages/orbit-components/transforms/__testfixtures__/translations-new.input.js
```diff
@@ -13,6 +13,8 @@ import Breadcrumbs, { BreadcrumbsItem } from "@kiwicom/orbit-components/lib/Brea
 /* prettier-ignore */
 
 const Component = () => {
+  const intl = useIntl();
```
DSil, line 16: Should this be on the input? 🤔 shouldn't the codemod handle this?

mainframev: I struggled to properly find a parent nodes, now should be fine 🤞🏻

DSil (approved):

PR #3860, opened 2023-05-25
docs: changed do/dont icon

majlan:
Updated guideline component to show different icons in the styled list


docs/src/components/Guidelines/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import styled, { css } from "styled-components";
-import { CheckCircle, CloseCircle } from "@kiwicom/orbit-components/icons";
+import { CheckCircle, CloseCircle, Check, Close } from "@kiwicom/orbit-components/icons";
```
DSil, line 3: Check and Close are not being used, so why import them? Is this intended?

DSil (commented):
Left a question. I will address lint issues that cause the pipeline to fail

docs/src/components/Guidelines/index.tsx
majlan: They are used in GuidelineItemIcons (new component I added)

DSil: Yes, but that is a different file, so they should only be imported there (and they are). So I will remove it from here while addressing the pipeline issues

majlan: Ok, I was probably trying to use them here as well as it might be needed in some cases but ended up not figuring it out.

DSil (approved):

PR #3865, opened 2023-05-31
fix(Select): fixed spacing between prefix and customValueText

DSil:
Reported [here](https://skypicker.slack.com/archives/CAMS40F7B/p1685526457132239).

During the inlineLabel and size refactoring of Select, this bug was introduced. The condition for prefix is no longer needed because of the way the customValueText is rendered

[ORBIT-2767](https://kiwicom.atlassian.net/browse/ORBIT-2767)

### DIFF
--- a/packages/orbit-components/src/Select/index.tsx
+++ b/packages/orbit-components/src/Select/index.tsx
@@ -244,10 +244,8 @@ SelectSuffix.defaultProps = {
   theme: defaultTheme,
 };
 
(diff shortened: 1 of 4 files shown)


mainframev:
@DSil could you also change the type of the label inside `Option` to a string only? I doubt people use it since it leads to `[Object object]`

mainframev:
LGTM, quick fix 👍🏻

mainframev (approved):

mainframev:
As from DM dicussion, let's solve then all that positioning / truncating stuff in another PR

PR #3869, opened 2023-06-06
fix(ModalFooter): wrap elements in StyledChild only when multiple

mainframev:
Getting back accidentally removed condition for checking if children element is array or not. 

[Slack request](https://skypicker.slack.com/archives/C9B1H0ACW/p1686053579542679)

### DIFF
--- a/packages/orbit-components/src/Modal/ModalFooter/__tests__/index.test.tsx
+++ b/packages/orbit-components/src/Modal/ModalFooter/__tests__/index.test.tsx
@@ -20,6 +20,16 @@ describe("ModalFooter", () => {
     });
   });
 
+  it("should not have StyledChild wrapper on single children", () => {
+    render(
+      <ModalFooter dataTest="footer">
+        <Button type="secondary">Cancel</Button>
+      </ModalFooter>,
+    );
+
+    expect(screen.findByTestId("footer-el-wrapper")).toMatchObject({});
+  });
+
   it("should render buttons with space-between", () => {
     render(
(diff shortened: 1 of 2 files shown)


DSil (commented):
LGTM but pipeline is broken

DSil:
Also, maybe we could add a test to this, so we can avoid a situation like this in the future 💭

mainframev:
> 

yep, will add then 👍🏻

packages/orbit-components/src/Modal/ModalFooter/__tests__/index.test.tsx
```diff
@@ -20,6 +20,16 @@ describe("ModalFooter", () => {
...
   });
 
+  it("should not have StyledChild wrapper on single children", () => {
+    render(
+      <ModalFooter dataTest="footer">
+        <Button type="secondary">Cancel</Button>
+      </ModalFooter>,
+    );
+
+    expect(screen.findByTestId("footer-el-wrapper")).toMatchObject({});
```
DSil, line 30: Just for consistency with other tests we have around the repo, why don't we do it like this?

```suggestion
    expect(screen.queryByTestId("footer-el-wrapper")).not.toBeInTheDocument();
```

DSil (approved):

packages/orbit-components/src/Modal/ModalFooter/__tests__/index.test.tsx
mainframev: It won't work for that case

PR #3871, opened 2023-06-07
Fix Select prefix and disabled styles issues

DSil:
This fixes what was reported [in Slack](https://skypicker.slack.com/archives/CAMS40F7B/p1685526457132239), regarding not being able to open the Select by clicking the Prefix.

This also addresses an issue with the disabled style that was spotted during the implementation of this task.

This is also waiting on a design decision about truncating the inlineLabel when it is too large. However, I feel that can be added later to unblock these bugs.

[ORBIT-2768](https://kiwicom.atlassian.net/browse/ORBIT-2768)

### DIFF
--- a/packages/orbit-components/src/Select/index.tsx
+++ b/packages/orbit-components/src/Select/index.tsx
@@ -161,12 +161,7 @@ const StyledSelect = styled(
 
     &:disabled {
       color: ${theme.orbit.colorTextInputDisabled};
-      background: ${theme.orbit.backgroundInputDisabled};
       cursor: not-allowed;
-
-      &:hover {
-        box-shadow: inset 0 0 0 1px ${theme.orbit.borderColorInput};
-      }
     }
 
     ${customValueText &&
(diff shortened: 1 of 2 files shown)


mainframev:
I renamed PR as it includes only solution to prefix bug and disabled state

created the ticket [ORBIT-2802](https://kiwicom.atlassian.net/browse/ORBIT-2802) to fix others. We will make them in different PR.

packages/orbit-components/src/Select/index.tsx
```diff
@@ -116,15 +116,18 @@ const StyledSelect = styled(
...
     appearance: none;
     background: transparent;
     cursor: pointer;
     color: ${filled ? theme.orbit.colorTextInput : theme.orbit.colorPlaceholderInput};
     font-family: ${theme.orbit.fontFamily};
     font-size: ${theme.orbit.fontSizeInputNormal};
     height: ${theme.orbit.heightInputNormal};
-    padding: ${rtlSpacing(`0 ${theme.orbit.spaceXXLarge} 0 ${theme.orbit.spaceSmall}`)};
+    padding: ${rtlSpacing(
+      (prefix && `0 ${theme.orbit.spaceXXLarge} 0 ${theme.orbit.paddingLeftSelectPrefix}`) ||
```
mainframev, line 128:
```suggestion
      (prefix ? `0 ${theme.orbit.spaceXXLarge} 0 ${theme.orbit.paddingLeftSelectPrefix}`) :
```

mainframev (changes requested):

mainframev (approved):

PR #3873, opened 2023-06-08
chore: fix and upgrade storybook

mainframev:
[Reported in Slack](https://skypicker.slack.com/archives/CAMS40F7B/p1686212928941389). Although I decided to fix that radically and just upgrade the storybook to the next major, cleaned up a bit the plugins for it, as now essentials include the ones we need. Replaced webpack with vite. 

Suppressed some of the lint issues related to addons, because we need to migrate to a new way of creating actions
commented out info parameters for stories, as the info addon seems outdated (last update 3 years ago), it's broken with that, so I just commented the info parameters, we may later get that back with docs addon


yarn.lock
DSil, line 1: This file is outdated, please update it by running `yarn` on your side

DSil (commented):
> because we need to migrate to a new way of creating actions
commented out info parameters for stories, as the info addon seems outdated (last update 3 years ago), it's broken with that, so I just commented the info parameters

Could you please clarify what broke exactly? The `parameters.info` content was being used by the `orbitDecorator`, I don't understand its connection to the addon that you mention nor the reason for removing it 😕 

Also, can we hide the "controls" tab? If it's empty I would say there is no need to display it on every story, especially as the default tab…

mainframev:
> > because we need to migrate to a new way of creating actions
> > commented out info parameters for stories, as the info addon seems outdated (last update 3 years ago), it's broken with that, so I just commented the info parameters
> 
> Could you please clarify what broke exactly? The `parameters.info` content was being used by the `orbitDecorator`, I don't understand its connection to the addon that you mention nor the reason for removing it 😕

This plugin is not compatible with the current version of Storybook, it's just not working together and the plugin is deprecated and has not get updated last 3 years, so I do not think we should continue using it as it's blocking us from upgrading the storybook.

I thought that it won't be working with info. As the plugin mentions that it has to be added in the same way. Anyway, I will get them back then and keep the plugin removed. Good catch.

DSil (commented):
Maybe I wasn't clear. When I meant to hide the controls tab I meant the tab only, not the whole options panel.
The controls tab is empty, so we could leave just the Actions and Knobs tabs, which are the ones that matter.
image

And `yarn.lock` is still producing a lot of changes when I run `yarn` locally 🤔

PR #3878, opened 2023-06-12
fix(Slider): value state update was not working correctly

DSil:
Reported [here](https://skypicker.slack.com/archives/CAMS40F7B/p1686147239545699)

The Slider component was passing callbacks to event listeners that could not access the latest value of `value`. For that, a ref should be used instead. The renderHandle function should also access the most up-to-date value of `value`, instead of the state stored one. State is still needed to force re-renders.

Tests were also updated.


mainframev:
LGTM, good job on fixing this 👍🏻

mainframev (approved):

PR #3879, opened 2023-06-13
Fix/InputFile icon label click

DSil:
When clicking the error/help icon on the label of an InputFile component, the native OS mobile input file will no longer be triggered.

Also adjusted the position of the tooltip, similarly to #3877.

`pointer-events: none` was also added to avoid "ghost clicks" on mobile. Even though the component wasn't visible, it could still be clicked on mobile, and it even caused a small "flickr". Adding `pointer-events: none`, it no longer happens. This was noticed while testing on the device itself, and not using the dev tools

[ORBIT-2809](https://kiwicom.atlassian.net/browse/ORBIT-2809)


mainframev (changes requested):
LGTM

PR #3894, opened 2023-06-20
fix(InputGroup): disabled state was not visually correct

DSil:
As reported [here](https://skypicker.slack.com/archives/CAMS40F7B/p1687177057766109), the disabled styles of the InputGroup were not being correctly applied.


packages/orbit-components/src/InputGroup/index.tsx
```diff
@@ -288,7 +287,7 @@ const InputGroup = React.forwardRef<HTMLDivElement, Props>(
           </FormLabel>
         )}
 
-        <FakeGroup label={label} error={errorReal} active={active} size={size}>
+        <FakeGroup label={label} error={errorReal} active={active} size={size} disabled={disabled}>
```
mainframev, line 290: could you prefix here the props with $? so they do not appear as attributes on FakreGroup

packages/orbit-components/src/InputGroup/index.tsx
```diff
@@ -62,12 +62,11 @@ const FakeGroup = styled.span<{
       border-radius: ${theme.orbit.borderRadiusNormal};
     `)};
 
-    &:hover {
-      box-shadow: inset 0 0 0
-        ${`${theme.orbit.borderWidthInput} ${
-          error ? theme.orbit.borderColorInputErrorHover : theme.orbit.borderColorInputHover
-        }`};
-    }
+    ${!disabled &&
```
mainframev, line 65: Instead add condition inside hover box-shadow, it won't be rendered to component if condition is false :)

mainframev (changes requested):
Just a couple small things. LGTM

PR #3899, opened 2023-06-25
chore: use nx

mainframev:
Since Nx and Lerna are working together, we can use it. Added `nx.json` config with minimal setup, which helps us to run scripts with caching. Updated CI actions and lint-staged with affected commands. 

Features branch now only will run what's affected, the master will run all the checks completely.

Storybook: https://orbit-mainframev-chore-nx.surge.sh


package.json
```diff
@@ -128,8 +134,6 @@
     "stylelint-config-styled-components": "^0.1.1",
     "tailwindcss": "^3.3.3",
     "ts-node": "^10.9.1",
-    "tsc-files": "^1.1.3",
-    "tsup": "^6.7.0",
```
DSil, line 132: I think tsup is needed, no?

mainframev: yes ;)

packages/orbit-components/package.json
```diff
@@ -12,17 +12,19 @@
...
     "build": "ts-node --esm config/build/index.mts",
+    "test": "jest",
     "prepublishOnly": "yarn build",
     "size:build": "yarn build --size",
     "size": "size-limit",
     "size:why": "size-limit --why",
     "fetch:icons": "ts-node --esm config/fetchIcons.mts",
     "check:icons": "ts-node --esm config/checkIconsCli.mts",
     "check:types": "tsc --p tsconfig.json",
+    "check:lin": "eslint . --report-unused-disable-directives",
```
DSil, line 23:
```suggestion
    "check:lint": "eslint . --report-unused-disable-directives",
```

no?

packages/orbit-design-tokens/output/tokens.less
```diff
@@ -304,32 +303,92 @@
...
-@button-bundle-basic-background: linear-gradient(to top right, #E13E3B 0%, #E87E09 100%);
-@button-bundle-basic-background-hover: linear-gradient(to top right, #BD2825 0%, #D67000 100%);
-@button-bundle-basic-background-active: linear-gradient(to top right, #9F1816 0%, #C36802 100%);
-@button-bundle-medium-background: linear-gradient(to top right, #3719AB 0%, #8539DB 100%);
-@button-bundle-medium-background-hover: linear-gradient(to top right, #BD2825 0%, #D67000 100%);
-@button-bundle-medium-background-active: linear-gradient(to top right, #250F79 0%, #5A3485 100%);
-@button-bundle-top-background: linear-gradient(to top right, #2D2D2E 0%, #696E73 100%);
-@button-bundle-top-background-hover: linear-gradient(to top right, #171718 0%, #51575C 100%);
-@button-bundle-top-background-active: linear-gradient(to top right, #101011 0%, #51575C);
+@button-link-secondary-background-hover: rgb(
```
DSil, line 358: These changes should not happen (the whole of them on this file, actually). This happened to me before. It is cause by saving the file manually during a rebase / conflict resolution. On save, VSCode will format this file, but we actually don't want it. So please build the tokens again and prevent all these changes to be included in this PR

DSil (commented):
It seems to be waiting for a pipeline that is never run, so we also need some adjustments there, no?

mainframev:
> It seems to be waiting for a pipeline that is never run, so we also need some adjustments there, no?

what kind of adjustments you mean? I guess this is happening because I remove the action and created a new one, and I do not think it will happen in new MR

mainframev:
closing it as it's not currently in our prio and quite old. Can be used later as inspiration.

PR #3900, opened 2023-06-26
Fix/select label

DSil:
On Select component, a long inline label now truncates automatically. The focus styles were also missing and were fixed.


mainframev (approved):
LGTM 👍🏻

PR #3903, opened 2023-06-27
fix(Slider): range Slider was not updating correctly

DSil:
defaultValue comparison was not being made correctly. Therefore, it was triggering a state update all the time, overcoming the actual value update from the user's interaction.

**Update:** During inspection with Search, another bug was noticed, that was actually present even before this branch was created. So it was impossible to slide over the currently selected value. This was fixed by removing the verifications of new values, which, from my understanding, were legacy to the older implementation without hooks. But review is appreciated 🙏 

Reported [here](https://skypicker.slack.com/archives/CAMS40F7B/p1687859299240369).


mainframev:
LGTM. Just make sure it works 🤞🏻

mainframev (approved):

DSil:
This will be tested in a pair session with Search before it is merged

PR #3909, opened 2023-06-30
Remove-ink-lighter-references

DSil:
Reported [here](https://skypicker.slack.com/archives/CAMS40F7B/p1688038702345299).
The inkLighter token was removed some months ago but some references were kept lost. This removes those references and also adds some missing inkDark references.
It also removes inkLighter and adds inkDark to Box accepted types for color.


packages/orbit-components/src/Box/types.d.ts
```diff
@@ -35,15 +35,15 @@ export type ColorTokens =
...
-  | "inkLighter"
-  | "inkLighterHover"
-  | "inkLighterActive"
   | "inkLight"
   | "inkLightHover"
   | "inkLightActive"
   | "inkNormal"
   | "inkNormalHover"
   | "inkNormalActive"
+  | "inkDark"
```
mainframev, line 44: could you reuse the logic that I made recently for Separator? Have a typescript.template file with the type, that will allow us to have aligned with tokens and no need to update manually :)

mainframev (changes requested):

packages/orbit-components/src/Box/types.d.ts
DSil: Had to do some "gymnastics" to avoid the breaking change, but I think it's ok 😅 🤸

mainframev (approved):

PR #3910, opened 2023-06-30
chore: setup orbit-components to TW

DSil:
This prepares orbit-components for tailwind


package.json
```diff
@@ -66,16 +66,23 @@
...
     "@types/lodash": "^4.14.194",
     "@types/prettier": "^2.7.2",
     "@types/react": "^17.0.26",
     "@types/react-dom": "^17.0.9",
+    "@types/testing-library__jest-dom": "^5.14.5",
     "@typescript-eslint/eslint-plugin": "^5.57.1",
     "@typescript-eslint/parser": "^5.57.1",
     "babel-jest": "^28.1.3",
     "browserslist": "^4.16.6",
+    "classnames": "^2.3.2",
```
mainframev, line 85: insted of classnames please use `clsx`. And remove it from root, as it's planned to be used only inside components package.

packages/orbit-components/package.json
```diff
@@ -12,6 +12,7 @@
     "cy:run": "start-server-and-test cy:dev http-get://localhost:3000 'cypress run'",
     "cy:open": "start-server-and-test cy:dev http-get://localhost:3000 'cypress open'",
     "build": "ts-node --esm config/build/index.mts",
+    "build:tailwind": "postcss ./src/tailwind.css -o ./lib/tailwind.css",
```
mainframev, line 15: why with postcss? `yarn tailwindcss -o` is not the command we need?

mainframev (changes requested):

packages/orbit-components/package.json
mainframev, line 15: why you decided to use postcss instead of default tailwindcss cli?

package.json
DSil: Changed to `clsx` but if I remove from toor devDeps, I keep getting the [eslint import/no-extraneous-dependencies error](https://github.com/import-js/eslint-plugin-import/blob/v2.27.5/docs/rules/no-extraneous-dependencies.md). Doesn't it make sense to keep as devDep on root?

mainframev: I think it should be moved to its package where it will be used as a normal dependency, not a dev, as it will be executed in runtime

DSil: Right, but I have it on orbit-components as dependency and it still complains 😞

packages/orbit-components/package.json
```diff
@@ -11,7 +11,8 @@
     "cy:dev": "vite cypress/integration --logLevel silent",
     "cy:run": "start-server-and-test cy:dev http-get://localhost:3000 'cypress run'",
     "cy:open": "start-server-and-test cy:dev http-get://localhost:3000 'cypress open'",
-    "build": "ts-node --esm config/build/index.mts",
+    "build": "yarn tailwind && ts-node --esm config/build/index.mts",
```
mainframev: you do need to modify that command here, as you already added that as a step to build inside index.mts :)

mainframev (changes requested):

packages/orbit-components/package.json
DSil: As agreed, it will change to use tailwind cli but storybook will keep postcss

packages/orbit-components/package.json
DSil: Yes, it was outdated as I did not add that change 👍 well spotted though

mainframev:
also the stylelint config should be adjusted to tailwind :)

DSil:
I have added [stylelint-config-tailwindcss](https://github.com/zhilidali/stylelint-config-tailwindcss) but I don't know if you're expecting anything else 🤔

package.json
```diff
@@ -20,12 +20,12 @@
...
     "loki:approve": "yarn components loki approve",
     "loki:test": "yarn components loki test",
-    "orbit-kiwi:build": "yarn && yarn tokens build && yarn components build && yarn themer build && yarn docs build",
+    "orbit-kiwi:build": "yarn && yarn tokens build && yarn tailwind-preset build && yarn components build && yarn themer build && yarn docs build",
     "postinstall": "lerna run postinstall --stream",
     "prepare": "husky install",
     "prettier:test": "prettier --check '**/*.md'",
     "publish-packages": "zx scripts/publish.mjs",
-    "size:build": "yarn tokens build && yarn components size:build",
+    "size:build": "yarn tokens build && yarn tailwind-preset build && yarn components size:build",
```
mainframev, line 28: I do not think we need to build a tailwind package for the size

DSil: We do. Otherwise components can't be built

mainframev:
> I have added [stylelint-config-tailwindcss](https://github.com/zhilidali/stylelint-config-tailwindcss) but I don't know if you're expecting anything else 🤔

LGTM, I think for the beginning should be enough

mainframev:
@DSil could you also please add an[ official prettier plugin?](https://www.npmjs.com/package/prettier-plugin-tailwindcss) 
and maybe we can also consider https://www.npmjs.com/package/eslint-plugin-tailwindcss, but it's not official

DSil:
I like the idea of having a prettier for that, but it makes my suggestion [here](https://github.com/kiwicom/orbit/pull/3914#discussion_r1269372585) of having context-specific classes first impossible to execute 😞 I will still add it, but just so we know one of the possible disadvantages of having it

mainframev (approved):

mainframev:
> I like the idea of having a prettier for that, but it makes my suggestion [here](https://github.com/kiwicom/orbit/pull/3914#discussion_r1269372585) of having context-specific classes first impossible to execute 😞 I will still add it, but just so we know one of the possible disadvantages of having it.
> 
> Regarding the ESLint plugin, a quick test showed me that their `classnames-order` rule is different from prettier 🤯, even though they say they follow it 🤷‍♂️ (but people [seem to disagree](https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/231)) so I have disabled that rule on ESLint, as we can rely on Prettier and auto-format on save for that, no?

eslint-plugin is not official, while prettier is, so if it's causing problems, we can just stop using it :)

packages/orbit-design-tokens/src/index.ts
```diff
@@ -5,6 +5,7 @@ export const defaultTheme = createTheme();
 
 export { default as fromPlainObject } from "./fromPlainObject";
 export { default as transparentColor } from "./js/transparentColor";
+export type { Tokens as paletteColors } from "./js/paletteColors";
```
mainframev, line 8: remove that from index.ts as export, that's why you are getting that error on flow ci :)

DSil: I can't remove. I need it, that is precisely why I created it…

mainframev: I did not mean to remove completely, just to remove from index file as export. Already did it for you.

PR #3914, opened 2023-07-05
Migrate country flag tailwind

DSil:
CountryFlag component migrated to Tailwind. Needed tokens and Tailwind configs were also added.
 Storybook: https://orbit-mainframev-migrate-CountryFlag-tailwind.surge.sh


packages/orbit-components/src/CountryFlag/__tests__/index.test.tsx
```diff
@@ -42,4 +43,36 @@ describe("CountryFlag", () => {
     const flag = screen.getByRole("img");
     expect(flag).toHaveAttribute("src", expect.stringContaining("us"));
   });
+  it("should have the correct styles", () => {
```
mainframev, line 46: it looks a bit dirty too me :) 

better will be avoid imperative style with let and cleanup. I suggest to split on two different tests, something like:
- it should have expected DOM output / styles 
- it should have correct size

packages/orbit-components/src/CountryFlag/index.tsx
```diff
@@ -1,77 +1,52 @@
...
+    data-test={dataTest}
+    src={src}
+    srcSet={srcSet}
+  />
+);
 
-StyledShadow.defaultProps = {
-  theme: defaultTheme,
-};
+const Shadow = () => (
```
mainframev, line 47: I think no need to have a component for that, you can directly use it as div with className inside CountryFlag :)

packages/orbit-components/src/CountryFlag/index.tsx
```diff
@@ -86,19 +61,16 @@ function getCountryProps(code?: string, name?: string) {
...
-        title={name}
-        code={code}
-        id={id}
-        data-test={dataTest}
-        size={size}
-      />
-      <StyledShadow />
-    </StyledCountryFlag>
+    <Wrapper size={size}>
+      <Image code={code} name={name} id={id} dataTest={dataTest} src={src} srcSet={srcSet} />
```
mainframev, line 71: the same for Image, seems like unnecessary abstraction :)

mainframev (changes requested):

packages/orbit-components/src/CountryFlag/index.tsx
DSil: Yeah I gave it a thought about that while implementing for the first time. I decided to keep the abstraction to improve readability. Having a div with a bunch of classes to apply a shadow makes it a bit hard to read / understand what that is. For the image I can agree and I am ok with removing the abstraction. But with the shadow… idk. I can accept either decision, but please consider the argument of readability and let me know what you think 🙂

mainframev: I do not think it's making it worse to read :) 

It would, in the case of the bigger component, but for that case, it looks ok to me to write directly there :)

mainframev: I will leave decision up to you, I do not want to insist on something like that :)

DSil: Yes, going through it, I think we can remove all abstraction. I will move the `shadow-country-flag` class to the start of the component, so it is easier to spot and understand that that div refers to a shadow 🙂

mainframev (approved):

mvidalgarcia (approved):

PR #3925, opened 2023-07-18
fix(Separator): fix background and border issue

mainframev:
[Reported in Slack](https://skypicker.slack.com/archives/CAMS40F7B/p1689588311958009)

### DIFF
--- a/packages/orbit-components/src/Separator/TYPESCRIPT_TEMPLATE.template
+++ b/packages/orbit-components/src/Separator/TYPESCRIPT_TEMPLATE.template
@@ -7,9 +7,19 @@ export type Align = "left" | "center" | "right";
 
 export type PaletteTokens =%TOKENS%
 
-export interface Props extends Common.SpaceAfter {
+export interface CommonProps extends Common.SpaceAfter {
   indent?: Indent;
-  type?: "solid" | "dashed" | "dotted" | "double" | "none";
-  color?: PaletteTokens;
   align?: Align;
 }
+
+type ConditionalSeparatorType =
+  | {
+      readonly color: PaletteTokens;
+      readonly type: "solid" | "dashed" | "dotted" | "double" | "none";
+    }
+  | {
+      readonly color?: false;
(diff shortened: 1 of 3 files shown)


packages/orbit-components/src/Separator/types.d.ts
```diff
@@ -83,9 +83,19 @@ export type PaletteTokens =
...
+export interface CommonProps extends Common.SpaceAfter {
   indent?: Indent;
-  type?: "solid" | "dashed" | "dotted" | "double" | "none";
-  color?: PaletteTokens;
   align?: Align;
 }
+
+type ConditionalSeparatorType =
+  | {
+      readonly color: PaletteTokens;
```
DSil, line 93: I think we want this
```suggestion
      readonly color?: PaletteTokens;
```

In the past, we were able to have a type without defining a color. If we don't make the `color` optional here, there will be an error requiring `color` to be defined.

DSil (changes requested):

packages/orbit-components/src/Separator/TYPESCRIPT_TEMPLATE.template
```diff
@@ -7,9 +7,19 @@ export type Align = "left" | "center" | "right";
...
+export interface CommonProps extends Common.SpaceAfter {
   indent?: Indent;
-  type?: "solid" | "dashed" | "dotted" | "double" | "none";
-  color?: PaletteTokens;
   align?: Align;
 }
+
+type ConditionalSeparatorType =
+  | {
+      readonly color: PaletteTokens;
```
DSil, line 17: I didn't know you'd accept the suggestion directly here 😅 so this one needs a change as well
```suggestion
      readonly color?: PaletteTokens;
```

mainframev: RIght :) 👍🏻

DSil (approved):
Squash commits and 🚀 👍

PR #3928, opened 2023-07-20
InputSelect improvements

DSil:
This PR addresses all the bugs reported on the document.

Some questions are still undecided, but review can start already.

[FEPLT-1607](https://kiwicom.atlassian.net/browse/FEPLT-1607)


packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -424,14 +447,14 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
                     <Box>
                       <Heading type="title2">{label}</Heading>
                     </Box>
-                    <ModalCloseButton onClick={handleClose} title="Close" />
+                    <ModalCloseButton onClick={_ev => handleClose()} title="Close" />
```
mainframev, line 450: unfortunately title here has to be prop, in order to provide translations

mainframev (changes requested):

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -424,14 +447,14 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
...
                   </Stack>
                 )}
                 {input}
               </StyledModalHeader>
               <StyledModalSection>{dropdown}</StyledModalSection>
               <ModalFooter flex="100%">
-                <Button type="secondary" fullWidth onClick={handleClose}>
+                <Button type="secondary" fullWidth onClick={_ev => handleClose()}>
```
mainframev, line 457: can you make a handleClose as a carried function that receives event? So you do not need to call it like that on multiple places

mainframev:
also, I guess you have to correct description, it leads to a wrong ticket inside jira, no?)

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -107,6 +100,25 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
...
+          setResults(groupedOptions);
+          setActiveIdx(0);
+        }
+      }
+
+      if (onClose && isOpened) onClose(selection || selectedOption);
+      setIsOpened(false);
+    };
+
+    const handleOutsideClick = () => {
```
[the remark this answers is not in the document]
DSil: Sometimes we need to call `handleClose` with a parameter of type `Option | null`. If we pass that function to `useClickOutside`, it will be type incompatible, because `useClickOutside` expects a function that receives an event. With this, and the reason mentioned below, I think this should be the reasonable approach. WDYT?

packages/orbit-components/src/InputSelect/index.tsx
DSil: For the same reason mentioned before, I can't. However, I was able to merge the function handled on outsideClick and the ones called onClick into one, because they both expect events as arguments (even though they do nothing with them) and they should both call `handleClose` without any parameter.

packages/orbit-components/src/InputSelect/index.tsx
DSil: Right, `labelClose` can be used

packages/orbit-components/src/InputSelect/index.tsx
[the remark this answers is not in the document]
mainframev: wdym by type incompatiable? adding it like that should be ok, no? :) 

*useClickOutside(labelRef, () => handleClose());*

mainframev:
⚠️ Clicking on enter does not fire onClose, as with clicking by mouse, could you also add? 🙏🏻

packages/orbit-components/src/InputSelect/index.tsx
mainframev: I meant it something like that: 

*const handleClose = (selection?: Option | null) => (_ev?: unkown) =>* 

then you can use it just like: 

*onClose={handleClose()}*

or *handleClose(option)*

packages/orbit-components/src/InputSelect/index.tsx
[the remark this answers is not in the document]
DSil: Yes, but then on all the onClick I would have to leave the same, WDYT? We win in one side to lose on the other 😅

DSil:
> ⚠️ Clicking on enter does not fire onClose, as with clicking by mouse, could you also add? 🙏🏻

In any particular scenario? Because when I select an option with Enter it triggers onClose 🤔

packages/orbit-components/src/InputSelect/index.tsx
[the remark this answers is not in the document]
mainframev: not sure if I understand wdym by leave the same. This function is used only once inside useClickOutside, no need for that, simply call it as I showed :)

mainframev:
> > ⚠️ Clicking on enter does not fire onClose, as with clicking by mouse, could you also add? 🙏🏻
> 
> In any particular scenario? Because when I select an option with Enter it triggers onClose 🤔

already deleted :) You replied too fast :D It was caused by my experiments and I forgot to stash

packages/orbit-components/src/InputSelect/index.tsx
mainframev: tried that with carried function, works ok, but I am ok with your approach also, so resolving

packages/orbit-components/src/InputSelect/index.tsx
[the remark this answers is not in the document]
DSil: Yes, sorry, I wasn't too clear. Besides `useOutsideClick` On the Modal we have an `onClose` prop, and we also have a Button for closing and a ModalCloseButton, both with `onClick`. They all need to do the same as `useOutsideClick`: call `handleClose` with no args.

As it is right now, I am defining `handleCloseClick` once that does exactly that, and I use that definition on all 4 scenarios. But if I do as you suggest, I will have to define `() => handleClose()` four different times, unnecessarily, because I could just reuse it.

mainframev (approved):
LGTM 👍🏻 Good job 🚀

PR #3930, opened 2023-07-25
Migrate Box tailwind

DSil:
Box component migrated to Tailwind.

Bundle size increased by a lot because it features a lot of common classes that can then be reused by other components.

~Tests are relying on styles (recommended) but also on classes (unrecommended) for the media queries, until a better solution is found.~


packages/orbit-components/src/common/tailwind/color.ts
```diff
@@ -0,0 +1,470 @@
...
+    redNormalHover: "mm:text-red-normal-hover",
+    redNormalActive: "mm:text-red-normal-active",
+    redDark: "mm:text-red-dark",
+    redDarkHover: "mm:text-red-dark-hover",
+    redDarkActive: "mm:text-red-dark-active",
+    redDarker: "mm:text-red-darker",
+    socialFacebook: "mm:text-social-facebook",
+    socialFacebookHover: "mm:text-social-facebook-hover",
+    socialFacebookActive: "mm:text-social-facebook-active",
+    white: "mm:text-white-normal",
```
mainframev, line 465: this should be removed, I guess. we have only white-normal tokens, it should be as the same tokens, we do not have red, only redNormal and etc.

packages/orbit-design-tokens/src/dictionary/utils/is.ts
```diff
@@ -5,6 +5,9 @@ const isTypeOf = ({ type, category }: DesignToken, typeName: string): boolean =>
 
 export const isColor = (prop: DesignToken): boolean => isTypeOf(prop, "color");
 
+export const isColorPalette = (prop: DesignToken): boolean =>
+  isColor(prop) && prop.attributes.name.includes("palette");
```
mainframev, line 9: use here startsWith, it's safer

mainframev (changes requested):

mainframev:
+11kb to bundle size, which is not good. Potentially can be the same, when we will reuse a common folder, but still not good. I think we will have to maintain it like that until we will rethink our API towards TW after the migration

packages/orbit-components/src/common/tailwind/color.ts
DSil: We do have `white`. Maybe we should remove it, but it is being generated on the types file, so we have it.
It is defined [here](https://github.com/kiwicom/orbit/blob/setup-orbit-components-tailwind/packages/orbit-design-tokens/src/dictionary/definitions/global/palette/white.json#L4). Should it be removed? If so, on a different PR, no?

mainframev: Yes, you can remove it another PR :)

packages/orbit-components/src/Box/index.tsx
```diff
@@ -46,23 +45,319 @@ const Box = React.forwardRef<HTMLDivElement, Props>(
...
+            typeof vpMargin === "object" &&
+            vpMargin.left &&
+            marginLeftClasses[viewport][vpMargin.left],
+        ];
+      });
+    }, [mediumMobile, largeMobile, tablet, desktop, largeDesktop]);
+
+    const vars = {
+      "--box-shrink": shrink,
+      "--ld-box-shrink": largeDesktop?.shrink,
```
mainframev, line 163: I thought we decided to use variables when it's hard to actually have a large object of classNames for such properties as max-height, width and etc. In case like shrink, grow -  I do not see a problem to have it with classNames, it later can be reused for Stack too. :)

packages/orbit-components/src/Box/tailwindClasses.ts
```diff
@@ -0,0 +1,224 @@
+import { QUERIES } from "../utils/mediaQuery/consts";
+
+enum WRAP {
+  WRAP = "wrap",
+  NO_WRAP = "nowrap",
+}
+
+export const wrapClasses: {
```
mainframev, line 8: this should be also part of the common folder, I guess

DSil: I have considered that. But we actually have no other space where we use the `wrap` prop as string.

We only have a `wrap` prop in one more component: Stack. And it uses it as a boolean, so the logic there won't be accessing an object, but probably using a ternary or something. So I didn't find any value in having this on the common.

I am fine in moving it there, ofc, as I also think it is something to be reusable, but just saying that, for now, it won't be used anywhere else.

packages/orbit-components/src/Box/index.tsx
mainframev: yes, so for now, I think we can use it as separate classes because of differences in components API. As you say, it's uncommon to use values greater than 1. In the tracking, nobody uses values > 1 (18 occurrences in total I found). 

What if we will unify that in a common folder to be a boolean and in Box it will be conditionally applied, like: 

```
typeof shrink !== "undefined" && shrinkClasses[Boolean(shrink)]
```

packages/orbit-components/src/Box/index.tsx
```diff
@@ -46,23 +45,319 @@ const Box = React.forwardRef<HTMLDivElement, Props>(
...
+      left,
+      color,
+      background,
+      padding,
+      margin,
     },
     ref,
   ) => {
-    const viewports = { smallMobile, mediumMobile, largeMobile, tablet, desktop, largeDesktop };
+    const viewportClasses = React.useMemo(() => {
```
mainframev, line 77: Don't you consider getting functions that we have in the initial implementation on search useful? It looks like it would make the component code clean. It's 233 lines of code compared to 363 here 🤔

packages/orbit-components/src/Box/index.tsx
DSil: That is still a breaking change on whoever is using it with a different value. It has no impact on our teams as far as our tracking tool can detect. We can accept it and consider it to be non-impactful on our primary users, but it should still be documented as a BC if we do it like that.

Or we can leave it as is and revisit when we plan to unify the Box and Stack APIs

mainframev: No problem. Document it as breaking change then ;)

packages/orbit-components/src/Box/__tests__/index.test.tsx
```diff
@@ -74,33 +102,55 @@ describe("#Box", () => {
       convertHexToRgba(hex, NaN).replace(", NaN", "").replace("rgba", "rgb");
 
     expect(el).toBeInTheDocument();
-    expect(el).toHaveStyle({ color: getOmittedHex(theme.orbit.paletteBlueDark) });
-    expect(el).toHaveStyle({ background: getOmittedHex(theme.orbit.paletteCloudLight) });
+    expect(el.tagName).toBe("DIV");
+    expect(el).toHaveStyle({ "box-sizing": "border-box" });
+    expect(el).toHaveStyle({ "font-family": theme.orbit.fontFamily });
+    expect(el).toHaveStyle({
+      color: expect.stringContaining(getOmittedHex(theme.orbit.paletteBlueDark)),
```
DSil, line 109: I have just noticed that this test always passes, regardless of the token provided as argument. So will have to figure how to test this.

packages/orbit-components/src/Box/index.tsx
mainframev: >  On Search it works because it never happens on their project.

That sounds strange :) How their setup is different? 

> Calling functions does not work if things change in runtime.

Depends on how you use them. Functions that are defined inside Box and Stack there work and will work for us too.

mainframev: > But calling a function in runtime to get a TW class from some prop value is the same as just building the class in runtime (with template strings, for example)

that's not the same, functions that return static objects like those defined in those components on Serch should work 🤔

packages/orbit-components/src/Box/__tests__/index.test.tsx
mainframev: It has to be something with the toHaveStyle matcher. What you can try -> is to manually add logs inside the node_modules folder to that matcher and see how it behaves. I tried a bit and it seems that it simply can't resolve the color 🤔 So either the matcher or our custom render function. I will try to look more at that

mainframev (changes requested):

packages/orbit-components/src/Box/index.tsx
DSil: Their setup is not different. Their usage of the Box component is just in a way that props never change in runtime.

The functions do not return a static object. They return a string value based on the prop value. If that prop changes in runtime, the string will be different. And if the string contains a class that did not exist on build time, it won't apply the correct style because Tailwind did not consider the existence of such class.

packages/orbit-components/src/Box/index.tsx
mainframev: Right, I meant the string value true/false or we can use numbers like 0 and 1. But anyway, I hope you got what I meant 🙂 

My point was to not limit the feature because nobody uses greater numbers and perhaps won't, so it's not limiting imo, but to unify that logic on our side. I will leave the solution up to you then, it's not critical and I won't insist on that 🙂

packages/orbit-components/src/Box/__tests__/index.test.tsx
DSil: I have removed the test until we find a way to assert it

packages/orbit-components/src/Box/index.tsx
mainframev: Okay, so this wont work then? 

```
const [isRed, setIsRed] = React.useState(false);

return (
  <>
  <Button onClick={() => setIsRed(prev => !prev)} />
  <Box  color={isRed ? "redNormal" : "blueNormal" />
  </>
)

// inside Box 
const getTailwindClass = (color: string, viewport: string) => {
  return viewport ? colorClasses[viewport][color] : colorClasses[color];
};

...
 className={cx(
   className,
   "orbit-box font-base box-border",
   color && getColorClasses(color),
   ..viewports.map(query => {
    if (!viewportSpecs[query]) return null;
    return getColorClasses(viewportSpecs[query]["color"], query);
)}
....
```

DSil: If the classes mapped by `redNormal` and `blueNormal` do not exist anywhere else on the code, yes, it won't. Because tailwind won't generate the appropriate classes for them

mainframev: I thought (can't find that thread) that we already agreed that it's working when the function returns a complete string based on static objects (as I mentioned above). 

We knew that we can't construct strings dynamically via template literals, but when the ternary condition resolves to not dynamic string it's ok, the result of the function should be accessible at build time

You can also take a look at [example here](https://flaviocopes.com/tailwindcss-cannot-generate-classes-dynamically/)

@oreqizer can you take a look at that here? 🙏🏻

oreqizer: https://tailwindcss.com/docs/content-configuration#dynamic-class-names


as long as classnames are unbroken strings, it **does not matter** where they're placed. whether in functions, objects, or just hovering in spaces. they just cannot be _constructed_ dynamically by slapping together multiple strings

DSil: > whether in functions, objects, or just hovering in spaces

I don't see a reference to functions on the link you have sent. And from my experiments, it does not work with the Box implementation we have on Search

mainframev: Can you maybe share the link to the branch with your experiments? I think it's quite important to make sure about that since it will affect further migration and how we write them

DSil: I stand corrected. As I was double-checking my example on the Search repo I noticed that the "problem" was actually on their config. They are missing tokens for classes that are defined on the Box implementation (again, it's not a problem because nobody is using them, that is why everything works fine there). But that is the ultimate reason why classes were not being generated.

Alright, I will refactor to function calls on our side 👍

mainframev:
@DSil 

LGTM, good work 💪🏻 

documentation looks broken a bit (where the Box component is used), and the Box component page itself is broken, we forgot to include tailwind inside docs during the setup MR. We can add it here or separately in the branch (I created the *feat/first-tailwind-migration* branch).

mainframev (approved):

oreqizer (approved):
I have weak context outside of the search migration, but looks good to me overall 👍

mainframev (approved):

mainframev (approved):

PR #3932, opened 2023-07-26
Fix country flag width token

DSil:
The token `widthCountryFlag` changed from 24px to 20px during the Styled Dictionary / Tailwind migration, unexpectedly.
The token is being used to get the src from images.kiwi and the test is only asserting that the src URL contains the correct country code, it does not assert the width, that is also part of the URL. So it silently broke.

This PR improves the test and fixes the token value.

[FEPLT-1627](https://kiwicom.atlassian.net/browse/FEPLT-1627)


mainframev:
Good catch, thanks! 🚀

PR #3934, opened 2023-07-27
fix(InputSelect): clearing the input now clears the selected option

DSil:
This feature was missing from the previous batch of fixes/features introduced.

InputSelect now clears any selected option if the text field is cleared.

### DIFF
--- a/packages/orbit-components/src/InputSelect/index.tsx
+++ b/packages/orbit-components/src/InputSelect/index.tsx
@@ -103,14 +103,17 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
 
     const handleClose = (selection?: Option | null) => {
       if (!selection) {
-        if (inputValue !== selectedOption?.title) {
+        if (inputValue === "") {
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -103,14 +103,17 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
 
     const handleClose = (selection?: Option | null) => {
       if (!selection) {
-        if (inputValue !== selectedOption?.title) {
+        if (inputValue === "") {
```
mainframev, line 106: I guess you can just check if it's falsy value, empty string is always falsy

mainframev (approved):

packages/orbit-components/src/InputSelect/index.tsx
DSil: I think I can, but is it any better? 🤔 I think I prefer to have it explicitly compared to the empty string. It seems more readable to me to understand that I want that to happen if the value of the field is empty? 💭

PR #3936, opened 2023-07-28
Remove white token

DSil:
Only `whiteNormal`, `whiteNormalActive` and `whiteNormalHover` should exist, as similar to other palette colors.


mainframev:
@DSil I just realized, that even though it's making sense in terms of having it the same as other palette colors, it's kinda huge breaking change. Could you please take an advantage of not breaking current API, deprecated the token inside the tokens package and keep changes inside our internal code 🙏🏻

DSil:
Closing until further decision and sync regarding naming of the token

PR #3938, opened 2023-07-28
chore(tokens): restore fetchBaseTokens script

mainframev:
I accidentally removed the script, when were removing the XML script 😭 So restoring that again


mainframev:
white color tokens are broken because of the figma library bug, [reported here](https://skypicker.slack.com/archives/GSGN9BN6Q/p1690559654676469)

DSil (approved):

PR #3940, opened 2023-08-01
fix(tokens): fix bundle tokens definitons

mainframev:
[Slack request.](https://skypicker.slack.com/archives/C7T7QG7M5/p1690814179553659) Some typos during the definition of button deprecated tokens.


DSil (commented):
LGTM. Pipeline is failing because it is missing the class definition for the new palette colors on `backgroundColor` and `color` under `common/tailwind` files. This ensures we always support all colors for the palette tokens we have defined 🙂

mainframev:
> LGTM. Pipeline is failing because it is missing the class definition for the new palette colors on `backgroundColor` and `color` under `common/tailwind` files. This ensures we always support all colors for the palette tokens we have defined 🙂
> 
> Building the components package will also update the types for Box and Separator props, as they accept all palette colors

Yeah, already fixing this 💪🏻

DSil (commented):
Shouldn't Box and Separator types be updated on this PR as well?

I also feel the second commit message should not be a fix, as I don't think it's something worth mentioning on the changelog (the first commit already covers the fix, imo)

mainframev:
> Shouldn't Box and Separator types be updated on this PR as well?
> 
> I also feel the second commit message should not be a fix, as I don't think it's something worth mentioning on the changelog (the first commit already covers the fix, imo)

I will still squash it

PR #3942, opened 2023-08-02
DEV: First Tailwind Migration

mainframev:
The dev branch to merge Tailwind refactored components, which later will be merged to master
 Storybook: https://orbit-mainframev-feat-fist-tailwind-migration.surge.sh


packages/orbit-components/src/common/tailwind/shrink.ts
DSil, line 1: This was introduced by accident on the Box PR but I see this is being used on Stack, so we can keep it here

docs/tailwind.config.ts
```diff
@@ -0,0 +1,11 @@
+import { orbitComponentsPreset } from "@kiwicom/orbit-tailwind-preset";
+import type { Config } from "tailwindcss";
+
+export default {
+  content: ["./src/**/*.{js,jsx,ts,tsx}"],
```
DSil, line 5: This is actually failing to inspect Orbit components using on those files. It is only checking for tailwind classes on the src documentation files. But it needs to go deeper and check the Orbit components.
So this should be
```suggestion
  content: [
    "./src/**/*.{js,jsx,ts,tsx}",
    path.join(path.dirname(require.resolve("@kiwicom/orbit-components")), "**/*.js"),
  ],
```

PR #3944, opened 2023-08-03
Migrate Text and TextLink to Tailwind

DSil:
Text and TextLink components migrated to Tailwind.

StyledText is deprecated, until it is no longer used in Itinerary.

StyledTextLink was not needed, so it was removed entirely. The `getLinkStyle` function is deprecated until Alert component is migrated to TW.
 Storybook: https://orbit-mainframev-migrate-Text-and-TextLink-tailwind.surge.sh


packages/orbit-components/src/Text/helpers/getMargin.ts
```diff
@@ -0,0 +1,34 @@
+import type { Props } from "../types";
+
+const getMargin = (margin: Props["margin"]): { vars: object; classes: string[] } => {
+  if (typeof margin === "string")
```
mainframev, line 4: Instead of actually conditionally returning it, you can create the helper function that will do it for you and just call it here inside :)

packages/orbit-components/src/Text/helpers/getMargin.ts
```diff
@@ -0,0 +1,34 @@
...
+    return { vars: { "--text-margin": margin }, classes: ["m-[var(--text-margin)]"] };
+
+  if (typeof margin === "number")
+    return { vars: { "--text-margin": `${margin}px` }, classes: ["m-var(--text-margin)]"] };
+
+  const { top, bottom, left, right } = margin || {};
+  const cssVar = {};
+  const marginClasses: string[] = ["m-0"];
+
+  if (top) {
```
mainframev, line 14: could it be something like? 🤔 : 

```
const result = [top, bottom, left, right].filter(Boolean).reduce(pos => {
 ...
}, cssVar: {}, classes: []})
```

packages/orbit-components/src/Text/__tests__/index.test.tsx
```diff
@@ -1,59 +1,146 @@
...
+    expect(textElement).toHaveAttribute("id", id);
+    expect(textElement).toBeInTheDocument();
     expect(screen.getByText(text)).toBeInTheDocument();
+    expect(textElement).toHaveStyle({
+      margin: 0,
+      "font-family": theme.orbit.fontFamily,
+      "font-size": theme.orbit.fontSizeTextNormal,
+      "line-height": theme.orbit.lineHeightTextNormal,
+      "text-align": "left",
+      color: expect.stringContaining(theme.orbit.paletteInkDark),
```
mainframev, line 65: I still think nested expects is a bit strange thing, let's avoid such practices :)

packages/orbit-design-tokens/src/dictionary/definitions/foundation/palette/ink.json
```diff
@@ -5,12 +5,14 @@
         "dark": { "value": "#252A31", "internal": true, "type": "color" },
         "dark-active": { "value": "#0B0C0F", "internal": true, "type": "color" },
         "dark-hover": { "value": "#181B20", "internal": true, "type": "color" },
+        "dark-transparency": { "value": "#252A311A", "internal": true, "type": "color" },
```
mainframev, line 8: so you added transparency tokens only for tests? 🤔

packages/orbit-components/cypress/integration/pages/text-styles.test.ts
```diff
@@ -0,0 +1,197 @@
+import { defaultFoundation } from "@kiwicom/orbit-design-tokens";
+import color from "onecolor";
```
mainframev, line 2: I think better will be add here Cypress screenshot test instead of actually adding a new dependency, or new tokens (that are supposed to be used only in tests from what I see), it looks too comprehensive imo :) 

So I suggest to: 
- remove onecolor dep
- remove those transparency tokens
- remove those color tests
- use screenshot testing instead

mainframev (changes requested):
I made initial review for that :) 

Except for the requested changes and questions -> the docs pages on cloudflare staging look broken and the Text there is not working correctly, so I would take a look also on that

In Storybook -> withBackground prop seems not to be working

packages/orbit-components/src/Text/helpers/getMargin.ts
DSil: I have considered that, but it can't be like that. If one of those properties is `0`, it will be filtered out. Besides that, I do find the code more readable with the explicit conditions and results applied.

packages/orbit-components/src/Text/helpers/getMargin.ts
DSil: I am not sure I understand. The helper function would be defined outside this one but still on the context of Text? I don't think it will be reused anywhere, so what would be the benefit here? 🤔

packages/orbit-components/src/Text/__tests__/index.test.tsx
DSil: I can actually remove this expect, since it is being asserted in Cypress

packages/orbit-design-tokens/src/dictionary/definitions/foundation/palette/ink.json
DSil: No. Tokens exist for that. A primary text with background will have the class `.bg-text-primary-background` to apply the background

DSil:
> In Storybook -> withBackground prop seems not to be working

It is working for me (and on the staging deploy of this branch). Did you build the tokens package?

packages/orbit-components/cypress/integration/pages/text-styles.test.ts
DSil: Tokens are not used just for tests, as mentioned above. I am not worried about using `onecolor` for color assertions. It's a devDep just for testing, so I honestly see no harm in here… and it seems faster and equally (or even more) reliable than screenshot testing, no?
By using screenshot you suggest using loki? Or just screenshot testing in cypress?

mainframev: Just screenshot with cypress. Cypress is already running on CI, Loki has to be run manually.

packages/orbit-components/src/Text/helpers/getMargin.ts
mainframev: It's just an example, you can easily filter only undefined values and let falsy values like 0 be there.

packages/orbit-components/src/Text/helpers/getMargin.ts
mainframev: What I meant is something like that: 

```
const getMarginValue = (margin: string | number): string =>
  typeof margin === "number" ? `${margin}px` : margin;

const getMargin = (margin: Props["margin"]): { vars: object; classes: string[] } => {
  if (typeof margin === "string" || typeof margin === "number")
    return {
      vars: { "--text-margin": getMarginValue(margin) },
      classes: ["m-[var(--text-margin)]"],
    };

  const { top, bottom, left, right } = margin || {};
  const cssVar = {};
  const marginClasses: string[] = ["m-0"];

  if (top) {
    marginClasses.push("mt-[var(--text-margin-top)]");
    cssVar["--text-margin-top"] = getMarginValue(top);
  }

  if (bottom) {
    marginClasses.push("mb-[var(--text-margin-bottom)]");
    cssVar["--text-margin-bottom"] = getMarginValue(bottom);
  }

  if (left) {
    marginClasses.push("ml-[var(--text-margin-left)]");
    cssVar["--text-margin-left"] = getMarginValue(left);
  }

  if (right) {
    marginClasses.push("mr-[var(--text-margin-right)]");
    cssVar["--text-margin-right"] = getMarginValue(right);
  }

  return { vars: cssVar, classes: marginClasses };
};
```

Basically just a suggestion just DRY, as you are repeating those condition on multiple places

packages/orbit-design-tokens/src/dictionary/definitions/foundation/palette/ink.json
mainframev: Okay, got It. What you should change there, IMO -> the palette tokens should not include those transparency values, as we won't use them somewhere else. Just add them as static values to component level file text-background 🤔

packages/orbit-components/src/Text/helpers/getMargin.ts
DSil: Got it, agreed. Will refactor into that

packages/orbit-components/src/Text/helpers/getMargin.ts
DSil: Right, but won't that require me to dynamically construct the class names? I am not seeing a more efficient and readable approach, honestly, but I am open to suggestions

packages/orbit-design-tokens/src/dictionary/definitions/foundation/palette/ink.json
mainframev: > And we can't define it on the component to be 10% of the opacity of the original token, can we?

Currently do not think so, except if you will use the rgba, but it still will be static 
Maybe a good idea to extend our tokens generation logic to support automatic conversion to transparency token

> Besides that, by having it defined as internal on the foundation, they are only part of the foundation, they're not available on the theme. Isn't the foundation supposed to be precisely for those situations (tokens that should exist for us but not for external use)

if it has an internal flag, then it will be only for us 

the problem with that is also that foundation colors are linked to Figma and Figma does not have those transparency tokens. So if we decide now to keep them as tokens of the palette and not statically for Text (I agree, that it might in future force us to manually change it) 

So we can just move them to global or have it statically saved as I suggested above and later add a feature that will transform the token automatically :)

DSil: I think it is possible to transform tokens and apply opacity. I'll investigate and try to find a solution

mainframev: why not to use rgba value? 🤔 I mean simply to add support for our tokens package, something like: 
```
{ type: "color-rgba", value: {foundation.palette.ink.dark}, opacity: 10 }

under the hood: convertHexToRgba(token, opacity)

output will be rgba token
```

DSil: Yes, that's what I ended up doing. I noticed we actually had some buggy internal transformation (for instance, `socialButtonAppleBorderColorFocus` should have transparency, but if you check on Docs and other token outputs it was missing the alpha value. So I also fixed that. I will add the fix on this branch on a separate commit, for the sake of simplifying

packages/orbit-components/cypress/integration/pages/text-styles.test.ts
DSil: Cypress [does not allow visual testing natively](https://docs.cypress.io/guides/tooling/visual-testing). It has some open-source plugins for that, but I can believe those extra dependencies are even bigger than just this one. I do agree we need visual regression testing asap, as even [Cypress mentions that asserting CSS is not ideal](https://docs.cypress.io/guides/tooling/visual-testing#:~:text=You%20could%20technically%20write%20a%20functional%20test%20asserting%20the%20CSS%20properties%20using%20the%20have.css%20assertion%2C%20but%20these%20may%20quickly%20become%20cumbersome%20to%20write%20and%20maintain%2C%20especially%20when%20visual%20styles%20rely%20on%20a%20lot%20of%20CSS%20styles.).

My suggestion is that, as we already have this for Box, Stack and now Text, we don't block this PR for this and we actually think of a testing solution to implement. Let me know what you think.

mainframev: I would test only functionality and media queries, testing of colors then I would completely skip. We might consider to try out some of the plugins, as it seems to me as better alternative than using Loki.

DSil: It is not about testing colors _per se_. It is about testing the effects of the `withBackground` and `type` props, that I feel should not be left untested as they're part of the core of the component

mvidalgarcia:
Here are some things I've noticed in the [docs site](https://978e0aab.orbit.pages.dev/components/text/text/):

image

In this very same playground, it seems like the Medium and Larg variants have the same size.

image

"Success" has to be lowercase.

It can be changed here I believe:
- [size](https://github.com/kiwicom/orbit/blob/fb2c88abea266489c34c0070a416b8304d9e45f4/docs/src/__examples__/Text/SIZES.tsx#L9)
- [success](https://github.com/kiwicom/orbit/blob/fb2c88abea266489c34c0070a416b8304d9e45f4/docs/src/__examples__/Text/COLOR.tsx#L21)

--- 
> It is working for me (and on the staging deploy of this branch). Did you build the tokens package?

Do you have a staging storybook? If not, perhaps we should invest time in it, WDYT? cc @RobinCsl @oreqizer

DSil:
> Do you have a staging storybook?

We do, but it's set to be deployed on PRs to master only. Maybe we should change that as well 💭

DSil:
Well spotted, @mvidalgarcia. Those bugs were already there, actually, and were not introduced by this branch. Fixed 🙂

mainframev:
> Yes, we do. It's fixed in the #3899 and if it will be merged (I hope) the staging link will appear automatically in the description of PR. Currently you can get the link only if you will manually open the storybook action and click on the link of staging.

mainframev:
> > Do you have a staging storybook?
> 
> We do, but it's set to be deployed on PRs to master only. Maybe we should change that as well 💭

Nope, it's not, it's deployed for branches too

mainframev:
LGTM and good work on this 🚀 

Just classic rebase and it can be merged 👍🏻

mainframev (approved):

PR #3948, opened 2023-08-08
Fix ServiceLogo height token

DSil:
The token `heightServiceLogoMedium` changed from 24px to 20px during the Styled Dictionary / Tailwind migration, unexpectedly.
The token is being used to get the src from images.kiwi and the test is only asserting that the src URL contains the correct country code, it does not assert the width, that is also part of the URL. So it silently broke.

This PR improves the test and fixes the token value.


mainframev (approved):

mainframev:
Just update the snapshots and it should be ok 👍🏻

oreqizer (approved):
looks okay

PR #3949, opened 2023-08-08
chore(tokens): remove timestamps

mainframev:
I decided to remove the timestamps, as it creates for us conflicts usually that we need to handle


DSil (approved):
Great! LGTM! I would keep the warning to not edit directly, and just remove the timestamp. Is it possible?

mainframev:
> Great! LGTM! I would keep the warning to not edit directly, and just remove the timestamp. Is it possible?

Unfortunately with current API seems like it's not supported, I found only this option 😞

PR #3950, opened 2023-08-09
feat: abstract away react-uid and remove it from deps

RobinCsl:
This is to allow for React 17/18 compatibility, with additional steps required for React 17 compat.

BREAKING CHANGE: Message to be amended

---

This is a proposal to use dependency injection and allow for React 17 compat while updating Orbit to React 18 primitives like useId.

 Storybook: https://orbit-mainframev-rcsl-abstract-react-uid.surge.sh

### DIFF
--- a/docs/src/documentation/05-development/04-migration-guides/v9.mdx
+++ b/docs/src/documentation/05-development/04-migration-guides/v9.mdx
@@ -62,3 +62,46 @@ Orbit v9 introduces a new prop on `OrbitProvider` called `useId`.
   ```
 
 One more note: the underlying implementation for how ids are generated is slightly different now. In particular, Orbit's `useRandomIdSeed` no longer relies on `react-uid`'s `useUIDSeed`.
+
+## Deprecations
+
+### ClickOutside
+
+The ClickOutside component is now deprecated and its usage should be replaced by the `useClickOutside` hook, already available in Orbit and documented [here](/utilities/useClickOutside/).
+
+**Before:**
+
+```jsx
+<ClickOutside onClickOuside={...}>
+  <div>
+    Some content
+  </div>
+</ClickOutside>
+```
+
+**Now:**
+
+```jsx
+useClickOutside(ref, ev => ...);
+
+<div ref={ref}>
+  Some content
+</div>
+```
+
+### Flow types
+
+Flow types are being deprecated in favor of TypeScript types. The Flow types definitions will be removed in a future version of Orbit.
+
+### Removed previously deprecated components
+
+The following deprecated components were removed in this version:
+
+- FormFeedback
+- InputField
+- InputFile
+- InputGroup
+- Select
+- Textarea
+
+Notice that for most of them there are already replacements available in Orbit, with the same name. The removed components were only the deprecated versions.
--- a/docs/src/documentation/05-development/04-migration-guides/v9.mdx
+++ b/docs/src/documentation/05-development/04-migration-guides/v9.mdx
@@ -1,0 +1,64 @@
+---
+title: V9
+description: How to migrate to Orbit 9.0.0
+redirect_from:
+  - /migration-guides/v9/
+---
+
+# Orbit Migration Guide v9
+
+This migration guide focuses on the process of migrating from Orbit v8.1.2 to v9.0, as some breaking changes were introduced.
+With this guide we aim to walkthrough all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortless.
+
+This version of Orbit is **still** compatible with React 17 and doesn't make use of new features only available in React 18.
+
+That being said, because v9 now recommends using React 18, some changes are required to allow for this transition.
+
+## Breaking changes
+
+### `react-uid` vs React.useId
+
+React 18 introduces this new hook to generate stable random ids called `useId`, while Orbit used to rely on `react-uid` for that purpose.
+
+`react-uid` is not compatible with React 18 (especially when Strict or Concurrent modes are enabled) and as a result was removed from Orbit's dependencies.
+
+The need to generate random ids is still present, however.
+
+Orbit v9 introduces a new prop on `OrbitProvider` called `useId`.
+
+- **React 18**: if you are using React 18, please pass in `React.useId`, e.g.
+
+  ```tsx
+  import * as React from "react";
+
+  export const Root = (props: Props) => {
+    return (
+      <OrbitProvider theme={theme} useId={React.useId}>
+        {children}
+      </OrbitProvider>
+    );
+  };
+  ```
+
+- **React 17**: if you are using React 17, you will need to:
+
+  - install `react-uid` in your dependencies (`yarn add react-uid`)
+  - pass `react-uid`'s `useRandomId` export in the `useId` prop, as well as mounting two providers (`UIDReset` and `UIDFork`), like so
+
+  ```tsx
+  import { useRandomId, UIDReset, UIDFork } from "react-uid";
+
+  export const Root = (props: Props) => {
+    return (
+      <UIDReset>
+        <UIDFork>
+          <OrbitProvider theme={theme} useId={useRandomId}>
+            {children}
+          </OrbitProvider>
+        </UIDFork>
+      </UIDReset>
+    );
+  };
+  ```
+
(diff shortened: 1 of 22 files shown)


mainframev:
@RobinCsl I think the idea is good, I would add a simple test to it to make sure the value is passed correctly. The pipeline fails because the react dependencies aren't updated to v18, but I guess you knew that ;)

RobinCsl:
> @RobinCsl I think the idea is good, I would add a simple test to it to make sure the value is passed correctly. The pipeline fails because the react dependencies aren't updated to v18, but I guess you knew that ;)

Yep :)

Just wanted to polish the rest in case we liked the idea

packages/orbit-components/src/OrbitProvider/types.d.ts
```diff
@@ -8,4 +8,5 @@ import type { Theme } from "../defaultTheme";
 export interface Props {
   readonly theme: Theme;
   readonly children: React.ReactNode;
+  readonly useId: typeof React.useId;
```
DSil, line 11: Is `react-uid` compliant with this type?

yarn.lock
```diff
@@ -2637,6 +2637,20 @@
   resolved "https://registry.npmjs.org/@kiwicom/browserslist-config/-/browserslist-config-4.0.0.tgz"
   integrity sha512-CkBXOf5noNBSh8nOQl6/qMBEz1GwYQpIxDTyImqJDwSiXgY/lWHgW9TQT8n9akhFMd4uB9tAFsWFPk8EhzogfA==
 
+"@kiwicom/orbit-design-tokens@5.0.0":
```
DSil, line 2640: Why is this here? 🤔

RobinCsl: No idea :D but I think it got removed now that I changed the base branch to be the react upgrade one.

packages/orbit-components/src/OrbitProvider/types.d.ts
RobinCsl: yeah, I'll simplify it to `() => string`.

packages/orbit-components/src/OrbitProvider/index.tsx
```diff
@@ -1,19 +1,31 @@
...
-          <QueryContextProvider>{React.Children.only(children)}</QueryContextProvider>
-        </StyledThemeProvider>
+        <OrbitProvider theme={theme} useId={useUID}>
+          {children}
+        </OrbitProvider>
       </UIDFork>
     </UIDReset>
+ * ```
+ *
+ */
```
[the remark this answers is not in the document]
RobinCsl: Yeah, I plan to add a proper migration guide for this, at least in the BREAKING CHANGE block of the commit message, but probably in a markdown file like you did elsewhere? (Could you please point me to an example migration guide/release note you did?)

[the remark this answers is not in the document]
DSil: The v7 migration guide is [here](https://github.com/kiwicom/orbit/blob/master/docs/src/documentation/05-development/04-migration-guides/v7.mdx).

This is visible in orbit.kiwi, [here](https://orbit.kiwi/development/migration-guides/v7/).

docs/src/documentation/05-development/04-migration-guides/v9.mdx
RobinCsl, line 1: @DSil @mainframev here's a first draft of the migration guide related to `react-uid`.

packages/orbit-components/src/hooks/useRandomId/index.ts
```diff
@@ -1,14 +1,17 @@
...
+import { useRandomId as useRandomIdContext } from "../../OrbitProvider/RandomId/Provider";
 
 type UseRandomIdSeed = () => (id: string) => string;
 type UseRandomId = () => string;
 
-export const useRandomIdSeed: UseRandomIdSeed = () => {
-  return useUIDSeed();
+export const useRandomId: UseRandomId = () => {
+  const useId = useRandomIdContext();
+  return useId();
```
DSil, line 8: Typescript doesn't like this.
> Cannot invoke an object which is possibly 'undefined'

Do we need to assert that it is defined before calling it? 💭 
In practice it should never happen because of the thrown error, but I guess TS doesn't catch that 🤔

docs/src/documentation/05-development/04-migration-guides/v9.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: V9
+description: How to migrate to Orbit 9.0.0
+redirect_from:
+  - /migration-guides/v9/
+---
+
+# Orbit Migration Guide v9
+
+This migration guide focuses on the process of migrating from Orbit v<LATEST 8 VERSION> to v9.0, as some breaking changes were introduced.
```
DSil, line 10: `v<LATEST 8 VERSION>` is this a placeholder to change before merge/release?

docs/src/documentation/05-development/04-migration-guides/v9.mdx
mainframev: LGTM :)

docs/src/documentation/05-development/04-migration-guides/v9.mdx
RobinCsl: Before release, yes. Though we could remove it in case it causes more trouble for us :) (or update it after the facts)

docs/src/documentation/05-development/04-migration-guides/v9.mdx
```diff
@@ -0,0 +1,64 @@
...
+Orbit v9 introduces a new prop on `OrbitProvider` called `useId`.
+
+- **React 18**: if you are using React 18, please pass in `React.useId`, e.g.
+
+  ```tsx
+  import * as React from "react";
+
+  export const Root = (props: Props) => {
+    return (
+      <OrbitProvider theme={theme} useId={React.useId}>
```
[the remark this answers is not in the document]
DSil: I had the same question myself. I am not sure about the consequences of that for a React 17 user that does not actually pass any prop there 🤔 isn't it safer to just embrace the Breaking Change and force it to explicitly be defined?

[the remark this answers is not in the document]
mainframev: I guess yes, safer will be explicitly defined, we can left it as is then 👍🏻

packages/orbit-components/src/hooks/useRandomId/index.ts
```diff
@@ -1,14 +1,17 @@
-import { useUIDSeed, useUID } from "react-uid";
+import { useRandomId as useRandomIdContext } from "../../OrbitProvider/RandomId/Provider";
 
 type UseRandomIdSeed = () => (id: string) => string;
 type UseRandomId = () => string;
 
-export const useRandomIdSeed: UseRandomIdSeed = () => {
-  return useUIDSeed();
+export const useRandomId: UseRandomId = () => {
```
mainframev, line 6: Let's add a jsDoc deprecation message here and move this hook maybe to deprecated folder 🤔

RobinCsl: I think it's okay to keep it for now, once we officially don't support React 17, we can clean this up and remove these hooks if they prove to be unnecessary (since it's basically just using `React.useId` anyway).

mainframev: > _We can decide to deprecate it, but what is the alternative? I think for deprecation we need to have a solution that works for both versions of React or officially deprecate all support for React <18, which is not something we are planning to do right now, right?_

The alternative is React.used hook that will replace that, deprecated does not mean that it's not in use, it means that it will be removed and it will be, and not necessarily in the next major version. The deprecation period may vary.

DSil: I get what you're saying, but I think we should only do that when we officially deprecate React 17 support entirely (which we are not doing yet, right?). I think, for now, the best option is to just add a note on the documentation that this hook should only be used by React <18 users. WDYT?

mainframev:
@RobinCsl Do not forget ammend the deprecation message ;)

mainframev (approved):

DSil:
@mainframev I am taking care of it, thank you for reminding. Done

PR #3952, opened 2023-08-10
feat(NavigationBar): migrate to Tailwind

DSil:
Migrate NavigationBar component to Tailwind
 Storybook: https://orbit-mainframev-migrate-NavigationBar-tailwind.surge.sh


packages/orbit-components/src/NavigationBar/index.tsx
```diff
@@ -101,8 +57,17 @@ const NavigationBar = ({
...
-    <StyledNavigationBar data-test={dataTest} id={id} shown={shown}>
-      <StyledNavigationBarContent>{children}</StyledNavigationBarContent>
+    <nav
+      data-test={dataTest}
+      id={id}
+      className={cx(
+        "bg-white-normal shadow-fixed p-sm z-navigation-bar fixed left-0 right-0 top-0 box-border flex w-full translate-x-0 items-center",
+        "duration-normal transition-transform ease-in-out",
+        "tb:h-[64px] h-[52px]", // As defined on the const above
+        shown ? "translate-y-0" : "tb:translate-y-[-64px] translate-y-[-52px]", // As defined on the const above
```
DSil, line 67: Wondering if these should be tokens 🤔 We could use them for the calculations above as well 💭

packages/orbit-components/src/NavigationBar/index.tsx
```diff
@@ -101,8 +57,17 @@ const NavigationBar = ({
...
 
   return (
-    <StyledNavigationBar data-test={dataTest} id={id} shown={shown}>
-      <StyledNavigationBarContent>{children}</StyledNavigationBarContent>
+    <nav
+      data-test={dataTest}
+      id={id}
+      className={cx(
+        "bg-white-normal shadow-fixed p-sm z-navigation-bar fixed left-0 right-0 top-0 box-border flex w-full translate-x-0 items-center",
+        "duration-normal transition-transform ease-in-out",
```
mainframev, line 65: adding `transform-gpu` woulnd't be a bad idea, we usually prefer to use transform3d and wll-change transform to make them more smooth.

packages/orbit-components/src/NavigationBar/index.tsx
mainframev: I am okay with having it like you did, not everything should be a token

packages/orbit-components/src/common/tailwind/wrap.ts
```diff
@@ -0,0 +1,33 @@
...
+  WRAP = "wrap",
+  NO_WRAP = "nowrap",
+}
+
+export const wrapClasses: {
+  [K in QUERIES | WRAP]: K extends QUERIES ? Record<WRAP, string> : string;
+} = {
+  [WRAP.WRAP]: "flex-wrap",
+  [WRAP.NO_WRAP]: "flex-nowrap",
+  [QUERIES.LARGEDESKTOP]: {
```
mainframev, line 13: Can you define queries as top->down approach? I saw it before already and I found it inconsistent, let's define them top->down from mediuMobile to largeDesktop, for better readability

mainframev (changes requested):

docs/package.json
```diff
@@ -122,5 +122,8 @@
     "uuid": "^8.3.2",
     "vfile-reporter": "^7.0.5",
     "webpack": "^5.76.0"
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
DSil: This was not installed by me 😅 it was here because of missing rebase, but is out of scope in this PR… Can be added to the base branch though

packages/orbit-components/cypress/integration/pages/navigation-bar.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { NavigationBar } from "@kiwicom/orbit-components";
+
+export default function BoxMediaProps() {
```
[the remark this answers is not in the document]
DSil: 🙈 copy-paste… I actually called it `NavigationBarScroll`, since is the behavior being tested here

packages/orbit-components/src/common/tailwind/wrap.ts
mainframev: Agree, just if it can be fixed during this change -> it will be also good :)

mainframev (approved):

mainframev:
just autofixup before the merge and 👍🏻

PR #3954, opened 2023-08-11
List to Tailwind

oreqizer:
Migrated `List` to Tailwind.

Closes [FEPLT-1662](https://kiwicom.atlassian.net/browse/FEPLT-1662)
 Storybook: https://orbit-mainframev-oreqizer-tailwind-list.surge.sh


mainframev:
@oreqizer could you please instead of adding Jira ticket with just a name add it as a link? 🙏🏻  :)

packages/orbit-components/src/List/ListItem/index.tsx
```diff
@@ -1,143 +1,25 @@
...
-      height: ${theme.orbit.heightIconSmall};
-      width: ${theme.orbit.widthIconSmall};
-    `,
-  };
-
-  if (!type) return null;
-  return tokens[type];
-};
-
-export const Item = styled(({ ...props }) => <li {...props} />)`
```
mainframev, line 63: Unfortunately for now you would have to keep it, as it's required in other our components :) we will remove it once they also will be migrated ;)

oreqizer:
I see there are some weird dependencies on the removed `Item` styled component. I will adjust that

packages/orbit-components/src/List/ListContext.ts
```diff
@@ -1,18 +0,0 @@
...
-
-import type { Size } from "../common/types";
-import type { Type } from "./types";
-
-interface ListContextProps {
-  readonly size?: Size | null;
-  readonly type?: Type | null;
-}
-
-const ListContext = React.createContext<ListContextProps>({
```
mainframev, line 11: Not sure if it's a good idea to remove the context here. It's being used in order not to pass those size and type properties to each ListItem individually, so better for users, as they can have a lot of them. 

And it's also a breaking change :) So I think better will be to keep it

mainframev (changes requested):

packages/orbit-components/src/List/ListContext.ts
mainframev, line 11: Please make sure that this removal won't be a breaking change :)

packages/orbit-components/src/List/ListItem/index.tsx
oreqizer: I solved it using `.orbit-list-item` in those other components

packages/orbit-components/src/List/ListContext.ts
oreqizer: it isn't, I checked with Daniel

packages/orbit-components/src/List/ListItem/index.tsx
```diff
@@ -1,143 +1,28 @@
...
-  const { size, type } = React.useContext(ListContext);
   return (
-    <Item data-test={dataTest} type={type}>
+    <li
+      data-test={dataTest}
+      className="orbit-list-item font-base mb-xxs flex last:m-0 last-of-type:m-0"
+    >
       {icon && (
-        <IconContainer type={type} size={size}>
+        <div className="orbit-list-item-icon me-xs [&>*]:h-icon-small [&>*]:w-icon-small [&>svg]:h-icon-small [&>svg]:w-icon-small [&>.orbit-carrier-logo]:h-icon-small [&>.orbit-carrier-logo]:w-icon-small [&>.orbit-carrier-logo>img]:h-icon-small [&>.orbit-carrier-logo>img]:w-icon-small flex flex-[0_0_auto] items-center">
```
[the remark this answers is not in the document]
DSil: Also, ESLint complains that those custom selectors are not Tailwind classes. Should we add a rule to avoid such warnings?

packages/orbit-components/src/List/index.tsx
```diff
@@ -1,76 +1,54 @@
...
-const getTypeToken = ({ theme, type }: { theme: Theme; type?: Type }): string | null => {
-  const typeTokens = {
-    [TYPES.PRIMARY]: theme.orbit.colorTextPrimary,
-    [TYPES.SECONDARY]: theme.orbit.colorTextSecondary,
-  };
-
-  if (!type) return null;
-
-  return typeTokens[type];
+const sizeTokens = {
```
DSil, line 18: For organization consistency, we tend to have a `helpers` folder with a `twClasses.ts` file where we store this kind of logic

packages/orbit-components/src/List/index.tsx
```diff
@@ -1,76 +1,54 @@
...
-    [SIZES.LARGE]: theme.orbit.fontSizeTextLarge,
-  };
-
-  if (!size) return null;
-
-  return sizeTokens[size];
+import type { Props } from "./types";
+import SPACINGS_AFTER from "../common/getSpacingToken/consts";
+
+const spaceAfterTokens = {
```
DSil, line 8: We already have these tokens-classes match in a common file. Under `src/common/tailwind/`

packages/orbit-components/src/List/__tests__/index.test.tsx
```diff
@@ -21,6 +21,5 @@ describe("List", () => {
 
     expect(screen.getByTestId(dataTest)).toBeInTheDocument();
     expect(screen.getByText(content)).toBeInTheDocument();
-    expect(screen.getByRole("list")).toHaveStyle({ marginBottom: "12px" });
```
DSil, line 24: Why remove this test? 🤔

packages/orbit-components/src/List/__tests__/index.test.tsx
DSil, line 1: We could take the opportunity to improve our tests and assert that props actually have effect on the component (and on its descendants, in this case) 💭

packages/orbit-components/src/List/__tests__/index.test.tsx
oreqizer: because it tests that the component is doing what it is doing. the assert has no value, outside of binding styling to unit (non-visual) tests

packages/orbit-components/src/List/index.tsx
oreqizer: while I appreciate the fact that this is how it is commonly done now, I wholeheartedly disagree with the `helpers` and `twClasses` pattern. it is unnecessary indirection that just makes things hard to find, makes us use redundant imports for one-off usages, makes CI slower and makes DX slower

https://softwareengineering.stackexchange.com/questions/422907/is-putting-general-use-functions-in-a-helpers-file-an-anti-pattern-or-code-sme

I challenge this pattern and propose we co-locate object-class maps directly to components where they're used, for:
- ease of discovery
- CI speed (less files and imports)
- less mental overhead and open files in editors

if there's a  **functional** binding between usages, then I propose using a **focused** file/directory where we store such bindings (usually objects with properties shared across multiple components)

and I propose banning files/folders with names including:
- `helpers`
- `utils`
- `classes`
- `tailwind` / `tw`
- `general`
- `shared`
- `common`

mainframev: Thanks for the proposal Boris :) 

while in general, I agree about CI speed, but I would still prefer to have those classes separated sometimes, I think it's more a personal preference or I do not think it's making the DX significantly worse. In the List example, it's completely ok to use it directly where they used it, but in bigger component, it would lead to a much bigger file and then need to scroll a lot :) 

Also, I doubt that our common folder can be completely banned, but we can reduce the helpers/twClasses pattern where it's possible

packages/orbit-components/src/List/ListItem/index.tsx
```diff
@@ -1,143 +1,28 @@
...
-  theme: defaultTheme,
-};
-
 const ListItem = ({ label, children, icon = <CircleSmall />, dataTest }: Props) => {
-  const { size, type } = React.useContext(ListContext);
   return (
-    <Item data-test={dataTest} type={type}>
+    <li
+      data-test={dataTest}
+      className="orbit-list-item font-base mb-xxs flex last:m-0 last-of-type:m-0"
```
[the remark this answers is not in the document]
oreqizer: imo no, as long as it looks okay I wouldn't add redundant styles. `flex-row` is the default, so not necessary

why does it need `StyledText` / `.orbit-text`? colour is handled level up, font here (though should not be necessary imo)

packages/orbit-components/src/List/ListItem/index.tsx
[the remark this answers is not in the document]
oreqizer: will adjust, thank 👍

packages/orbit-components/src/List/ListItem/index.tsx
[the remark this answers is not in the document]
DSil: > imo no, as long as it looks okay I wouldn't add redundant styles

I agree. My concern is that I am not sure if it wasn't there in the first place because it was needed because of something else. But I am in favor of avoiding extra redundant styles.

Regarding Text, it was applying `inherit` to `line-height` and `font-size` properties so that it is consistent with the size passed to List, I assume

packages/orbit-components/src/List/index.tsx
DSil: I can believe about CI speed, but I could argue about DX, which, in the end, can be measured by personal preference, as Viktor mentioned. Especially when we have classes that are reused by multiple components, like the spaceAfter, for example. If one changes, I would prefer not to change it manually everywhere it is being used if I can avoid that.

Just like sometimes one could have the `.styled.ts` with all the Styled Components, we could also have the TW classes on a separate file, so we separate styling from logic / behavior and we avoid files with hundreds of lines of code. Or we can do the complete opposite if we find that beneficial. Personally, I would prefer the code more separated and organized (though we can define a different organization) than an all-in-one file. But that's a decision for the whole team.

oreqizer: as I said in

> if there's a **functional** binding between usages, then I propose using a focused file/directory where we store such bindings (usually objects with properties shared across multiple components), with a file name that accurately represents the semantics of its contents

we can for example create a `src/spaceAfter.ts` file (or nest focused utilities under an e.g. `layout` folder) and use this. I have no problem with that

as you Viktor mentioned, if we have a massive object that would pollute the `Component/index.tsx` file, there are two things:
- this object should not exist in the first place (to be solved later)
- we indeed can place *huge* objects outside next to the file, or to a `modificators` folder, with a **focused** name, for example `colors.ts`. point is, people _know_ what this folder/file is for. it is not `common`, `shared`, `classes`, or other ambiguous things 😄 I hope you get my drift

mainframev: Agree about that :) Objects will be removed later, yes

we have `utils/layout`, we can break this file into smaller ones, like: 
```
-- utils/layout
-- -- getSpacingToken.ts
-- -- someOther.ts
```

> _**we can for example create a src/spaceAfter.ts file (or nest focused utilities under an e.g. layout folder) and use this. I have no problem with that**_

I think you do not mind if we can keep the `utils` folder if all of them will be with a focused folder and name, right? :D

I will make up some new structure that will follow your suggestion (although it will breaking, as I know for sure that people using our common folder, for instance)

oreqizer: we can move this discussion elsewhere, for now I cleaned up my implementation for the scope of this MR

packages/orbit-components/src/List/__tests__/index.test.tsx
oreqizer: then you're unit-testing that CSS works like it should 😅 this is in the scope of visual testing, and that will be covered separately

mainframev (approved):
@oreqizer LGTM 👍🏻 could you rebase and drop the first commit with Box? It seems that conflicts were not resolved correctly and those files are not anymore inside the Box folder, they are in common and used in Stack and Box.

PR #3961, opened 2023-08-13
remove deprecated components

mainframev:
These components have been deprecated more than a year ago. From what I see from our tracking, they aren't being used anywhere. Maybe we can double-check via source graph, but still, they were deprecated too long and all of them have a replacement


DSil (approved):
These components were no longer being exported in `index.ts` from `orbit-components`, so is it really a breaking change? 💭 🤔

PR #3966, opened 2023-08-16
feat(Loading): migrate to Tailwind

mainframev:
[FEPL-1674](https://kiwicom.atlassian.net/browse/FEPLT-1674)
 Storybook: https://orbit-mainframev-tw-migrate-loading.surge.sh

### DIFF
--- a/packages/orbit-components/src/Loading/index.tsx
+++ b/packages/orbit-components/src/Loading/index.tsx
@@ -10,7 +10,7 @@ import useTheme from "../hooks/useTheme";
 const CircleLoader = ({ animationDelay }: { animationDelay?: string }) => {
   return (
     <div
-      className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px] [&>*:nth-child(3)]:my-0"
+      className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px] [&:nth-child(3)]:m-0"
       style={{ animationDelay }}
     />
   );
@@ -99,7 +99,7 @@ const Loading = ({
             type === TYPE_OPTIONS.BOX_LOADER && "h-[80px] justify-center",
             type === TYPE_OPTIONS.PAGE_LOADER && "h-[120px] flex-col justify-center",
           ])}
-          style={{ height: `${customSize}px` }}
+          style={{ height: customSize }}
           data-test={dataTest}
           id={id}
         >
--- a/packages/orbit-tailwind-preset/src/presets/__tests__/__snapshots__/configs.test.ts.snap
+++ b/packages/orbit-tailwind-preset/src/presets/__tests__/__snapshots__/configs.test.ts.snap
@@ -483,7 +483,7 @@ exports[`orbitPreset should match snapshot 1`] = `
     },
     "animation": {
       "bounce": "bounce 1s infinite",
-      "loader": "loader 1s 1.25s infinite ease-in-out",
+      "loader": "loader 1.25s infinite ease-in-out",
       "none": "none",
       "ping": "ping 1s cubic-bezier(0, 0, 0.2, 1) infinite",
       "pulse": "pulse 1.5s infinite",
--- a/packages/orbit-tailwind-preset/src/presets/components/index.ts
+++ b/packages/orbit-tailwind-preset/src/presets/components/index.ts
@@ -202,7 +202,7 @@ const cfg = (options?: Options): Config => {
         animation: {
           "pulse-slow": "slow-pulse 2s ease-in-out 0.5s infinite",
           spinner: "spinner 0.75s linear infinite",
-          loader: "loader 1s 1.25s infinite ease-in-out",
+          loader: "loader 1.25s infinite ease-in-out",
           pulse: "pulse 1.5s infinite",
         },
         textColor: Object.entries(getForegroundColors()).reduce((acc, [key, value]) => {


packages/orbit-components/src/Loading/index.tsx
```diff
@@ -1,191 +1,70 @@
...
     return (
-      <StyledLoader>
-        <StyledLoaderCircle />
-        <StyledLoaderCircle />
-        <StyledLoaderCircle />
-      </StyledLoader>
+      <div className="flex items-center justify-center">
+        <div className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px]" />
+        <div
+          className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px]"
```
DSil, line 43: I see this `animate-loader` class applies `loader 1s 1.25s infinite ease-in-out` but it used to be `loader 1.25s infinite ease-in-out`. This must be changed on the tailwind-preset package.

packages/orbit-components/src/Loading/index.tsx
```diff
@@ -1,191 +1,70 @@
...
   return (
-    <StyledSpinner viewBox="0 0 40 40">
-      <StyledSpinnerCircle cx="50%" cy="50%" r="18" type={type} />
-    </StyledSpinner>
+    <svg viewBox="0 0 40 40" className="orbit-loading-spinner animate-spinner h-xxl w-xxl">
+      <circle
+        cx="50%"
+        cy="50%"
+        fill="transparent"
+        strokeWidth={3}
```
DSil, line 60: Same question as above

packages/orbit-components/src/Loading/index.tsx
```diff
@@ -1,191 +1,70 @@
...
+        viewBox={`0 0 ${customSize} ${customSize}`}
+        className="orbit-loading-spinner animate-spinner h-xxl w-xxl"
+      >
+        <circle
           cx="50%"
           cy="50%"
-          r={customSize ? customSize / 2 - 2 : 18}
-          customSize={customSize}
+          fill="transparent"
+          strokeWidth={3}
```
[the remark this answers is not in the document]
mainframev: it won't make any difference, so it's ok

packages/orbit-components/src/Loading/index.tsx
```diff
@@ -200,15 +79,38 @@ const Loading = ({
...
+      {Boolean(children) && !loading ? (
         children
       ) : (
-        <StyledLoading type={type} dataTest={dataTest} id={id}>
+        <div
+          className={cx([
+            "flex",
+            "items-center",
+            "justify-center",
+            "h-[120px]",
```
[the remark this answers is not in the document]
mainframev: it will be a default and if type has it's own height -> it will override that value above, but agree that better to be explicit here and define for each type.

packages/orbit-components/.eslintrc.cjs
```diff
@@ -30,6 +30,10 @@ module.exports = {
       files: ["src/**/*.ts?(x)"],
       rules: {
         "@typescript-eslint/consistent-type-imports": "error",
+        "tailwindcss/no-custom-classname": [
+          1,
+          { whitelist: ["(orbit\\-).*"], config: "./tailwind.config.ts" },
```
DSil, line 35: Why this change? I believe that if we want to use custom components in our docs, for example, where we have some, we will face same problem. So why can't we have it global?

packages/orbit-components/src/Loading/index.tsx
```diff
@@ -1,194 +1,76 @@
...
-    margin: 0;
-  }
-`;
-
-StyledLoaderCircle.defaultProps = {
-  theme: defaultTheme,
+const CircleLoader = ({ animationDelay }: { animationDelay?: string }) => {
+  return (
+    <div
+      className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px] [&>*:nth-child(3)]:my-0"
```
DSil, line 13:
```suggestion
      className="animate-loader rounded-circle bg-cloud-dark h-xs w-xs me-[6px] [&:nth-child(3)]:m-0"
```

You were targeting descendants with the nth(3) property, but you wanna target the actual element. If it is the third child of its parent, it should have `margin: 0` (and not `margin-y: 0` 😉).

packages/orbit-components/.eslintrc.cjs
mainframev: because it needs the reference to tailwind.config.ts, which is not global (it can't find our preset classes without it and eslint throws warnings) 
Screenshot 2023-09-07 at 12 16 27

mainframev: I can try to have it globally and just import the object from our preset

DSil: I don't think that is the reason. The warning is thrown every time a new class is used for the first time, until CSS is built. So if you see that warning what you can do is build the CSS `yarn components tailwind` and, if needed, restart the ESLint server on VSCode. Those steps have been enough to solve those warnings to me

mainframev: I thought it would eliminate the need to build the file

mainframev: yep, we can

DSil (approved):

PR #3969, opened 2023-08-20
chore(Stack): fix stack issues after refactoring to Tailwind

mainframev:
There were two problems that caused those issues reported by Marco: 

- inline just had wrong condition instead of checking it to false it was applying full-width when inline is true.
- the other problem was a bit more complex and I just reused Search solution with having getProperty function, otherwise spacing and directions are not working correctly with media query properites 
 Storybook: https://orbit-mainframev-fix-stack-issue.surge.sh


DSil (approved):
LGTM. The new function (that came from Search) is a little bit hard to read / maintain for me personally… but I don't have any suggestion for improving it at the moment without some bigger refactor, which I think we can avoid for now

PR #3972, opened 2023-08-22
Adjust tailwind preset export

DSil:
The tailwind config preset can be simplified into just one, as it happens that **both** are always needed by our users.

Since both are always needed, we can export just the most complete one, but document only the foundation. Everything else available can be used at the users' own risk.

This is a breaking change, since we change the default export. But we're on an early stage of adoption, so the impact will be low.

I have also updated the documentation accordingly.
 Storybook: https://orbit-mainframev-adjust-tailwind-preset-export.surge.sh


packages/orbit-components/tailwind.config.js
```diff
@@ -1,11 +1,10 @@
-const { default: orbitPreset, orbitComponentsPreset } = require("@kiwicom/orbit-tailwind-preset");
+const orbitPreset = require("@kiwicom/orbit-tailwind-preset");
```
mainframev, line 1: can you refactor it to ts file in the same way as it's #3942 branch? 🙏🏻

mainframev:
LGTM, just one suggestion :)

mainframev:
also @DSil as soon it's as a breaking changes, let's maybe merge it to #3942 so it will be available with v10 release, wdyt?

DSil:
The commit with the BC only affects the tailwind-preset package, so it doesn't make a difference where we merge it, as it will only affect that package anyway. Also, the time between releases, ideally, won't be too different, so I don't know if it's worth it. I'm ok with any decision though

PR #3974, opened 2023-08-23
fix(Stack): add gap and legacy margins

mainframev:
[As noticed here](https://skypicker.slack.com/archives/C05FDP4UM3N/p1692720206376469), would be better to use gaps and legacy margins under legacy prop
 Storybook: https://orbit-mainframev-fix-stack-gap-space-issue.surge.sh


DSil (commented):
I have one question: we were applying the `margin` property [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Stack/helpers/getGap.ts#L49) and the `safari` variant was even added. But then why don't we do it as well here? Or am I missing something?

The first commit should be deleted, though. We must also reword the commit message that document the breaking change in Stack, removing the `legacy` prop. Because it is not removed after all

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -206,7 +206,11 @@ const cfg = (options?: Options): Config => {
           addVariant("not-first", "&:not(:first-child)"),
           addVariant("type-even", "&:nth-of-type(even)"),
           addVariant("type-odd", "&:nth-of-type(odd)"),
-          addVariant("target-blank", "&[target='_blank']")
+          addVariant("target-blank", "&[target='_blank']"),
+          addVariant(
+            "safari",
```
DSil, line 211: We add this but we don't do anything with it… Then, why? 💭

mainframev: I planned to add that after initial review of adding gap

packages/orbit-components/src/common/tailwind/spacing.ts
```diff
@@ -177,29 +177,276 @@ export const verticalTokens = {
...
+    [SPACING.medium]: "ld:gap-md",
+    [SPACING.large]: "ld:gap-lg",
+    [SPACING.XLarge]: "ld:gap-xl",
+    [SPACING.XXLarge]: "ld:gap-xxl",
+    [SPACING.XXXLarge]: "ld:gap-xxxl",
+  },
+};
+
+const safariVerticalTokens = {
+  [SPACING.none]: "safari:[&>*:not(:last-child)]:safari-y-none",
```
DSil, line 254: I'm not sure I understand this. I understand the `safari:` that will apply `@supports (-webkit-touch-callout: none) and (not (translate: none))` but then it should be `space-y-none`, right? I can't test it effectively, because I don't have the device and browser, but I think this is a typo, right?

DSil (changes requested):
I can't effectively test it, but I think it looks ok, if the supposed typo is fixed.

packages/orbit-components/src/common/tailwind/spacing.ts
mainframev: right, it's typo, thanks 👍🏻

PR #3975, opened 2023-08-25
Card to Tailwind

oreqizer:
[FEPLT-1663](https://kiwicom.atlassian.net/browse/FEPLT-1663)
 Storybook: https://orbit-mainframev-oreqizer-card-tw.surge.sh


oreqizer:
failing tests are unrelated

DSil:
@oreqizer there is one snapshot testing related to this. And since I had to rebase the base branch with master, can you please rebase yours and update the snapshot testing for the new tailwind tokens? 🙏

packages/orbit-components/src/Card/index.tsx
```diff
@@ -48,46 +27,34 @@ const Card = ({
...
 
   return (
-    <StyledCard
-      // TODO: remove spaceAfter in the next major version
-      margin={spaceAfter ? { bottom: getSpacingToken({ spaceAfter, theme }) } : margin}
-      data-test={dataTest}
+    <div
       id={id}
+      className={cx(
+        "[&>*]:border-elevation-flat-border-color [&>*:first-child]:rounded-t-normal [&>*:last-child]:rounded-b-normal [&>*:first-child]:border-t-[1px]",
```
DSil, line 50: I assume this dev replaces the `StyledCard` but I don't recognize these styles there. I see a different set of styles, not matched here
```
    width: 100%;
    box-sizing: border-box;
    position: relative;
    font-family: ${theme.orbit.fontFamily};
```

packages/orbit-components/src/Card/index.tsx
```diff
@@ -99,52 +66,16 @@ const Card = ({
...
-                ) : (
-                  renderSection(item)
-                )}
-              </SectionProvider>
-            );
-          })
-        : null}
-    </StyledCard>
+      {loading && (
+        <div className="border-x-[1px] border-b-[1px]">
```
DSil, line 75: If this is replacing `<CardWrapper noPadding>` this is missing `py-md`. Also, when a border is of 1px, can't we ommit its custom value? And just be `border-x border-b`?

DSil (changes requested):
This was a very tough component to review, and I imagine to migrate as well. I have pointed some points of improvement and asked a few questions, for clarity.
Also, the component is visually different on mobile. I have seen some media queries on styled components that I haven't seem applied on the migrated version, so that might be the cause.

This branch:
image


Master:
image

(Screenshots have the same height, so you can see how much space is different in both versions)

DSil:
Also, forgot to mention, please consider changes on #3985

packages/orbit-components/src/Card/__tests__/index.test.tsx
```diff
@@ -66,7 +60,7 @@ describe("Card", () => {
 
     render(
       <Card>
-        <CardSection title="kek" onClick={onClick} />
+        <CardSection title="kek" expandable onClick={onClick} />
```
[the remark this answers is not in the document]
oreqizer: `onClick` shouldn't work when `CardSection` is non-expandable. components shouldn't have extraneous event handlers for best practice reasons, and because it breaks RSCs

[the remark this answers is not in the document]
DSil: I believe a Card can be clickable for some action (eg: route to another page).
Our tracking tool shows no usages of `onClick` on `CardSection`, though.

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -133,6 +144,12 @@ const cfg = (options?: Options): Config => {
           "country-flag-small": defaultTokens.countryFlagSmallWidth,
           "country-flag-medium": defaultTokens.countryFlagMediumWidth,
         },
+        minWidth: {
+          "icon-small": defaultTokens.widthIconSmall,
```
mainframev, line 148: those tokens are deprecated. We deprecated all tokens starting with width/height. Instead should be used: 
```
widthIconMedium -> iconMediumSize
widthIconLarge -> iconLargeSize
```

and so on, if there will be a mistake it should be fixed and added appropriate token in the same way as for others :)

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -125,6 +125,17 @@ const cfg = (options?: Options): Config => {
           "country-flag-small": defaultTokens.countryFlagSmallHeight,
           "country-flag-medium": defaultTokens.countryFlagMediumHeight,
         },
+        minHeight: {
+          "icon-small": defaultTokens.heightIconSmall,
+          "icon-medium": defaultTokens.heightIconMedium,
+          "icon-large": defaultTokens.heightIconLarge,
+          "input-normal": defaultTokens.heightInputNormal,
+          separator: defaultTokens.heightSeparator,
+          "button-small": defaultTokens.heightButtonSmall,
```
mainframev, line 134: those tokens are also deprecated, instead should be used: 
```
 heightButtonSmall -> formBoxSmallHeight
 heightButtonNormal -> formBoxNormalHeight
```

packages/orbit-components/src/Card/index.tsx
```diff
@@ -48,46 +27,34 @@ const Card = ({
...
-      // if (item.props.children && item.type.name !== "CardSection") {
-      //   return React.createElement(CardSection, {
-      //     ...item.props.children.props,
-      //     key: index,
-      //   });
-
-      return React.cloneElement(item);
+}: Props) {
+  const marginStyles = (() => {
+    if (margin == null) {
```
mainframev, line 32: let's not add curly braces if this can be one line return :) 
Not a big deal and more like a matter of personal preference, we do not have some eslint rules to enforce that and do not think we should, but this is how we get used to write in this project :)

packages/orbit-components/src/Card/components/Header.tsx
```diff
@@ -0,0 +1,93 @@
...
+import type * as Common from "../../common/types";
+import type { As } from "../../Heading/types";
+import ChevronDown from "../../icons/ChevronDown";
+import Stack from "../../Stack";
+import Heading from "../../Heading";
+import Text from "../../Text";
+import Close from "../../icons/Close";
+import ButtonLink from "../../ButtonLink";
+
+type Props = {
```
mainframev, line 13: I do not mind having it as a type, but we get used to defining those types with interfaces. Here is no intersection, so there is no difference in what to use, it just looks different and not like the majority in our project. So we either do this way or another. Maybe you could justify why you find that it's better, but for me, it just looks like you do it in the way you get used to :)

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -125,6 +125,17 @@ const cfg = (options?: Options): Config => {
           "country-flag-small": defaultTokens.countryFlagSmallHeight,
           "country-flag-medium": defaultTokens.countryFlagMediumHeight,
         },
+        minHeight: {
+          "icon-small": defaultTokens.heightIconSmall,
+          "icon-medium": defaultTokens.heightIconMedium,
+          "icon-large": defaultTokens.heightIconLarge,
+          "input-normal": defaultTokens.heightInputNormal,
```
mainframev, line 132:
``` suggestion
          "input-normal": defaultTokens.formBoxNormalHeight,
```

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -133,6 +144,12 @@ const cfg = (options?: Options): Config => {
           "country-flag-small": defaultTokens.countryFlagSmallWidth,
           "country-flag-medium": defaultTokens.countryFlagMediumWidth,
         },
+        minWidth: {
+          "icon-small": defaultTokens.widthIconSmall,
+          "icon-medium": defaultTokens.widthIconMedium,
+          "icon-large": defaultTokens.widthIconLarge,
+          checkbox: defaultTokens.widthCheckbox,
```
mainframev, line 151:
```suggestion
          checkbox: defaultTokens.iconSmallSize,
```

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -1,17 +1,15 @@
...
-import SectionHeader from "./components/SectionHeader";
-import SectionContent from "./components/SectionContent";
 import useRandomId from "../../hooks/useRandomId";
 import { ELEMENT_OPTIONS } from "../../Heading/consts";
 import type { Props } from "./types";
+import Header from "../components/Header";
+import Expandable from "./components/Expandable";
 
-const CardSection = ({
+export default function CardSection({
```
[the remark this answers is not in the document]
oreqizer: no besides that I tried to sneak this in since there's so many changes already 😅 

I prefer this because `function` declarations > arrow functions for non-binding function definitions. also, you can one-line `export default` here

packages/orbit-components/src/Card/components/Header.tsx
oreqizer: Orbit is the first project I have seen using `interface` instead of `type` 😅 `interface` provides no benefit and is more verbose. since there's no ESlint rule, I assumed both syntaxes are correct so I went with the more "terse" and more common one

packages/orbit-components/src/Card/index.tsx
mainframev: I would enforce that with multi-option, meaning, that it will add curlies in case it's multiline. We have used it here for years and no bugs were related to that.

packages/orbit-components/src/Card/components/Header.tsx
mainframev: I hope you do not mind if I will enforce it with [consistent-type definitions](https://typescript-eslint.io/rules/consistent-type-definitions/) in a separate request, wdyt? :) In order to avoid such kind of cases 🙂 

I feel like we should use interfaces for object kind of type definitions, it has also the advantage of throwing an error in case of a conflict, while type alias will just merge it. And use the type aliases for everything else.

packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
```diff
@@ -1,10 +1,31 @@
...
-import { getBorder, getBorderRadius } from "../../../Card/helpers/borders";
 import type { Props } from "./types";
 import defaultTheme from "../../../defaultTheme";
+import type { Theme, ThemeProps } from "../../../defaultTheme";
+import mq from "../../../utils/mediaQuery";
+
+export const getBorder = ({ theme }: ThemeProps): string =>
+  `${theme.orbit.borderWidthCard} ${theme.orbit.borderStyleCard} ${theme.orbit.borderColorCard}`;
+
+export const getBorderRadius = ({ theme }: ThemeProps): string =>
```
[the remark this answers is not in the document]
oreqizer: I copied them from `Card` and didn't want to touch em

packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
```diff
@@ -1,10 +1,31 @@
...
+  `${theme.orbit.borderRadiusNormal}`;
+
+const CardElement = css`
```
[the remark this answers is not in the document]
oreqizer: no because it is not a styled component, but a mixin

packages/orbit-components/src/Card/components/Header.tsx
```diff
@@ -0,0 +1,93 @@
...
+        color="secondary"
+      />
+    )}
+    {actions && (
+      <Stack inline grow={false} justify="end">
+        {actions}
+      </Stack>
+    )}
+    {onClose && !actions && (
+      <div className="absolute right-0 top-0 rtl:left-0">
```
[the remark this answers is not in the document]
oreqizer: adjusted, though I find `z-index: 1` to be redundant

packages/orbit-components/src/Card/index.tsx
oreqizer: added font-family. other styles are not needed

packages/orbit-components/src/Card/index.tsx
oreqizer: adjusted, thank, TIL

packages/orbit-components/src/Card/components/Header.tsx
oreqizer: aight let's do it in a separate request

packages/orbit-components/src/Card/index.tsx
oreqizer: let's solve it separately, though imo we should do `all` for maximum consistency and readability

packages/orbit-components/src/Card/__tests__/index.test.tsx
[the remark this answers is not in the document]
oreqizer: ah okay I misunderstood the purpose of `onClick` then. alright I will return `onClick` as it was before, though normally this would be solved by composition to also allow hrefs, for example:

```
<Card>
  <Link href={`/product/${id}`}>
    <CardSection title={title}>
      <Product id={id} />
    </CardSection>
  </Link>
</Card>
```

packages/orbit-components/src/Layout/LayoutColumn/index.tsx
```diff
@@ -18,14 +17,6 @@ const StyledColumn = styled.div<{ spanEntireRow?: boolean; hideOn?: Devices[] }>
     css`
       grid-column: 1 / -1;
     `};
-
-    @media (max-width: ${+getBreakpointWidth(QUERIES.LARGEMOBILE, theme, true) - 1}px) {
```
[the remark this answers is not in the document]
oreqizer: it is not needed, it was bcs of those weird negative margins in `CardSegment` that are no longer present

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -26,108 +24,85 @@ const CardSection = ({
...
-    <CardWrapper
-      dataTest={dataTest}
-      roundedTop={roundedBorders.top}
-      roundedBottom={roundedBorders.bottom}
-      expanded={opened}
-      onClick={onClick}
-      expandable={expandable}
-      noBorderTop={noBorderTop}
+    <div
+      className={cx(
```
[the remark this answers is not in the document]
oreqizer: adjusted

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -26,108 +24,85 @@ const CardSection = ({
...
+      {(title != null || header != null) && (
+        <button
+          type="button"
+          className={cx("p-lg w-full", expandable && "hover:bg-white-normal-hover cursor-pointer")}
+          aria-expanded={expandable ? opened : undefined}
+          aria-controls={expandable ? slideID : undefined}
+          disabled={expandable !== true}
+          onClick={handleClick}
+        >
+          <Header
```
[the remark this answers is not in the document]
oreqizer: it was never present here. it is used in `Card`, not `CardSection`

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -26,108 +24,85 @@ const CardSection = ({
...
-          expandable={expandable}
-          expanded={opened}
-          actions={actions}
-          isContent={!!children}
-          onClick={expandable ? handleClick : undefined}
-          description={description}
-          handleKeyDown={handleKeyDown}
-        />
+      {(title != null || header != null) && (
+        <button
```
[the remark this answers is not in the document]
oreqizer: adjusted

DSil (commented):
Small detail but the height of the component is slightly different. I don't know how it can affect complex layouts that might be using Card. I think the change is caused by the changes on the borders, but I was unable to find the real cause.

Master:
image

Here:
image

oreqizer:
@DSil the change in size is caused by removing the negative margins. it was +24 padding -25 margin, causing the component to be 1px smaller

I cannot match the component size without using custom sizings, which would be a bad practice, so accepted this slight change

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -125,13 +125,21 @@ const cfg = (options?: Options): Config => {
...
           "country-flag-medium": defaultTokens.countryFlagMediumHeight,
         },
-        minWidth: {
-          "dialog-width": defaultTokens.dialogWidth,
+        minHeight: {
+          "icon-small": defaultTokens.iconSmallSize,
+          "icon-medium": defaultTokens.iconMediumSize,
+          "icon-large": defaultTokens.iconLargeSize,
+          "input-normal": defaultTokens.formBoxNormalHeight,
+          separator: defaultTokens.heightSeparator,
```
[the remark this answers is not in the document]
mainframev: for now, can be just as a static value in the preset

mainframev:
Screenshot 2023-09-08 at 9 31 39
Screenshot 2023-09-08 at 9 31 42

There are visual differences in RTL. I guess it should be the same as in the current master, now RTL'd properly only titles

packages/orbit-components/src/Card/index.tsx
```diff
@@ -99,52 +64,16 @@ const Card = ({
...
-            const topRoundedBorder =
-              expandedSections.indexOf(key - 1) !== -1 || expandedSections.indexOf(key) !== -1;
-            const bottomRounderBorder =
-              expandedSections.indexOf(key + 1) !== -1 || expandedSections.indexOf(key) !== -1;
-
-            // This is used for the case when user wants to map sections and change their order
-            // related issue: #1005
-            // @ts-expect-error TODO
-            const index = Number(item.key) || key;
+      {!loading && children}
```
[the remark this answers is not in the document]
oreqizer: do you imagine `loading ? null : children` rather than this?

oreqizer:
the RTL changes are not due to this MR, it is a part of `Text`

packages/orbit-tailwind-preset/src/presets/components/index.ts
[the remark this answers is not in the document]
oreqizer: thanks, adjusted. I somehow badly rebased it and the changes disappeared

packages/orbit-components/src/Card/index.tsx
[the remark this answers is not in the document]
mainframev: no no, I mean turning a loading condition into ternarny below: 

```
  {loading && (
        <div className="lm:border-x border-b">
          <Loading loading={loading} type="boxLoader" />
        </div>
      )}
```

into: 

```
  {loading ? (
        <div className="lm:border-x border-b">
          <Loading loading={loading} type="boxLoader" />
        </div>
      ) : children}
```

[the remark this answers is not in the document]
mainframev: and the condition `{!loading && children}` can be removed after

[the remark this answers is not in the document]
oreqizer: ah okay. I made it like this for two reasons:
1. the ` : children` part is visually buried because of `prettier` formatting
2. this way it is much easier to extend conditions as necessary, e.g. `loading && loadingDebounce && <div className=...">`

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -26,108 +23,102 @@ const CardSection = ({
...
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+    <div
+      className={cx(
+        "duration-fast lm:border-x border-b transition-all ease-in-out",
+        opened && "my-xs rounded-normal shadow-action-active [&+*]:border-t",
+        onClick != null && "hover:bg-white-normal-hover cursor-pointer",
+      )}
+      data-test={dataTest}
+      role={onClick == null ? undefined : "button"}
+      tabIndex={onClick == null ? undefined : -1}
```
mainframev, line 62: Sorry, do not want to block and bother again, but this seems broken 😞  It's having onClick it will have tabIndex -1? I guess it should be focusable, otherwise, tab navigation is broken (in master it's focusable and works correctly)

mainframev (changes requested):

packages/orbit-components/src/Card/CardSection/index.tsx
oreqizer: https://kiwicom.github.io/orbit/?path=/story/card--card-with-sections can you focus the first one?

oreqizer: ah okay I originally thought it is jsut for event bubbling. sec I will adjust

mainframev: the value of tabIndex should be `0`, so the order is natural ;)

mainframev: also still missing the onKeyDown: 

```
onKeyDown={handleKeyDown(onClick)}
```

the same as in [Tile component](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Tile/index.tsx#L59) ;)

oreqizer: fixed

mainframev (approved):

DSil (approved):

PR #3976, opened 2023-08-25
docs: extend contributing.md about tokens addition

mainframev:
Just quickly added a few sentences about how to add tokens. Might be useful to have it in contributing guidelines if someone will struggle with adding new tokens, it refers to SCHEMA file and add basic info about how to add them.

Storybook: https://orbit-mainframev-docs-extend-contribution-guidelines.surge.sh

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -152,7 +152,7 @@ Refer to the token schema documentation for a comprehensive understanding of tok
 
 - Categorize Tokens: Organize your tokens under appropriate categories, either globally or component-specific. This structure ensures clarity and maintainability.
 
-- Generate Tokens: Once your token schema is ready, execute command:
+Generate Tokens: Once your token schema is ready, execute the `yarn tokens build` command. This triggers style-dictionary to transform your definitions into a usable output.
 
 ```bash
 yarn tokens build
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -119,6 +120,52 @@ yalc push
...
+
+Refer to the token schema documentation for a comprehensive understanding of token properties.
+
+- Categorize Tokens: Organize your tokens under appropriate categories, either globally or component-specific. This structure ensures clarity and maintainability.
+
+- Generate Tokens: Once your token schema is ready, execute command:
+
+```bash
+yarn tokens build
+```
```
DSil, line 159: Although I don't disagree with what this paragraph says, it feels a bit distracting from the objectiveness that I expect from a technical documentation 😄 I won't block this, at all, but it feels weird to me. Team's decision is fine to me

DSil (approved):
Just two non-blocking suggestions 👍

.github/CONTRIBUTING.md
DSil: This should be on line 165, the second to last paragraph (weird Github mobile app)

mainframev: > Although I don't disagree with what this paragraph says, it feels a bit distracting from the objectiveness that I expect from a technical documentation 😄 I won't block this, at all, but it feels weird to me. Team's decision is fine to me

Ofc it can be always improved, I am not a professional technial writer and particaulary in that case I prefer to have something done on that front rather do not have anything at all ;)

PR #3981, opened 2023-08-30
Fix Tailwind config and OrbitProvider children

DSil:
This PR addresses three things:
- The typo in the tailwind.config.ts inside `orbit-components`
- ~The brokens examples caused by #3979 because OrbitProvider expects a single child~ This was replaced by allowing OrbitProvider to receive multiple children
- The tailwind config does not include pre-flight for testing 

With this, there are possible future work:
- Type checks should run for the tailwind.config files as well
- ~The type of prop `children` on `OrbitProvider` should ensure it only gets a single child.~
 Storybook: https://orbit-mainframev-fix-tw-config-and-docs.surge.sh

### DIFF
--- a/packages/orbit-components/tailwind.config.ts
+++ b/packages/orbit-components/tailwind.config.ts
@@ -6,7 +6,7 @@ export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [
     orbitPreset({
-      disabledPreflight: false,
+      disablePreflight: false,
     }),
   ],
 } satisfies Config;
--- a/packages/orbit-components/tailwind.config.ts
+++ b/packages/orbit-components/tailwind.config.ts
@@ -6,7 +6,7 @@ export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [
     orbitPreset({
-      disablePreflight: false,
+      disablePreflight: process.env.NODE_ENV === "test",
     }),
   ],
 } satisfies Config;
--- a/packages/orbit-components/package.json
+++ b/packages/orbit-components/package.json
@@ -15,7 +15,7 @@
     "cy:run": "start-server-and-test cy:dev http-get://localhost:3000 'cypress run'",
     "cy:open": "start-server-and-test cy:dev http-get://localhost:3000 'cypress open'",
     "build": "ts-node --esm config/build/index.mts",
-    "tailwind": "npx tailwindcss -i ./src/tailwind.css -o ./lib/tailwind.css",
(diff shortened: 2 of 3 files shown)


mvidalgarcia:
> The type of prop `children` on `OrbitProvider` should ensure it only gets a single child.

@oreqizer do you have any idea how to solve this? We've been investigating but we haven't found anything helpful.

mvidalgarcia (approved):

oreqizer:
> > The type of prop `children` on `OrbitProvider` should ensure it only gets a single child.
> 
> @oreqizer do you have any idea how to solve this? We've been investigating but we haven't found anything helpful.

the type should be `React.ReactNode | React.ReactNode[]`. it has no reason for it not to be an array

DSil:
> the type should be React.ReactNode | React.ReactNode[]. it has no reason for it not to be an array

@oreqizer It expects a single child. If it has multiple children it throws an error
> Error: React.Children.only expected to receive a single React element child.

packages/orbit-components/tailwind.config.ts
```diff
@@ -6,7 +6,7 @@ export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [
     orbitPreset({
-      disablePreflight: false,
+      disablePreflight: process.env.NODE_ENV === "test",
```
mvidalgarcia, line 9: Won't this always be `true` for us? Does it mean we change the behavior we had?

DSil: When running storybook, the NODE_ENV is not `test` so this is false, and therefore preflight is enabled.
When we execute `yarn components tailwind`, that should be executed before testing, the NODE_ENV is set to `test` and so preflight will be disabled

oreqizer:
> > the type should be React.ReactNode | React.ReactNode[]. it has no reason for it not to be an array
> 
> @oreqizer It expects a single child. If it has multiple children it throws an error. So that's why we would like TS to avoid the possibility to have multiple children
> 
> > Error: React.Children.only expected to receive a single React element child.

🤔 why are we doing `React.Children.only` then? let me check code

oreqizer:
alright I propose that, instead of wrapping things in a fragment, we remove `React.Children.only` from `OrbitProvider`, and adjusting types as I proposed. it literally serves no purpose there

DSil:
@oreqizer I have removed the single child restriction, but having `React.ReactNode` as children already accepts multiple children. So why do we need to have the type defined as `React.ReactNode | React.ReactNode[]`?

oreqizer:
does it? 🤔 I thought it doesn't... then okay, no need to change types!

oreqizer (approved):

PR #3982, opened 2023-08-31
fix(Text): default value for align is now start

DSil:
With Tailwind, default value can't be `left`, but should be `start`, so that it is correctly aligned in RTL.
 Storybook: https://orbit-mainframev-fix-text-alignment-default.surge.sh

### DIFF
--- a/packages/orbit-components/src/Text/Text.stories.tsx
+++ b/packages/orbit-components/src/Text/Text.stories.tsx
@@ -139,11 +139,15 @@ Playground.story = {
   },
 };
 
-export const Rtl = () => (
-  <RenderInRtl>
-    <Text align="left">Lorem ipsum dolor sit amet</Text>
-  </RenderInRtl>
-);
+export const Rtl = () => {
+  const align = select("Align", Object.values(ALIGN_OPTIONS), ALIGN_OPTIONS.START);
+
+  return (
+    <RenderInRtl>
+      <Text align={align}>Lorem ipsum dolor sit amet</Text>
+    </RenderInRtl>
+  );
+};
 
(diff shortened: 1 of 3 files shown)


mainframev:
@DSil Just noticed, that RTL in master in Text story and Text behaves differently

DSil:
@mainframev well spotted. It was actually the story messing up, because it was explicitly applying `align="left"`. I have changed it to use a knob for alignment and defaulted the value to `"center"`

mainframev:
> @mainframev well spotted. It was actually the story messing up, because it was explicitly applying `align="left"`. I have changed it to use a knob for alignment and defaulted the value to `"center"`

good job 👍🏻  I think it's still missing the rules like rtl:right (when it's aligned left) 🤔 

The best will be ofc to left only start and end, but since we already had them and removing them will be breaking, I think it's neccesseray to add

DSil:
That's a fair point. But if we do that we are removing the possibility of forcing text to be left, even in RTL. I think we should not be that limitative.
Up until now, the `align="left"` was being used expecting the alignment to be at start. And if we change it, it's a breaking change, yes. However, I believe that in the vast majority of the cases, the `align` prop is not even defined, defaulting to "left", and now will default to "start", which is no breaking change for those usages.
Tracking tool shows 4 usages of value "left", 12 usages of value "right" and 3 conditional expressions, over a universe of 1300+ instances. So yes, it should be documented as a breaking change, but the impact is minimal. And I don't think we should apply logic to force "left" to behave as "start".

mainframev:
> 

Okay, just wanted to pointed that out, agree with that 👍🏻

PR #3985, opened 2023-09-01
fix(Card): header actions are now vertically center aligned

DSil:
As reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1693468169219649), actions should always be vertically center aligned.

@oreqizer consider this change on the TW migration as well 🙏 
 Storybook: https://orbit-mainframev-fix-card-actions-alignment.surge.sh


mvidalgarcia (approved):
LGTM

image

PR #3989, opened 2023-09-04
docs: use correct hook in react@17 example

icaromh:
This Pull Request aims to fix the migration guide for orbit-components@v9.
In the current docs, it's suggested to use `useRandomId` but this hook is not part of `react-uid` package.

While in the component description, it is correctly stated to use `useUID` instead. 
https://github.com/kiwicom/orbit/blob/015f84a9a74f0b5f05de30f4b747158d1126b249/packages/orbit-components/src/OrbitProvider/index.tsx#L12-L21


DSil (approved):
Thank you!

PR #3991, opened 2023-09-04
chore: eslint package changes

mainframev:
I marked SC-related rules as deprecated because we won't need them in the form they currently exist.

Also migrated the package from @babel/types and @babel/eslint-parser to @typescript-eslint/parser and its utils, we had a lot of suppressed errors because the typings were not correct. Before we used types from eslint directly and it's using by default espree, so with babel-types it had conflicts. Now it's properly typed and has the advantage of using the utility functions from @typescript-eslint that help to build the rules faster and in a more convenient way

Also added an internal rule called 'use-client-directive' in order to prevent missing 'use client' directives in our components

Temporary disabled the rules after the *@typescript-eslint* update: 
- "@typescript-eslint/no-explicit-any": "error",
- "@typescript-eslint/consistent-type-imports": "error"

Suppressed the unused vars where we destructuring the props win SC, could be done in a different way, but since we still migrating to Tailwind -> it does not matter imo. 
 
The "any" usage should be avoided, but this rule was missing for ts files, and with the update, it started to be a part of the recommended config, so I just set it ti warn and it should be handled in a separate request. 

consistent-type-imports has the issue after the update with typescript (added the link and todo comment), so also disabled for now.

Storybook: https://orbit-mainframev-chore-depercate-eslint-rules.surge.sh


packages/orbit-components/.eslintrc.cjs
```diff
@@ -26,12 +26,13 @@ module.exports = {
...
-      files: ["src/**/*.ts?(x)"],
-      rules: {
-        "@typescript-eslint/consistent-type-imports": "error",
-      },
-    },
+    // TODO: temporary disabled because of the issue: https://github.com/typescript-eslint/typescript-eslint/issues/4608
+    // {
+    //   files: ["src/**/*.ts?(x)"],
+    //   rules: {
+    //     "@typescript-eslint/consistent-type-imports": "warn",
```
DSil, line 33: It's unfortunate that we lose this rule tbh. Just to be clear, this became buggy after the update? And was never fixed?

packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
```diff
@@ -3,6 +3,7 @@ import { select, text } from "@storybook/addon-knobs";
 
 // @ts-expect-error TODO
 import { NAMES } from "./consts.mts";
```
DSil, line 5: Removing the comment triggers the error
`An import path can only end with a '.mts' extension when 'allowImportingTsExtensions' is enabled.` but then why don't we just enable this rule?

packages/orbit-components/config/fetchIcons.mts
```diff
@@ -38,6 +38,10 @@ interface FigmaComponents {
   i18n: null;
 }
 
+interface Images {
```
DSil, line 41: Just to clarify, any reason for being interface and not type?

packages/orbit-components/src/InputSelect/helpers.ts
```diff
@@ -1,13 +1,17 @@
 import type { Option } from "./types";
 
+type Groups = {
```
DSil, line 3: Same question as before, but opposite 😄 any reason for being type and not interface?

packages/orbit-components/src/Radio/index.tsx
```diff
@@ -166,6 +166,7 @@ StyledInput.defaultProps = {
   theme: defaultTheme,
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
DSil, line 169: Instead of this, can't we add the `_` to the unused vars? Same for other components

packages/orbit-components/src/Stack/Stack.stories.tsx
```diff
@@ -206,7 +206,7 @@ export const TabletProperties = () => {
   };
 
   return (
-    <Stack tablet={tablet}>
+    <Stack tablet={tablet} spaceAfter={spaceAfter}>
```
DSil, line 209: This does not belong here

mainframev: It's just had unused spaceAfter knob ;)

packages/orbit-components/cypress/integration/index.tsx
```diff
@@ -58,6 +58,6 @@ function App() {
 }
 
 const container = document.getElementById("app");
-// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-const root = createRoot(container!); // createRoot(container!) if you use TypeScript
+// @ts-expect-error TODO
+const root = createRoot(container); // createRoot(container!) if you use TypeScript
```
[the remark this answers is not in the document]
DSil: At the same time, I see the no-null-assertion verification is no longer working. Do we want that?

packages/eslint-plugin-orbit-components/src/__fixtures__/file.ts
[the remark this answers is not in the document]
mainframev: yep, it's a part of [how @typescript-eslint/parser should be setup :) ](https://typescript-eslint.io/packages/rule-tester#type-aware-testing)

packages/orbit-components/.eslintrc.cjs
mainframev: As far as I know it's still not fixed, I will check more on that 🤔

packages/orbit-components/src/Stack/Stack.stories.tsx
DSil: It was before the `spaceAfter` was added to the `tablet` object defined above, as it was missing and we added a few days ago 😄 so we can (and should) remove it from here, as all other media query stories don't feature it

packages/orbit-components/config/fetchIcons.mts
mainframev: we decided to enforce interfaces, so I will keep that :)

packages/orbit-components/src/InputSelect/helpers.ts
mainframev: we decided to enforce interfaces, so I will rewrite that :)

packages/orbit-components/src/Radio/index.tsx
mainframev: yep, it can, it does not make sense to me as it still soon all will be rewritten to Tailwind

DSil: Alright, non-blocking

packages/eslint-plugin-orbit-components/src/__tests__/useClient.test.ts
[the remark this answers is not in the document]
mainframev: I will check why this is not working, probably because eslint config is not applied correctly

packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
mainframev: Good point. I think when I was adding mts -> typescript did not support yet that. [It appeared quite recently ](https://github.com/microsoft/TypeScript/issues/53316) :)

packages/orbit-components/src/Stack/Stack.stories.tsx
```diff
@@ -190,7 +190,6 @@ export const TabletProperties = () => {
   const align = select("Align", Object.values(ALIGNS), ALIGNS.END);
   const justify = select("Justify", Object.values(JUSTIFY), undefined);
   const spacing = select("Spacing", Object.values(SPACINGS), undefined);
-  const spaceAfter = select("spaceAfter", Object.values(SPACINGS_AFTER), undefined);
```
DSil, line 193: This is to stay, just like the other deletion. What was not to add was the prop to `<Stack>`. This file is to be left untouched in this PR

.eslintrc.cjs
```diff
@@ -111,15 +104,31 @@ module.exports = {
...
         "react/function-component-definition": "off",
         "react/no-unknown-property": ["error", { ignore: ["css"] }],
         "@typescript-eslint/no-unused-expressions": "error",
         "no-use-before-define": "off", // disable conflicting rule
-        "@typescript-eslint/no-use-before-define": "error",
+        "@typescript-eslint/no-use-before-define": [
+          "error",
+          {
+            typedefs: false,
+            ignoreTypeReferences: true,
```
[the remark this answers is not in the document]
mainframev: I think I was testing that and forgot to remove 👍🏻

DSil (commented):
Please rebase and resolve conflicts so I can do the final review 🙏 it seems almost good to go

mainframev:
> Please rebase and resolve conflicts so I can do the final review 🙏 it seems almost good to go

Yep, working on it, I plan to check first unresolved issue with applying the rule to the codebase

DSil (commented):
LGTM, just resolve the conflict and rebase 🙏

DSil (approved):

PR #3992, opened 2023-09-05
Add defaultValue to input and Textarea

DSil:
As determined [here](https://skypicker.slack.com/archives/C05FDP4UM3N/p1693836772851779), `defaultValue` is needed in InputField. Took the chance to also add it to Textarea.
 Storybook: https://orbit-mainframev-feat-add-defaultValue-input-textarea.surge.sh


packages/orbit-components/src/InputField/index.tsx
```diff
@@ -433,6 +434,7 @@ const InputField = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
           name={name}
           type={getDOMType(type)}
           value={value}
+          defaultValue={defaultValue}
```
RobinCsl, line 437: Given what you wrote above, should we alter the logic to pass either `value` or `defaultValue` and not both at the same time?

DSil: I have had that question in my life in the past and ended up concluding that the responsibility of knowing when to use `value` + `onChange` or `defaultValue` (controlled vs uncontrolled) is the developer's.

The console will show a warning in dev mode if someone tries to use both at the same time.
image

So I would prefer not to add that restrictive logic to our component.

RobinCsl: But from the typing system point of view, it's still wrong since these are incompatible, right?

IMO, static analysis tooling should help developers not rely on runtime warnings/errors.

DSil: I tried adding conditional types, but honestly… I couldn't. I could define the type (I had to convert some interfaces to types, but all should be fine), but then passing the props to this `Input` component started throwing me errors because of type incompatibility. I might be missing something 😓

oreqizer: I wouldn't bother. ideally we'd not specify these types **at all**, we'd just extend `React.InputHTMLAttributes<HTMLInputElement>` with any custom props we have, such as `error`, and be done with it

DSil: As agreed irl, we won't add conditional types

packages/orbit-components/src/Textarea/__tests__/index.test.tsx
```diff
@@ -54,20 +51,23 @@ describe("Textarea", () => {
     expect(textarea).not.toBeInvalid();
     expect(textarea.parentElement).toHaveStyle({ marginBottom: "12px" });
     expect(textarea).toHaveStyle({ padding: "12px" });
-
-    await act(() => user.type(textarea, "kek"));
-    expect(onChange).toHaveBeenCalled();
   });
 
-  it("should have focus and blur", async () => {
```
oreqizer, line 62: out of scope, but this is a prime example of a redundant test 😅 normally we'd pass native props to the `textarea` element, and we really don't need to test whether `textarea` works as a `textarea`. we'd maybe have one user-like test to check that you can type stuff in there and that it reacts to changes, that's it

janpanschab:
@DSil @RobinCsl would be able to add `defaultChecked` to the checkbox and radio inputs? To cover all cases and be able to use uncontrolled inputs fully.

packages/orbit-components/src/Textarea/__tests__/index.test.tsx
DSil: Yes… kinda. Because our `Textarea` component does not simply render a `<textarea>`. It renders a label, a span, a textarea (through `styled.textarea`) and possibly even a tooltip with more stuff inside.
I fully understand what you mean, but I think we are just asserting that callbacks are not lost in the middle of the inner logic of the component and that they are passed to the correct element.
So I don't think the problem is the test, but the implementation that makes the test reasonably justifiable. But, as you said, that is (even more) out of scope of this PR. Let's address this soon™ 🤞

DSil:
@janpanschab as agreed IRL, we will address that on a separate PR later. Thank you for raising awareness!

RobinCsl (approved):

PR #3996, opened 2023-09-06
chore: enforce interfaces for object-like types

mainframev:
[as it was discussed here ](https://github.com/kiwicom/orbit/pull/3975#discussion_r1315770132)-> enforcing it for consistency 
 Storybook: https://orbit-mainframev-chore-enforce-eslint-rules.surge.sh

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -21,10 +21,7 @@ module.exports = {
       "error",
       {
         groups: [["builtin", "external"], ["parent", "sibling"], "index"],
-        pathGroups: [
-          { pattern: "@kiwicom/**", group: "external" },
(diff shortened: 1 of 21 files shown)


DSil:
This is better to be reviewed after #3991 (or have the branch rebased on top of it) as I see some duplicated and/or conflicting code

mainframev:
> This is better to be reviewed after #3991 (or have the branch rebased on top of it) as I see some duplicated and/or conflicting code

I think it's not stopping from the code review, just skip the part of this change with `no-unused-vars` addititional params, I will remove the similar to the mentioned PR

DSil (changes requested):
Overall 👍 just remove that file and fix pipeline 🙏

DSil (approved):

PR #3997, opened 2023-09-06
chore: automate package release

mvidalgarcia:
- Create a GitHub Action workflow to release packages manually

Closes FEPLT-1675


mvidalgarcia:
@mainframev as the one with most experience making releases, I would like you to take a look at this PR 🙏 

I'm sure many things will break in the first attempts, but I can't find any other way to test it, perhaps you have a better idea?

mainframev:
> I'm sure many things will break in the first attempts, but I can't find any other way to test it, perhaps you h

That's what I am afraid of. Unfortunately, if something fails (as I mentioned before), let's say it will fail and commits will be pushed, but no npm and GitHub release will happen, which would require us to do that manually and maybe even rewrite the commit history. 

What if we create a completely empty project with some dummy packages (or we can even fork Orbit and change the package names), apply our actions there, and test it? A safe place for running that kind of thing without touching our real project, wdyt?

mvidalgarcia:
> or we can even fork Orbit and change the package names

That's what I tried at the beginning but I faced many problems with the build and types and couldn't make it work. You can see multiple failures I experienced [here](https://github.com/mvidalgarcia/orbit/actions?query=is%3Afailure). We can try again, you may have a better idea.

> What if we create a completely empty project with some dummy packages

That's probably easier, we can do that.

lerna.json
```diff
@@ -3,13 +3,16 @@
   "version": "independent",
   "npmClient": "yarn",
   "useWorkspaces": true,
```
mainframev, line 5: I think as a part of this request, we can also raise the version of Lerna. As #3899 probably won't be merged in the near future. 
Upgrade to v6 is non-breaking. From v6 to v7 commands: *add*, *link*, *bootstrap* are dropped. 
and *useWorkspaces* too. Running *lerna repair* on v7 should remove this flag. [Here is the changelog for v7](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#700-2023-06-08)

mainframev: It also include some fixes for publish package, which is good for us

mainframev:
Also, the thing that comes to my mind. Instead of having an action that will run the Lerna script, the best would be to have the action that will create a release PR for us, that we can walk through again, see the changelog changes and versions bumped, and after merging this to master will happen the release. That would be probably the best, but I do not mind starting with something smaller if it will work. Maybe it's worth it for us to take a look at existing github actions that do smth like I described above ;)

mvidalgarcia:
@mainframev Thanks for your comments

> That would be probably the best, but I do not mind starting with something smaller if it will work

Let's start simple as you say. In frontend/core we don't have a setup to create release MRs and it has been working perfectly fine. I think it's a matter of having everything properly configured.

I already have a dummy repo with the publish job configured. It's working as expected and both creating GitHub releases and publishing npm versions.

https://github.com/mvidalgarcia/dummy-monorepo

I'll propagate the changes to Orbit now.

.github/workflows/publish.yml
```diff
@@ -6,5 +6,18 @@ jobs:
...
+
+      - name: Publish
+        env:
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          SLACK_API_POST_RELEASE_MESSAGE: ${{ secrets.SLACK_API_POST_RELEASE_MESSAGE }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" >> .npmrc
+          echo "hello publish job!"
+          yarn publish-packages
```
mainframev, line 23: Just for safety reasons, since we keeping `publish.mjs` file, I suggest using here in action directly `lerna publish` command as it will be needed to run it with `--yes` flag to have autoprompt on a question about the release, which we do not want to have in publish.mjs file if we would like to keep it :)

mainframev: can be abstracted from the publish mjs and included here as "zx postChangelogToSlack" smth like that :)

mainframev: it's still not posting correctly the version of the package, so maybe you can fix also this 🤔

scripts/post-changelog.mjs
```diff
@@ -0,0 +1,167 @@
...
+import dotenv from "dotenv-safe";
+import conventionalChangelog from "conventional-changelog";
+import { getPackages } from "@lerna/project";
+import slackify from "slackify-markdown";
+import { simpleGit } from "simple-git";
+import gitDiffParser from "gitdiff-parser";
+
+/* eslint-disable no-console */
+
+// const CHANNEL = `orbit-react`;
```
mvidalgarcia, line 11: Fix me

.github/workflows/publish.yml
```diff
@@ -5,6 +5,32 @@ on: workflow_dispatch
...
+        with:
+          fetch-depth: 0 # fetch tags
+
+      - name: Node
+        uses: ./.github/actions/node
+
+      - name: Setup git config
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
```
mvidalgarcia, line 22: We can create a bot account if we want to keep track of release commits.

lerna.json
```diff
@@ -6,12 +6,12 @@
   "command": {
     "version": {
       "message": "chore: publish",
-      "allowBranch": "master",
+      "allowBranch": ["master", "marco/automate-release"],
```
mvidalgarcia, line 9: remove before merging

.github/workflows/publish.yml
```diff
@@ -1,10 +1,51 @@
...
+        if: ${{ github.event.inputs.dryrun == 'false' }}
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "scope=@kiwicom" > ~/.npmrc
+          echo "access=public" >> ~/.npmrc
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+          echo "//registry.npmjs.org/:always-auth=true" >> ~/.npmrc
+          yarn lerna publish --no-private --conventional-commits --create-release github
```
mainframev, line 48:
```suggestion
          yarn lerna publish --no-private --conventional-commits --create-release github --yes
```

mvidalgarcia: Oh yes, forgot to add it back. Thanks!

mainframev (approved):
LGTM, good job on that 👍🏻 

Before merging, I guess you will remove the "delete me" commit and change the path of the slack channel to "orbit-react", right? :)

mvidalgarcia:
@mainframev Yes, I left it like that in case you wanted to try the dry run.

Removing [this commit](https://github.com/kiwicom/orbit/pull/3997/commits/f20ac760b3564f38ea496ae35f857a0d0aa14a36), should be enough.

.github/CONTRIBUTING.md
```diff
@@ -212,6 +213,12 @@ Check [this section](contribution/component-design.md) for a complete descriptio
 
 Please read our [guidelines for authoring commits](contribution/commits.md).
 
+## Publish releases
```
DSil, line 216: Should we include here that the NPM_TOKEN should be defined on the env file?

lerna.json
```diff
@@ -2,14 +2,16 @@
...
-  "packages": [
-    "packages/*",
-    "docs"
-  ],
+  "packages": ["packages/*", "docs"],
   "command": {
     "version": {
-      "message": "chore: publish"
+      "message": "chore: publish",
+      "allowBranch": "master",
```
DSil, line 9: If we consider to release RC/pre-releases from different branches in the future we might need to remove this rule, no?

.github/CONTRIBUTING.md
mvidalgarcia: It's defined in the repo [secrets](https://github.com/kiwicom/orbit/settings/secrets/actions). I think we're using @mainframev's one for now. We can consider creating an NPM service account in the future.

lerna.json
mvidalgarcia: Yes, not remove but adapt. We could have something like `["main", "pre/*", "rc/*"]`.

mainframev:
@mvidalgarcia can be squashed and merged, I guess

mvidalgarcia:
@mainframev squashed, we can go ahead.

PR #4001, opened 2023-09-07
feat(Breadcrumbs): migrate Breadcrumbs to Tailwind

mainframev:
[FEPL-1738](https://kiwicom.atlassian.net/browse/FEPLT-1738)
 Storybook: https://orbit-mainframev-tw-migrate-breadcrumbs.surge.sh

### DIFF
--- a/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.tsx
+++ b/packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.tsx
@@ -1,63 +1,48 @@
 "use client";
 
 import * as React from "react";
-import styled, { css } from "styled-components";
+import cx from "clsx";
 
-import defaultTheme from "../../defaultTheme";
-import ChevronForward from "../../icons/ChevronForward";
 import type { Props } from "./types";
+import ChevronForward from "../../icons/ChevronForward";
 
-const StyledBreadcrumbsItem = styled.li`
-  display: flex;
-  align-items: center;
-`;
+const Component = ({
+  href,
+  onClick,
+  component: Comp,
+  children,
+  ...props
+}: Props &
+  Pick<
+    React.HTMLAttributes<HTMLDivElement | HTMLAnchorElement | HTMLButtonElement>,
+    "itemScope" | "itemType" | "itemProp" | "itemID" | "className"
+  >) => {
+  const isClickable = onClick != null || href != null;
+  if (!isClickable) return <div {...props}>{children}</div>;
 
-const StyledBreadcrumbsItemAnchor = styled(({ component, children, isClickable, ...props }) => {
-  const Component = isClickable ? component : "div";
-  return <Component {...props}>{children}</Component>;
-})`
-  ${({ theme, $active, isClickable }) => css`
-    font-weight: ${$active ? theme.orbit.fontWeightBold : theme.orbit.fontWeightMedium};
-    color: ${theme.orbit.paletteInkDark};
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/Breadcrumbs/index.tsx
```diff
@@ -1,78 +1,52 @@
...
-                  active: key === React.Children.count(children) - 1,
-                  contentKey: key + 1,
-                });
-              }
-              return null;
+              if (!React.isValidElement(item)) return null;
+              return React.cloneElement<BreadcrumbsItemProps>(item, {
+                active: key === React.Children.count(children) - 1,
+                contentKey: key + 1,
+              });
```
DSil, line 43: Non-blocking, but it could even been a ternary, no?

packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.tsx
```diff
@@ -68,32 +45,53 @@ const BreadcrumbsItem = ({
...
       >
         <span itemProp="name">{children}</span>
-      </StyledBreadcrumbsItemAnchor>
+      </Component>
+
       <meta itemProp="position" content={String(contentKey)} />
-      {!active && (
-        <StyledBreadcrumbsItemIcon ariaHidden reverseOnRtl size="small" color="tertiary" />
+
+      {active !== true && (
```
DSil, line 85: If `active=false` by default, can't this be kept as `!active`? Are we moving away from these kind of assertions for any reason?

packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.tsx
```diff
@@ -1,63 +1,40 @@
...
+  children,
+  ...props
+}: Props &
+  Pick<
+    React.HTMLAttributes<HTMLDivElement | HTMLAnchorElement | HTMLButtonElement>,
+    "itemScope" | "itemType" | "itemProp" | "itemID" | "className"
+  >) => {
+  if (onClick == null || href == null) return <div {...props}>{children}</div>;
+  if (Comp != null) return <Comp {...props}>{children}</Comp>;
+  if (href == null)
```
DSil, line 22: Line 22 is unreachable, no? If `href == null` it will return the `div` on line 20 and never reach line 22

packages/orbit-components/src/Breadcrumbs/index.tsx
mainframev: I prefer to have it like that instead of ternary :)

DSil (changes requested):

packages/orbit-components/src/Breadcrumbs/BreadcrumbsItem/index.tsx
```diff
@@ -1,63 +1,48 @@
...
+    return (
+      <button type="button" {...props} onClick={onClick}>
+        {children}
+      </button>
+    );
 
-StyledBreadcrumbsItemIcon.defaultProps = {
-  theme: defaultTheme,
+  return (
+    <a href={href} {...props}>
```
DSil, line 38: I think we should pass onClick here as well. An anchor can have both, right? 🤔

mainframev: yes, you are right 👍🏻

DSil (approved):

PR #4003, opened 2023-09-08
Fix text alignment default

DSil:
Saves #3982 
 Storybook: https://orbit-mainframev-fix-text-alignment-default.surge.sh


DSil:
Merging as it was approved in #3982

PR #4007, opened 2023-09-08
Adjust TextLink + Alert styles

DSil:
A TextLink inside an Alert should have different styles, based on its own type and also the Alert's type. With TW migration that logic was lost. This gets it back.
 Storybook: https://orbit-mainframev-adjust-textlink-alert-styles.surge.sh


oreqizer (approved):
lgtm 👍

PR #4015, opened 2023-09-12
Button to Tailwind

oreqizer:
[WEBC-1679](https://kiwicom.atlassian.net/browse/FEPLT-1679)
 Storybook: https://orbit-mainframev-oreqizer-tw-button.surge.sh


packages/orbit-components/src/Button/index.tsx
```diff
@@ -1,35 +1,89 @@
...
+  criticalSubtle:
+    "bg-button-critical-subtle-background hover:bg-button-critical-subtle-background-hover active:bg-button-critical-subtle-active-background disabled:bg-button-critical-subtle-background focus:bg-button-critical-subtle-background-focus text-button-critical-subtle-foreground focus:text-button-critical-subtle-foreground-focus active:text-button-critical-subtle-foreground-active hover:text-button-critical-subtle-foreground-hover disabled:text-button-critical-subtle-foreground focus:shadow-product-normal/50 active:shadow-button-active-pale",
+  white:
+    "bg-button-white-background hover:bg-button-white-background-hover active:bg-button-white-background-active disabled:bg-button-white-background focus:bg-button-white-background-focus text-button-white-foreground focus:text-button-white-foreground-focus active:text-button-white-foreground-active hover:text-button-white-foreground-hover disabled:text-button-white-foreground focus:shadow-red-normal/50 active:shadow-button-active-pale",
+  bundleBasic:
```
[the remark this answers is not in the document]
DSil: It is applying `background-color` css property, but it has to be `background-image` because that is how gradients are applied to background. We have two options, the correct/difficult and the quickest but hacky/buggy/quickest: 

**Correct/difficult**: on the tokens definition and tailwind config, bundle tokens should not be treated as colors, but as images 🤷‍♂️ they won't work as colors anywhere else anyway, so having them defined as that is just wrong

**Hacky/buggy/quickest**: classes for bundle buttons must be `[background-image:button-bundle-basic-background]` (and so on)

I very much prefer the first one, since we're not on a rush for this component and it's an issue we need to address anytime soon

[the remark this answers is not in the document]
oreqizer: I had a third idea - create two separate tokens, gradient from and gradient to, for each tier. then use this with the native Tailwind gradient API

do you think this'd work?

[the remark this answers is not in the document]
mainframev: @oreqizer We have available in the palette two basic bundle colors, this will work, but unfortunately, you will have to create a lot of new tokens, as it should also have hover/active variants and two tokens for each, meaning you will need to create 12 new tokens, so I would prefer the fix with background-image :)

mainframev:
Doc buttons are broken 😢 
Screenshot 2023-09-15 at 13 41 54

packages/orbit-tailwind-preset/src/presets/foundation/theme/button.ts
```diff
@@ -0,0 +1,17 @@
+import { convertHexToRgba, defaultTokens } from "@kiwicom/orbit-design-tokens";
+
+export const opacity = {
+  "button-disabled": defaultTokens.buttonDisabledOpacity,
+};
+
+export const boxShadow = {
```
mainframev, line 7: This file and those tokens should be a part of the component preset, not the default one :)

packages/orbit-tailwind-preset/src/presets/foundation/theme/index.ts
```diff
@@ -4,4 +4,6 @@ import boxShadow from "./boxShadow";
 import font from "./font";
 import duration from "./duration";
 
+export * as button from "./button";
```
mainframev, line 7: Also shouldn't be a part of this config :)

packages/orbit-tailwind-preset/src/presets/__tests__/__snapshots__/configs.test.ts.snap
```diff
@@ -2517,14 +2477,17 @@ exports[`orbitPreset should match snapshot 1`] = `
     },
     "padding": {
       "0": "0px",
+      "button-large": "52px",
```
mainframev, line 2480: I think instead of introducing those new tokens you can use `formBox` tokens, the values are the same 

If you take a look at the `formBox` json definition, it sets height tokens for buttons as deprecated tokens :)

mainframev:
Circled in the branch vs. Circled in the master 🐛 

Screenshot 2023-09-15 at 13 55 14
Screenshot 2023-09-15 at 13 55 22

packages/orbit-components/src/Button/index.tsx
```diff
@@ -1,35 +1,89 @@
...
-      />
+        className={cx(
+          "orbit-button space-x-xs p-button-padding focus:shadow-button-focus disabled:shadow-none",
+          sizeStyles[size],
+          typeStyles[type],
+          !props.circled && iconLeft == null && iconRight == null && paddingNoIconsStyles[size],
+          !props.circled && iconLeft != null && iconRight == null && paddingLeftIconStyles[size],
+          !props.circled && iconLeft == null && iconRight != null && paddingRightIconStyles[size],
+          !props.circled && iconLeft != null && iconRight != null && paddingBothIconsStyles[size],
+          props.circled === true && circledStyles[size],
```
mainframev, line 82: why not just 🤔 : 

```props.circled && circledStyles[size]```

File: packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx [hunk 6]
```diff
@@ -1,263 +1,283 @@
...
+          >
+            {iconLeft}
+          </div>
+        )}
+        {children != null && (
+          <div
+            className={cx(
+              "orbit-button-primitive-content inline-block",
+              (iconLeft != null || iconRight != null) && "text-start",
+              loading === true && "invisible",
```
mainframev, line 246: I can understand usage of *!= null* or *== null* equality checks 

but for such cases with true, I would prefer just to have *&&* 

I do not mind having all of these, but personally, I found it harder to read and the benefit of it is questionable, there are cases when it's especially useful, but dunno if it's a good idea to promote it everywhere. :)

packages/orbit-components/src/Button/index.tsx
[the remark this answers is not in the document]
mainframev: @oreqizer Yes, that will work, we can do it that way, in addition to our bundle tokens, and create more granular ones. You can create them inside *foundation/global/palette/gradients*

packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
oreqizer: I am used to having [strict boolean expressions(https://typescript-eslint.io/rules/strict-boolean-expressions/) in place and use the code style even when the rule is not enabled. it imo has a lot of explicitness value and I'd argue it makes things _more_ readable, because you don't have to second-guess types of expressions

mainframev: later we can think about unifying this, since you already refactored several components and the code % with this will be increasing

packages/orbit-components/src/Button/index.tsx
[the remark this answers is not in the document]
oreqizer: solved it by specifying these colors as `backgroundImage` in the TW theme

oreqizer:
TODOs:
- [ ] loading on pale backgrounds

mainframev:
Content in the circled Button is not properly centered 😢 

Screenshot 2023-09-19 at 10 50 49
Screenshot 2023-09-19 at 10 50 51

mainframev:
Seems like the font-weight issue is still present: 
Screenshot 2023-09-19 at 10 57 04
Screenshot 2023-09-19 at 10 57 08

oreqizer:
@mainframev I checked and both versions' font weights are `500`

packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
DSil, line 1: I couldn't blame an exact line, but something is off regarding the color on hover.

Master
image

This branch
image

I notice that on hover it is applying just the foreground var, but shouldn't it be applying the hover var?
image

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -170,6 +173,9 @@ const cfg = (options?: Options): Config => {
         borderRadius: {
           "dialog-desktop": defaultTokens.dialogBorderRadiusDesktop,
           "dialog-mobile": defaultTokens.dialogBorderRadiusMobile,
+          "button-circled-small": defaultTokens.heightButtonSmall,
+          "button-circled-normal": defaultTokens.heightButtonNormal,
+          "button-circled-large": defaultTokens.heightButtonLarge,
```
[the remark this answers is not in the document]
oreqizer: these are not for non-circled buttons, that is why the word is there

packages/orbit-components/src/Button/index.tsx
```diff
@@ -1,35 +1,96 @@
...
-        disabled={disabled}
+        iconLeft={iconLeft}
+        iconRight={iconRight}
         {...props}
-        {...buttonStyles}
-        {...commonProps}
-        {...icons}
-      />
+        className={cx(
+          "orbit-button space-x-xs focus:shadow-button-focus disabled:shadow-none",
```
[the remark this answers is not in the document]
oreqizer: yea I see that now. it was present in the old code so I added it, but it probably isn't working

packages/orbit-components/src/Button/index.tsx
```diff
@@ -1,35 +1,96 @@
...
+};
+
+const paddingNoIconsStyles: Record<Size, string> = {
+  small: "p-button-small-padding",
+  normal: "p-button-normal-padding",
+  large: "p-button-large-padding",
+};
+
+const paddingLeftIconStyles: Record<Size, string> = {
+  small: "p-button-small-left-icon-padding",
```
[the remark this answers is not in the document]
oreqizer: I totally agree, though imo it is a potential followup token cleanup endeavour we could do. I'd keep it out of scope of this MR. there's already plenty of changes

packages/orbit-components/src/Button/index.tsx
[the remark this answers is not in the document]
oreqizer: removed it. the code was there previously but didn't work

packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
oreqizer: found it, gonna fix

packages/orbit-components/src/Button/index.tsx
```diff
@@ -1,35 +1,96 @@
...
+  criticalSubtle:
+    "bg-button-critical-subtle-background hover:bg-button-critical-subtle-background-hover active:bg-button-critical-subtle-active-background disabled:bg-button-critical-subtle-background focus:bg-button-critical-subtle-background-focus text-button-critical-subtle-foreground focus:text-button-critical-subtle-foreground-focus active:text-button-critical-subtle-foreground-active hover:text-button-critical-subtle-foreground-hover disabled:text-button-critical-subtle-foreground focus:shadow-product-normal/50 active:shadow-button-active-pale",
+  white:
+    "bg-button-white-background hover:bg-button-white-background-hover active:bg-button-white-background-active disabled:bg-button-white-background focus:bg-button-white-background-focus text-button-white-foreground focus:text-button-white-foreground-focus active:text-button-white-foreground-active hover:text-button-white-foreground-hover disabled:text-button-white-foreground focus:shadow-red-normal/50 active:shadow-button-active-pale",
```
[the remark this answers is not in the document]
oreqizer: removed the shadow focus. it _was_ there before but didn't work

packages/orbit-components/src/Button/index.tsx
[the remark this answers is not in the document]
oreqizer: actually I had to fix it here, due to RTL

packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
```diff
@@ -1,263 +1,297 @@
...
+        {...props}
+        disabled={isDisabled}
+        className={cx(
+          className,
+          "orbit-button-primitive font-base duration-fast group relative max-w-full cursor-pointer items-center border-none text-center transition-all [&>*]:align-middle [&_.orbit-loading-spinner]:stroke-[currentColor]",
+          fullWidth && "w-full",
+          centered || children == null ? "justify-center" : "justify-[var(--button-content-align)]",
+          circled !== true && "rounded-large tb:rounded-normal",
+          isDisabled
+            ? "opacity-button-disabled cursor-not-allowed"
```
DSil, line 167: Non-blocking but just wondering if we really need this token... We have `opacity-30` available, and it's much more explicit and one less token for us...

DSil (commented):
Overall LGTM :+1: just one non-blocking question

packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
oreqizer: imo we don't, but again, it exists so I used it. gonna remove its usage in TW and Button, we can clean it up later

DSil: It's part of the component tokens, so it can be removed with no breaking change notice. I'd just remove it now and leave only the deprecated one, to be removed on the next major tokens release. But up to you 👍

DSil (approved):

mainframev (approved):
Good job on that :cacodemon: 🚀 component!

PR #4016, opened 2023-09-12
chore(tracking): fix the glob pattern and update react-scanner

mainframev:
Our data for September was corrupted because of the slightly wrong glob pattern, it was collecting a much smaller amount (as only js files). Also included a new version of our fork of react-scanner, that includes an update to node >= 18 and a fix that throws warnings instead of exceptions, if the script fails to parse a file (happens on a couple of files on Balkan that do not include Orbit components) 
 Storybook: https://orbit-mainframev-chore-fix-glob-tracking.surge.sh

### DIFF
--- a/packages/orbit-tracking/package.json
+++ b/packages/orbit-tracking/package.json
@@ -57,7 +57,7 @@
     "dotenv": "^16.0.1",
     "execa": "^6.1.0",
     "filedirname": "^2.7.0",
-    "react-scanner-orbit": "^2.0.0",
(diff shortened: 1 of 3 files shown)


DSil (commented):
Please rerun `yarn` as it produces `yarn.lock` changes

DSil (approved):

mainframev:
@DSil it happens from time to time sometimes randomly, locally it works. Do not think it's related. I will return just to make sure

PR #4018, opened 2023-09-13
Fix Button bundleMedium hover

DSil:
As reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1694600566693979).

I have made one of the commit messages to refer to Button even though it is not on the orbit-components package. This will make the changelog message to refer to the correct component affected. And since version bump is the same level (patch) it is not problematic.
 Storybook: https://orbit-mainframev-fix-button-bundle-medium-hover.surge.sh


mainframev (approved):
LGTM 🚀

PR #4019, opened 2023-09-13
V10 migration guide

DSil:
Migration guide for v10

 Storybook: https://orbit-mainframev-v10-migration-guide.surge.sh

### DIFF
--- a/docs/src/documentation/02-foundation/11-tailwind/01-installation.mdx
+++ b/docs/src/documentation/02-foundation/11-tailwind/01-installation.mdx
@@ -1,6 +1,8 @@
 ---
 title: Installation
 description: Orbit tailwind preset is the abstraction above Orbit design tokens package. It provides the Orbit design tokens as Tailwind CSS configuration.
+redirect_from:
+  - /foundation/tailwind/installation/
 ---
 
 import README from "@kiwicom/orbit-tailwind-preset/README.md";
--- a/docs/src/documentation/05-development/04-migration-guides/01-v7.mdx
+++ b/docs/src/documentation/05-development/04-migration-guides/01-v7.mdx
@@ -1,5 +1,5 @@
 ---
-title: V7
+title: v7
 description: How to migrate to Orbit 7.0.0
 redirect_from:
   - /migration-guides/v7/
@@ -8,7 +8,7 @@ redirect_from:
 # Orbit Migration Guide v7
 
 This migration guide focuses on the process of migrating from Orbit v6.4 to v7.0, as some breaking changes were introduced.
-With this guide we aim to walkthrough all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortless.
+With this guide, we aim to walk through all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortlessly.
 
 ## Breaking changes
 
--- a/docs/src/documentation/05-development/04-migration-guides/02-v9.mdx
+++ b/docs/src/documentation/05-development/04-migration-guides/02-v9.mdx
@@ -1,5 +1,5 @@
 ---
-title: V9
+title: v9
 description: How to migrate to Orbit 9.0.0
 redirect_from:
   - /migration-guides/v9/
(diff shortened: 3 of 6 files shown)


mainframev:
@DSil good job! 

Let's just fix the order of migration guides that appear in the sidebar, It should have the same file name logic as we do for other mdx files: 

```
01-v10.mdx
02-v9.mdx
```

👍🏻

docs/src/documentation/05-development/04-migration-guides/v10.mdx
```diff
@@ -0,0 +1,101 @@
...
+title: V10
+description: How to migrate to Orbit 10.0.0
+redirect_from:
+  - /migration-guides/v10/
+---
+
+# Orbit Migration Guide v10
+
+This migration guide focuses on the process of migrating from Orbit v9.1 to v10.0, as some breaking changes were introduced.
+With this guide we aim to walkthrough all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortless.
```
mainframev, line 11:
```suggestion
With this guide, we aim to walk through all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortlessly.
```

docs/src/documentation/05-development/04-migration-guides/v10.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+This version of Orbit includes the first batch of components styled with Tailwind, instead of Styled Components. Note that you can keep using Styled Components, especially on your custom components. But we recommend start migrating them to Tailwind as well.
+
+Components in Tailwind should have the same API as before, so no breaking changes are expected in that regard. However, and since some of the migrated components went through a deep refactoring, we recommend being on the lookout for unexpected and undocumented changes. Those should be reported so that we can fix them.
```
mainframev, line 15:
```suggestion
Components in Tailwind should have the same API as before, so no breaking changes are expected in that regard. However, since some of the migrated components went through deep refactoring, we recommend being vigilant for unexpected and undocumented changes. Those should be reported so that we can fix them.
```

docs/src/documentation/05-development/04-migration-guides/v10.mdx
```diff
@@ -0,0 +1,101 @@
...
+### Tailwind requirement
+
+As mentioned before, this version of Orbit includes the first batch of components styled with Tailwind. Therefore, Tailwind must be installed and configured in your project for you to be able to use these components.
+Please refer to the [official Tailwind documentation](https://tailwindcss.com/docs/installation) and the [Orbit's Tailwind setup documentation](/foundation/tailwind/installation) for more information on how to do that.
+
+As a summary, you should have Tailwind installed and configured in your project, as well as the Orbit preset for Tailwind included on the config file.
+
+### StyledBox removed
+
+The `StyledBox` component is no longer exported. This allowed custom Styled Components to target the `Box` components from Orbit and apply styles to it. Therefore, things like the following code will not work anymore:
```
mainframev, line 28:
```suggestion
The `StyledBox` component is no longer exported. This allowed custom Styled Components to target the `Box` components from Orbit and apply styles to it. Therefore, things like the following code will no longer work:
```

docs/src/documentation/05-development/04-migration-guides/v10.mdx
```diff
@@ -0,0 +1,101 @@
...
+    padding: 10px;
+  }
+`;
+```
+
+### Align prop in Text
+
+The `Text` component features an `align` prop that directly matches the `text-align` CSS property. This property used to have a default value of `"left"` and Styled Components was automatically dealing with RTL and converting `"left"` to `"right"` when the browser was presenting a page in RTL.
+
+Now, this is no longer the case. Therefore, we have changed the default value to be `"start"` so that it keeps the same behavior as before. However, if you were using the `align` prop with the value `"left"` or `"right"` and relying on an automatic RTL convertion, you should now use `"start"` instead.
```
mainframev, line 59:
```suggestion
Now, this is no longer the case. Therefore, we have changed the default value to be `"start"` so that it keeps the same behavior as before. However, if you were using the `align` prop with the value `"left"` or `"right"` and relying on an automatic RTL conversion, you should now use `"start"` instead.
```

docs/src/documentation/05-development/04-migration-guides/v10.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+Notice that `"left"` and `"right"` are still valid values, but they will not adjust to RTL anymore.
+
+### Icon prop in Card
+
+The `icon` prop was already deprecated in the `Card` and `CardSection` components and was finally removed on this version.
```
mainframev, line 65:
```suggestion
The `icon` prop was already deprecated in the `Card` and `CardSection` components and has finally been removed in this version.
```

mainframev (changes requested):

docs/src/documentation/05-development/04-migration-guides/v10.mdx
DSil: Changed on the others as well, as this is a common paragraph to all migration guides

mainframev (approved):

PR #4022, opened 2023-09-14
chore: add tailwind folder as temporary storage for refactored compon…

mainframev:
As we agreed, creating the temporary folder for storing refactored TW components

To test:
- run *yarn components build* and check the output of the es/lib folders 
- or look at the result of compressed-size-action

 Storybook: https://orbit-mainframev-chore-tailwind-folder.surge.sh


DSil (commented):
How can we try these components on the docs if we want? 🤔

mainframev:
> How can we try these components on the docs if we want? 🤔

why we would like to try them on doc? Locally you can try that before merging, but if the storybook works, then it should be enough :) Before releasing them we will get them out of the folder

DSil:
To try on the Playground. But even locally, the only way to import is using relative paths, right?

mainframev:
> To try on the Playground. But even locally, the only way to import is using relative paths, right?

do not think it will work, as it’s excluded from build, only if the relative path to src folder

DSil (approved):

PR #4023, opened 2023-09-14
docs: return the colors to TW docs

mainframev:


docs/src/components/TailwindClassnames/index.tsx
```diff
@@ -104,6 +122,12 @@ const TailwindClassnames = () => {
             <Heading type="title2" spaceAfter="normal">
               {group}
             </Heading>
+            {group === "Colors" && (
+              <Text type="info" weight="medium" spaceAfter="medium">
```
DSil, line 126: Just a 💅 but since we have `spaceAfter` deprecated, can we use the `margin` prop instead here? One less place to change after we finally remove `spaceAfter`. WDYT?

docs/src/components/TailwindClassnames/index.tsx
```diff
@@ -104,6 +122,12 @@ const TailwindClassnames = () => {
...
+                Please remember to use default Tailwind prefixes with those classnames: bg, text,
```
DSil, line 127: I really like that we're saying this. I would just rephrase it and also add it to other classes as well. Something like

```
To be used with Tailwind classes that apply a color property (eg: `text-blue-dark`, `bg-red-light`, etc).
```

And then for the screens something like
```
To be used as a Tailwind prefix for defining media queries (eg: `sm:opacity-1`, `tb:hidden`, etc).
```

For spacing, something like
```
To be used with Tailwind classes that apply spacing properties (eg: `p-sm`, `m-xl`, `mb-0`, etc).
```

And finally, a note on zIndex (since those are being extended, and not overwritten) saying something like
```
Default Tailwind zIndex classes are also available.
```

For all the other properties I don't think we need to add any message. Plus, the text I've suggested here is not a hard suggestion, just a content suggestion, to be coherent across all properties and informative rather than "a reminder request" (_please remember_).

mainframev: Yes, thanks for the suggestion 🚀

docs/src/components/TailwindClassnames/index.tsx
mainframev: Actually I tried that and found a bug, that only in case of providing a px value it applies the margin. If I add something like *margin={{bottom: medium}}* -> nothing happens

mainframev: I hope the bug is not present in TW branch

docs/src/components/TailwindClassnames/index.tsx
```diff
@@ -52,6 +69,49 @@ const getCategory = (name: string) => {
...
+    return acc;
+  }, {});
+};
+
+const CategoryDescriptions = ({ category }: { category: Category }) => {
+  return (
+    <>
+      {category === "Colors" && (
+        <Text type="info" weight="medium" margin={{ bottom: "12px" }}>
+          To be used with Tailwind classes that apply a color property (eg: `text-blue-dark`,
```
DSil, line 91: Apparently the `` don't have any effect (too used to MDX 😅) can we maybe use `<code>` or something to have it formatted? Or maybe just _italic_? WDYT?

docs/src/components/TailwindClassnames/index.tsx
```diff
@@ -52,6 +69,49 @@ const getCategory = (name: string) => {
...
+      )}
+      {category === "Screens" && (
+        <Text type="info" weight="medium" margin={{ bottom: "12px" }}>
+          To be used as a Tailwind prefix for defining media queries (eg: `sm:opacity-1`,
+          `tb:hidden`, etc).
+        </Text>
+      )}
+      {category === "Spacing" && (
+        <Text type="info" weight="medium" margin={{ bottom: "12px" }}>
+          To be used as a Tailwind prefix for defining margin and padding (eg: `m-1`, `p-2`, etc).
```
DSil, line 103:
```suggestion
          To be used as a Tailwind prefix for defining margin and padding (eg: `m-sm`, `p-xl`, etc).
```

We should use our tokens 😄

docs/src/components/TailwindClassnames/index.tsx
```diff
@@ -52,6 +69,49 @@ const getCategory = (name: string) => {
...
+        </Text>
+      )}
+      {category === "ZIndex" && (
+        <Text type="info" margin={{ bottom: "12px" }}>
```
DSil, line 107: All others have `weight="medium"` except this one. Which leads me to think…
Couldn't we have something like

```
 <Text type="info" weight="medium" margin={{ bottom: "12px" }}>
  {category === ...}
</Text>
```

So that we only write the `<Text>` code once and the logic of the string is calculated inside? Or maybe abstract it before the return… This way we can avoid the fragment at all, no?

docs/src/components/TailwindClassnames/index.tsx
mainframev: I've just copied your suggestion :D

docs/src/components/TailwindClassnames/index.tsx
mainframev: do not think anything is needed here, simple string is enough

docs/src/components/TailwindClassnames/index.tsx
DSil: Hm, no 🤔 😅 
image

mainframev: Hmmm, maybe copilot messed that up :D

DSil (commented):
LGTM 👍 Please squash before merge

DSil (approved):

PR #4026, opened 2023-09-15
fix(ButtonLink): fix incorrect hover backgroud color on primary

mainframev:
Accidently noticed on #4015 , the hover background token was not correct for ButtonLink primary. 
 Storybook: https://orbit-mainframev-fix-button-link-type.surge.sh


DSil (commented):
LGTM, good catch. Just update the snapshots before approval 🙏

DSil (approved):

PR #4027, opened 2023-09-15
ci: add loki gh action workflow

mvidalgarcia:


mvidalgarcia:
We're experimenting with Chromatic, closing as this is not relevant anymore.

PR #4040, opened 2023-09-19
Migrate ErrorFormTooltip to Tailwind

DSil:
[FEPLT-1722](https://kiwicom.atlassian.net/browse/FEPLT-1722)
 Storybook: https://orbit-mainframev-migrate-error-form-tooltip-tailwind.surge.sh


mainframev:
The position of tooltip is slightly off on RTL (in master in one side, in this branch in another) but still it's a 🐛

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,68 @@ const ErrorFormTooltip = ({
...
+        )}
         ref={setArrowRef}
-        inlineLabel={inlineLabel}
-        placement={attrs.popper && attrs.popper["data-popper-placement"]}
       />
-      <StyledTooltipContent ref={contentRef}>{children}</StyledTooltipContent>
+      <div
+        className={cx(
+          "font-base text-normal leading-small text-white-normal flex h-full items-center font-normal",
+          "[&>.orbit-text]:text-white-normal [&>.orbit-text]:text-normal hover:[&>.orbit-text]:text-white-normal focus:[&>.orbit-text]:text-white-normal [&>.orbit-text]:font-normal",
```
mainframev, line 142: We have to be mindful about those [&>.class] selectors. For instance, here it won't work for all text-links, if the user decides to pass more than one text-link, it will apply only to the first one

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,68 @@ const ErrorFormTooltip = ({
...
+        "lm:w-auto tb:rounded-normal",
+        "[&>img]:max-w-full",
+      )}
+      style={cssVars as React.CSSProperties}
     >
-      <StyledArrow
-        isHelp={isHelp}
+      <div
+        className={cx(
+          "absolute start-[8px]",
```
mainframev, line 131:
```suggestion
          "absolute start-xs",
```

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,68 @@ const ErrorFormTooltip = ({
...
+          isVertical ? "bottom-[2px]" : "top-[2px]",
```
mainframev, line 132:
```suggestion
          isVertical ? "bottom-xxxs" : "top-xxxs",
```

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,68 @@ const ErrorFormTooltip = ({
...
+          inlineLabel && "rtl:start-0",
+          isHelp ? "before:bg-blue-normal" : "before:bg-red-normal",
+          "before:absolute before:h-[8px] before:w-[8px] before:-translate-x-1/2 before:-translate-y-1/2 before:rotate-45",
```
mainframev, line 135:
```suggestion
          "before:absolute before:h-xs before:w-xs before:-translate-x-1/2 before:-translate-y-1/2 before:rotate-45",
```

mainframev (changes requested):

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
mainframev: Should be "[&_a]:text-white-normal"

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,68 @@ const ErrorFormTooltip = ({
     };
   }, [onShown, isHelp, helpClosable]);
 
+  const isVertical =
+    (attrs.popper && attrs.popper["data-popper-placement"] === "top-start") ||
+    (attrs.popper && attrs.popper["data-popper-placement"] === "top-end")
+      ? "bottom"
+      : "top";
```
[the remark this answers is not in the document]
oreqizer: `attrs.popper?["data-popper-placement"] === "top-end"`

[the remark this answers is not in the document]
DSil: @oreqizer suggestion did not work. Was throwing error
```
This comparison appears to be unintentional because the types 'string[]' and 'string' have no overlap.ts(2367)
This condition will always return 'false' since JavaScript compares objects by reference, not value.ts(2839)
```

Ended up doing
```js
const isVertical =
  attrs.popper &&
  (attrs.popper["data-popper-placement"] === "top-start" ||
    attrs.popper["data-popper-placement"] === "top-end")
    ? "bottom"
    : "top";
```

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
DSil: Right, my bad, thanks

DSil:
@mainframev regarding the position, it's funny because I was checking with dev tools open and it was working fine. Turns out you checked it without dev tools and saw it buggy. Why? Because the FormLabel element has a different padding on desktop and above. With dev tools open, the viewport was of size tablet, and so I was unable to spot the bug, and so I only fixed it for the other one that we had in Master… But anyway, it's fixed now, thanks 👍

mainframev (approved):

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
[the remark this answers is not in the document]
oreqizer: 🤔 really? that's weird, this didn't work?

```ts
const isVertical =
    attrs.popper?.["data-popper-placement"] === "top-start" ||
    attrs.popper?.["data-popper-placement"] === "top-end"
      ? "bottom"
      : "top";
```

I mean it is just syntax here since we're not enforcing [prefer nullish coalescing](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md) in ESlint, so no problem, I just find it interesting

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -213,32 +89,66 @@ const ErrorFormTooltip = ({
...
         ref={setArrowRef}
-        inlineLabel={inlineLabel}
-        placement={attrs.popper && attrs.popper["data-popper-placement"]}
       />
-      <StyledTooltipContent ref={contentRef}>{children}</StyledTooltipContent>
+      <div
+        className={cx(
+          "font-base text-normal leading-small text-white-normal flex h-full items-center font-normal",
+          "[&_.orbit-text]:text-white-normal [&_.orbit-text]:text-normal hover:[&_.orbit-text]:text-white-normal focus:[&_.orbit-text]:text-white-normal [&_.orbit-text]:font-normal",
+          "[&_.orbit-list-item]:text-white-normal [&_.orbit-list-item]:text-normal hover:[&_.orbit-list-item]:text-white-normal focus:[&_.orbit-list-item]:text-white-normal [&_.orbit-list-item]:font-normal",
```
mainframev, line 141: do we need these focus and hover classes if the color is not changing?

mainframev: besides that nothing. LGTM and can be merged :) @oreqizer I though we were waiting for you approval on that also :)

DSil: I believe it is here to enforce the color. If we use the Text inside, the default colors of the text are given by classes that will always be overwritten by this. It's true, we don't seem to have styles on hover which means we could probably remove it. I would personally not risk it at this moment tbh. But fair question

oreqizer (approved):
lgtm 👍

PR #4041, opened 2023-09-20
ci: fix post-changelog script

mvidalgarcia:
- read release commit diff instead of using `conventional-changelog`
- same approach as in frontend/core


scripts/post-changelog.mjs
```diff
@@ -1,17 +1,15 @@
 import { $, fetch, argv } from "zx";
 import dotenv from "dotenv-safe";
-import conventionalChangelog from "conventional-changelog";
```
mvidalgarcia, line 3: This dep is still used in the _old_ publish script, otherwise, it could be removed. It's a dev dep, so not a big deal.

scripts/post-changelog.mjs
```diff
@@ -1,17 +1,15 @@
...
-import { getPackages } from "@lerna/project";
```
mvidalgarcia, line 4: Same here.

scripts/post-changelog.mjs
```diff
@@ -1,17 +1,15 @@
...
 
 /* eslint-disable no-console */
 
-const CHANNEL = `orbit-react`;
-const COLOR_CORE = `#00A58E`;
-const PACKAGES = ["orbit-components", "orbit-tailwind-preset"];
+const CHANNEL = "orbit-react";
 const PACKAGE_PREFIX = "@kiwicom";
+const COLOR_CORE = "#00A58E";
+const PACKAGES = ["orbit-components", "orbit-tailwind-preset"];
```
mvidalgarcia, line 12: This can be increased as we like. I believe these two are the most relevant to end users, but correct me if I'm wrong.

mainframev: orbit-design-tokens should be included also then :)

scripts/post-changelog.mjs
```diff
@@ -1,17 +1,15 @@
...
 import { simpleGit } from "simple-git";
 import gitDiffParser from "gitdiff-parser";
 
 /* eslint-disable no-console */
 
-const CHANNEL = `orbit-react`;
-const COLOR_CORE = `#00A58E`;
-const PACKAGES = ["orbit-components", "orbit-tailwind-preset"];
+const CHANNEL = "orbit-react";
 const PACKAGE_PREFIX = "@kiwicom";
```
mainframev, line 10: It's used only once, maybe does not make sense to have a constant for that

better will be to adjust the function make it completely pure: 
```
function format(str, package_, prefix = "@kiwicom") { 
```

scripts/post-changelog.mjs
```diff
@@ -39,6 +38,19 @@ function adjustChangelog(str) {
   return output;
 }
 
+function getChangelogFromDiff(files) {
+  // Only one file as we're only looking at the changelog
+  const [changelogFile] = files;
+  const changelog = changelogFile.hunks
+    .flatMap(hunk => {
+      return hunk.changes.filter(({ isInsert }) => isInsert).map(({ content }) => content);
```
mainframev, line 46: Just 💅🏻 

```suggestion
hunk.changes.filter(({ isInsert }) => isInsert).map(({ content }) => content)
```

scripts/post-changelog.mjs
mainframev: Feel free to remove as a part of this PR

mainframev (approved):
I ran that only locally with dry, it did not show the changelog as split to packages as usual in the core package it shows. Should it work like that, right? Otherwise the code LGTM

mvidalgarcia:
@mainframev What do you mean?

For me it's showing the proper changelog, which is what's present in the last [release commit](https://github.com/kiwicom/orbit/commit/5ba032a71f3e2034c7be4860d74bf55ac0ab7c57).

```sh
yarn run v1.22.17
$ /Users/marco/code/github.com/kiwicom/orbit/node_modules/.bin/zx scripts/post-changelog.mjs --dry
No changes in orbit-design-tokens
# _@kiwicom/orbit-components_ [9.3.0](https://github.com/kiwicom/orbit/compare/@kiwicom/orbit-components@9.2.0...@kiwicom/orbit-components@9.3.0) (2023-09-19)


### Features 🆕

* **Illustration:** add Damage and Wheelchair illustrations ([04c78b6](https://github.com/kiwicom/orbit/commit/04c78b6bfadf8f90df297a7daddaf4182601e354))
No changes in orbit-tailwind-preset
Done in 2.56s.
```

mainframev:
@mvidalgarcia feel free to squash and merge

PR #4043, opened 2023-09-20
chore: split general checks into multiple jobs

DSil:
General job was executing 3 types os checks. They all now have their own job each.

Also removed the TODO as we _should_ run the pipeline for every PR. We can just have less restrict required checks on non-master PRs, but that can be configured somewhere else 

After this is approved and before is merged, I will change the required checks on the branch rules to reflect the new jobs.

**Time changes** (compared with [the latest run](https://github.com/kiwicom/orbit/actions/runs/6237496256)):
| Scenario | Test workflow time (excluding Cypress and Docs) | Jobs duration |
|---|---|---|
| This branch | 8m 50s | Types (4m 37s), Lint (8m 50s), Unit (8m 5s) |
| Before | 15m 59s | General (15m 59s) |

Not only we can detect failures earlier (if a test fails, we don't have to wait for types and lint to run to see it failing), the time to execute also decreases drastically.

Storybook: https://orbit-mainframev-split-pipelines-jobs.surge.sh

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,5 @@
 name: Test
 
-#TODO remove the all branches condition later
 on:
   push:
     branches:
@@ -17,7 +16,7 @@ env: # Set environment variables for every job and step in this workflow
   STREAMLINE_FAMILIES: ${{ secrets.STREAMLINE_FAMILIES }}
 
 jobs:
-  general:
+  types:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


mainframev:
@DSil 

Could you also attach to the PR description the comparison of the speed of the initial General job and the refactored one? 🙏🏻

mainframev:
@DSil  he build steps are repeated in multiple job tasks. Wdyt, will that make sense to have an artifact for that, so it's built only once and reused later in jobs? 🤔

DSil:
I thought about that but I did not know if it would be possible, as it was not being done in other jobs (cypress and docs).

I was going to try it anyway, but only after making sure this separation was working. I'll try it now

DSil:
@mainframev I tried your proposal of using artifacts but I was unable to have it working. I agree it could be optimized and avoid building multiple times, but maybe I am lacking something, I just couldn't make it work 😞 

So we can either dig a little more on this or leave that optimization for later, as we already observe a good gain in times

mainframev (approved):
LGTM 🚀

.github/workflows/test.yml
```diff
@@ -39,13 +38,51 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Node
+        uses: ./.github/actions/node
+
+      - name: Build
+        env:
+          GITLAB_TOKEN: ${{ secrets.GITLAB_TOKEN }}
+        run: |
```
mainframev, line 53: I do not think we need to build everything for the lint job. I think only tokens here have to be built. Definitely themer, tracking can be excluded. Es and lib folders are ignored, so components also can be excluded from build 🤔

DSil: I tried it locally and it worked but not on CI. It needs all the packages built 
image

PR #4044, opened 2023-09-20
ButtonLink to Tailwind

oreqizer:
[WEBC-1680](https://kiwicom.atlassian.net/browse/FEPLT-1680)
 Storybook: https://orbit-mainframev-oreqizer-tw-button-link.surge.sh


oreqizer:
TODO:
- [ ] text color of secondary compact link

packages/orbit-components/src/tmp-tailwind/ButtonLink/README.md
DSil, line 1: I'd say that if the file is untouched (unchanged) we can avoid having a copy on the tmp folder...

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.js.flow
DSil, line 1: I'd avoid moving this as well

packages/orbit-components/src/tmp-tailwind/ButtonLink/types.d.ts
DSil, line 1: And also this. We can import the types from the original file, no?

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    ref,
+  ) => {
+    return (
+      <ButtonPrimitive
+        ref={ref}
+        iconLeft={iconLeft}
+        iconRight={iconRight}
+        {...props}
+        className={cx(
+          "orbit-button space-x-xs rtl:space-x-reverse",
```
DSil, line 69: Is `orbit-button` expected here? Do we need the class at all? I don't think this was being targeted anywhere 💭

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+          compact !== true &&
+            children != null &&
+            iconLeft == null &&
+            iconRight != null &&
+            paddingRightIconStyles[size],
+          compact !== true &&
+            children != null &&
+            iconLeft != null &&
+            iconRight != null &&
+            paddingBothIconsStyles[size],
```
DSil, line 94: This is not correct. And I have just noticed we have the same problem on Button 😞 As an example, a large button with icon on both sides should not have a 24px horizontal padding, but a 16px.

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ref={ref}
+        iconLeft={iconLeft}
+        iconRight={iconRight}
+        {...props}
+        className={cx(
+          "orbit-button space-x-xs rtl:space-x-reverse",
+          sizeStyles[size],
+          compact !== true && disabled !== true && typeStyles[type],
+          compact === true && disabled !== true && typeCompactStyles[type],
+          compact !== true && disabled === true && typeDisabledStyles[type],
```
DSil, line 73: Disabled styles are not being correctly applied.

On Button, the `cursor: not-allowed` is also missing when it is disabled

DSil (changes requested):
Noticed some styles are off and bugs on the already implemented Button / ButtonPrimitive

DSil (changes requested):
Noticed some styles are off and bugs on the already implemented Button / ButtonPrimitive

DSil (changes requested):
Noticed some styles are off and spotted bugs on the already implemented Button / ButtonPrimitive

packages/orbit-components/src/primitives/ButtonPrimitive/sizes.ts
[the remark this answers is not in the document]
oreqizer: imo `sizes` makes sense, since the intention of the file is to contain styles related to the `Size` type. when sb will be looking for styles that are affected by the `size` prop, this will be quite intuitive to look into. does this help with the context?

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
oreqizer: adjusted, nice catch, yea it was difficult to get these right, and are not immedatelly obvious which ones are correct and which are not

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
oreqizer: fixed 👍

mainframev:
@oreqizer I think on hover missing the transition, as in master it looks a bit smoother 🤔

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+        {...props}
+        className={cx(
+          "space-x-xs rtl:space-x-reverse",
+          sizeStyles[size],
+          compact !== true && disabled !== true && typeStyles[type],
+          compact === true && disabled !== true && typeCompactStyles[type],
+          compact !== true && disabled === true && typeDisabledStyles[type],
+          compact === true && disabled === true && typeCompactDisabledStyles[type],
+          compact !== true &&
+            children != null &&
```
mainframev, line 77: this condition is repetitive, could you abstract that to a single variable instead? 🤔 

```
     compact !== true &&
            children != null &&
            iconLeft == null &&
            iconRight == null &&
```

packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+            iconLeft != null &&
+            iconRight == null &&
+            paddingLeftIconStyles[size],
+          compact !== true &&
+            children != null &&
+            iconLeft == null &&
+            iconRight != null &&
+            paddingRightIconStyles[size],
+          compact !== true &&
```
mainframev, line 91: this condition seems to be repetitive, could you assign it to a variable? 

```
 compact !== true &&
            children != null &&
            ```

Comment (oreqizer) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx:92]: while this `compact !== true && children != null` bit is used four times, imo it is better to be explicit with these low level conditions than abstracting them away. with these you _immediately_ know they're props and what they represent

if we added _another_ variable that is composed of this, e.g. `isNotCompactWithoutChildren`, that is almost as many characters as the explicit condition 😅 and you have more cognitive load to bear

Comment (mainframev) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx:92]: @oreqizer You may try smth like that, looks shortened and more readable imo 
```
    const isIconPaddingStyle = compact !== true && children != null;

    return (
      <ButtonPrimitive
        ref={ref}
        iconLeft={iconLeft}
        iconRight={iconRight}
        disabled={disabled}
        {...props}
        className={cx(
          "space-x-xs rtl:space-x-reverse",
          sizeStyles[size],
          compact === true && [
            disabled === true typeDisabledStyles[type],
            disabled !== true && typeCompactStyles[type],
          ],
          compact !== true && [
            disabled === true && typeDisabledStyles[type],
            disabled !== true && typeStyles[type],
          ],
          isIconPaddingStyle && [
            iconLeft == null &&
            iconRight == null &&
            paddingNoIconsStyles[size],
            iconLeft != null &&
            iconRight == null &&
            paddingLeftIconStyles[size],
            iconLeft == null &&
            iconRight != null &&
            paddingRightIconStyles[size],
            iconLeft != null &&
            iconRight != null &&
            paddingBothIconsStyles[size]
          ],
          props.circled === true && circledStyles[size],
          props.circled === true && children == null && circledIconOnlyStyles[size],
        )}
      >
        {children}
      </ButtonPrimitive>

```

Comment (oreqizer) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx:92]: I don't know man 😕 I just very much prefer less words and less nesting in code

I'd understand if it was like 20+ cases, but this is really compact and explicit, even when unrolled

cognitive load here seems much bigger, too. when I see `isIconPaddingStyle` I have _no idea_ where is this coming from. I have to jump up and see what does it represent

Comment (mainframev) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx:92]: > while this compact !== true && children != null bit is used four times, imo it is better to be explicit with these low level conditions than abstracting them away. with these you immediately know they're props and what they represent

Not sure if I can agree with that, when there are a lot of lines I feel like am struggling when I am reading that, especially when they look almost the same and there are multiple conditions in one single line and you have them like > 4 rows. Assigning it to a variable you can call "isSomething" and it also says immediately what is it if it's properly named. Plus (as I showed above) maybe we should consider using arrays to group them.

Comment (oreqizer) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx:92]: aight I can go with the array syntax, without an extra variable though. is that good enoguh of a compromise for you?

Comment (mainframev) [reply packages/orbit-components/src/tmp-tailwind/ButtonLink/index.tsx]: yep :)

**Review by DSil — COMMENTED** [review]
Issue about tokens and please chore the commits as talked IRL

File: packages/orbit-tailwind-preset/src/presets/components/index.ts [hunk 6]
```diff
@@ -258,10 +257,16 @@ const cfg = (options?: Options): Config => {
...
-        }, {}),
+        textColor: {
+          ...Object.entries(getForegroundColors()).reduce((acc, [key, value]) => {
+            const name = key.replace("text-", "");
+            return { ...acc, [name]: value };
+          }, {}),
+          "button-link-secondary-compact-foreground-hover":
+            defaultTokens.colorTextButtonLinkSecondaryCompactHover,
+          "button-link-secondary-compact-foreground-active":
+            defaultTokens.colorTextButtonLinkSecondaryCompactActive,
```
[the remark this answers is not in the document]
oreqizer: right 👍 I need to get in to the habit of checking for deprecation. these are actually the same as `product-normal` colours so I used those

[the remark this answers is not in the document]
oreqizer: done them commits

oreqizer:
thank @DSil , fixed 👌

DSil (approved):
Rebase and 🚀

mainframev (approved):

PR #4050, opened 2023-09-24
chore(InputField): migrate to Tailwind

mainframev:
Closes [FEPLT-1692](https://kiwicom.atlassian.net/browse/FEPLT-1692)

 Storybook: https://orbit-mainframev-tw-migrate-inputfield.surge.sh

### DIFF
--- a/packages/orbit-components/src/tmp-tailwind/InputField/InputTags/index.tsx
+++ b/packages/orbit-components/src/tmp-tailwind/InputField/InputTags/index.tsx
@@ -1,0 +1,76 @@
+import React from "react";
+import cx from "clsx";
+
+const InputTags = ({ children }: { children: React.ReactNode }) => {
+  const tagsRef = React.useRef<HTMLDivElement | null>(null);
+
+  React.useEffect(() => {
+    const handleMouseMove: React.MouseEventHandler<HTMLDivElement> = event => {
+      if (tagsRef && tagsRef.current) {
+        // @ts-expect-error TODO
+        const { isDragging } = tagsRef.current;
+        if (isDragging && event.movementX) {
+          tagsRef.current.scrollLeft -= event.movementX;
+        }
+      }
+    };
+
+    const handleMouseUp: React.MouseEventHandler<HTMLDivElement> = () => {
+      if (tagsRef && tagsRef.current) {
+        // @ts-expect-error TODO
+        tagsRef.current.isDragging = false;
+      }
+    };
+
+    const tags = tagsRef.current as HTMLElement;
+
+    if (tags) {
+      // @ts-expect-error TODO
+      tags.addEventListener("mousemove", handleMouseMove);
+      // @ts-expect-error TODO
+      tags.addEventListener("mouseup", handleMouseUp);
+    }
+
+    return () => {
+      if (tags) {
+        // @ts-expect-error TODO
+        tags.removeEventListener("mousemove", handleMouseMove);
+        // @ts-expect-error TODO
+        tags.removeEventListener("mouseup", handleMouseUp);
+      }
+    };
+  }, [tagsRef]);
+
+  return (
+    <div
+      className={cx(
+        "relative",
+        "ms-sm",
+        "flex flex-initial items-center",
+        "h-full min-w-[50px]",
+        "overflow-hidden",
+        "z-[2]",
+      )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
+      <div
+        className={cx(
+          "white-space-nowrap overflow-x-scroll",
+          "space-x-xs flex items-center",
+          "scrollbar-none",
+        )}
+        ref={tagsRef}
+        onMouseDown={() => {
+          if (tagsRef && tagsRef.current) {
+            // @ts-expect-error TODO
+            tagsRef.current.isDragging = true;
+          }
+        }}
+      >
+        {children}
+      </div>
+    </div>
+  );
+};
+
+export default InputTags;
--- a/packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
+++ b/packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
@@ -1,0 +1,324 @@
+"use client";
+
+import * as React from "react";
+import cx from "clsx";
+
+import { TYPE_OPTIONS } from "../../InputField/consts";
+import InputTags from "./InputTags";
+import getFieldDataState from "../../common/getFieldDataState";
+import useRandomId from "../../hooks/useRandomId";
+import ErrorFormTooltip from "../../ErrorFormTooltip";
+import AlertCircle from "../../icons/AlertCircle";
+import InformationCircle from "../../icons/InformationCircle";
+import FormLabel from "../../FormLabel";
+import useErrorTooltip from "../../ErrorFormTooltip/hooks/useErrorTooltip";
+import type { Props } from "../../InputField/types";
+import { spaceAfterClasses } from "../../common/tailwind";
+
+const getDOMType = (type: Lowercase<keyof typeof TYPE_OPTIONS>) => {
+  if (type === TYPE_OPTIONS.PASSPORTID) return TYPE_OPTIONS.TEXT;
+  return type;
+};
+
+export const FakeInput = ({
+  error,
+  disabled,
+  readOnly,
+}: Pick<Props, "error" | "disabled" | "readOnly">) => (
+  <div
+    className={cx(
+      "orbit-input-field-fake-input",
+      "h-form-box-normal text-form-element-normal-font-size z-[1]",
+      "absolute left-0 top-0",
+      "duration-fast transition-all ease-in-out",
+      "rounded-large tb:rounded-normal box-border w-full",
+      "peer-focus:outline-blue-normal peer-focus:outline peer-focus:outline-2",
+      error ? "shadow-form-element-error" : "shadow-form-element",
+      disabled || readOnly
+        ? "bg-form-element-disabled-background"
+        : [
+            "bg-transparent",
+            error
+              ? "peer-hover:shadow-form-element-error-hover"
+              : "peer-hover:shadow-form-element-hover",
+          ],
+    )}
+  />
+);
+
+const Prefix = ({ children }: { children: React.ReactNode }) => (
+  <div
+    className={cx(
+      "text-form-element-prefix-foreground",
+      "ps-sm pointer-events-none z-[3] flex h-full items-center justify-center",
+      "[&>svg]:text-icon-tertiary-foreground",
+      "[&_*_svg]:w-icon-medium [&>svg]:w-icon-medium",
+      "[&_*_svg]:h-icon-medium [&>svg]:h-icon-medium",
+      "[&_.orbit-button-primitive-icon]:text-icon-secondary-foreground",
+    )}
+  >
+    {children}
+  </div>
+);
+
+const Suffix = ({
+  disabled,
+  children,
+}: {
+  disabled: Props["disabled"];
+  children: React.ReactNode;
+}) => (
+  <div
+    className={cx(
+      "h-form-box-normal text-form-element-prefix-foreground",
+      "z-[3] flex shrink-0 items-center justify-center",
+      "[&_.orbit-button-primitive-icon]:text-icon-secondary-foreground",
+      "[&_.orbit-service-logo]:pe-sm [&_.orbit-service-logo]:h-md",
+      disabled && "pointer-events-none",
+    )}
+  >
+    {children}
+  </div>
+);
+
+const InputField = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
+  const {
+    disabled,
+    type = TYPE_OPTIONS.TEXT,
+    label,
+    inlineLabel,
+    dataTest,
+    required,
+    error,
+    name,
+    prefix,
+    onChange,
+    onFocus,
+    onBlur,
+    onSelect,
+    onMouseUp,
+    onMouseDown,
+    onKeyUp,
+    onKeyDown,
+    placeholder,
+    minValue = -Infinity,
+    maxValue = Infinity,
+    minLength,
+    maxLength,
+    suffix,
+    help,
+    helpClosable = true,
+    value,
+    defaultValue,
+    tags,
+    tabIndex = 0,
+    readOnly,
+    list,
+    autoComplete,
+    ariaAutocomplete,
+    ariaActiveDescendant,
+    ariaHasPopup,
+    ariaExpanded,
+    ariaControls,
+    autoFocus,
+    spaceAfter,
+    id,
+    width = "100%",
+    inputMode,
+    insideInputGroup,
+    dataAttrs,
+    role,
+  }: Props = props;
+
+  const forID = useRandomId();
+  const inputId = id || forID;
+
+  const hasTooltip = Boolean(error || help);
+
+  const {
+    tooltipShown,
+    tooltipShownHover,
+    setTooltipShown,
+    setTooltipShownHover,
+    labelRef,
+    iconRef,
+    handleFocus,
+  } = useErrorTooltip<HTMLInputElement, HTMLDivElement>({ onFocus, hasTooltip });
+
+  const shown = tooltipShown || tooltipShownHover;
+  const fieldRef = React.useRef(null);
+  const Component = label ? "label" : "div";
+
+  return (
+    <Component
+      className={cx(
+        "font-base relative block flex-1 basis-full",
+        spaceAfter && spaceAfterClasses[spaceAfter],
+      )}
+      style={{ width }}
+      ref={fieldRef}
+      htmlFor={label ? inputId : undefined}
+      onMouseEnter={() => (disabled && inlineLabel ? setTooltipShownHover(true) : undefined)}
+      onMouseLeave={() => (disabled && inlineLabel ? setTooltipShownHover(false) : undefined)}
+    >
+      {label && !inlineLabel && (
+        <FormLabel
+          inlineLabel={inlineLabel}
+          filled={!!value}
+          required={required}
+          error={!!error}
+          help={!!help}
+          labelRef={labelRef}
+          iconRef={iconRef}
+          onMouseEnter={() => setTooltipShownHover(true)}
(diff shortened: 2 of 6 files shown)


packages/orbit-components/src/tmp-tailwind/InputField/README.md
```diff
@@ -0,0 +1,138 @@
+# InputField
```
oreqizer, line 1: pls remove redundant files from `tmp-tailwind`, keep only the story and the main component file 🙏

mainframev: I do not think it's a good idea remove __tests__ folder, I can remove the types and README

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+          className={cx(
+            "font-base p-form-element-normal-padding text-inherit",
+            "z-[2] appearance-none border-none shadow-none",
+            "box-border h-full w-full min-w-0",
+            "cursor-auto",
+            "bg-transparent",
+            "flex-[1_1_20%]",
+            "[&::-webkit-inner-spin-button,&::-webkit-outer-spin-button]:appearance-none",
+            "[&::-webkit-inner-spin-button,&::-webkit-outer-spin-button]:margin-0",
+            "[&[data-com-onepassword-filled]]:bg-inherit",
```
DSil, line 221: These selectors do not work in Tailwind. The margin prefix is `m-`.

```suggestion
            "[&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:appearance-none",
            "[&::-webkit-inner-spin-button]:m-0 [&::-webkit-outer-spin-button]:m-0",
```

DSil, line 221: I was unable to test this property, but it used to have the `!important` so I guess we should have it here as well.

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+            "peer focus:outline-none",
+            "[&::placeholder]:opacity-100",
+            "[&::placeholder]:text-form-element-foreground",
+            "[&::-ms-input-placeholder]:text-form-element-foreground",
+            "[&::-ms-clear]:display-none [&::-ms-reveal]:display-none",
```
DSil, line 226: I couldn't test this as this is an Edge exclusive property, but `display` is not a valid Tailwind prefix.

```suggestion
            "[&::-ms-clear]:hidden [&::-ms-reveal]:hidden",
```

packages/orbit-components/src/tmp-tailwind/InputField/InputTags/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+        tags.removeEventListener("mouseup", handleMouseUp);
+      }
+    };
+  }, [tagsRef]);
+
+  return (
+    <div
+      className={cx(
+        "relative",
+        "ps-sm",
```
DSil, line 48: It used to be margin. Why the change to padding?

packages/orbit-components/src/tmp-tailwind/InputField/InputTags/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+        "flex flex-initial items-center",
+        "h-full min-w-[50px]",
+        "overflow-hidden",
+        "z-[2]",
+      )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
+      <div
+        className={cx(
+          "white-space-nowrap overflow-scroll",
```
DSil, line 58:
```suggestion
          "white-space-nowrap overflow-x-scroll",
```

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
DSil: Wrong line suggestion. Should be on the lines above

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -86,7 +86,7 @@ const cfg = (options?: Options): Config => {
           "button-normal": defaultTokens.buttonNormalFontSize,
           "button-small": defaultTokens.buttonSmallFontSize,
           "form-element-normal": defaultTokens.formElementNormalFontSize,
-          "form-element-small": defaultTokens.formElementSmallFontSize,
+          "form-element-small": defaultTokens.formElementNormalFontSize,
```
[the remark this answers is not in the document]
mainframev: Remove it

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+        )}
+        {label && inlineLabel && (
+          <div
+            className={cx(
+              "orbit-input-field-inline-label",
+              "flex items-center justify-center",
+              "pointer-events-none h-full",
+              "[&>span]:mb-0",
+              "[&>span]:text-form-box-normal [&>span]:whitespace-nowrap [&>span]:leading-normal",
+              "[&>span]:z-[3]",
```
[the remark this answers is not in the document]
mainframev: It can be span only there. As it's from the label which is a span. 
Actually, we should avoid adding classes everywhere, as it opens pandora's box for overwriting our components

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+const Prefix = ({ children }: { children: React.ReactNode }) => (
+  <div
+    className={cx(
+      "orbit-input-field-prefix",
+      "text-form-element-prefix-foreground",
+      "ps-sm pointer-events-none z-[3] flex h-full items-center justify-center",
+      "[&>svg]:text-icon-tertiary-foreground",
+      "[&_*_svg,&>svg]:w-icon-normal",
+      "[&_*_svg,&>svg]:h-icon-normal",
```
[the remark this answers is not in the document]
mainframev: Thanks, well spotted. Visually it's hard to detect, as the icon size is 20px itself, so no difference, only via inspector

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+          iconRef={iconRef}
+          onMouseEnter={() => setTooltipShownHover(true)}
+          onMouseLeave={() => setTooltipShownHover(false)}
+        >
+          {label}
+        </FormLabel>
+      )}
+      <div
+        className={cx(
+          "orbit-input-field-input-container",
```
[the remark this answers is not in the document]
mainframev: I think we should do that in InputGroup refactoring, not related to this request. @oreqizer is working on that, so he will target that class ;)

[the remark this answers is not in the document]
DSil: Agreed. Just concerned if the two refactors happen at the same time. But I think it's a scenario that we can decide on the go. So no need to do anything now 🙂

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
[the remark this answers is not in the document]
DSil: I agree with overwriting styles. But ultimately this approach breaks the abstraction barrier on the FormLabel. If we decide to change the rendered element in `orbit-form-label` to a `p` or a `div` (it's unlikely, but for whatever reason it can happen, and it's just a general example here) then this selector will not have the effect we want it to have

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+          labelRef={labelRef}
+          iconRef={iconRef}
+          onMouseEnter={() => setTooltipShownHover(true)}
+          onMouseLeave={() => setTooltipShownHover(false)}
+        >
+          {label}
+        </FormLabel>
+      )}
+      <div
+        className={cx(
```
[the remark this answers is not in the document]
mainframev: do you think it makes sense? the hover is working fine, isn't it redundant to add another hover?

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+            "flex-[1_1_20%]",
+            "[&::-webkit-inner-spin-button,&::-webkit-outer-spin-button]:appearance-none",
+            "[&::-webkit-inner-spin-button,&::-webkit-outer-spin-button]:margin-0",
+            "[&[data-com-onepassword-filled]]:bg-inherit",
+            "peer focus:outline-none",
+            "[&::placeholder]:opacity-100",
+            "[&::placeholder]:text-form-element-foreground",
+            "[&::-ms-input-placeholder]:text-form-element-foreground",
+            "[&::-ms-clear]:display-none [&::-ms-reveal]:display-none",
+            disabled && "[-webkit-text-fill-color]:text-form-element-disabled-foreground",
```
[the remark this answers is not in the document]
DSil: Didn't notice on first time, but this also needs to be applied if `readOnly` is true

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+            inlineLabel ? "font-medium" : "font-normal",
+            type === TYPE_OPTIONS.PASSPORTID && "tabular-nums",
+            type === TYPE_OPTIONS.PASSPORTID && "letter-spacing-[2px]",
```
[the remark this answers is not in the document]
DSil: This is correct here, but I just noticed that on Storybook these styles are actually not being applied 🤔 but it might be a bug there, no?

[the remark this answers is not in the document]
mainframev: Ah, you meant on original InputField. Then yes, I think now it's correct

DSil (commented):
Just one thing on the readOnly state and please remove the untouched files (except tests, as agreed)

DSil (commented):
LGTM, please fix pipeline 🙏

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,315 @@
...
+          // eslint-disable-next-line jsx-a11y/no-autofocus
+          autoFocus={autoFocus}
+          id={inputId}
+          inputMode={inputMode}
+          {...dataAttrs}
+        />
+        {suffix && <Suffix disabled={disabled}>{suffix}</Suffix>}
+        <div
+          className={cx(
+            "orbit-input-field-fake-input",
```
mvidalgarcia, line 279: This component is exported and used in other places such as Select component.

https://github.com/kiwicom/orbit/blob/951112cb6574bd96ea4a43a62f0ecd989f62b89d/packages/orbit-components/src/InputField/index.tsx#L68

mainframev: @mvidalgarcia I know, it was not removed from the original InputField

mvidalgarcia: I need this component in Select. After this change is not exported :)

https://github.com/kiwicom/orbit/blob/951112cb6574bd96ea4a43a62f0ecd989f62b89d/packages/orbit-components/src/Select/index.tsx#L427

mainframev: import that from normal InputField, from: "src/InputField", not tmp.

mvidalgarcia: Right, but the src/InputField will be replaced by the content in tmp. And tmp doesn't export it...

mainframev: Simply reuse the classes from InputField PR then, I do not think it's the right thing to export that anyway, use the primitive to unify all those stylings between InputField/InputFile/Select, but currently, we do not have one.

mvidalgarcia: I'm sorry but I can't agree to that. What's the benefit of keeping duplicated code in two different components? If we change a style here we have to remember to change it in the Select component as well, and I'm almost certain that will forget about it since if there's no component, it won't be easily searchable.

What's stopping us from creating a simple component with the snippet you shared and exporting it?

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+import useErrorTooltip from "../../ErrorFormTooltip/hooks/useErrorTooltip";
+import type { Props } from "../../InputField/types";
+import { spaceAfterClasses } from "../../common/tailwind";
+
+const getDOMType = (type: Lowercase<keyof typeof TYPE_OPTIONS>) => {
+  if (type === TYPE_OPTIONS.PASSPORTID) return TYPE_OPTIONS.TEXT;
+  return type;
+};
+
+export const fakeInputSnippet = ({
```
mvidalgarcia, line 23: Could we have just a simple `FakeInput` component (as we used to have) rendering a div with these classes? Or is there any reason why this is has to be function?

mainframev: I think with Tailwind it's better for such cases to use the function as if needed, you can adjust it, use any element you want etc., more flexible.

mvidalgarcia: I agree it's more flexible, but this doesn't necessarily mean that's better. Following that criterion, we should make all simple components the same way, and I don't recall seeing this kind of function around, but I might be wrong.

In addition, we could potentially introduce an `as` prop defaulting to `div` if we need this flexibility.

mainframev: I do not think we should continue arguing about that, it helps you to achieve what you want and it can be in form of a component or a function. Feel free to refactor it to a component, if you want later in your request. I do not think adding 'as' prop will be a good thing, this looks simpler to me. Let's not block this MR more :)

mvidalgarcia: I'm not blocking it, just expressing my opinion.

> it helps you to achieve what you want

This is not about what I want, it's not personal. I'm just discussing what I believe is better for the project. But you're right, it's not worth arguing more.

In any case, this PR will break the Select component once is moved to the src folder, I guess this is a downsize of using a tmp folder instead of branches. But no worries, I'll adapt it on my side.

DSil: I can't agree with this being a function. This was, and should continue to be, a component. It receives props and renders an HTML element with classes and some logic based on those props. That's the React abstraction working.
The `Select` component uses this and should not need to worry about its internal implementation, including which HTML element it uses. The only thing users of this FakeInput should know is its class so they can target it: `.orbit-input-field-fake-input` (as it is right now, but it can be any other thing, like simply `.orbit-fake-input`).

Furthermore, this FakeInput is using `peer` related classes. If it's a reusable component (or function), at first, I am not sure if it is a good approach (I know I was the one suggestind it at first, but the context is now broader... and it eventually can play in our favor). This means that if we use this as a sibling of any other component with `peer` we might get undesirable side effects. However, and looking at the current scenario, it seems it can actually play in our favor as we have things right now:
The `Select` component can take advantage of this and simplify its logic by removing all the targets to the `.orbit-input-field-fake-input` class. Applying the `peer` class to the `select` element rendered inside should be enough to ensure the very same visual styles that we have in InputField, which is what we want.

So, to me, this should be a component. And for now it is kept inside this `InputField` folder with the proper class and the peer targets, as we have. And we will be able to migrate Select taking advantage of the class selector and the peer classes that already exist.
In the future, we can (probably should) reconsider extracting this component to its own folder, with its own tests, and independent implementation. And we can evaluate the effort of not relying on `peer` classes on a component that is meant to be reused in more than one place. This does not mean we will export the component, but it can be kept as an internal component that is still capable of livingo on its own and suits the needs of whoever uses it.

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
```diff
@@ -0,0 +1,318 @@
...
+    "absolute left-0 top-0",
+    "duration-fast transition-all ease-in-out",
+    "rounded-large tb:rounded-normal box-border w-full",
+    "peer-focus:outline-blue-normal peer-focus:outline peer-focus:outline-2",
+    error ? "shadow-form-element-error" : "shadow-form-element",
+    disabled || readOnly
+      ? "bg-form-element-disabled-background"
+      : [
+          "bg-transparent",
+          error ? "peer-hover:shadow-form-element-error" : "peer-hover:shadow-form-element-hover",
```
DSil, line 40:
```suggestion
          error
            ? "peer-hover:shadow-form-element-error-hover"
            : "peer-hover:shadow-form-element-hover",
```

packages/orbit-components/src/tmp-tailwind/InputField/index.tsx
mainframev: I will refactor it to the component in order just to stop this discussion and continue with migration, but all these arguments above don't convince me that it should be a component

_> It receives props and renders an HTML element with classes and some logic based on those props._
There's not much of a difference between receiving a prop or function parameter in this situation. 

Before it was the component just because you can't do it the other way with SC. Now we can't trigger that via class names and such things can be considered as snippets or mixins that can be applied with function and extended or slightly changed if needed (with the component you can't do that) 

This whole paragraph about the peer is interesting, but not related to this situation, isn't it? 
 
 > In the future, we can (probably should) reconsider extracting this component to its own folder, with its own tests, and independent implementation.  
 
No, we shouldn't. Tested should be only input/select native elements, those visuals do not need to be abstracted as components and tested. They will be tested in visual regression within the whole component.

DSil (approved):

PR #4051, opened 2023-09-24
chore: update the fetchIcons script to prevent false positives

mainframev:
[Context](https://skypicker.slack.com/archives/C05FDP4UM3N/p1695563894902239)
 Storybook: https://orbit-mainframev-chore-fix-false-positives-icons.surge.sh


packages/orbit-components/package.json
```diff
@@ -139,6 +140,8 @@
     "svgicons2svgfont": "^10.0.0",
     "tailwindcss": "^3.3.2",
     "ttf2woff2": "^4.0.1",
+    "pixelmatch": "^5.3.0",
+    "pngjs": "^7.0.0",
```
DSil, line 144: Please don't add deps manually to package.json. Use `yarn add` so it keeps the package.json file sorted

DSil (commented):
LGTM, just one 💅

oreqizer (approved):
sort `package.json` deps, otherwise good 👍

oreqizer (changes requested):
_accidentally approved_

packages/orbit-components/package.json
mainframev: I did it because of the error I recieving, so added that for quickness.

mainframev: before I could add deps with lerna add, but since we upgraded the version of lerna recently this command is not available anymore.

mainframev: the error was: https://github.com/yarnpkg/yarn/issues/7807, but with @typescript/cypress for orbit-components.

mainframev: cleaning the cache seemed to help, will run again then

mainframev: we need to migrate to pnpm asap, we are still at old version of 1, when there is yarn v3 and pnpm is just conceptually much better

mainframev: If y'all will suggest how to resolve -> I will be thankfull, what I tried: 
- regenerating yarn.lock
- deleting the cache via yarn cache clean
- adding @testing-library/cypress again 

```
error An unexpected error occurred: "expected workspace package to exist for \"@testing-library/cypress\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/rejuv/orbit/packages/orbit-components/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

DSil (approved):

PR #4052, opened 2023-09-24
fix: replace size tailwind tokens with correct

mainframev:
There are some button-size classes defined that are completely redundant and have to be replaced with "form-box" tokens. Applied them where it's necessary, plus removed previously defined ones that are not needed anymore. 
 Storybook: https://orbit-mainframev-fix-tailwind-tokens.surge.sh


oreqizer (commented):
lgtm 👍

oreqizer (approved):

PR #4053, opened 2023-09-24
chore(Textarea): migrate Textarea to Tailwind

mainframev:
[Closes 1697](https://kiwicom.atlassian.net/browse/FEPLT-1697)
 Storybook: https://orbit-mainframev-tw-migrate-textarea.surge.sh


packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+          "w-full appearance-none",
+          "font-base",
+          "rounded-normal relative box-border block overflow-auto",
+          "min-h-form-box-normal",
+          "duration-fast transition-shadow ease-in-out",
+          "border border-transparent",
+          size === SIZE_OPTIONS.SMALL && "leading-small text-small px-sm py-xs",
+          size === SIZE_OPTIONS.NORMAL && "text-normal p-sm leading-normal",
+          error
+            ? "shadow-form-element-error hover:shadow-form-element-error"
```
DSil, line 96: It should not be the same shadow on hover and on normal state

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+        placeholder={String(placeholder)}
+        maxLength={maxLength}
+        onChange={onChange}
+        rows={rows}
+        onFocus={handleFocus}
+        onBlur={onBlur}
+        tabIndex={tabIndex ? Number(tabIndex) : undefined}
+        readOnly={readOnly}
+        ref={ref}
+        style={{ resize }}
```
DSil, line 123: I think we can avoid this and use tailwind classes

`resize === RESIZE_OPTIONS.VERTICAL ? "resize-y" : "resize-none"`

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <textarea
+        className={cx(
+          "w-full appearance-none",
+          "font-base",
+          "rounded-normal relative box-border block overflow-auto",
+          "min-h-form-box-normal",
+          "duration-fast transition-shadow ease-in-out",
+          "border border-transparent",
+          size === SIZE_OPTIONS.SMALL && "leading-small text-small px-sm py-xs",
+          size === SIZE_OPTIONS.NORMAL && "text-normal p-sm leading-normal",
```
DSil, line 94: Font size does not change with size. They were using fontSizeInputSmall and fontSizeInputNormal, but, as we observed, they were actually the same font-size... So it ahould just be `text-form-element-normal`, regardless of element size.

What's weird to me is that the line-height and padding also change. But maybe that's a bug...

DSil (changes requested):

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
mainframev: good, solved in #4047, will add the class for now, but will work after the merge of Radio

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
mainframev: Waiting for designers reply, might be not necessary to support small size at all

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -117,17 +117,37 @@ const cfg = (options?: Options): Config => {
           "icon-small": defaultTokens.iconSmallSize,
           "icon-medium": defaultTokens.iconMediumSize,
           "icon-large": defaultTokens.iconLargeSize,
+<<<<<<< HEAD
```
DSil, line 120: Please fix this

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
mainframev: So we should have only normal-sized textarea

DSil: Give that it's a breaking change, I'd add it to this release right now (aka skip the tmp folder) and include it on this PR. WDYT?

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+          error
+            ? "shadow-form-element-error hover:shadow-form-element-error"
+            : "shadow-form-element",
+          disabled ? "cursor-not-allowed" : "cursor-text",
+          fullHeight && "h-full flex-1",
+          disabled || readOnly
+            ? "bg-form-element-disabled-background"
+            : "bg-form-element-background",
+          "[&::placeholder]:form-element-placeholder",
+          !(disabled || readOnly) && "hover:shadow-form-element-hover",
```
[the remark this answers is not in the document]
DSil: This is still not applied

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
mainframev: Nope, I won't do that right now, I will prefer it to be released with other input components

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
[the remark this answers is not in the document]
mainframev: Should be ok now :)

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+          size === SIZE_OPTIONS.NORMAL && "text-normal p-sm leading-normal",
+          error
+            ? "shadow-form-element-error hover:shadow-form-element-error-hover"
+            : "shadow-form-element hover:shadow-form-element-hover",
+          disabled ? "cursor-not-allowed" : "cursor-text",
+          fullHeight && "h-full flex-1",
+          disabled || readOnly
+            ? "bg-form-element-disabled-background"
+            : "bg-form-element-background",
+          "[&::placeholder]:text-form-element-foreground",
```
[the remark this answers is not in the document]
DSil: Just realized this suggestion is messy… I changed line 96 and 103 on the original file removing the hover effects from the `error` condition and adding them inside an extra condition inside the `disabled || readOnly` condition

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -229,8 +229,8 @@ const cfg = (options?: Options): Config => {
           "country-flag": defaultTokens.countryFlagShadow,
           "form-element": defaultTokens.formElementBoxShadow,
           "form-element-error": defaultTokens.formElementBoxShadowError,
+          "form-element-error-hover": defaultTokens.formElementErrorFocusBoxShadow,
```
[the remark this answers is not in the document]
mainframev: You are right, actually, to me, it even visually looks better :) As it's not almost visible, but I will change that because of consistency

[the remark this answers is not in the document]
DSil: I kinda agree. I don't like any of them, tbh. One looks odd to me (two shades of red) and the other is barely noticeable. But that's something we can address designers in the future and easily change the tokens accordingly 🙂

packages/orbit-components/src/tmp-tailwind/Textarea/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+          size === SIZE_OPTIONS.NORMAL && "text-normal p-sm leading-normal",
+          disabled ? "cursor-not-allowed" : "cursor-text",
+          fullHeight && "h-full flex-1",
+          disabled || readOnly
+            ? "bg-form-element-disabled-background"
+            : [
+                "bg-form-element-background",
+                error
+                  ? "shadow-form-element-error hover:shadow-form-element-error-hover"
+                  : "shadow-form-element hover:shadow-form-element-hover",
```
DSil, line 104: This can't be like this. This is saying the error shadow only appears when it's not disabled nor readOnly, which is not true. Only hover shadow is affected disabled || readOnly.

The suggestion [here](https://github.com/kiwicom/orbit/pull/4053#discussion_r1345877876) has it correctly. It adds the the shadow depending on error state (line 107) and then the hover shadow depends on the disabled || readOnly (line 114).

mainframev: it can be disabled and have the error shadow?

DSil (approved):

PR #4056, opened 2023-09-25
test: remove act wrappers for user event actions

mainframev:
Having an act wrapper for user event actions seems a redundant thing to have, tests are working fine without 

We also started to receive warnings about the wrong usage of the act and [I later found this comment](https://github.com/testing-library/react-testing-library/issues/1061#issuecomment-1419824786)
 Storybook: https://orbit-mainframev-chore-fix-test-suits.surge.sh


DSil (approved):
LGTM 👍

PR #4057, opened 2023-09-26
fix(BadgePrimitive): fix min-height class

mvidalgarcia:


DSil (commented):
LGTM, but this should not be a fix commit because we don't want it on the changelog. Not only this is a change on the tmp folder, it is also fixing something that is not yet released. Therefore, the bug never existed, so it does not deserve a mention on the changelog.

DSil (approved):

PR #4058, opened 2023-09-27
deprecate components

mainframev:
Resolves: 
- [FEPLT-1715](https://kiwicom.atlassian.net/browse/FEPLT-1715)
- [FEPLT-1714](https://kiwicom.atlassian.net/browse/FEPLT-1714)
- [FEPLT-1713](https://kiwicom.atlassian.net/browse/FEPLT-1713)
 Storybook: https://orbit-mainframev-deprecated-components.surge.sh


docs/src/components/Footer.tsx
```diff
@@ -61,7 +61,7 @@ const StyledFooterLink = styled(Link)`
 `;
 
 function Dot() {
-  return <Desktop>·</Desktop>;
+  return <Hide on={["smallMobile", "mediumMobile", "largeMobile", "tablet"]}>·</Hide>;
```
oreqizer, line 64: do we want to also deprecate `Hide` down the road? we should just use TW classes, e.g. `className="hidden dt:block"`

mainframev: nope, Hide is still useful and should be probably deprecated after migration is fully over and users will also use the tailwind.

oreqizer: ok let's deprecate `Hide` later

oreqizer (approved):

PR #4059, opened 2023-09-28
chore(InputFile): migrate to tailwind

mvidalgarcia:
Closes [FEPLT-1693](https://kiwicom.atlassian.net/browse/FEPLT-1693)
 Storybook: https://orbit-mainframev-marco-inputfile-tailwind.surge.sh


DSil (changes requested):
Besides what is reported, while reviewing this, I noticed two small bugs (alredy present, so we can just solve them later):

- contrary to other inputs (InputField, Checkbox, etc) this one has hover effect even when disabled. I believe it should not
- the required prop triggers no visual change

Again, these are already existent bugs. If their resolution is not straightforward, I say we create tasks for it. Your decision. Non-blocking.

mvidalgarcia:
Thanks for the comments! I think I've addressed them all.

> * contrary to other inputs (InputField, Checkbox, etc) this one has hover effect even when disabled. I believe it should not
> 
> * the required prop triggers no visual change

I also fixed these two bugs.

DSil (approved):

PR #4066, opened 2023-10-03
fix(Dialog): hide scroll bar, but keep scrolling behaviour

mainframev:
[Reported on Slack](https://skypicker.slack.com/archives/C7T7QG7M5/p1696338848003649)
 Storybook: https://orbit-mainframev-fix-tooltip-overflow.surge.sh


DSil (changes requested):
This is removing the expected scrollbar on mobile when the content is scrollable (although it's barely visible anyway… but it's there).
I tried wrapping the CSS rules you added inside a media query selector for desktop and it seemed to solve the issue, hiding the white space on desktop but showing the scrollbar on mobile… but can you please try and confirm yourself?

Before:
https://github.com/kiwicom/orbit/assets/6265045/57de15a3-9cf6-493a-b4d5-21bf7c2a1c42

Now:
https://github.com/kiwicom/orbit/assets/6265045/14b3da4c-388d-4503-a06e-31294db7715b

Videos taken in Safari iOS

DSil (approved):

PR #4067, opened 2023-10-03
migrate Switch to Tailwind

DSil:
For height and width values I decided not to use tokens because tokens are associated with spacing and those size values are very much independent from any spacing scale.
 Storybook: https://orbit-mainframev-tw-switch.surge.sh

### DIFF
--- a/packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
+++ b/packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
@@ -1,0 +1,59 @@
+"use client";
+
+import * as React from "react";
+import cx from "clsx";
+
+import Circle from "../../icons/Circle";
+import handleKeyDown from "../../utils/handleKeyDown";
+import type { Props } from "../../Switch/types";
+
+const Switch = React.forwardRef<HTMLInputElement, Props>(
+  ({ onChange, checked, dataTest, id, icon, onBlur, onFocus, disabled, ariaLabelledby }, ref) => {
+    return (
+      <label className="inline-block">
+        <div
+          className={cx(
+            "duration-fast relative flex h-[20px] w-[40px] items-center justify-between rounded-[100px] transition-colors",
+            disabled ? "opacity-50" : "cursor-pointer",
+            checked ? "bg-blue-normal" : "bg-cloud-dark",
+          )}
+        >
+          <input
+            className="peer absolute inset-0 m-0 h-full w-full p-0 opacity-0"
+            ref={ref}
+            checked={checked}
+            disabled={disabled}
+            aria-checked={checked}
+            role="switch"
+            aria-labelledby={ariaLabelledby}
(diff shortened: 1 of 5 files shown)


packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import cx from "clsx";
+
+import Circle from "../../icons/Circle";
+import handleKeyDown from "../../utils/handleKeyDown";
+import type { Props } from "../../Switch/types";
+
+const Switch = React.forwardRef<HTMLInputElement, Props>(
+  ({ onChange, checked, dataTest, id, icon, onBlur, onFocus, disabled, ariaLabelledby }, ref) => {
+    return (
+      <div className="inline-block">
```
mvidalgarcia, line 13: Isn't this a label? Probably doesn't matter as we're overriding the `display` attr.

packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <div
+          className={cx(
+            "duration-fast relative flex h-[20px] w-[40px] items-center justify-between rounded-[100px] transition-colors",
+            disabled ? "opacity-50" : "cursor-pointer",
+            checked ? "bg-blue-normal" : "bg-cloud-dark",
+          )}
+        >
+          <input
+            className="peer absolute inset-0 h-full w-full opacity-0"
```
mvidalgarcia, line 22: Here you omitted `margin` and `padding` set to 0. That's fine, as it's 0 in the computed styles due to the preflight styles (I might be mistaken on this, not 100% sure).

image

What if consumers disable the preflight setting? Won't this affect them? As input checkbox has margin and padding by default. [Playground](https://codi.link//PHA+CjxpbnB1dCB0eXBlPSJjaGVja2JveCIgcm9sZT0ic3dpdGNoIiAvPmZvbwo8L3A+CjxwPgo8aW5wdXQgY2xhc3M9ImlucHV0IiB0eXBlPSJjaGVja2JveCIgcm9sZT0ic3dpdGNoIiAvPmJhcgo8L3A+Cg==|LmlucHV0IHsKICBtYXJnaW46IDA7CiAgcGFkZGluZzogMDsKfSA=|).

mvidalgarcia (approved):
Just a couple of minor details.

Looks great and works smoothly 💅 Great job!

packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
DSil: Honestly, I don't remember thinking about that when (not) adding the margin and padding classes.
I also don't know if they are not somehow useful for when using a Switch inside other layout components…
I prefer to have them rather than relying on preflight. Adding them back. Well spotted.

PR #4069, opened 2023-10-04
chore: fix unit tests

mainframev:
Adjusted the timeout for failing tests and removed steps. Removed act warnings for one of the tests inside docs.


mainframev:
@DSil docs are still failing, but even it was changed, so I assume that it should be fine after merge

DSil:
The Check Links step is stalling. I cancelled the job after more than 3 hours and now restarted it and it's not ending. Something is wrong.

To speed up and unblock can the changes on docs tests be addressed later on a separate PR and fix this one on unit test?

mainframev:
> The Check Links step is stalling. I cancelled the job after more than 3 hours and now restarted it and it's not ending. Something is wrong.
> 
> To speed up and unblock can the changes on docs tests be addressed later on a separate PR and fix this one on unit test?

Yep, I know, should be ok right now 🤔

DSil (approved):

PR #4070, opened 2023-10-04
chore: add clear cache action

mainframev:


.github/workflows/clear-cache.yml
```diff
@@ -0,0 +1,31 @@
+name: Clear branch cache
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch_number:
+        description: "Branch number"
```
DSil, line 7:
```suggestion
        description: "PR number"
```

The cache is associated with the PR number. So calling it PR number makes it clearer, imo

mainframev: agree, although it's kinda WIP :)

DSil: Yes, yes, it was not a formal review, just a suggestion

mainframev:
@DSil this action seems to be not needed, as we can delete it manually in managements: 

Screenshot 2023-10-05 at 11 47 12

PR #4071, opened 2023-10-04
fix(Stack): set flex back to false by default

mainframev:
I think I found a bug in Stack. Looks like we made a mistake, as before it had flex by default as false. So with the current Stack, there are cases when it's visually broken, when users do not add flex prop, only *spacing* and it relies on margins. Take a look for [instance here](https://kiwicom.github.io/orbit/?path=/story/itinerary--segment) -> now it's with flex and gap and the content is not shrinking, but overflowing out of the viewport when decreasing the window size.

So it was a breaking change and not documented. I am also worrying about how much it will produce bugs, so I would prefer to return it as it was.   
 Storybook: https://orbit-mainframev-fix-stack.surge.sh


DSil (commented):
Well spotted! I left a comment because this fix introduces another potential issue, because of class orders. Please take a look on that.

packages/orbit-components/src/Stack/index.tsx
```diff
@@ -132,17 +139,19 @@ const Stack = (props: Props) => {
...
       typeof spaceAfter !== "undefined" && getSpaceAfterClasses(spaceAfter, viewport),
-      typeof align !== "undefined" && getAlignItemsClasses(align, viewport),
-      typeof align !== "undefined" && getAlignContentClasses(align, viewport),
-      typeof wrap !== "undefined" && getWrapClasses(wrap, viewport),
-      typeof grow !== "undefined" && getGrowClasses(grow, viewport),
-      typeof shrink !== "undefined" && getShrinkClasses(shrink, viewport),
-      typeof justify !== "undefined" && getJustifyClasses(justify, viewport),
-      getDirectionClasses(direction, viewport),
+      (flex || inline) && [
+        flex && "flex",
```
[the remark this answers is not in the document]
mainframev: also, it's a good candidate for adding visual tests for chromatic. Probably the best one.

packages/orbit-components/src/Stack/index.tsx
```diff
@@ -132,17 +139,19 @@ const Stack = (props: Props) => {
...
+        typeof wrap !== "undefined" && getWrapClasses(wrap, viewport),
+        typeof grow !== "undefined" && getGrowClasses(grow, viewport),
+        typeof shrink !== "undefined" && getShrinkClasses(shrink, viewport),
+        typeof justify !== "undefined" && getJustifyClasses(justify, viewport),
+        getDirectionClasses(direction, viewport),
+      ],
       getSpacingClasses(spacing, viewport, direction, legacy),
-      inline && getDisplayInlineClass(inline, viewport),
+      inline != null && getDisplayInlineClass(inline, viewport),
       inline === false && "w-full",
```
DSil, line 154: Does this have to explicitly be false? Can't it be undefined to apply `w-full`?

packages/orbit-components/src/Stack/index.tsx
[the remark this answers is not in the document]
DSil: That can't be like that either. If one adds the prop `flex`, it will have both `flex` and `block` classes, which are incompatible.

My suggestion was covering that.

If `inline || flex`:
- If `true`, check if inline:
  - if `true`, apply `inline-flex`
  - if `!true`, apply `flex`
  - Apply all the other classes normally, that we currently have inside the array
- If `false`, apply `block`

This would go with something like

```js
flex || inline
  ? [
      inline ? "inline-flex" : "flex",
      typeof align !== "undefined" && getAlignItemsClasses(align, viewport),
      typeof align !== "undefined" && getAlignContentClasses(align, viewport),
      typeof wrap !== "undefined" && getWrapClasses(wrap, viewport),
      typeof grow !== "undefined" && getGrowClasses(grow, viewport),
      typeof shrink !== "undefined" && getShrinkClasses(shrink, viewport),
      typeof justify !== "undefined" && getJustifyClasses(justify, viewport),
      getDirectionClasses(direction, viewport),
    ]
  : "block",
```

**However**, while writing this, I also spotted that we are not taking the flex prop into consideration for viewports (never were)… So it seems to require more refactor 🤔 

IMO, the following would work:

Have a `getDisplayClass(flex, inline, viewport)` and inside the function get the correct class based on the `flex` and `inline` arguments:
- If `inline` => class `inline-flex`
- If `flex` => class `flex`
- else => class `block`
with the corresponding viewport prefixes, as the object already does.

After having this, we unconditionally call this function.

packages/orbit-components/src/Stack/index.tsx
```diff
@@ -132,17 +139,21 @@ const Stack = (props: Props) => {
...
+            typeof grow !== "undefined" && getGrowClasses(grow, viewport),
+            typeof shrink !== "undefined" && getShrinkClasses(shrink, viewport),
+            typeof justify !== "undefined" && getJustifyClasses(justify, viewport),
+            getDirectionClasses(direction, viewport),
+          ]
+        : "block",
       getSpacingClasses(spacing, viewport, direction, legacy),
-      inline && getDisplayInlineClass(inline, viewport),
+      inline && getDisplayClasses("inline-flex", viewport),
+      flex && getDisplayClasses("flex", viewport),
```
DSil, line 155: This might cause a conflict if both `inline` and `flex` props are true, as it will apply both classes. I had thought about that and that is why I suggested that the function received the props as arguments. Why did you decide to do it differently?

Also, docs state, about the `inline` prop:
> If `true`, the Stack will have `display` set to `inline-flex`, otherwise it will be `flex`. 

But this is incorrect. I would remove everything after the comma.

mainframev: that does not make sense to apply them both

DSil: It might not but:
- Nothing prevents them from doing it (and we're not adding conditional types)
- There was no problem in the past because SC would resolve the conflict, but now nothing resolves the conflicts, so we must do it ourselves

packages/orbit-components/src/Stack/index.tsx
DSil: Just missing this

mainframev: I think it's actually correct. Take a look on it [here](https://github.com/kiwicom/orbit/blob/dependabot/npm_and_yarn/remark-frontmatter-4.0.1/packages/orbit-components/src/Stack/helpers/getWidth.js#L5) 

So it has to be defined, but have a false 🤔

DSil: Ok, thank you 👍

DSil (approved):

PR #4072, opened 2023-10-05
chore(Switch): use proper classes

mainframev:
Had no time to look at Switch PR. Here are few small 💅🏻 
 Storybook: https://orbit-mainframev-chore-switch-correct-classes.surge.sh

### DIFF
--- a/packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
+++ b/packages/orbit-components/src/tmp-tailwind/Switch/index.tsx
@@ -14,7 +14,7 @@ const Switch = React.forwardRef<HTMLInputElement, Props>(
         <div
           className={cx(
             "duration-fast relative flex h-[20px] w-[40px] items-center justify-between rounded-[100px] transition-colors",
-            disabled ? "opacity-50" : "cursor-pointer",
+            disabled ? "cursor-not-allowed opacity-50" : "cursor-pointer",
             checked ? "bg-blue-normal" : "bg-cloud-dark",
           )}
         >
@@ -38,7 +38,7 @@ const Switch = React.forwardRef<HTMLInputElement, Props>(
             className={cx(
               "rounded-circle bg-white-normal duration-fast shadow-switch absolute box-border inline-flex h-[24px] w-[24px] items-center justify-center",
               "peer-focus:outline-blue-normal peer-focus:outline peer-focus:outline-2",
-              "[&_svg]:h-[12px] [&_svg]:w-[12px]",
+              "[&_svg]:h-icon-small [&_svg]:w-icon-small",
               !disabled && "active:shadow-action-active",
               !checked && (icon ? "[&_svg]:text-ink-normal" : "[&_svg]:text-cloud-dark"),
               checked
@@ -46,7 +46,7 @@ const Switch = React.forwardRef<HTMLInputElement, Props>(
                 : "left-[-3px]",
             )}
           >
-            {icon || <Circle />}
+            {icon || <Circle size="small" />}
           </div>
         </div>
       </label>


DSil (changes requested):
To be honest, I only agree with `cursor-not-allowed`, even though it was not on the original component, I agree, it can be added.

Regarding the background, there is not obvious relation between the Switch "ball" with Inputs. I would not relate them via a token. It's white, and white should be kept. Inputs can change color without affecting the Switch color.

Regarding the height and width of the "ball", the answer is the same. It's part of the Switch component. It is not an icon so it should not use an icon token. The dimensions happen to be the same, but the relationship is not tied. Icons can change size without affecting the Switch component.

mainframev:
@DSil it's also the form element, so it should have the background token from form element

mainframev:
@DSil also just found, that it should have circle 16px, not 12px. So both should have token icon, as it corresponds to circle icon and if the icon will change the size -> it's also should be changed

DSil:
Ok, so the issue is different here. The component was wrong / misaligned with Figma. This is not a problem of the migration. It was a problem we had before and that I was not concerned as I did a direct migration of what we had. I am ok with fixing it and adding the cursor and the icon tokens, as it refers to the svg inside (although it feels weird to target SVG, as it breaks the Icon component abstraction).

Regarding the background color, the concept of input is very different here. Yes, the Switch is an input and a visually different checkbox. But if we look at the token and its usage, it refers to the background of the place where the user actually inputs stuff. The place where the user writes or checks (in case of Checkbox and Radio).
The handle of a Switch is not the place to input stuff. It's a point of interaction, to click and toggle a state.
I honestly can't see the relationship between changing the background of a text input and wanting it to immediately affect the background of a Switch handle…

mainframev:
> Ok, so the issue is different here. The component was wrong / misaligned with Figma. This is not a problem of the migration. It was a problem we had before and that I was not concerned as I did a direct migration of what we had. I am ok with fixing it and adding the cursor and the icon tokens, as it refers to the svg inside (although it feels weird to target SVG, as it breaks the Icon component abstraction).
> 
> Regarding the background color, the concept of input is very different here. Yes, the Switch is an input and a visually different checkbox. But if we look at the token and its usage, it refers to the background of the place where the user actually inputs stuff. The place where the user writes or checks (in case of Checkbox and Radio). The handle of a Switch is not the place to input stuff. It's a point of interaction, to click and toggle a state. I honestly can't see the relationship between changing the background of a text input and wanting it to immediately affect the background of a Switch handle…

You can either target it via class as we are doing right now or do cloneElement and apply size prop to small. As there can be any of our icons. But it is what it is and I just used the class. 

> Regarding the background color, the concept of input is very different here. Yes, the Switch is an input and a visually different checkbox. But if we look at the token and its usage, it refers to the background of the place where the user actually inputs stuff. The place where the user writes or checks (in the case of Checkbox and Radio).

Not sure if I agree with that, for now, every form element by default should have a white background and it should be under this token, not depending on if the user types there, or if it's a checkbox, radio, or switch. Switch is kinda similar to Checkbox and Radio, we use this token there, so should use it also here. I think you are thinking too much about that I appreciate it, but let's just approach it a bit simpler and use it everywhere ;)

DSil:
Problem is that, precisely, we're not targeting via class. That would not break the abstraction barrier. We're targeting via HTML element. If, for whatever reason, we change how we render icons (font-based approach, for instance, without svg element), this will break.

I don't think it's a problem worth any action right now, though, as that change is currently not expected at any time.

On another note, and since this actually fixes bugs on the component, can you please reword the commit to call it a fix? I'd go with something like `fix(Switch): adjust icon size and cursor on disabled`, or something similar.

DSil:
Yes, it's complicated, I also realized that. It's just that when we move it out of the tmp folder we either do it in two separate commits (that's weird 😅) or we just comment it as a feat and never document the fix on the changelog.

Honestly, I'm fine with any approach, so non-block on this because of that .

DSil (approved):

PR #4073, opened 2023-10-05
fix(ItinerarySegmentDetail): fix overflowing content issue

mainframev:
[Reported in Slack](https://skypicker.slack.com/archives/C7T7QG7M5/p1695212720944469)

I tested that on Booking directly by applying the calc logic
 Storybook: https://orbit-mainframev-fix-itinerary-details.surge.sh

### DIFF
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
@@ -21,7 +21,9 @@ import { usePart } from "../context";
 import { useWidth } from "../../context";
 import AirplaneDown from "../../../icons/AirplaneDown";
 import type { Props } from "./types";
+import type { SpaceAfterSizes } from "../../../common/types";
 import ItineraryIcon from "../ItineraryIcon";
+import getSpacingToken from "../../../common/getSpacingToken";
 
 const StyledWrapper = styled.div`
   width: 100%;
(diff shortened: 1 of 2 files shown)


packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
```diff
@@ -128,6 +130,20 @@ const StyledIcon = styled.div<{ isFirst?: boolean; isLast?: boolean }>`
   `}
 `;
 
+const StyledItemWrapper = styled.div<{ $offset: number; spaceAfter: SpaceAfterSizes }>`
+  ${({ theme, $offset }) => css`
+    display: flex;
+    align-items: center;
+    flex-direction: column;
+    width: calc(100% - ${parseInt(theme.orbit.spaceXSmall, 10) + $offset}px);
```
oreqizer, line 138: wouldn't it be better to do `width: 100%` here and put `px-xs` here to achieve the desired effect?

mainframev: Unfortunately It is not possible, because offset is calculated for each Itinerary, and with width: 100% it will be overflowing as before on smallMobile :)

oreqizer: oh right there's the offset thing. okay

oreqizer (approved):

PR #4076, opened 2023-10-09
docs: add restore to default button to playground

mainframev:
As we already have multiple requests about clearing the local storage in playground, adding that feature there too, as in the examples. 
 Storybook: https://orbit-mainframev-docs-add-refresh-button.surge.sh


mvidalgarcia (approved):
@mainframev In my opinion this helps and I think it's better to have this button than not having it at all.

However, I think this does not solve the root problem. As a user, if I access the playground or the examples and get an output error, I wouldn't know that I have to press the refresh button.

We can discuss this further in one of our syncs. Approving as this is already an improvement.

PR #4077, opened 2023-10-09
feat(Hide): migrate to Tailwind

mainframev:
Decided not to use the tmp folder for this component, as it can be directly migrated to Tailwind imo, it's small. The same how we did it for FormLabel :) 

[FEPLT-1727](https://kiwicom.atlassian.net/browse/FEPLT-1727)

 Storybook: https://orbit-mainframev-feat-migrate-hide.surge.sh


mvidalgarcia:
There's something fishy.

image

Desktop viewport
Hide on tablet
Chevron is not visible

Can you reproduce it?

mvidalgarcia (commented):
There's something fishy.

image

Desktop viewport
Hide on tablet
Chevron is not visible

Can you reproduce it?

mainframev:
> Hide on tablet

are u sure? I've checked, seems to work fine 
ezgif com-video-to-gif

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -26,6 +26,8 @@ const COLORS: Partial<ExportedComponentLevelTokens>[] = [
   "countryFlag",
 ];
 
+const px = (n: number) => `${n}px`;
```
mvidalgarcia, line 29: I believe these changes should be in a different commit with scope `tailwind`

packages/orbit-components/src/Layout/LayoutColumn/index.tsx
```diff
@@ -4,13 +4,30 @@ import * as React from "react";
 import styled, { css } from "styled-components";
 
 import defaultTheme from "../../defaultTheme";
-import getViewportHideStyles from "../../Hide/helpers/getViewportHideStyles";
```
mvidalgarcia, line 7: Wouldn't it be better to keep this function for now until the Layout component is migrated?

packages/orbit-tailwind-preset/src/presets/components/index.ts
mainframev: do not think so, it's still will be a chore that later will be squashed

packages/orbit-components/src/Layout/LayoutColumn/index.tsx
mainframev: I found that better to remove it and just directly past it inside the Layout. It would make sense to keep inside Stack, if it were used on multiple places

packages/orbit-components/src/Hide/index.tsx
```diff
@@ -1,26 +1,23 @@
...
-StyledHide.defaultProps = {
-  theme: defaultTheme,
-};
 
 const Hide = ({ on = [], block, children }: Props) => (
-  <StyledHide $on={on} $block={block}>
+  <div
+    className={cx({
+      block,
+      "inline-block": block == null,
```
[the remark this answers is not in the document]
mainframev: I simplified that to *block ? "block" : "inline-block"*

packages/orbit-components/src/Layout/LayoutColumn/index.tsx
mvidalgarcia: Ok, I see, not really strong opinion on this.

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -73,6 +75,25 @@ const cfg = (options?: Options): Config => {
...
+          },
+          "tb-de": {
+            min: px(defaultTokens.breakpointTablet),
+            max: px(defaultTokens.breakpointDesktop - 1),
+          },
+          "de-ld": {
+            min: px(defaultTokens.breakpointDesktop),
+            max: px(defaultTokens.breakpointLargeDesktop - 1),
+          },
+        },
```
mvidalgarcia, line 96: Do these classes belong to the `components` preset instead of `foundation`? The rest of the classes I see in the `components` preset have a prefix with the component name, but in this case they haven't.

mvidalgarcia: Wouldn't they fit better [here](https://github.com/kiwicom/orbit/blob/feat/migrate-hide/packages/orbit-tailwind-preset/src/presets/foundation/theme/screens.ts)?

mainframev: Yes, it has no component prefix, but it's part of the component preset (I know that we merged them), we document only the foundation, so that's why it's under chore. Do you think we should move that to the foundation and allow these range mq? I feel like we should avoid it and use only a mobile-first approach and these range queries might be useful in components like Hide 🤔

mvidalgarcia: I don't know, to be honest. I just wanted to have a little bit more context on the difference between both presets. I guess that consumers can achieve the same behavior via the Hide component, so no need to document those classes, is this correct?

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -73,6 +75,25 @@ const cfg = (options?: Options): Config => {
     },
     theme: {
       extend: {
+        screens: {
+          "sm-mm": { max: px(defaultTokens.breakpointMediumMobile) },
```
mvidalgarcia, line 79:
```suggestion
          "sm-mm": { max: px(defaultTokens.breakpointMediumMobile - 1) },
```

mvidalgarcia (approved):
LGTM after the changes 👍 

PS. Remember to rebase and squash.

packages/orbit-tailwind-preset/src/presets/components/index.ts
mainframev: Yep, for hiding the content they can use Hide and do not need these range classes. To me, it's two different things and till now Orbit provided only mobile-first media queries, without ranges. So I think we should not introduce that now, they can add it to their configs if they want, but I guess it's quite a rare use case when it's needed.

PR #4079, opened 2023-10-09
Add visual test stories

oreqizer:
.
 Storybook: https://orbit-mainframev-oreqizer-test-stories.surge.sh

### DIFF
--- a/packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
+++ b/packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
@@ -36,14 +36,15 @@ Playground.story = {
   },
 };
 
-export const ListOfAllAirportIllustrations = () => {
+export const AirportIllustrationVisualTest = () => {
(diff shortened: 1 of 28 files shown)


packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
```diff
@@ -36,14 +36,15 @@ Playground.story = {
...
   return <IllustrationPrimitiveList nameOfComponent="AirportIllustration" images={NAMES} />;
 };
 
-ListOfAllAirportIllustrations.story = {
+AirportIllustrationVisualTest.story = {
   name: "List of all AirportIllustrations",
 
   parameters: {
     info: "Explore our new set of Airportillustrations for Kiwi.com.",
+    chromatic: { disableSnapshot: false },
```
DSil, line 48: I'd add a mode here so it only triggers snapshot for largeDesktop. No need to do it for all viewports

DSil: Agreed IRL that I'd do it

DSil (approved):

PR #4085, opened 2023-10-10
feat(Stepper): migrate Stepper to Tailwind

mainframev:
[FEPLT-1696](https://kiwicom.atlassian.net/browse/FEPLT-1696)
 Storybook: https://orbit-mainframev-tw-migrate-stepper.surge.sh

[FEPLT-1696]: https://kiwicom.atlassian.net/browse/FEPLT-1696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-components/src/Stepper/StepperStateless/index.tsx
+++ b/packages/orbit-components/src/Stepper/StepperStateless/index.tsx
@@ -1,118 +1,33 @@
 "use client";
 
 import * as React from "react";
-import styled, { css } from "styled-components";
+import cx from "clsx";
 
-import mq from "../../utils/mediaQuery";
 import Minus from "../../icons/Minus";
 import Plus from "../../icons/Plus";
-import defaultTheme from "../../defaultTheme";
-import Button from "../../primitives/ButtonPrimitive";
+import ButtonPrimitive from "../../primitives/ButtonPrimitive";
 import useTheme from "../../hooks/useTheme";
-import { defaultFocus } from "../../utils/common";
 import type { Props } from "./types";
 
-const getMaxWidth = ({ maxWidth }) => {
+const getMaxWidth = ({ maxWidth }: { maxWidth: string | number }) => {
   if (typeof maxWidth === "string") return maxWidth;
-  return `${parseInt(maxWidth, 10)}px`;
+  return `${parseInt(maxWidth.toString(), 10)}px`;
 };
 
-const StyledStepper = styled.div`
-  display: flex;
-  width: 100%;
-  max-width: ${getMaxWidth};
-  flex: 1 1 auto;
-`;
-
-const iconMixin = css<{ isActive?: boolean; isDisabled?: boolean }>`
-  ${({ theme, isActive, isDisabled }) => css`
-    padding: 2px;
-    height: 20px;
-    width: 20px;
-    background: ${isActive ? theme.orbit.paletteBlueNormal : theme.orbit.paletteCloudNormal};
-    border-radius: ${theme.orbit.borderRadiusCircle};
-    ${mq.desktop(css`
-      height: 16px;
-      width: 16px;
-    `)};
-
-    ${!isDisabled &&
-    css`
-      &:hover {
-        background: ${isActive
-          ? theme.orbit.paletteBlueNormalHover
-          : theme.orbit.paletteCloudNormalHover};
-      }
-
-      &:focus,
-      &:active {
-        box-shadow: inset 0 0 0 2px
-          ${isActive ? theme.orbit.paletteBlueLightActive : theme.orbit.paletteCloudNormalActive};
-      }
-    `};
-  `}
-`;
-
-const StyledMinusIcon = styled(Minus)`
-  ${iconMixin};
-`;
-
-StyledMinusIcon.defaultProps = {
-  theme: defaultTheme,
-};
-
-const StyledPlusIcon = styled(Plus)`
-  ${iconMixin};
-`;
-
-StyledPlusIcon.defaultProps = {
-  theme: defaultTheme,
-};
-
-const StyledStepperInput = styled.input`
-  ${({ theme, disabled }) => css`
-    width: 100%;
-    height: 44px;
-    padding: 0;
-    border: 0;
-    font-size: ${theme.orbit.fontSizeTextLarge};
-    font-weight: ${theme.orbit.fontWeightMedium};
-    color: ${theme.orbit.paletteInkDark};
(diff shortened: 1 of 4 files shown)


mainframev:
TODO: [reported here](https://skypicker.slack.com/archives/C7T7QG7M5/p1697558935054119?thread_ts=1697532364.353809&cid=C7T7QG7M5), after ButtonPrimitive refactoring to Tailwind

mvidalgarcia:
Remember to update the Tailwind migration status YAML file 😃

mainframev:
> Remember to update the Tailwind migration status YAML file 😃

Yeah, I did, just forgot to push 😭 Now I am looking into the Button issue above, but it's ready for review in Stepper, I will probably do it in separated PR

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -1,118 +1,34 @@
...
-    }
-  }
-`;
-
-StyledStepperButton.defaultProps = {
-  theme: defaultTheme,
-};
+const stepperButtonMixin = ({ disabled, active }: { disabled: boolean; active?: boolean }) =>
+  cx(
+    "bg-transparent",
```
mvidalgarcia, line 19: Minor: Is this necessary? 

It doesn't harm, but I believe Tailwind has transparent bg in buttons by default.

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -1,118 +1,34 @@
...
-          ? theme.orbit.paletteBlueNormalHover
-          : theme.orbit.paletteCloudNormalHover};
-      }
-
-      &:focus,
-      &:active {
-        box-shadow: inset 0 0 0 2px
-          ${isActive ? theme.orbit.paletteBlueLightActive : theme.orbit.paletteCloudNormalActive};
-      }
-    `};
```
mvidalgarcia, line 52: Are these styles translated to Tailwind?

When clicking on the button on master there's a inline ring, I guess that's missing

image

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -151,20 +62,34 @@ const StepperStateless = ({
...
         title={titleDecrement}
         icons={iconStyles}
-        {...commonButtonStyles}
       />
-      <StyledStepperInput
+      <input
+        className={cx(
+          "h-form-box-normal w-full min-w-0 border-0",
+          "p-0 text-center",
+          "text-form-element-large text-form-element-filled-foreground font-medium",
```
mvidalgarcia, line 87: I believe `text-form-element-large` does not exist.

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -151,20 +62,34 @@ const StepperStateless = ({
     disabled || disabledIncrement || (typeof value === "number" && value >= +maxValue);
 
   return (
-    <StyledStepper data-test={dataTest} id={id} maxWidth={maxWidth}>
-      <StyledStepperButton
+    <div
+      data-test={dataTest}
+      id={id}
+      className="flex w-full flex-1"
```
mvidalgarcia, line 68:
```suggestion
      className="flex w-full flex-auto"
```

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -151,20 +62,34 @@ const StepperStateless = ({
...
+      style={{ maxWidth: getMaxWidth({ maxWidth }) }}
+    >
+      <ButtonPrimitive
+        className={stepperButtonMixin({ disabled: isPlusDisabled, active })}
```
mvidalgarcia, line 72:
```suggestion
        className={stepperButtonMixin({ disabled: isMinusDisabled, active })}
```

mvidalgarcia (commented):
Found some minor issues, let me know if it makes sense.

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
mainframev: hm, not tailwind, but our primitive, so ok, can be removed, thanks!

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
```diff
@@ -1,118 +1,33 @@
...
+    "[&_svg]:p-xxxs",
+    "de:[&_svg]:h-icon-small de:[&_svg]:w-icon-small [&_svg]:h-icon-medium [&_svg]:w-icon-medium",
+    "[&_svg]:rounded-circle",
+    "focus:outline-0",
+    "[&_svg]:focus:outline-blue-normal [&_svg]:focus:outline [&_svg]:focus:outline-2",
+    !disabled && [
+      active
+        ? "[&_svg]:bg-blue-normal hover:[&_svg]:bg-blue-normal-hover [&_svg]:active:focus:shadow-[inset_0_0_0_2px_var(--palette-blue-light-active)]"
+        : "[&_svg]:bg-cloud-normal hover:[&_svg]:bg-cloud-normal-hover [&_svg]:active:focus:shadow-[inset_0_0_0_2px_var(--palette-cloud-normal-active)]",
+    ],
```
mvidalgarcia, line 28: I think that with these classes it has to be both `active` **and** `focus` at the same time to apply the style, as opposed to the current master where the style is applied when is either `active` **or** `focus`.

Plus, the pseudo-classes here target the button instead of the svg (icon), and then correctly apply the style to the svg.

I believe there's not much difference in terms of behavior when interacting with the component, but it differs from the original components.

mvidalgarcia: Said that, I believe we can keep it as it is if changing it would complicate things a lot. In the end, I believe it will be the same for the final user.

mvidalgarcia (approved):
Approving as it's good enough from the user's perspective, to prevent it from being blocked.

packages/orbit-components/src/Stepper/StepperStateless/index.tsx
mainframev: Thanks! I'd say it's ok to have it like that, if there's no difference

PR #4086, opened 2023-10-12
feat(tailwind): add foundation option and css variables

mainframev:
- [x] create palette css variables output for tokens
- [x] add foundation option to tailwind config 
- [x] update readme 
- [x] add css variables to palette colors 
- [x] add utility function to design tokens to transform object into css vars

There are several problems with our current Tailwind preset: 

- We do not provide the option of customizing the foundation. It was possible before with styled-components, where users could use the [getTokens](https://github.com/kiwicom/orbit/tree/master/packages/orbit-design-tokens#generating-your-theme) function from the tokens package and simply put it into OrbitProvider and it's documented in [Theming section](https://github.com/kiwicom/orbit/blob/master/.github/theming.md)

- The second problem, which is partially related to the previous, is related to our whitelabels, since the migration to Tailwind is already happening, whitelabels should be customizable the same as it's currently with styled components. However, there is a problem because we can use only the single Tailwind build and can't change it in runtime ([more info here](https://skypicker.slack.com/archives/C05FDP4UM3N/p1697113091295059)). TLDR -> we can use the CSS variables for our palette colors to make it work.  

 Storybook: https://orbit-mainframev-feat-tailwind-use-css-vars.surge.sh


packages/orbit-design-tokens/src/dictionary/utils/is.ts
```diff
@@ -6,7 +6,7 @@ const isTypeOf = ({ type, category }: DesignToken, typeName: string): boolean =>
 export const isColor = (prop: DesignToken): boolean => isTypeOf(prop, "color");
 
 export const isColorPalette = (prop: DesignToken): boolean =>
-  isColor(prop) && prop.attributes.name.startsWith("palette");
+  isColor(prop) && prop.path.includes("palette");
```
RobinCsl, line 9: Why is this change necessary? It seems like all the color palette tokens do start with `palette`.

packages/orbit-design-tokens/src/dictionary/formatters/jsFoundation.ts
```diff
@@ -27,7 +27,7 @@ import { falsyString } from "../utils/string.js";
 const genericFactory = (allProperties: DesignToken[], platform: "javascript" | "typescript") => {
   const upperFirst = determinateUpperFirst(platform);
   const withExport = platform !== "javascript";
-  const variableType = platform !== "javascript" ? "type" : "const";
+  const variableType = platform !== "javascript" ? "interface" : "const";
```
RobinCsl, line 30: Could you explain the reason for this change? Why are interfaces preferable to types?

packages/orbit-design-tokens/src/dictionary/utils/is.ts
mainframev: I feel like it's better to get them via path, rather than name and attributes.

packages/orbit-design-tokens/src/dictionary/formatters/jsFoundation.ts
mainframev: Because we have a eslint rule *"@typescript-eslint/consistent-type-definitions": ["error", "interface"]* and when it was creating types it had to be formatted with eslint and prettier to have it correct.

packages/orbit-tailwind-preset/src/presets/fns.ts
RobinCsl, line 1: The name of the file is not very descriptive. Could you please rename it?

packages/orbit-tailwind-preset/src/presets/foundation/transformToCssVariables.ts
RobinCsl, line 1: Would it make sense to add a test for this functionality?

packages/orbit-design-tokens/src/utils/tokensToCssVars.ts
```diff
@@ -0,0 +1,17 @@
...
+type TokensObject = Record<string, string | Record<string, string>>;
+
+const getCssVars = <T = TokensObject>(object: T, prefix = `-`) =>
+  Object.entries(object).reduce((css, [key, value]) => {
+    const newPrefix = kebabCase(`${prefix}-${key}`);
+    if (typeof value !== "object") return `${css + newPrefix}: ${value};\n`;
+    return css + getCssVars(value, newPrefix);
+  }, ``);
+
+const tokensToCssVars = <T = TokensObject>(tokens: T, cssClass?: boolean) => {
```
RobinCsl, line 12: Could we change the signature to be
```ts
const tokensToCssVars = <T = TokensObject>({tokens, cssClass = ":root"}: { tokens: T, cssClass?: string }) => {
```

packages/orbit-design-tokens/src/dictionary/utils/is.ts
RobinCsl: Okay, but is it contributing to the success of this contribution? If not, I would say it's out of scope and could/would be introduced in a different PR as it is a distraction.

packages/orbit-design-tokens/src/dictionary/formatters/jsFoundation.ts
RobinCsl: Okay, but why is this change happening now? What does it have to do with the goal of this PR?

mainframev: I personally do not mind having such kind of improvements alongside, if there aren't too many and they committed well. Will do it separately then

RobinCsl: It's not about you, @mainframev.

The PR has +2,500 -1,900 lines of code, that's a lot of changes to review.

Anything that is off-topic should be avoided, it just adds noise and confusion to what you're trying to accomplish.

It makes reviewing harder, it increases the risk of this PR being delayed, etc.

packages/orbit-tailwind-preset/src/presets/foundation/transformToCssVariables.ts
mainframev: not needed anymore, removing that after Boris suggestion

packages/orbit-tailwind-preset/src/presets/fns.ts
mainframev: Renamed to "utils". As Boris usually says, those generic names for files aren't good. I could take them out of the file and create a separate file for each or you may suggest a better name, nothing better comes to me

packages/orbit-design-tokens/src/__tests__/tokensToCssVars.ts
```diff
@@ -0,0 +1,45 @@
...
+      ":root {
+        --palette-product-normal: #ff0000;
+        --palette-product-normal-hover: #e50000;
+        --palette-product-normal-active: #cc0000;
+
+      }"
+    `);
+  });
+
+  it("should convert nested object", () => {
```
oreqizer, line 30: do we really need this?

packages/orbit-design-tokens/src/__tests__/tokensToCssVars.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const tokens = {
+  paletteProductNormal: "#ff0000",
+  paletteProductNormalHover: "#e50000",
+  paletteProductNormalActive: "#cc0000",
+};
+
+describe("transformToCssVars", () => {
+  it("should convert object into css variables string", () => {
+    expect(transformToCssVars({ tokens, cssClass: false })).toMatchInlineSnapshot(`
```
oreqizer, line 11: we imo do not need the functionality offered when `cssClass: true`. just the list of var defs is enough

packages/orbit-design-tokens/src/utils/tokensToCssVars.ts
```diff
@@ -0,0 +1,23 @@
+export const kebabCase = (str: string) => str.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase();
```
oreqizer, line 1: let's pls _not_ do these complex mappers. we **just need the palette**. let's keep it simple and maintain an object of `Record<PaletteObject, CssVar>` and keep the function _really_ simple

Screenshot 2023-10-13 at 17 25 14

this is what I mean. this is what modules have, and this is the _only_ thing that is themable currently. let's keep an object that maps these values to CSS variables

if the palette will be extended, to e.g. `productInk` colours, we can simply add them to this object

same as in `fromPlainObject`:

```ts
interface ThemePaletteColors {
    productLight: string;
    productLightHover: string;
    productLightActive: string;
    productNormal: string;
    productNormalHover: string;
    productNormalActive: string;
    productDark: string;
    productDarkHover?: string;
    productDarkActive?: string;
    productDarker?: string;
}
interface FromPlainObject {
    (themePaletteColors: ThemePaletteColors): Tokens$1;
}
```

packages/orbit-tailwind-preset/src/presets/foundation/__tests__/createCustomFoundation.test.ts
```diff
@@ -0,0 +1,177 @@
+import { defaultFoundation } from "@kiwicom/orbit-design-tokens";
+
+import { createCustomFoundation } from "../createCustomFoundation";
+
+describe("createCustomFoundation", () => {
```
oreqizer, line 5: I don't know what all these tests and functions are for 😕 they all seem rather extra to what the scope of this MR is

packages/orbit-tailwind-preset/src/presets/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const kebabCase = (str: string) => str.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase();
```
oreqizer, line 1: pls let's be more explicit instead of doing these wild string mapping things

oreqizer (changes requested):
please let's simplify these changes Viktor to what the scope of the PR is — adding vars to the **palette only**, and the one small utility for mapping the **palette** (and nothing else) to CSS vars. 🙏 

honestly, I didn't even go thru all the changes bcs there's _just so many_

packages/orbit-design-tokens/src/__tests__/tokensToCssVars.ts
mainframev: it's might be needed, palette is nested object

packages/orbit-design-tokens/src/utils/tokensToCssVars.ts
mainframev: fromPlainObject is not meant to be used like that. It will return you all the tokens, not just the flat object of the foundation. So if you would like to do something like: 

```
tokensToCssVars(fromPlainObject(foundation));
```

you will get the large object with all tokens transformed into css variables.

Besides that, as I already mentioned, if we would like to have this utility function to be the part of tokens package, then it should support nested objects, to make it fully compatible with our tokens and foundation. If you need a simple wrapper that will produce the css var from a plain object only, then it can be implemented right where the WL foundation is defined. 

_About complexity:_ 
  The function logic is 5 lines of code. It uses reduce and recursion, a helper function to transform from camelCase to kebabCase, sorry, but it looks quite simple and not complex at all. It's not even a cost to have a good function that fits our API and it won't be dependent on the side logic of normalization of objects from nested to flat.

packages/orbit-tailwind-preset/src/presets/foundation/__tests__/createCustomFoundation.test.ts
mainframev: Partially yes, it can be split on two MRs, but the logic inside will have to be still adjusted to make it themeable. The difference is whether the config will receive an additional parameter or not. (this can be abstracted to separated MR). 

The current problem -> when we define the palette with css variables, it won't have any effect on other tokens that are related to components, because they are now defined with `defaultTokens.<token>`. So the logic has to be adjusted to use `getTokens(cssVarsFoundation)` and passed down to both presets, so all the tokens will inherit the correct foundation. So that's why the PR eventually is bigger than it was expected, it's not just rewriting the values of the palette to css variables. Plus I am removing in this PR the things that are imo overengineered like this [one](https://github.com/kiwicom/orbit/blob/master/packages/orbit-tailwind-preset/src/presets/foundation/helpers.ts#L64C14-L64C34). I understand that it's a lot of change at once, so I will try to split it, but we have to also agree, that sometimes things can be significantly smaller as the problem is not that easy as it might look :)

mainframev:
Closing in favor of #4087 

Thanks guys for the review ❤️ 

The new PR solves only the scope of adding the css variables foundation to the tailwind preset. Nothing more. Although it's still big, it is what it is and I doubt it can be smaller because the problem is more complex than just changing palette colors

The css utility function will be added in a separate MR and the option for tailwind preset for custom foundation will be added separately :)

PR #4088, opened 2023-10-15
chore(tokens): add tokensToCssVars utility function

mainframev:
[Doc](https://docs.google.com/document/d/13q8UrRqCzvnQ9Pv6wH5QS_7SG4eeC7hdXTtvqNhWc9k/edit?usp=sharing) 
[Slack thread
](https://skypicker.slack.com/archives/C05FDP4UM3N/p1697428801266619)
 Storybook: https://orbit-mainframev-feat-tokens-add-css-variables-utility.surge.sh


packages/orbit-components/src/OrbitProvider/index.tsx
```diff
@@ -22,6 +23,17 @@ import type { Props } from "./types";
  *
  */
 const OrbitProvider = ({ theme, children, useId }: Props) => {
+  React.useEffect(() => {
+    const ID = "orbit-theme-css-vars";
+    const style = document.createElement("style");
+    style.setAttribute("id", ID);
+    style.innerText = tokensToCssVars({ tokens: theme.orbit }).replace(/\s/g, "");
```
DSil, line 30: This doesn't work. It needs a selector.

```suggestion
    style.innerText = `
    :root {
      ${tokensToCssVars({ tokens: theme.orbit }).replace(/\s/g, "")}
    }
    `;
```

DSil (changes requested):

packages/orbit-components/src/OrbitProvider/index.tsx
```diff
@@ -22,6 +23,20 @@ import type { Props } from "./types";
  *
  */
 const OrbitProvider = ({ theme, children, useId }: Props) => {
+  React.useEffect(() => {
+    const ID = "orbit-theme-css-vars";
+    const style = document.createElement("style");
+    style.setAttribute("id", ID);
+    style.innerText = tokensToCssVars({ tokens: theme.orbit, cssClass: ":root" }).replace(
+      /\s/g,
+      "",
```
DSil, line 32: I would prefer this replace to be inside the function, and not here, if possible.

DSil (commented):
LGTM, just one question.

Also, this should be a fix commit, no? This is a broken feature we currently have.

mainframev:
> LGTM, just one question.
> 
> Also, this should be a fixed commit, no? This is a broken feature we currently have.

Do you mean the OrbitProvider adjustment? no, it's a chore, for the user API is not changed 🤔 Under the hood are css variables, but they won't need any additional setup.

DSil:
Orbit was broken for WL. This change fixes that problem. Therefore, it's a fix. A fix commit does not require an API change.

DSil (approved):

PR #4090, opened 2023-10-17
FEPLT-1690: feat(ListChoice): migrate to Tailwind

mvidalgarcia:
Closes [FEPLT-1690](https://kiwicom.atlassian.net/browse/FEPLT-1690)

 Storybook: https://orbit-mainframev-tw-list-choice.surge.sh

[FEPLT-1690]: https://kiwicom.atlassian.net/browse/FEPLT-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-components/src/ListChoice/index.tsx
+++ b/packages/orbit-components/src/ListChoice/index.tsx
@@ -1,89 +1,13 @@
 "use client";
 
 import * as React from "react";
-import styled, { css } from "styled-components";
+import cx from "clsx";
 
 import Heading from "../Heading";
-import Checkbox, { StyledLabel as Label } from "../Checkbox";
+import Checkbox from "../Checkbox";
 import Text from "../Text";
-import defaultTheme from "../defaultTheme";
-import { getSize } from "../Icon";
-import { right } from "../utils/rtl";
 import handleKeyDown from "../utils/handleKeyDown";
 import type { Props } from "./types";
-
-const StyledListChoiceIcon = styled.div`
-  ${({ theme }) => css`
-    display: flex;
-    align-items: center;
-    align-self: flex-start;
-    flex: 0 0 auto;
-    margin-${right}: ${theme.orbit.spaceXSmall};
-    height: ${theme.orbit.lineHeightTextNormal};
-
-    svg {
-      align-self: center;
-      width: ${getSize("medium")};
-      height: ${getSize("medium")};
-      color: ${theme.orbit.colorIconPrimary};
-      transition: color ${theme.orbit.durationFast} ease-in-out;
-    }
-  `}
-`;
-
-StyledListChoiceIcon.defaultProps = {
-  theme: defaultTheme,
-};
-
-export const StyledListChoice = styled.div<Partial<Props>>`
-  ${({ theme, disabled }) => css`
-    display: flex;
-    align-items: center;
-    box-sizing: border-box;
-    width: 100%;
-    padding: ${`${theme.orbit.spaceSmall} ${theme.orbit.spaceMedium}`};
-    border-bottom: 1px solid ${theme.orbit.paletteCloudNormal};
-    background-color: ${theme.orbit.paletteWhite};
-    transition: background-color 0.15s ease-in-out;
-    cursor: ${disabled ? "not-allowed" : "pointer"};
-
-    &:hover {
-      outline: none;
-      button {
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/ListChoice/index.tsx
```diff
@@ -108,32 +32,52 @@ const ListChoice = React.forwardRef<HTMLDivElement, Props>(
...
     return (
-      <StyledListChoice
+      <div
+        className={cx(
+          "orbit-list-choice",
+          "py-sm px-md box-border flex w-full items-center",
+          "border-b-cloud-normal bg-white-normal border-b-solid border-b",
+          "duration-fast transition-[background-color] ease-in-out",
+          disabled ? "cursor-not-allowed" : "cursor-pointer",
+          "hover:outline-none [&:hover_button]:bg-transparent",
```
DSil, line 42: TIL that `[&:hover_button]:` is the same as `[&_button]:hover:`. We have been using the last one in every other components.

I honestly have no preference. But for the sake of consistency I'd advise for a change here.

I even tried if this would work with the storybook pseudo-states addon, but it also doesn't 😞

packages/orbit-components/src/ListChoice/index.tsx
```diff
@@ -108,32 +32,52 @@ const ListChoice = React.forwardRef<HTMLDivElement, Props>(
...
+          !disabled && "hover:bg-cloud-light group",
```
DSil, line 43: I'd just add this to the string of the disabled ternary above 💅

packages/orbit-components/src/ListChoice/index.tsx
```diff
@@ -108,32 +32,52 @@ const ListChoice = React.forwardRef<HTMLDivElement, Props>(
...
         role={role || (selectable ? "checkbox" : "button")}
         {...conditionalProps}
       >
-        {icon && <StyledListChoiceIcon>{icon}</StyledListChoiceIcon>}
-        <StyledListChoiceContent>
+        {icon && (
+          <div
+            className={cx(
+              "group-hover:[&_svg]:text-icon-primary-foreground",
+              "me-xs h-icon-medium flex shrink-0 grow-0 basis-auto items-center self-start",
```
DSil, line 62:
```suggestion
              "me-xs h-icon-medium flex flex-none items-center self-start",
```

DSil (commented):
Overall LGTM, nothing seems to be breaking. Just left a few code suggestions.

Also, visual tests should now be included. This might require some procedure now:
- Add visual tests on a separate commit
- Push the branch with the visual tests commit but without the TW migration commit
- Deploy chromatic and accept snapshots
- Add the TW migration commit
- Push branch with both commits
- Deploy chromatic again and no changes should be spotted

packages/orbit-components/src/ListChoice/index.tsx
mvidalgarcia: Whoops, actually took me a while to come up with that. Don't why I thought `hover:` must be always at the beginning of the class.

packages/orbit-components/src/ListChoice/index.tsx
```diff
@@ -108,32 +32,52 @@ const ListChoice = React.forwardRef<HTMLDivElement, Props>(
...
         aria-selected={selected}
         role={role || (selectable ? "checkbox" : "button")}
         {...conditionalProps}
       >
-        {icon && <StyledListChoiceIcon>{icon}</StyledListChoiceIcon>}
-        <StyledListChoiceContent>
+        {icon && (
+          <div
+            className={cx(
+              "group-hover:[&_svg]:text-icon-primary-foreground",
```
mvidalgarcia, line 61: Just realised this has no effect. `theme.orbit.colorIconPrimary` is used both for the icon color and the icon color on hover. @DSil could you confirm this?

DSil: That's true. Because we are overriding the icon's color no matter what with the class `[&_svg]:text-icon-primary-foreground`, I'd say this is not needed.

DSil:
LGTM. Please rerun visual tests 👍

DSil (approved):

PR #4092, opened 2023-10-17
Migrate Layout to Tailwind

oreqizer:
Needed to work with Grid.

 Storybook: https://orbit-mainframev-oreqizer-fix-layout.surge.sh

### DIFF
--- a/packages/orbit-components/src/Layout/index.tsx
+++ b/packages/orbit-components/src/Layout/index.tsx
@@ -1,12 +1,9 @@
 "use client";
 
 import * as React from "react";
-import styled, { css } from "styled-components";
 
-import defaultTheme from "../defaultTheme";
 import Grid from "../utils/Grid";
 import { LAYOUT_SETTINGS } from "./consts";
-import mq from "../utils/mediaQuery";
 import type { Props, Type } from "./types";
 import type { Props as LayoutColumnProps } from "./LayoutColumn/types";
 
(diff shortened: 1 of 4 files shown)


mainframev (approved):

mainframev (changes requested):
it's visually different than in master (width)

packages/orbit-components/src/Layout/index.tsx
```diff
@@ -18,30 +15,20 @@ const getChildrenProps = (type: Type, key: string) => {
...
-StyledLayout.defaultProps = {
-  theme: defaultTheme,
-};
-
 const Layout = ({ children, type, dataTest }: Props) => (
-  <StyledLayout {...LAYOUT_SETTINGS[type]} dataTest={dataTest}>
+  <Grid
+    {...LAYOUT_SETTINGS[type]}
+    width="100%"
+    className="px-md de:p-lg mx-auto my-0 box-border py-0"
```
mainframev, line 22: StyledLayout before had "width: 100%", not it does not have this class, could you add? That should fix it

oreqizer: tried that, didn't work

mainframev: found the [old build from may](https://9ca42447.orbit.pages.dev/components/layout/layout/) (where it's supposed to be ok), so I guess it's correct, that on tablet and above it does not have full width

mainframev (approved):
LGTM

packages/orbit-components/src/utils/Grid/types.d.ts
```diff
@@ -18,6 +18,7 @@ export interface BasicProps extends Common.SpaceAfter {
 
 export interface Props extends Common.Globals, Common.SpaceAfter, BasicProps {
   readonly as?: string;
+  readonly className?: string;
```
mainframev, line 21: separeted feat commit 🙏🏻 and do not forget also to add that to flow file ;)

oreqizer: done 👌

PR #4094, opened 2023-10-17
feat(BadgeList): migrate to Tailwind

mvidalgarcia:
Closes FEPLT-1685

 Storybook: https://orbit-mainframev-tw-badge-list.surge.sh

### DIFF
--- a/packages/orbit-components/src/BadgeList/BadgeListItem/index.js.flow
+++ b/packages/orbit-components/src/BadgeList/BadgeListItem/index.js.flow
@@ -1,6 +1,5 @@
 // @flow
 import * as React from "react";
-import type { StyledComponent } from "styled-components";
 
 import type { Globals } from "../../common/common.js.flow";
 
@@ -17,8 +16,5 @@ export type Props = {|
 |};
 
 declare export var getIconColor: (type: Type) => string;
-declare export var StyledBadgeListItem: StyledComponent<any, any, HTMLDivElement>;
-declare export var StyledVerticalBadge: StyledComponent<any, any, HTMLDivElement>;
-declare export var StyledBadgeContent: StyledComponent<any, any, HTMLDivElement>;
 
 declare export default React.ComponentType<Props>;
--- a/packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
+++ b/packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
@@ -1,28 +1,20 @@
 "use client";
 
 import * as React from "react";
-import styled, { css } from "styled-components";
+import cx from "clsx";
 
 import Text from "../../Text";
 import { TYPE_OPTIONS, SIZE_OPTIONS } from "../consts";
-import defaultTheme from "../../defaultTheme";
 import { ICON_COLORS } from "../../Icon/consts";
-import { right } from "../../utils/rtl";
 import type { Props, Type } from "./types";
 import type { Props as IconProps } from "../../Icon/types";
 
-const getBackground = ({ theme, $type }: { theme: typeof defaultTheme; $type?: Type }) => {
-  const tokens = {
-    [TYPE_OPTIONS.NEUTRAL]: theme.orbit.paletteCloudLight,
-    [TYPE_OPTIONS.INFO]: theme.orbit.paletteBlueLight,
-    [TYPE_OPTIONS.SUCCESS]: theme.orbit.paletteGreenLight,
-    [TYPE_OPTIONS.WARNING]: theme.orbit.paletteOrangeLight,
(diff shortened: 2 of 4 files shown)


packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
```diff
@@ -30,57 +30,43 @@ export const getIconColor = (type: Type) => {
...
-    .orbit-tooltip-wrapper .orbit-text {
-      font-weight: ${theme.orbit.fontWeightMedium};
-    }
-  `};
-`;
-
-StyledBadgeContent.defaultProps = {
-  theme: defaultTheme,
-};
+export const StyledBadgeContent = ({ children, ...props }) => (
```
DSil, line 62: This requires more work to keep compatibility in `ItineraryBadgeListItem`. The usage of this component there uses the `css` prop, that is accepted by styled components, but not here… So we have two options here:
- we open the style prop and spread it AND in `ItineraryBadgeListItem` we adjust accordingly
OR
- we create a new prop that applies that style that `ItineraryBadgeListItem` needs and we add that prop the component usage there

I'm fine with any approach. And the same suggestion regarding naming.

DSil (changes requested):

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
```diff
@@ -30,57 +30,43 @@ export const getIconColor = (type: Type) => {
...
-    display: flex;
-    flex-direction: row;
-    width: 100%;
-    & + & {
-      margin-top: ${theme.orbit.spaceXXSmall};
-    }
-  `};
-`;
+export const StyledBadgeListItem = ({ children, ...props }) => (
+  <li className="[&_+_&]:mt-xxs flex w-full flex-row" {...props}>
```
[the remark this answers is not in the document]
mvidalgarcia: The problem is that `StyledBadgeListItem` was receiving `data-test` as prop (or HTML attr), and I can't unpack this (I don't really know if having a variable named "data-test" is possible, probably not in JS). I can change it to `dataTest` and adapt it on `ItineraryBadgeListItem`, but since this component is exported, it would be a breaking change. It's probably not a big deal since, as you said, it is only used there and not exported outside.

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
```diff
@@ -30,57 +30,43 @@ export const getIconColor = (type: Type) => {
...
+}) => {
+  const theme = useTheme();
+  return (
+    <div
+      className={cx(
+        "me-xs h-icon-large w-icon-large rounded-circle flex flex-shrink-0 items-center justify-center",
+        "[&_svg]:h-icon-small [&_svg]:w-icon-small",
+      )}
+      style={{ background: getBackground({ theme, $type }) ?? undefined }}
+      {...props}
```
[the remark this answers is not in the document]
mvidalgarcia: Okay, thanks for clarifying. I was just trying to make it as flexible as possible.

packages/orbit-components/src/BadgeList/BadgeListItem/index.tsx
mvidalgarcia: I didn't even realised of the `css` prop, thanks for the thorough check.

I've created a `style` prop in `BadgeContent` and adapted `ItineraryBadgeListItem` to pass the values with not styled-components' `css` function. Is this what you had in mind?

mainframev (changes requested):

packages/orbit-components/src/BadgeList/BadgeListItem/index.js.flow
```diff
@@ -17,8 +17,8 @@ export type Props = {|
 |};
 
 declare export var getIconColor: (type: Type) => string;
-declare export var StyledBadgeListItem: StyledComponent<any, any, HTMLDivElement>;
-declare export var StyledVerticalBadge: StyledComponent<any, any, HTMLDivElement>;
-declare export var StyledBadgeContent: StyledComponent<any, any, HTMLDivElement>;
+declare export var ItemWrapper: StyledComponent<any, any, HTMLDivElement>;
+declare export var VerticalBadge: StyledComponent<any, any, HTMLDivElement>;
+declare export var BadgeContent: StyledComponent<any, any, HTMLDivElement>;
```
DSil, line 22: These can be removed. We're not exporting them to the outside and we don't use flow internally (and we're even removing them from Orbit entirely soon™) so I'd just remove them now.

DSil (commented):
Overall good. Just the background classes 💅

DSil (approved):

mainframev (approved):

PR #4095, opened 2023-10-17
FEPLT-1777: chore(tailwind): make setup instructions more robust

RobinCsl:
The code snippet for the tailwind.config.js file is expanded to account for monorepo setups where the Orbit components node module might be hoisted elsewhere.

Relates to [FEPLT-1777](https://kiwicom.atlassian.net/browse/FEPLT-1777)

 Storybook: https://orbit-mainframev-rcsl-expand-orbit-preset-docs.surge.sh

[FEPLT-1777]: https://kiwicom.atlassian.net/browse/FEPLT-1777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-tailwind-preset/README.md
+++ b/packages/orbit-tailwind-preset/README.md
@@ -23,12 +23,17 @@ The `orbitComponentsPreset` function accepts an optional object with one propert
 In your `tailwind.config.js` file (or equivalent), add the following:
 
 ```js
+const path = require("node:path");
 const orbitComponentsPreset = require("@kiwicom/orbit-tailwind-preset");
+// the following ensures the correct path is found (especially within monorepos)
+const orbitComponentsPath = require
+  .resolve("@kiwicom/orbit-components")
+  .replace("/lib/index.js", "");
 
 module.exports = {
   content: [
     // ...
-    "./node_modules/@kiwicom/orbit-components/**/*.js",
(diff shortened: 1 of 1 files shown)


packages/orbit-tailwind-preset/README.md
```diff
@@ -23,12 +23,17 @@ The `orbitComponentsPreset` function accepts an optional object with one propert
 In your `tailwind.config.js` file (or equivalent), add the following:
 
 ```js
+const path = require("node:path");
 const orbitComponentsPreset = require("@kiwicom/orbit-tailwind-preset");
+// the following ensures the correct path is found (especially within monorepos)
+const orbitComponentsPath = require
```
mainframev, line 29: I know that it was documented as CSJ before, but wdyt, maybe the better way will be to document that right with *ts* with the ESM example? Our users use typescript mainly and the config will be most likely in ts 🤔

RobinCsl: I'd say it's out of scope for this PR.

The only point was to make the instructions work with monorepos.

If we want to use a different format and extension, let's do it separately.

mainframev (approved):

PR #4096, opened 2023-10-17
Migrate Alert to Tailwind

DSil:
[FEPLT-1698](https://kiwicom.atlassian.net/browse/FEPLT-1698)

Visual Test story was updated and snapshots in Chromatic were also updated **before** the Tailwind migration. Visual regression testing to be made after PR's approval and before merge. **Edit:** Visual tests passed.

[FEPLT-1698]: https://kiwicom.atlassian.net/browse/FEPLT-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-tw-migrate-alert.surge.sh

### DIFF
--- a/packages/orbit-components/src/Alert/Alert.stories.tsx
+++ b/packages/orbit-components/src/Alert/Alert.stories.tsx
@@ -289,15 +289,7 @@ export const AlertVisualTest = () => {
             <Alert type={type} title="Title">
               <div className="space-x-xs">
                 <TextLink>Same type link</TextLink>
-                <TextLink dataTest="TextLinkHover">Same type link hover</TextLink>
-                <TextLink dataTest="TextLinkActive">Same type link active</TextLink>
                 <TextLink type="secondary">Secondary type link</TextLink>
-                <TextLink dataTest="TextLinkHover" type="secondary">
-                  Secondary type link hover
-                </TextLink>
-                <TextLink dataTest="TextLinkActive" type="secondary">
-                  Secondary type link active
-                </TextLink>
               </div>
             </Alert>
           </div>
@@ -353,9 +345,9 @@ AlertVisualTest.story = {
       disableSnapshot: false,
     },
     pseudo: {
-      hover: ["[data-test='AlertButtonHover']", "[data-test='TextLinkHover']"],
+      hover: ["[data-test='AlertButtonHover']"],
       focus: ["[data-test='AlertButtonFocus']"],
-      active: ["[data-test='AlertButtonActive']", "[data-test='TextLinkActive']"],
+      active: ["[data-test='AlertButtonActive']"],
     },
   },
 };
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/TextLink/helpers/twClasses.ts
```diff
@@ -23,3 +24,30 @@ export const typeClasses: Record<TYPE_OPTIONS, string> = {
   [TYPE_OPTIONS.WHITE]:
     "text-link-white-foreground hover:text-link-white-foreground-hover active:text-link-white-foreground-active",
 };
+
+export const alertDescendantClasses: Record<ALERT_TYPE, string[]> = {
```
oreqizer, line 28: should these be in `TextLink`? 🤔 they're `Alert` specific, imo makes more sense to have em there. would just put them inside the main component for observability

packages/orbit-components/src/Alert/index.tsx
```diff
@@ -290,37 +128,56 @@ const Alert = (props: Props) => {
     labelClose,
   } = props;
   return (
-    <StyledAlert
-      type={type}
-      icon={icon}
+    <div
+      className={cx(
+        "rounded-large text-ink-dark font-base text-normal p-sm relative box-border flex w-full border border-t-[3px] leading-normal",
```
oreqizer, line 133: do we maybe want a token for `border-t-[3px]`?

packages/orbit-components/src/TextLink/helpers/twClasses.ts
DSil: Good point. I thought thoroughly about this. In the end, I thought the opposite way: when we ever change TextLink styles, if we don't have these classes here… we _will_ forget about the impact such change can have in Alert. We won't remember looking for them inside Alert, will we?

Alert only changes _some_ of the TextLink styles, not all of them. So I am really not sure about the best approach here.

> changing Alert links will show up as changes in TextLink in Git history

I would worry about that if we had package per component versioning / changelog. Since it's not the case, I don't think that's problematic in the practical way (but I agree and understand your concern here).

oreqizer: well the long term vision is to _not_ have this kind of cross-referencing in the future anyway, so let's not bother then. consider this resolved!

packages/orbit-components/src/Alert/Alert.stories.tsx
```diff
@@ -289,15 +289,7 @@ export const AlertVisualTest = () => {
             <Alert type={type} title="Title">
               <div className="space-x-xs">
                 <TextLink>Same type link</TextLink>
-                <TextLink dataTest="TextLinkHover">Same type link hover</TextLink>
```
oreqizer, line 292: why did we remove these link tests?

oreqizer: DM'd

packages/orbit-components/src/Alert/index.tsx
DSil: As agreed on DMs, it's currently not justifiable to have a token here. If the usage of a 3px border evolves, we will reconsider

oreqizer (approved):

PR #4099, opened 2023-10-18
Migrate Modal to Tailwind

DSil:
[FEPLT-1684](https://kiwicom.atlassian.net/browse/FEPLT-1684)

[FEPLT-1684]: https://kiwicom.atlassian.net/browse/FEPLT-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-tw-modal.surge.sh


packages/orbit-design-tokens/src/dictionary/definitions/component/modal/modal-border.json
```diff
@@ -1,9 +1,26 @@
 {
   "component": {
+    "modal": {
+      "border-color": {
```
mainframev, line 4: hm, why we are adding this if we said, that the replacement for `borderColorModal` is `elevationFlatBackground` (I think the correct one is elevationFlatBorderColor, as it's for border and background is white). This was introduced for such kind of cases, where these borders will be used on multiple places and we do not need to have component tokens. So I think we can use for modal `elevationFlatBorderColor`

packages/orbit-design-tokens/src/dictionary/definitions/component/modal/modal-border.json
```diff
@@ -1,9 +1,26 @@
...
+        "value": 9
+      }
+    },
     "border-color": {
       "modal": {
         "type": "color",
-        "value": "{foundation.palette.cloud.normal}"
+        "value": "{foundation.palette.cloud.normal}",
+        "deprecated": true,
+        "deprecated-replace": "{global.elevation.flat.background}",
```
mainframev, line 22: should be `global.elevation.flat.border-color`

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -250,6 +264,8 @@ const cfg = (options?: Options): Config => {
           "form-element-hover": tokens.formElementBoxShadowHover,
           "form-element-focus": tokens.formElementFocusBoxShadow,
           switch: `inset 0 0 1px 0 rgba(7, 64, 92, 0.1),${tokens.boxShadowAction}`,
+          "modal-scrolled": `inset 0 1px 0 ${tokens.paletteCloudNormal}, ${tokens.boxShadowFixedReverse}`,
+          modal: `inset 0 0 0 transparent, ${tokens.boxShadowFixedReverse}`,
```
mainframev, line 268:
```suggestion
          elevation-flat: tokens.elevationFlatBorderColor,
```

mainframev (changes requested):

packages/orbit-components/src/Modal/Modal.stories.tsx
```diff
@@ -562,3 +562,43 @@ Rtl.story = {
...
 };
+
+export const VisualTest = () => {
+  const { Container, onClose } = useModal();
+  const content = "Lorem ispum dolor sit amet.";
+
+  return (
+    <Container>
+      <Modal onClose={onClose} mobileHeader size={SIZES.NORMAL}>
+        <ModalHeader
```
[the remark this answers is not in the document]
DSil: Agreed. Tbh I didn't go into all the possible variations of Modal because, in the end, it is impossible to verify all possibilities of a Modal in one story. But this one is definitely possible, so I've added as you suggested, thanks!

mainframev:
I think Modal story with removable section looks different, is this expected? 🤔 

master: 
Screenshot 2023-10-23 at 16 57 01


current:
Screenshot 2023-10-23 at 16 58 44

packages/orbit-design-tokens/src/dictionary/definitions/component/modal/modal-border.json
DSil: Yes, my typo

packages/orbit-design-tokens/src/dictionary/definitions/component/modal/modal-border.json
DSil: Agreed. Will remove the token. But I think we can have a tailwind class that uses the token, so the class will be `border-modal`. But instead of using a token, we define on the TW config that `modal: tokens.elevationFlatBorderColor`.
I think it even looks better (currently the class was `border-modal-border-color`).

I'm not strongly against using the class `border-elevation-flat-border-color`, but I think that we have no harm in abstracting internally to `modal`. WDYT?

packages/orbit-tailwind-preset/src/presets/components/index.ts
DSil: I don't understand this suggestion 😕 can you please explain?

packages/orbit-design-tokens/src/dictionary/definitions/component/modal/modal-border.json
mainframev: Yep, it's no harm as you say, I just want it to be consistent, as for instance our Card after refactoring uses these flat elevations 🙂 
Screenshot 2023-10-23 at 17 14 28

DSil: Hmm ok, I'll just go with that then 👍

DSil:
Found the cause of the visual bug, fixed it. It was a single space 🙃 

In the meantime, I've found another bug: Modal does not close when clicking outside. So I'll address that.

packages/orbit-tailwind-preset/src/presets/components/index.ts
mainframev: ah, I thought it's border color, np, ignore this one then :)

mainframev:
Now different visual bug 🤔 (Without section story)
Screenshot 2023-10-26 at 11 39 06 


Screenshot 2023-10-26 at 11 39 02

DSil:
@mainframev the first one was the same issue: a space on a class. The second one was fixed just by rebasing, so I am guessing it was some issue with the Stack, that we already fixed.

mainframev (approved):
LGTM 👍🏻 good job!

mainframev:
@DSil also do not forget to update the status yaml file

PR #4102, opened 2023-10-19
fix(OrbitProvider): replace useEffect with style tag

mainframev:
[Slack discussion](https://skypicker.slack.com/archives/C05FDP4UM3N/p1697715842011039)
 Storybook: https://orbit-mainframev-docs-extend-docs-about-theming.surge.sh


packages/orbit-components/src/OrbitProvider/index.tsx
```diff
@@ -34,19 +34,11 @@ const getCssVarsForWL = (theme: typeof defaultTokens) =>
...
-    style.innerText = tokensToCssVars({ tokens: getCssVarsForWL(theme.orbit), cssClass: ":root" });
-    const head = document.getElementsByTagName("head")[0];
-    if (!document.getElementById(ID)) {
-      head.appendChild(style);
-    }
-  }, [theme]);
-
   return (
     <RandomIdProvider useId={useId}>
+      <style id="orbit-theme-css-vars">
```
oreqizer, line 39: have you tested this? is this soon enough in the DOM that it won't cause flickering?

also, this doesn't correlate with the docs. docs say to embed the styles in the `<head>` (where they should be)

mainframev: @oreqizer how i'd add that to head without useEffect and other magic? That should be done in Overlord or inside the FE module as they handle head, SSR and so on 

Yep, I've tested that on home via yalc, works fine, no flickering, can send you a video on slack.

oreqizer: alright let's keep it like this for now. but pls change docs to reflect the intended behaviour 🙏

mainframev: I added a note, saying:  **Note: This approach is recommended only if you are not using OrbitProvider. We manage the addition of CSS variables within OrbitProvider until the migration to Tailwind is complete**.

Feel free to suggest different sentence, if this does not look descriptive enough :)

oreqizer (approved):
lgtm as a tmp solution 👍

PR #4106, opened 2023-10-23
docs: fix fetching issue for tracking

mainframev:
fixing the issue, when in tracking not repositories are shown
 Storybook: https://orbit-mainframev-tracking-fix-fetch-issue.surge.sh

### DIFF
--- a/docs/src/components/Dashboard/Difference.tsx
+++ b/docs/src/components/Dashboard/Difference.tsx
@@ -49,8 +49,9 @@ const DataDiffTemplate = ({ location }: PageRendererProps) => {
   `);
 
   const { nodes } = allTracking;
-  const first = nodes.slice(0, 8);
-  const last = nodes.slice(-8);
+  const TRACKED_REPOSITORIES_COUNT = 11;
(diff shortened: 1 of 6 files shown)


docs/src/components/Dashboard/Difference.tsx
```diff
@@ -49,8 +49,9 @@ const DataDiffTemplate = ({ location }: PageRendererProps) => {
   `);
 
   const { nodes } = allTracking;
-  const first = nodes.slice(0, 8);
-  const last = nodes.slice(-8);
+  const TRACKED_REPOSITORIES = 11;
```
mvidalgarcia, line 52: What do you think about naming this variable `TRACKED_REPOSITORIES_COUNT` or similar?

mvidalgarcia:
@mainframev I've tried to check the wiki staging but I don't have access. Do you know how I can test it?

Screenshot 2023-10-23 at 10 19 20

Or maybe you can share screenshot of the difference?

mainframev:
@mvidalgarcia 
Screenshot 2023-10-23 at 11 01 06

mvidalgarcia (approved):

PR #4107, opened 2023-10-23
docs: add tailwind status to tracking

mainframev:
[FEPL-1778](https://kiwicom.atlassian.net/browse/FEPLT-1778)
 Storybook: https://orbit-mainframev-tracking-tw-status.surge.sh


mvidalgarcia (changes requested):
This PR doesn't do what's described in the [Jira issue](https://kiwicom.atlassian.net/browse/FEPLT-1778).

> Display an additional column in the Orbit Tracking dashboards stating whether a component was migrated to TW or not

It's creating a whole new page instead, that's unnecessary. We agreed to modify /dashboard/tracking/<repo> instead.

> This is irrespective of the version currently in use in the frontends and whether the component was released (being in master and migrated is enough)

We don't care if it's released or not.

mainframev:
> This PR doesn't do what's described in the [Jira issue](https://kiwicom.atlassian.net/browse/FEPLT-1778).
> 
> > Display an additional column in the Orbit Tracking dashboards stating whether a component was migrated to TW or not
> 
> It's creating a whole new page instead, that's unnecessary. We agreed to modify /dashboard/tracking/ instead.
> 
> > This is irrespective of the version currently in use in the frontends and whether the component was released (being in master and migrated is enough)
> 
> We don't care if it's released or not.

have you checked the tracking pages? It's there, so PR solves it 

It's using single yaml file. I just thought, that it will be good to create it for both, for tracking and as a normal page, so users also will be able to see our progress.

mainframev:
Screenshot 2023-10-23 at 13 20 14

mvidalgarcia:
> So I just thought, that it will be good to create it for both, for tracking and as a normal page, so users also will be able to see our progress.

We, as a team, haven't agreed on this. So we don't need more code to maintain than what's strictly needed. This tracking is for our internal usage, if I understood it correctly.

It doesn't seem correct to have `+389 −7` for a simple column, that is what we originally wanted. But I guess we will need to escalate this to the rest of the team.

mainframev:
> > So I just thought, that it will be good to create it for both, for tracking and as a normal page, so users also will be able to see our progress.
> 
> We, as a team, haven't agreed on this. So we don't need more code to maintain than what's strictly needed. This tracking is for our internal usage, if I understood it correctly.
> 
> It doesn't seem correct to have `+389 −7` for a simple column, that is what we originally wanted. But I guess we will need to escalate this to the rest of the team.

I will ask in the chat about this

docs/src/components/Dashboard/components/ComponentList.tsx
```diff
@@ -40,12 +41,12 @@ const Sort = ({
...
+  const [data, setData] = React.useState<ComponentData[]>(components);
   const [isSorted, setSorted] = React.useState(false);
-
-  React.useEffect(() => {
-    setData(components);
-  }, [setData, components]);
+  const tailwindStatusData = tailwindMigrationData.reduce((acc, { component, migrated }) => {
+    acc[component] = migrated;
+    return acc;
+  }, {});
```
mvidalgarcia, line 49: Wouldn't it be easier to store the data in the expected format instead of doing this data manipulation?

```yaml
- Alert: true
- Box: true
...
```

No strong opinion, just a thought.

mvidalgarcia (approved):
Approving to avoid blocking.

PS. I think the commit message body is not needed.

docs/src/components/Dashboard/components/ComponentList.tsx
mainframev: thanks, added

mvidalgarcia (approved):
:shipit:

PR #4109, opened 2023-10-24
Fix text tailwind classes

mvidalgarcia:
- chore(Breadcrumbs): remove useless class
- fix(Loading): use correct font family class

 Storybook: https://orbit-mainframev-marco-fix-tw-classes.surge.sh


mainframev (approved):
Good catch :)

PR #4110, opened 2023-10-24
fix icon id collisions 

mainframev:
[Slack thread](https://skypicker.slack.com/archives/CAMS40F7B/p1698140695934119)
 Storybook: https://orbit-mainframev-chore-fix-icon-id.surge.sh
 
 Regenerated the icon ids for icons that had collisions with others. 
 Run `yarn components check:icons` to test
 
 Need approval also [here](https://gitlab.skypicker.com/frontend/images.kiwi.com/-/merge_requests/300)


DSil (commented):
LGTM, but are the commits worth a mention on the changelog? I'd just mark them as chore, as the change is not relevant for our users.

mainframev:
@DSil Yep, I was not sure what it should be, as it affects icons on the build and when they builded with incorrect ids the native team can't use them.

DSil (approved):

PR #4111, opened 2023-10-24
feat(tailwind): add foundation option

mainframev:
Added `foundation` option to main preset to make it customizable for users 

Storybook: https://orbit-mainframev-feat-add-foundation-option.surge.sh


mainframev:
Closing for now, until we will agree if we need this.

PR #4113, opened 2023-10-25
Migrate NotificationBadge to Tailwind

DSil:
Migrate NotificationBadge to Tailwind

[FEPLT-1719](https://kiwicom.atlassian.net/browse/FEPLT-1719)

[FEPLT-1719]: https://kiwicom.atlassian.net/browse/FEPLT-1719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-tw-notification-badge.surge.sh

### DIFF
--- a/packages/orbit-components/src/NotificationBadge/NotificationBadge.stories.tsx
+++ b/packages/orbit-components/src/NotificationBadge/NotificationBadge.stories.tsx
@@ -186,3 +186,30 @@ Rtl.story = {
     info: "This is a preview of this component in RTL setup.",
   },
 };
+
+export const VisualTest = () => {
+  const Icon = getIcon(getIcons("Airplane"));
+  const types = Object.values(TYPE_OPTIONS);
+
+  return (
+    <div className="space-x-xs flex items-center">
+      {types.map(type => (
+        <React.Fragment key={type}>
+          <NotificationBadge type={type}>1</NotificationBadge>
+          <NotificationBadge type={type}>10</NotificationBadge>
+          <NotificationBadge type={type} icon={<Icon />}>
+            10
+          </NotificationBadge>
+        </React.Fragment>
+      ))}
+    </div>
+  );
+};
+
+VisualTest.story = {
+  parameters: {
+    chromatic: {
+      disableSnapshot: false,
+    },
+  },
+};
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/NotificationBadge/NotificationBadge.stories.tsx
```diff
@@ -186,3 +186,30 @@ Rtl.story = {
...
+  const types = Object.values(TYPE_OPTIONS);
+
+  return (
+    <div className="space-x-xs inline-flex">
+      {types.map(type => (
+        <React.Fragment key={type}>
+          <NotificationBadge type={type}>1</NotificationBadge>
+          <NotificationBadge type={type}>10</NotificationBadge>
+          <NotificationBadge type={type} icon={<Icon />}>
+            10
```
oreqizer, line 201: does this `10` have an effect here?

DSil: I added it to check it with two chars. Because the padding and width overrides are there to avoid the "expansion" of the badge as the content grows, so I wanted to test it with one char and two chars

packages/orbit-components/src/NotificationBadge/NotificationBadge.stories.tsx
```diff
@@ -186,3 +186,30 @@ Rtl.story = {
     info: "This is a preview of this component in RTL setup.",
   },
 };
+
+export const VisualTest = () => {
+  const Icon = getIcon(getIcons("Airplane"));
+  const types = Object.values(TYPE_OPTIONS);
+
+  return (
+    <div className="space-x-xs inline-flex">
```
mainframev, line 195: is this expected to be slightly lifted? Maybe better will be "flex items-center" instead of inline-flex?

DSil: Ok, will add it after running the visual tests over tailwind, so that I don't get a false positive (because there will be a visual difference)

DSil: Done ✅

mainframev (approved):
LGTM 👍🏻

PR #4114, opened 2023-10-25
Move migrated components from tmp to src

mvidalgarcia:
Closes FEPLT-1789

 Storybook: https://orbit-mainframev-marco-tmp-to-src.surge.sh
 
##### TODO
- [x] Select's `readOnly` prop is gone, should we make a BC?


packages/orbit-components/src/Textarea/index.tsx
```diff
@@ -186,25 +81,43 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, Props>((props, ref) => {
...
-      <StyledTextArea
+      <textarea
+        className={cx(
+          "w-full appearance-none",
+          "font-base",
+          "rounded-normal relative box-border block overflow-auto",
+          "min-h-form-box-normal",
+          "duration-fast transition-shadow ease-in-out",
+          "border border-transparent",
+          // TODO: remove when will be migrated from tmp-folder
```
mvidalgarcia, line 92: @mainframev Could you please explain this? Do we need to do something?

mainframev: our inputs like InputField, and InputFile -> no longer provide size prop and it was removed, but looks like it was missed for Textarea. In Figma, there is only one size for this component (normal). So it should be removed imo. I didn't want to do that under the chore commit in the temp folder. I asked designers and they said, that "in Figma we have one size, but it should support resize" (which is supported via resize prop)

packages/orbit-components/src/InputField/__tests__/index.test.tsx
```diff
@@ -142,14 +142,14 @@ describe("InputField", () => {
...
-      await user.tab();
+      await act(() => user.tab());
       expect(screen.queryByTestId("help")).not.toBeInTheDocument();
       expect(screen.getByTestId("error")).toBeInTheDocument();
     });
   });
 
   describe("error forms", () => {
-    it("should show the correct error tooltips while tabbing between fields", async () => {
+    it("shold show the correct error tooltips while tabbing between fields", async () => {
```
mainframev, line 152:
```suggestion
    it("should show the correct error tooltips while tabbing between fields", async () => {
```

packages/orbit-components/src/InputField/__tests__/index.test.tsx
```diff
@@ -142,14 +142,14 @@ describe("InputField", () => {
       expect(input).toHaveAttribute("data-state", "error");
       expect(input).toBeInvalid();
 
-      await user.tab();
+      await act(() => user.tab());
```
mainframev, line 145: I think it should be the opposite, we removed recently act wrappers 
```suggestion
      await user.tab();
```

packages/orbit-components/src/InputField/__tests__/index.test.tsx
```diff
@@ -64,7 +64,7 @@ describe("InputField", () => {
     expect(screen.getByTestId("test")).toBeInTheDocument();
     expect(screen.getByTestId("prefix")).toBeInTheDocument();
     expect(screen.getByTestId("suffix")).toBeInTheDocument();
-    await user.tab();
+    await act(() => user.tab());
```
mainframev, line 67:
```suggestion
    await user.tab();
```

packages/orbit-components/src/Textarea/__tests__/index.test.tsx
```diff
@@ -42,7 +42,7 @@ describe("Textarea", () => {
     expect(screen.getByDisplayValue("value")).toBeInTheDocument();
     expect(screen.getByPlaceholderText(placeholder)).toBeInTheDocument();
 
-    await user.tab();
+    await act(() => user.tab());
```
mainframev, line 45:
```suggestion
    await user.tab();
```

packages/orbit-components/src/Textarea/__tests__/index.test.tsx
```diff
@@ -66,17 +66,17 @@ describe("Textarea", () => {
     expect(onFocus).toHaveBeenCalled();
     await user.tab();
     expect(onBlur).toHaveBeenCalled();
-    await user.type(textarea, "kek");
+    await act(() => user.type(textarea, "kek"));
```
mainframev, line 69:
```suggestion
    await user.type(textarea, "kek");
```

packages/orbit-components/src/Textarea/__tests__/index.test.tsx
```diff
@@ -66,17 +66,17 @@ describe("Textarea", () => {
...
     expect(onChange).toHaveBeenCalled();
   });
 
   it("should have error", async () => {
     render(<Textarea error="error" size="small" />);
     const textarea = screen.getByRole("textbox");
 
-    await user.tab();
+    await act(() => user.tab());
```
mainframev, line 77:
```suggestion
    await user.tab();
```

mainframev (changes requested):
some tests leftovers

mainframev:
@mvidalgarcia Also do not forget about removing the ignore patterns [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/config/build/consts.mts#L15) :) 🙏🏻

mvidalgarcia:
@mainframev Thanks for all the catches! 👁️ 

The reason these problems existed is that in some components there were new changes introduced in src, and in other cases in tmp, so it wasn't enough to copy from tmp to src, but _merge_ them instead.
This proves that the tmp folder is not the best solution as there might be lots of side-effects hard to detect.

packages/orbit-components/src/Textarea/index.tsx
mainframev: @mvidalgarcia I see that you kept this. No problem, can be changed any time, it's supposed to breaking, so just change the comment to something like: "TODO: remove size small for Textarea"

mvidalgarcia: Just pushed a commit addressing it

packages/orbit-components/src/Textarea/types.d.ts
```diff
@@ -9,7 +9,6 @@ type Resize = "vertical" | "none";
 
 export interface Props extends Common.Globals, Common.SpaceAfter, Common.DataAttrs {
   readonly ref?: React.ForwardedRef<HTMLTextAreaElement>;
-  readonly size?: Common.InputSize;
```
mainframev, line 12: Good! Then it should be removed also from "docs/src/__examples__/Textarea/DEFAULT.tsx", from README.md and from the "index.js.flow" file 👍🏻

mainframev (changes requested):

packages/orbit-components/src/Textarea/types.d.ts
mvidalgarcia: 🤦 I always forget those, thanks!!

mainframev: I think "feat", at least we mostly awalys put it under feat, from what I see in the changelog

mvidalgarcia: Done!

mainframev (approved):
Good job! 🚀

PR #4116, opened 2023-10-25
Stack: Fix classes not applied in viewports

DSil:
On #4071 we started to verify a condition on viewport properties that should not be required. The condition is now only required for the classes that it actually affects.

Commit is labeled as a chore as this is a fix to an unreleased bug.
 Storybook: https://orbit-mainframev-fix-stack-viewports.surge.sh


mainframev (approved):

mainframev:
@DSil LGTM, good catch 

I think you do not mind if I will merge it now for fixing our docs 🙏🏻

PR #4117, opened 2023-10-26
Fix TileGroup hover states and bg colors

oreqizer:
from https://github.com/kiwicom/orbit/pull/4105
 Storybook: https://orbit-mainframev-oreqizer-tw-tile-group-fix.surge.sh


DSil (commented):
Changes LGTM, but I'd discard the test change, as we've seen it does nothing for us…

About the commit, it should be a chore, as it fixes a problem that was never released, so it should have no mention on the changelog.

DSil (approved):

PR #4118, opened 2023-10-26
chore: extend tracking scope

mainframev:
Closes [FEPLT-1790](https://kiwicom.atlassian.net/browse/FEPLT-1790)

[FEPLT-1790]: https://kiwicom.atlassian.net/browse/FEPLT-1790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-chore-extend-tracking-data.surge.sh

### DIFF
--- a/docs/src/components/Dashboard/AllRepositories.tsx
+++ b/docs/src/components/Dashboard/AllRepositories.tsx
@@ -10,7 +10,7 @@ const AllRepositories = ({ location }: PageRendererProps) => {
   const { allTracking }: SchemeTrackingNode = useStaticQuery(graphql`
     query AllRepositoriesTracking {
       # get only the latest
-      allTracking(sort: { fields: createdAt, order: DESC }, limit: 11) {
+      allTracking(sort: { fields: createdAt, order: DESC }, limit: 21) {
         nodes {
           fields {
             currentComponents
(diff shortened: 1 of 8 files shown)


docs/src/components/Dashboard/AllRepositories.tsx
```diff
@@ -10,7 +10,7 @@ const AllRepositories = ({ location }: PageRendererProps) => {
   const { allTracking }: SchemeTrackingNode = useStaticQuery(graphql`
     query AllRepositoriesTracking {
       # get only the latest
-      allTracking(sort: { fields: createdAt, order: DESC }, limit: 11) {
+      allTracking(sort: { fields: createdAt, order: DESC }, limit: 21) {
```
DSil, line 13: It feels weird to manually update these values. Can't we use the PROJECTS constant and take the size from it?

I also couldn't find this function's signature, to understand what this `limit` is, can you please point me to it? 🙏

mainframev: I thought about this, and no, it can't be like that. At least inside useStaticQuery, which does not allow to use of any kind of a variable inside. 

To understand what it is, you can run it locally open graphiql and apply that graphql param :) 

It's needed for fetching only the latest 21 repositories data, as it's sorted by createdAt and in desc order. As it also fetches data from the previous periods. (currently i removed all previous month data, as they do not include new repositories and it does not make sense to compare them, so difference page is empty now)

mainframev: I can think about improving that, splitting the queries and creating a different one to fetch only the latest, that way it won't require the usage of this graphql param, but this is out of the scope of this PR

DSil (approved):

PR #4122, opened 2023-10-27
Playwright component testing

oreqizer:
Adds Playwright component testing tests inside `Component.ct.tsx`

The component to be tested is in a `Component.ct-story.tsx` file bcs of https://playwright.dev/docs/test-components#test-stories
 Storybook: https://orbit-mainframev-oreqizer-try-playwright-ct.surge.sh


packages/orbit-components/playwright-ct.config.ts
```diff
@@ -0,0 +1,64 @@
...
+    trace: "on-first-retry",
+    testIdAttribute: "data-test",
+    ctPort: 3100,
+  },
+  projects: [
+    {
+      name: "Small Mobile",
+      use: {
+        ...devices["iPhone SE"],
+        viewport: { ...devices["iPhone SE"].viewport, width: 320 },
```
mainframev, line 25: I think you can use token values here for width, so it's not hardcoded :)

mainframev:
@oreqizer could you also add `.ct` files to build ignore and type declaration build ignore?  es and lib folders include those files, I do not think they should be a part of the output :) 

Add them [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/config/build/consts.mts) and check output ;)

mainframev (changes requested):

packages/orbit-components/.gitignore
```diff
@@ -18,3 +18,7 @@ orbit-icons-font.zip
 orbit-svgs.zip
 cypress/screenshots
 cypress/videos
+node_modules/
```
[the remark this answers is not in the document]
oreqizer: cleaned up

packages/orbit-components/playwright-ct.config.ts
oreqizer: adjusted

packages/orbit-components/config/build/consts.mts
```diff
@@ -14,7 +14,8 @@ export const COMPILE_IGNORE_PATTERNS = [
   // TODO: remove that after TW migration, tailwind is temporary folder
   "**/tmp-tailwind/**/*.{ts?(x),d.ts}",
   "**/*.d.ts",
-  "**/*.stories.*",
```
mainframev, line 17: this shouldn't be removed, we still have storybook and stories shouldn't be in the builded output 🙂

packages/orbit-components/config/build/consts.mts
```diff
@@ -14,7 +14,8 @@ export const COMPILE_IGNORE_PATTERNS = [
...
+  "**/*.ct.*",
+  "**/*.ct-story.*",
```
mainframev, line 18: 🙏🏻 add also to ignore for type declarations [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/config/build/consts.mts#L24)

packages/orbit-components/config/build/consts.mts
oreqizer: ah lol don't know why it was removed

packages/orbit-components/config/build/consts.mts
oreqizer: fixed

mainframev (approved):
LGTM 👍🏻

PR #4123, opened 2023-10-27
fix(InputSelect): make root label block

ConorHaining:
Fixes a bug in FF & Safari for `<InputSelect>` where the child widths collapse as <label> is usually inline.

When focusing on the input element of the `<InputSelect>` component, the list of options isn't clearly visible to users of: Firefox (MacOS & Windows) and Safari (MacOS) while the viewport is is >576px. I think the cause was due to the top level  <label> being inline so it's children were collapsing.

### DIFF
--- a/packages/orbit-components/src/InputSelect/InputSelect.styled.ts
+++ b/packages/orbit-components/src/InputSelect/InputSelect.styled.ts
@@ -10,6 +10,7 @@ import { Field } from "../InputField";
 export const StyledLabel = styled.label<{ spaceAfter?: Props["spaceAfter"] }>`
   position: relative;
   margin-bottom: ${getSpacingToken};
+  display: block;
 `;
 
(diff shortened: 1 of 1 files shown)


mainframev:
hi @ConorHaining , thank you for contribution! 🚀

mvidalgarcia:
I've tried the fix locally and it fixes the issue.


https://github.com/kiwicom/orbit/assets/8863238/b9d2f25e-c2e4-42f7-8e5c-ee298970ceab

mvidalgarcia:
@mainframev @DSil the pipeline will never pass as the contribution comes from a repo fork. How should we proceed with this?

The change is good from my point of view and it's so simple that we could merge it.

mvidalgarcia (approved):

mvidalgarcia:
Merged! thanks @ConorHaining

PR #4125, opened 2023-10-30
feat(Separator): migrate to Tailwind

mainframev:

I've decided to introduce a few small breaking changes during this refactoring:

- Renamed *color* prop to *borderColor* and made it accept tailwind class instead of a token. It does not make sense to create a large object similar to a Box background colors with border colors for Separator, when it can accept an appropriate tailwind classname. On our prod both properties needed for custom color (type and color) and they are used only 2 times from more than 300+ instances, so it's a **minimal breaking change.** 
- *type* prop now by default is *solid*, not *none*. Before it was a bit confusing, as it was combining background and border-color for styles, so it had to accept *type* explicitly for applying border styles instead of background, which is a bit confusing. So I removed that and now it uses only borderColor, so no need to have it by default as *type=none* , also removed the conditional type, because it's not needed anymore

### For tailwind migrations:

- [x] Visual regression test suite has been run after the PR approval and no visual changes were identified.

[FEPLT-1718](https://kiwicom.atlassian.net/browse/FEPLT-1718)
 Storybook: https://orbit-mainframev-tw-migrate-separator.surge.sh

[FEPLT-1718]: https://kiwicom.atlassian.net/browse/FEPLT-1718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DSil:
I am still reviewing this, but I don't find the visual test in Chromatic. It should have been executed **before** the migration to Tailwind.

Can you please do that? You'd need to reset commits, keep only the visual test commit, force push, run chromatic, accept the changes, push the migration commits again and after PR review run Chromatic again and verify there are no changes

packages/orbit-components/src/Separator/types.d.ts
```diff
@@ -5,98 +5,88 @@ import type * as Common from "../common/types";
...
+  | "border-white-hover"
+  | "border-white-active";
 
-export interface CommonProps extends Common.SpaceAfter {
-  indent?: Indent;
-  align?: Align;
+export interface Props extends Common.SpaceAfter {
+  readonly indent?: Indent;
+  readonly align?: Align;
+  readonly borderColor?: BorderColorClass;
```
DSil, line 90: I agree with the change to accept classes instead of the tokens as strings. But honestly I'd keep the prop name as `color`. It's less confusing for the migration and it's very much explicit, as the `color` prop refers to the `Separator` components, so it's, intuitively, the "color of the Separator".

packages/orbit-components/src/Separator/types.d.ts
```diff
@@ -5,98 +5,88 @@ import type * as Common from "../common/types";
...
+  readonly type?: "solid" | "dashed" | "dotted" | "double" | "none";
```
DSil, line 91: This is a good simplification and a good change. However, there is one catch:
In the past, if someone had explicitly `type="none"` without any `color` defined, the separator would be visible. But now it is not.

I am not against the change. But we need this change to be clear and documented, besides the others that you mention on the PR description.

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
-    <StyledContainer $align={align} $indent={indent}>
-      <StyledSeparator
-        $type={type}
-        spaceAfter={spaceAfter}
-        $color={color && (theme.orbit[color] as Props["color"])}
+    <div className={cx("box-border w-full", getIndentAmount(indent, align))}>
+      <hr
+        className={cx(
+          "orbit-separator",
+          // h-0 and border-t are the same as in default tailwind preflight
```
DSil, line 69: I'd say we don't need this comment. I'm ok with relying on preflight. I am concerned about how it can affect who is not using it, though. Nothing to change here, just so we are aware.

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
+          "mt-0 box-border h-0 border-t",
+          borderColor || "border-elevation-flat-border-color",
+          type && BorderTypes[type],
```
DSil, line 72: We don't need the `type` verification as we have a default value for the prop, right?

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
-import getSpacingToken from "../common/getSpacingToken";
-import { left, right } from "../utils/rtl";
-import type { Props, Indent, Align } from "./types";
-import useTheme from "../hooks/useTheme";
+import type { Props, Align } from "./types";
+import { getSpaceAfterClasses } from "../common/tailwind";
 
-function capitalize(string: string) {
-  return string.charAt(0).toUpperCase() + string.slice(1);
+enum BorderTypes {
```
DSil, line 9: Why not an object like in other places? And I'd call it `borderTypeClasses`

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
-  $indent: Indent;
-}) {
-  return $indent === "none" ? 0 : theme.orbit[`space${capitalize($indent)}`];
+enum Indent {
+  none = "none",
+  small = "small",
+  medium = "medium",
+  large = "large",
+  XLarge = "XLarge",
+  XXLarge = "XXLarge",
```
DSil, line 23: I'd say we don't need this enum 🤔 We can say the function's parameter is of the same type as the prop

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
-    width: 100%;
-    padding-${right}: ${$align === "left" && getIndentAmount};
-    padding-${left}: ${$align === "right" && getIndentAmount};
-    padding: ${$align === "center" && `0 ${getIndentAmount}`};
-  `};
-`;
+function getIndentAmount(indent: keyof typeof Indent, align: Align) {
+  const classes = {
+    left: {
+      [Indent.none]: "0",
```
DSil, line 29: Shouldn't it be `pe-0`? And same for others

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -1,77 +1,79 @@
...
-const StyledContainer = styled.div<{ $align: Align; $indent: Indent }>`
-  ${({ $align }) => css`
-    box-sizing: border-box;
-    width: 100%;
-    padding-${right}: ${$align === "left" && getIndentAmount};
-    padding-${left}: ${$align === "right" && getIndentAmount};
-    padding: ${$align === "center" && `0 ${getIndentAmount}`};
-  `};
-`;
+function getIndentAmount(indent: keyof typeof Indent, align: Align) {
```
DSil, line 26: While reviewing, I noticed that this is very non-intuitive… If we say `align="left" indent="large"` we suddenly have a big padding… on the right side. That is the exact opposite of what's expected. An indentation refers to a spacing at the start, and if it's aligned to the left, the indentation space should be a padding on the left.

This is a problem that already existed, so it's not introduced by this PR. I'd say it could be done later, but we're introducing a breaking change already so we could avoid two BC in a row… Also, to fix it is not so costly… so I'd consider doing it. We either rename the prop (I can't propose a good name, tbh) or we switch the classes applied, which can be more confusing for migrating, but will actually be semantically correct.

DSil (changes requested):

packages/orbit-components/src/Separator/index.tsx
mainframev: actually, it's the opposite, by adding this we aren't relying on preflight, because I feel that in case if they won't use it -> the component will be broken, so I explicitly added it here :)

DSil:
I also forgot to mention we need these breaking changes to be documented on the migration guide, that needs to be in sync with the one from #4114 (still TO DO)

packages/orbit-components/src/Separator/index.tsx
mainframev: I did it in the same way as we have it in `common/tailwind` folder

There is one thing for the future, that's out of the scope of this PR, but this can be useful for future Orbit optimization with TW, but there are two things that need to be improved and keep in mind: 

- types like this (Indent above) should be `const enum`, not just `enum`
- we need to add `optimizeConstEnums: true` to our `@babel/preset-typescript`

so after build we will have output like this (inlined primitive values, but in the code we still can have enums, that will be removed after build)  
Screenshot 2023-11-01 at 20 03 07

packages/orbit-components/src/Separator/index.tsx
mainframev: I would say it's rather a problem of naming than a problem of behavior. It should work like that if we say `align="left"`, the padding should be on the right side and vice versa, so after the Separator actually. So I think it's correct in terms of behavior. Maybe we just need to rename it, so it's more descriptive for behavior 🤔

packages/orbit-components/src/Separator/Separator.stories.tsx
```diff
@@ -22,12 +22,18 @@ Default.story = {
 export const Playground = () => {
   const spaceAfter = select("spaceAfter", Object.values(SPACINGS_AFTER), SPACINGS_AFTER.LARGEST);
   const align = select("align", ["left", "right", "center"], "left");
-  const indentOptions: Indent[] = ["none", "small", "medium", "large", "XLarge", "XXLarge"];
-  const type = select("type", ["solid", "dashed", "dotted", "double", "none"], "none");
-  const color = text("color", "paletteBlueNormal") as PaletteTokens;
-  const indent = select("indent", indentOptions, "none");
+  const sideOffsetOptions: SideOffset[] = ["none", "small", "medium", "large", "XLarge", "XXLarge"];
+  const type = select("type", ["solid", "dashed", "dotted", "double", "none"], "solid");
+  const color = text("color", "border-blue-normal") as BorderColorClass;
```
mvidalgarcia, line 27: Can we make this a select like Skeleton?

```
  const color = select(
    "color",
    Object.keys(defaultTheme.orbit).filter(t => t.startsWith("palette")),
    "",
  );
  ```

File: packages/orbit-components/src/Separator/index.tsx [hunk 10]
```diff
@@ -1,77 +1,78 @@
...
-    <StyledContainer $align={align} $indent={indent}>
-      <StyledSeparator
-        $type={type}
-        spaceAfter={spaceAfter}
-        $color={color && (theme.orbit[color] as Props["color"])}
+    <div className={cx("box-border w-full", getSideOffsetAmount(sideOffset, align))}>
+      <hr
+        className={cx(
+          "orbit-separator",
+          "mt-0 box-border h-0 border-t",
```
mvidalgarcia, line 69: Is the `h-0` correct?
In SC we had `height: ${theme.orbit.heightSeparator};` and I see its value is 1px, isn't it?
I think it doesn't have any visual effect though 🤔

packages/orbit-components/src/Itinerary/ItinerarySeparator/types.d.ts
```diff
@@ -5,5 +5,5 @@ import type { Props as SeparatorProps } from "../../Separator/types";
 export interface Props {
   readonly children?: React.ReactNode;
   readonly type?: SeparatorProps["type"];
-  readonly color?: SeparatorProps["color"];
+  readonly color?: string;
```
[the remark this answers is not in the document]
mainframev: because SeparatorProps accept Tailwind class, while ItinerarySeparator still token

packages/orbit-components/src/Separator/index.tsx
mainframev: I made it the same as Tailwind preflight default styles for hr element, so it's explicitly set, so in case if Tailwind preflight is used or not -> it's the same. We do not need the height anymore as it's not based on background as before and based on border :)

mvidalgarcia (approved):

PR #4126, opened 2023-10-30
Migrate Toast to Tailwind

oreqizer:
https://kiwicom.atlassian.net/browse/FEPLT-1711
 Storybook: https://orbit-mainframev-oreqizer-tw-toast.surge.sh

couldn't really make visual tests in Chromatic bcs we can't block animations there, and it would be broken all the time


packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -122,15 +79,26 @@ const ToastMessage = ({
           setPaused(false);
         }}
       >
+        <div
+          className="rounded-large bg-white-normal absolute left-0 top-0 z-[10] h-full w-full animate-[toast-light_var(--toast-message-duration)_linear] opacity-10 will-change-transform"
```
mainframev, line 83: before z-index was 1, not 10, don't think it will change something, but let's keep it as it was :) 

```suggestion
          className="rounded-large bg-white-normal absolute left-0 top-0 z-[1] h-full w-full animate-[toast-light_var(--toast-message-duration)_linear] opacity-10 will-change-transform"
```

mainframev:
@oreqizer the commit about adding toast animation should be a chore, not feature :) It's for our internal API, we do not document it for users ;)

mainframev (changes requested):
LGTM, there are few small things 👍🏻

mainframev:
The animation in RTL is not working properly. In current master it goes from right to left, in this PR it's keeping to move from left to right 
Screenshot 2023-11-02 at 16 09 22
Screenshot 2023-11-02 at 16 09 15

packages/orbit-components/src/Toast/ToastMessage.tsx
oreqizer: adjusted

packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -96,23 +26,50 @@ const ToastMessage = ({
...
-      offsetY={offset}
-      offsetX={swipeOffset}
-      placement={placement}
+    <div
+      aria-live={ariaLive}
+      role="status"
+      className={cx(
+        "z-onTop duration-normal absolute left-0 right-0 flex cursor-grab cursor-grab transition-all ease-in-out will-change-transform",
+        "translate-x-[var(--toast-message-offset-x)] translate-y-[var(--toast-message-offset-y)] opacity-[var(--toast-message-opacity)]",
+        (placement === "top-right" || placement === "bottom-right") && "justify-end",
```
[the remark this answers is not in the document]
oreqizer: adjusted, good suggestion 👍

packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -96,23 +26,50 @@ const ToastMessage = ({
...
-      opacity={swipeOpacity}
-      visible={visible}
-      offsetY={offset}
-      offsetX={swipeOffset}
-      placement={placement}
+    <div
+      aria-live={ariaLive}
+      role="status"
+      className={cx(
+        "z-onTop duration-normal absolute left-0 right-0 flex cursor-grab cursor-grab transition-all ease-in-out will-change-transform",
```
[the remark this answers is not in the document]
oreqizer: ah thank, fixed

packages/orbit-components/src/Toast/ToastRoot.tsx
```diff
@@ -43,13 +25,21 @@ const ToastRoot = ({
...
-      $bottom={bottomOffset}
-      $left={leftOffset}
-      $right={rightOffset}
+      className={cx(
+        "z-onTop pointer-events-none fixed",
+        "bottom-[var(--toast-root-bottom)] end-[var(--toast-root-end)] start-[var(--toast-root-start)] top-[var(--toast-root-top)]",
+      )}
+      style={
+        {
+          "--toast-root-top": `${topOffset}px`,
```
mainframev, line 37: Found some more. Positioning is not correct compared to master. If you'd set the position bottom -> it should be at the bottom of the viewport, now it's not. I guess this is happening because of this: (as it has position fixed and it should be ok) 

Screenshot 2023-11-03 at 11 33 16

mainframev:
> couldn't really make visual tests in Chromatic bcs we can't block animations there, and it would be broken all the time

This can be solved by adding a new prop, something like `disableAnimation` that will turn it off -> that way it won't be broken in visual testing. That would help us to have positions tested under visual regression. This is slightly out of the scope, so I will leave the decision to you :)

mainframev (approved):
good job 🚀

PR #4129, opened 2023-10-31
feat(Skeleton): migrate to Tailwind

mvidalgarcia:
- chore(Skeleton): add visual tests
- chore(Skeleton): use more meaningful rtl story
- feat(Skeleton): migrate to Tailwind

Closes FEPLT-1720

### For tailwind migrations:

 Storybook: https://orbit-mainframev-tw-skeleton.surge.sh


DSil:
I am still reviewing this, but I don't find the visual test in Chromatic. It should have been executed before the migration to Tailwind.

Can you please do that?

packages/orbit-components/src/Skeleton/helpers.ts
DSil, line 1: I think we can get rid of this file entirely and have these functions on the component file, since it's the only place they're used. WDYT?

packages/orbit-components/src/Skeleton/helpers.ts
```diff
@@ -1,6 +1,3 @@
-import type { FlattenSimpleInterpolation } from "styled-components";
-import { keyframes, css } from "styled-components";
-
 export const resolveValue = (value: string | number): string => {
```
DSil, line 1: This function is weird. It function accepts a string or number. Checks types and returns accordingly. But then it returns "100%" if both conditions fail... which should be impossible to happen...
So should 100% be the default value, that was being ignored, as a bug? Or that's just an unnecessary return?

It has implications on the functions usage, since it's being used with a trenary to return undefined.

I see for width it's not problematic because it has the default value of "100%", which is ultimately the same of this function. But for height and max-height it's slightly different.
Height can be undefined, but it passes by `resolveHeight` before reaching this. If it is undefined, `resolveHeight` immediately returns `"100%"`.
But if maxHeight is undefined, this function should return "100%". However, on the trenary, you're doing `maxHeight != null ? resolveValue(maxHeight) : undefined`. That won't apply the expected "100%" to the maxHeight property.

I also see `resolveHeight` might have a problem with `calculatedHeight` when it is 0. As it evaluates to false but maybe we want it to return `0px`. If not, then the function is not clearly readable...

Can you please take a look at those potential issues? :pray:

mainframev: @DSil Yep, agree it's weird 🚀 but It becomes weird with TW, with SC it was kinda ok to have it like that. In case if height prop is undefined, then it returns 100% for the height value in SC :) So it's possible to happen

It can be improved by setting height as default `100%`, the same as for width, and change the function to simply this: 

```

export const resolveHeight = ({
  calculatedHeight,
  height,
}: {
  calculatedHeight: number;
  height: number | string;
}): string => {
  if (calculatedHeight) return `${calculatedHeight}px`;
  return resolveValue(height);
};
``` 

or we can completely omit this helper function and do it directly inside, which is even better imo: 

```
      style={{
        height: calculatedHeight > 0 ? `${setCalculatedHeight}px` : resolveValue(height),
        ...
      }}

```

> I also see resolveHeight might have a problem with calculatedHeight when it is 0. As it evaluates to false but maybe we want it to return 0px

I feel this is correct. I do not think we need the function to return `0` for calculatedHeight, the goal of `calculatedHeight` is to get the height of the rows, so we do not need it, if there are no rows and no height, so then the value should be based on height prop :)

mainframev (changes requested):

packages/orbit-components/src/Skeleton/helpers.ts
mvidalgarcia: Thanks for your comments, I applied the suggested changes. I also renamed `calculatedHeight` to `calculatedRowsHeight` to make it more clear. Let me know.

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -93,6 +100,40 @@ export const Presets = () => {
   return <Skeleton preset={presets} />;
 };
 
+export const VisualTest = () => {
```
mainframev, line 103: One last thing. I think we shoud set `animte=false` explicitly, so it won't fail on visual regression 🤔

mainframev (approved):
LGTM 🚀

PR #4131, opened 2023-10-31
docs: use splat route for dashboard

mainframev:
That should fix the problem with broken layout in dashboard pages. [Splat routes in gatsby docs](https://www.gatsbyjs.com/docs/reference/routing/file-system-route-api/#splat-routes)
 Storybook: https://orbit-mainframev-docs-splat-route.surge.sh


oreqizer:
how does it fix the problem? just a brief description pls

mainframev:
> how does it fix the problem? just a brief description pls

I believe it should fix it, as it's CSR and before routing was not correct, because it had the main page with routes under "dashboard/index*, and when we were opening the link like "dashboard/tracking/allrepositories" it was broken because it had to check first if the user is logged in and routing is not working in that case as it's not applied for that path, only when you are opening dashboard page without the path after trailing slash

Now the routing mechanism will be applied to all paths after dashboard

oreqizer (approved):

PR #4132, opened 2023-11-02
migrate Table to Tailwind

oreqizer:

### For tailwind migrations:

- [x] Visual regression test suite has been run after the PR approval and no visual changes were identified.

 Storybook: https://orbit-mainframev-oreqizer-tw-table.surge.sh

https://kiwicom.atlassian.net/browse/FEPLT-1769


packages/orbit-components/src/Table/TableBody/index.tsx
```diff
@@ -1,27 +1,11 @@
-"use client";
```
mvidalgarcia, line 1: Do we want to delete this? The same applies to all the sub-components.

packages/orbit-components/src/Table/TableHead/index.tsx
```diff
@@ -1,31 +1,14 @@
...
 
 import type * as Common from "../../common/types";
-import defaultTheme from "../../defaultTheme";
-import { StyledTableCell } from "../TableCell";
 
-export const StyledTableHead = styled(({ children, className, dataTest }) => (
-  <thead className={className} data-test={dataTest}>
+const TableHead = ({ children, dataTest }: React.PropsWithChildren<Common.Globals>) => (
+  <thead
+    className="border-b-table-head font-table-head w-full whitespace-nowrap border-b"
```
mvidalgarcia, line 7: I think `font-table-head` should only target table cell children. Although I believe having this style in the parent doesn't hurt, I can't see any side effects. Let me know if it was intentional.

mvidalgarcia: I would add `border-solid` explicitly here as well.

packages/orbit-components/src/Table/index.tsx
```diff
@@ -144,20 +47,42 @@ const Table = ({
...
     >
-      <StyledTableInner ref={inner} onScroll={handleScroll} showShadows={shadows}>
-        <StyledTable striped={striped} compact={compact} type={type} ref={table}>
+      <div
+        className={cx("w-full", shadows && "overflow-x-auto")}
+        onScroll={handleScroll}
+        ref={inner}
+      >
+        <table
+          className={cx(
```
mvidalgarcia, line 69: Just to confirm, we don't need `display: table`, right?

packages/orbit-components/src/Table/index.tsx
```diff
@@ -144,20 +47,42 @@ const Table = ({
...
+            "w-full border-collapse border-spacing-0 whitespace-nowrap",
+            "[&_tbody>tr]:bg-table hover:[&_tbody>tr]:bg-table-hover [&_tbody>tr]:border-b-table [&_tbody>tr]:duration-fast [&_tbody>tr]:border-b [&_tbody>tr]:transition-colors [&_tbody>tr]:ease-in-out last:[&_tbody>tr]:border-b-0",
+            striped === true && "[&_tbody>tr:nth-of-type(even)]:bg-table-even",
```
mvidalgarcia, line 72: Is it possible to use our custom variant here? – `"type-even:[&_tbody>tr]:bg-table-even"`

https://github.com/kiwicom/orbit/blob/2bcce2314b3e2681eaf931b0873f53f7987e8fff/packages/orbit-tailwind-preset/src/presets/components/index.ts#L347

packages/orbit-components/src/Table/index.tsx
```diff
@@ -144,20 +47,42 @@ const Table = ({
...
+            "[&_td]:px-sm [&_th]:px-sm [&_td]:leading-normal [&_th]:leading-normal",
+            compact === true
+              ? // TODO: remove 10px and 6px with new tokens
```
mvidalgarcia, line 75: Is this to be done now?

docs/src/data/tailwind-migration-status.yaml
```diff
@@ -88,7 +88,7 @@ StopoverArrow: false
 Switch: true
 Tab: false
 TabPanel: false
-Table: false
+Table: true
 TableBody: false
```
mvidalgarcia, line 92: Please update all the sub-components as well.

packages/orbit-components/src/Table/TableBody/index.tsx
oreqizer: yep, it is not necessary. the component contains no hooks/state, or other constructs incompatible with RSCs

packages/orbit-components/src/Table/TableBody/index.tsx
```diff
@@ -1,27 +1,11 @@
...
 import * as React from "react";
-import styled from "styled-components";
 
 import type * as Common from "../../common/types";
-import defaultTheme from "../../defaultTheme";
 
-export const StyledTableBody = styled(({ children, className, dataTest }) => (
-  <tbody className={className} data-test={dataTest}>
+const TableBody = ({ children, dataTest }: React.PropsWithChildren<Common.Globals>) => (
+  <tbody className="border-b-table w-full whitespace-nowrap border-b" data-test={dataTest}>
```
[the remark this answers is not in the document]
oreqizer: for maintainability, we should strive to utilise sensible defaults and cascading over overly specifying styles. I'd avoid specifying styles we don't need to write explicitly

it would also open a Pandora's box — which styles do we _not_ specify? we would have to specify e.g. `box-border block` on every `<div>`

packages/orbit-components/src/Table/TableHead/index.tsx
oreqizer: see my answer above — cascading (in case of `font-table-head` on `thead`) and sensible defaults (`border-solid` is the default) is the most scalable and maintainable way

packages/orbit-components/src/Table/index.tsx
oreqizer: yea, that's the default for `table`

packages/orbit-components/src/Table/index.tsx
oreqizer: TIL 👍

docs/src/data/tailwind-migration-status.yaml
oreqizer: right 🤦‍♀️ didn't notice those

packages/orbit-components/src/Table/TableBody/index.tsx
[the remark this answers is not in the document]
mvidalgarcia: I agree overall. I was just following @mainframev's practice, as he's adding some styles in case preflight is disabled. And since this case, it was explicit in styled-components I thought we could include it. The search team is facing some issues because of this very same reason, but it should be their responsibility.

packages/orbit-components/src/Table/TableCell/index.tsx
```diff
@@ -1,61 +1,39 @@
...
   return (
-    <StyledTableCell
-      verticalAlign={verticalAlign}
-      whiteSpace={whiteSpace}
-      align={align}
-      dataTest={dataTest}
+    <Component
+      className={cx(
+        "font-base text-normal text-ink-dark box-border",
+        "whitespace-[var(--table-cell-white-space)] align-[var(--table-cell-vertical-align)]",
```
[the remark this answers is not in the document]
oreqizer: right. enum'd these both. I didn't check their types

_point for keeping types next to component definitions_ ☝️

packages/orbit-components/src/Table/index.tsx
oreqizer: didn't want to change things beyond the migration itself. there aren't any related tokens atm, and didn't know whether these are to be specified by designers, or I could create them myself

packages/orbit-components/src/Table/TableBody/index.tsx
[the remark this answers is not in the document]
oreqizer: yep that's the point. we can't consider every wild thing consumers are doing 😄 we should to write the minimum we need, and let the rest cascade and default nicely

mvidalgarcia (approved):
Nice works splitting the commits, it made the review process much easier 👍 

:shipit: 


PS. Remember to run Chromatic tests.

PR #4142, opened 2023-11-09
fix: tailwind preset to support rgb opacity

mainframev:
[Slack report](https://skypicker.slack.com/archives/C04HMMQ6VNE/p1699355684758489), [Slack thread ](https://skypicker.slack.com/archives/C05FDP4UM3N/p1699379061340519) 

To be ready for review at the beginning of SPRINT-26 

Giving solution to [FEPLT-1801](https://kiwicom.atlassian.net/browse/FEPLT-1801)


packages/orbit-tailwind-preset/src/presets/foundation/cssVarsFoundation.ts
```diff
@@ -1,102 +1,111 @@
-import { defaultTokens } from "@kiwicom/orbit-design-tokens";
+import { defaultTokens, parseToRgba } from "@kiwicom/orbit-design-tokens";
+
+const getChannels = (color: string) => parseToRgba(color).slice(0, -1).join(", ");
+const rgba = (color: string, fallback: string) =>
```
DSil, line 4: Nitpick and non blocking suggestion:
I was confused about this when reading the code below, on the class definition. Because I though you were using directly the css property `rgba` but then you were using the string directly, instead of the `var(--`. Only after I found out that was actually this function, so an abstraction.

My suggestion is to rename this function for better and undoubtable reading of the code. I wouldn't mind calling it `generateRGBA`, but up to you.

packages/orbit-tailwind-preset/src/presets/__tests__/colors.test.tsx
DSil, line 1: Does this mean we're not testing the color classes?

packages/orbit-design-tokens/src/js/transparentColor.ts
```diff
@@ -1,11 +1,16 @@
...
+
 const transparentColor = (color: string, opacity: number): string => {
   // tailwind package is using css variables
   if (color.match(/var\(/)) {
     const regex = /var\(([^,]+),\s*([^)]+)\)/;
     const [, , colorValue] = color.match(regex);
-    return transparentColor(colorValue, opacity);
+    const value = colorValue as string;
+    const isValidColor = isHex(value) || isRgb(value);
+    return transparentColor(isValidColor ? value : `rgb(${value})`, opacity);
```
DSil, line 13: I understand this code, but why is this necessary now?

packages/orbit-design-tokens/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import { parseToRgba } from "color2k";
```
DSil, line 1: I don't think we should be doing this. We are reexporting this inside our package just to be used in another package. If that's the case, the package that needs it should import it.

packages/orbit-tailwind-preset/src/presets/__tests__/colors.test.tsx
mainframev: yep, I think not needed, covered by visual testing and it was broken, because of this change

packages/orbit-design-tokens/src/js/transparentColor.ts
mainframev: because it's broken without, it recieves "255, 255, 255" and need to parse this value

packages/orbit-design-tokens/src/index.ts
mainframev: this isn't working like that, in case if the package will import this -> it will need to have it as dependency. I do not want to add additional dependency to tailwind preset. It suits tokens API, where we export those functions like for rgba and transparent color

DSil: > I do not want to add additional dependency to tailwind preset.

It is actually being used in `orbit-components`. And I don't agree with this "masking" dependency. If it's needed there, it's a dependency there. It doesn't feel right to be reexporting functions from other packages.

Either that or we find a way to export the data already processed (I'm not sure if it's possible) so we don't use the function there.

packages/orbit-tailwind-preset/src/presets/foundation/cssVarsFoundation.ts
mainframev: yeah, good suggestion, thanks

packages/orbit-design-tokens/src/index.ts
mainframev: 👍🏻  can you give more arguments here why it's not right in your opinion? I feel like we do not have strong arguments against it. Reexporting is fine in this case imo, we have a tokens package and related functions to it, if the function is used in the tokens package and reexported -> it's fine to me, I do not see much value in adding it on both places (tailwind and components) and adding additional dep to components because of single usage that might be removed later :) 

We are reexporting a single function, that will be used on two other packages. I am not sure about how much it will increase bundle size, if we will add this package as a dependency to other two packages, I am also not comfortable with adding a new dependency because of single function, that might be removed later :) 

This solution looks less invasional to me, wdyt? 🙂

DSil (commented):
LGTM, just one issue with the commit messages. The first commit says
`fix(tailwind):` but it touches three different packages. I think we should split those. Those that are unnoticeable / irrelevant changes to the user are just chores. And then the actual fix just touching the tailwind-preset package

packages/orbit-design-tokens/src/index.ts
```diff
@@ -11,4 +13,4 @@ export { default as tokensToCssVars } from "./tokensToCssVars";
 export { default as defaultFoundation } from "./js/defaultFoundation";
 export { default as convertHexToRgba } from "./js/convertHexToRgba";
 export { default as convertRgbaToHex } from "./js/convertRgbaToHex";
-export { createTheme, PaletteColors };
+export { createTheme, PaletteColors, parseToRgba };
```
DSil, line 16: Sorry didn't notice it before. But now we don't need to export this, right?

DSil (commented):
LGTM, just one last question that I missed on the previous review

packages/orbit-design-tokens/src/index.ts
mainframev: right 👍🏻

DSil (approved):

PR #4144, opened 2023-11-10
fix(Card): only capture keys on expandable sections

oreqizer:
https://skypicker.slack.com/archives/C7T7QG7M5/p1699604167098469
 Storybook: https://orbit-mainframev-oreqizer-fix-card-interaction.surge.sh

### DIFF
--- a/packages/orbit-components/src/Card/CardSection/CardSection.ct.tsx
+++ b/packages/orbit-components/src/Card/CardSection/CardSection.ct.tsx
@@ -1,0 +1,16 @@
+import * as React from "react";
+import { test, expect } from "@playwright/experimental-ct-react";
+
+import { CardSectionInput } from "./CardSection.ct-story";
+
+test.describe("interaction", () => {
+  test("typing spaces works on non-expandable sections", async ({ mount }) => {
+    const component = await mount(<CardSectionInput />);
+    const input = component.getByTestId("input");
+
+    await input.focus();
+    await input.press("Space");
+
+    expect(await input.inputValue()).toBe(" ");
+  });
+});
--- a/packages/orbit-components/src/Card/CardSection/index.tsx
+++ b/packages/orbit-components/src/Card/CardSection/index.tsx
@@ -65,7 +65,7 @@ export default function CardSection({
       tabIndex={onClick == null ? undefined : 0}
       onClick={onClick}
       // Not needed once we can use <button> or <a> like we should
-      onKeyDown={handleKeyDown(onClick)}
+      onKeyDown={onClick == null ? undefined : handleKeyDown(onClick)}
     >
       {(title != null || header != null) && expandable && (
         <button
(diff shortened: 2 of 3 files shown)


packages/orbit-components/src/Card/CardSection/CardSection.ct.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { test, expect } from "@playwright/experimental-ct-react";
+
+import { CardSectionInput } from "./CardSection.ct-story";
+
+test.describe("interaction", () => {
+  test("typing spaces works on non-expandable sections", async ({ mount }) => {
+    const component = await mount(<CardSectionInput />);
+    const input = component.getByTestId("input");
+
+    await input.focus();
```
mainframev, line 11: isn't this should be normally tested with testing-library?

oreqizer: we don't have `CardSection` Jest tests, so I created this. and it has one big benefit — it tests the actual browser environment, not a simulated one

mainframev: hm, can we add it to a jest then? it's a bit strange to have it under visual regression 🤔 And don't think it matters here what environment it is. Sorry for complications, I just feel it's not the purpose of visual regression to test cases like this 🙂

oreqizer: if we want to agree to use Playwright strictly for visual tests, then yes, I can move em, for consistency. though I personally see value in _any_ component tests being in it

mainframev: Yep, let's use it for visual testing only for now. We have testing-library and I do not think we will get rid of it in near feature in favor of playwright, having it tested half here and half there -> isn't a good idea probably. I like playwright experimentals, it seems a better solution than chromatic :) Until now I thought we have it as a replacement for chromatic, not for the whole testing 🙂

packages/orbit-components/src/Card/CardSection/index.tsx
```diff
@@ -65,7 +65,7 @@ export default function CardSection({
       tabIndex={onClick == null ? undefined : 0}
       onClick={onClick}
       // Not needed once we can use <button> or <a> like we should
-      onKeyDown={handleKeyDown(onClick)}
+      onKeyDown={onClick == null ? undefined : handleKeyDown(onClick)}
```
mainframev, line 68: the probolem is not fixed completely, if you will pass onClick, it's still will be broken with InputField. The problem is in event bubbling and ev.preventDefault inside handleKeyDown, it's prevented :)

packages/orbit-components/src/Card/CardSection/CardSection.ct.tsx
mainframev: Resolved. Jest is not failing on this case, playwright does. We can have it in playwright.

mainframev (approved):

PR #4148, opened 2023-11-10
Update @kiwicom/browserslist-config

oreqizer:
includes fix for Safari <= 15
[Slack issue reported](https://skypicker.slack.com/archives/C7T7QG7M5/p1699626858575329). We need to get back Safari iOS 14 <= 15. 
 Storybook: https://orbit-mainframev-oreqizer-fix-babel-fancy-pants-class-props.surge.sh

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,5 +1,10 @@
 [production]
 extends @kiwicom/browserslist-config
 
+ios_saf 14
+ios_saf 15
+safari 14
+safari 15
+
 [test]
 current node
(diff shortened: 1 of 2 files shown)


mainframev:
@oreqizer seems like it can't fetch the package from private registry

mainframev (approved):

mainframev:
@oreqizer Versions 9.3.3 and 10.3.1 are released 

As the change is not releated to any of the packages, Lerna generated only a bump version changelog, but it should include correct build under fixed browserlist

PR #4155, opened 2023-11-16
Add space-x-reverse class for rtl in InputTag

DSil:
The `rtl:space-x-reverse` class was missing so that spacing is correctly applied on RTL
 Storybook: https://orbit-mainframev-input-field-missing-rtl-spacing-class.surge.sh


oreqizer (approved):
👌

PR #4160, opened 2023-11-21
Fix InputField and Radio

oreqizer:
https://kiwicom.atlassian.net/browse/FEPLT-1830
 Storybook: https://orbit-mainframev-oreqizer-fix-form-things.surge.sh

I recommend going commit-by-commit as I
1. generated broken screenshots first
2. fixed the issue
3. regenerated correct screenshots after

after both components were fixed, I generated correct Linux screenshots (and created a convenience `pancake.json` task)

has one unrelated commit that decreases the pixel contrast threshold for it to count as a difference. it was causing false negatives when regenerating screenshots also in this MR so I added it here


packages/orbit-components/src/Radio/index.tsx
```diff
@@ -69,6 +68,7 @@ const Radio = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
             "active:scale-95",
             "peer-focus:outline-blue-normal peer-focus:outline peer-focus:outline-2",
             disabled && "bg-form-element-disabled-background border-transparent",
+            !disabled && "bg-form-element-background",
```
DSil, line 71: Can't it be?
```suggestion
            disabled ? "bg-form-element-disabled-background border-transparent" : "bg-form-element-background"
```

DSil (commented):
I am not in favor of the last two commits belonging to this PR, but I won't block because of that. Just one minor suggestion

DSil (approved):

PR #4163, opened 2023-11-23
fix(Breadcrumbs): the spaceAfter prop

jozef-mikusinec-kiwicom:
# Fix Breadcrumbs#spaceAfter
I highly recommend setting `noImplicitAny` to true in tsconfig. It would have prevented this error. (I'm certainly turning it on in fe/mmb bc wtf :D )

### For tailwind migrations:
N/A

### For new components:
N/A


mainframev (approved):
Thanks for contribution 👍🏻

PR #4166, opened 2023-11-24
fix(Stack): fix height issue applied when no direction column

mainframev:
Hotfix after a recent update, it was adding height: 100% every time, when no column and with the column. I changed it more specific, only when flex and column and when `justify` is added  
 Storybook: https://orbit-mainframev-fix-stack-height.surge.sh


mainframev:
@DSil Sorry for merging that without approval, I just feel it's better to merge it right now in order to not have the broken docs. I feel that the fix is ok and I added the screenshot test for the specific scenario when it was broken, if anything -> we can add it on Monday. Sorry again. 😢 I just expected that we solve it during Friday, but maybe you were busy

PR #4169, opened 2023-11-25
fix(Stack): remove height: 100% for justify and column

mainframev:
Sorry for that mess. I just rethought it again, while it was a change to have expected behavior with a direction column and justify and height, it may lead to visual breaking changes as in our docs, where `justify` was added with a column (but it was not doing anything) and got broken after this change. I think better would be better not to touch it right now, as it's already quite a lot of changes for users. At least we never received any request about this, so maybe it's not needed for users 🤷🏻 

 Storybook: https://orbit-mainframev-fix-stack-height.surge.sh


mainframev:
@DSil I will merge it to fix the broken documentation pages and our components. Basically, here it's the previous state of Stack, including the fix, just removed that idea of having 100% height, as described in the description of the PR. I apologize for merging without approve, just for the sake of not having broken docs. And it's just roll back to the previous state which was ok, so I do not think it needs code review, we may discuss later if this behaviour of justify requires our attention 🙂

PR #4172, opened 2023-11-29
fix(List): fix ListItem icon size

mainframev:
fixes [FELPT-1836](https://kiwicom.atlassian.net/browse/FEPLT-1836) 

 Storybook: https://orbit-mainframev-fix-list-item.surge.sh


DSil (approved):
Non-blocking but could we have a visual test for this?

mainframev:
> Non-blocking but could we have a visual test for this?

added ;)

DSil (approved):

PR #4173, opened 2023-11-30
fix(Popover): links inside tooltips inside content are now clickable

DSil:
The outside click hook was triggering the close before the event was captured by interactable components. This was changed just for the Popover, so no breaking behavior should occur.

[FEPLT-1841](https://kiwicom.atlassian.net/browse/FEPLT-1841)

[FEPLT-1841]: https://kiwicom.atlassian.net/browse/FEPLT-1841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-fix-clik-outside-popup-with-link.surge.sh


mainframev:
@DSil can this behaviour be covered by a simple unit test? 🙏🏻

PR #4174, opened 2023-11-30
fix(Stepper): fix icon circle accoring to Figma

mainframev:
Stepper icon circle fix, should be 20px circle on desktop, responsive should have 24px according to [Figma](https://www.figma.com/file/Y79I18H9w8HJZVxPdAkAHb/%5BOrbit%5D-Desktop-Components?node-id=5925%3A12621&mode=dev)

Also added 44px for ButtonPrimitive as it was before, but that does not have any affect and should be fixed in ButtonPrimitive 
 Storybook: https://orbit-mainframev-fix-stepper-width.surge.sh


DSil (approved):
LGTM. We should have a visual test for this

PR #4183, opened 2023-12-06
Make "smallMobile" of the Hide component hide stuff on screens of size 0-320px

jozef-mikusinec-kiwicom:
"smallMobile" is the smallest screen size that Orbit recognizes. It ought to hide everything for screens of size 0px - 413px.

The alternative would be to add a new screen size smaller than "smallMobile".

Also removes deprecated components Desktop and Mobile from docs
 Storybook: https://orbit-mainframev-jozef-hide-small-mobile.surge.sh

### DIFF
--- a/docs/src/documentation/03-components/12-layout/layout/01-guidelines.mdx
+++ b/docs/src/documentation/03-components/12-layout/layout/01-guidelines.mdx
@@ -15,8 +15,6 @@ To choose the best component for your use case, see our [developing layouts guid
 
 ## When not to use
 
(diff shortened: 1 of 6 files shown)


mainframev:
@jozef-mikusinec-kiwicom tbh I don't think it's a big issue, as unlikely there are screens like 200px, but thanks for contribution and it will be more correct now 👍🏻

mainframev:
@jozef-mikusinec-kiwicom just update the snapshots and it's good

PR #4188, opened 2023-12-11
chore(tokens): fix node issue with ts-node

mainframev:
This should fix the issue with error `UNKNOWN FILE EXTENESION .ts` 

Found issue on GitHub, quite the same with our [here](https://github.com/TypeStrong/ts-node/issues/1997). Tried in on both, node v18 and node v20 - works. 
 Storybook: https://orbit-mainframev-fix-ci-fail.surge.sh


mainframev:
@DSil I think it should be ok after merge, could not reset the action even after deleting the cache

DSil (approved):

PR #4191, opened 2023-12-12
docs(Dialog): fix bad description for onClose prop

DSil:
Reported live by @ArtemKutafin 
 Storybook: https://orbit-mainframev-fix-dialog-readme-onClose-prop.surge.sh


mainframev (approved):
good catch 👍🏻

PR #4194, opened 2023-12-14
fix(OrbitProvider): be more defensive when transforming palette color…

RobinCsl:
…s to RGBA

There was an issue that some input colors had extraneous spaces by mistake, which caused color2k's toRgba function to throw an exception.

### For tailwind migrations:

- [ ] Visual regression test suite has been added and run before the migration.
- [ ] `tailwind-migration-status.yaml` file has been updated with the migrated component(s).

### For new components:

- [ ] New Components have both `.flow` and `d.ts` files and are exported in `index.js.flow` and `index.d.ts`

 Storybook: https://orbit-mainframev-rcsl-try-catch-color2k.surge.sh

### DIFF
--- a/packages/orbit-components/src/OrbitProvider/index.tsx
+++ b/packages/orbit-components/src/OrbitProvider/index.tsx
@@ -2,9 +2,8 @@
 
 import * as React from "react";
 import { ThemeProvider as StyledThemeProvider } from "styled-components";
-import { tokensToCssVars } from "@kiwicom/orbit-design-tokens";
+import { tokensToCssVars, defaultTokens } from "@kiwicom/orbit-design-tokens";
 import { parseToRgba } from "color2k";
-import type { defaultTokens } from "@kiwicom/orbit-design-tokens";
 
 import QueryContextProvider from "./QueryContext/Provider";
 import RandomIdProvider from "./RandomId/Provider";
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/OrbitProvider/index.tsx
```diff
@@ -14,9 +13,18 @@ import type { Props } from "./types";
 const getCssVarsForWL = (theme: typeof defaultTokens) =>
   Object.keys(theme).reduce((acc, key) => {
     if (key.startsWith("palette")) {
-      acc[key] = parseToRgba(theme[key] as string)
-        .slice(0, -1)
-        .join(", ");
+      try {
+        acc[key] = parseToRgba(theme[key] as string)
```
mainframev, line 17: wdyt, could we have some normalization here? kinda same as yours in Nitro, for removing whitespaces

mainframev (commented):
LGTM, could you also maybe add a simple test for covering it? 🙏🏻

packages/orbit-components/src/OrbitProvider/index.tsx
RobinCsl: You mean, on top of the try/catch?

I guess it wouldn't hurt. 🤔 

@oreqizer what do you think?

mainframev: Yep, on top of it, just to avoid the same problem as reported and fixed elsewhere, I can imagine that it might occur, there might be unwanted whitespaces 👍🏻

oreqizer: that is out of scope of Orbit though 😕 this is a Nest-specific bug with the whitespace. I wouldn't leak it into Orbit tbh

mainframev: ok, resolving then this 👍🏻

mainframev (approved):

PR #4195, opened 2023-12-14
fix(Itinerary): some text wasn't truncating properly

DSil:
Truncate component was only targeting text and heading classes, not contemplating the TemporaryText styled component.

Reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1702566892095489).
 Storybook: https://orbit-mainframev-fix-temporary-text-truncate.surge.sh


mvidalgarcia (approved):
Master:

image


This branch:

image

👍

PR #4196, opened 2023-12-18
fix(ButtonPrimitive): add missing shrink-0 and grow-0

mainframev:
Reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1699451085622079) 
Before migration to Tailwind `ButtonPrimitive` had `flex: 0 0 auto` 

Added a small story to replicate the behavior, where you can see that it's fixed 

Giving solution to [FEPLT-1804](https://kiwicom.atlassian.net/browse/FEPLT-1804)

[FEPLT-1804]: https://kiwicom.atlassian.net/browse/FEPLT-1804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-fix-button-missing-flex-and-shrink.surge.sh

### DIFF
--- a/packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
+++ b/packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
@@ -160,6 +160,7 @@ const ButtonPrimitive = React.forwardRef<HTMLButtonElement | HTMLAnchorElement, 
         className={cx(
           className,
           "orbit-button-primitive font-base duration-fast group relative max-w-full items-center border-none text-center leading-none transition-all [&>*]:align-middle [&_.orbit-loading-spinner]:stroke-[currentColor]",
+          "flex-none",
           fullWidth && "w-full",
           centered || children == null
             ? "justify-center"


packages/orbit-components/src/primitives/ButtonPrimitive/index.tsx
```diff
@@ -160,6 +160,7 @@ const ButtonPrimitive = React.forwardRef<HTMLButtonElement | HTMLAnchorElement,
         className={cx(
           className,
           "orbit-button-primitive font-base duration-fast group relative max-w-full items-center border-none text-center leading-none transition-all [&>*]:align-middle [&_.orbit-loading-spinner]:stroke-[currentColor]",
+          "flex shrink-0 grow-0 basis-auto",
```
DSil, line 163: The `flex` can't be there. It's already defined below (line 172).
The other three classes can be simplified into `flex-none`.
So it's just
```suggestion
          "flex-none",
```

packages/orbit-components/src/Button/Button.stories.tsx
```diff
@@ -43,6 +43,17 @@ export const BasicButtons = () => {
...
+  return (
+    <div style={{ width: 300 }}>
+      <Stack direction="row" align="stretch">
+        I am a llama and I like to llama
+        <Button>To be or not to be</Button>
+      </Stack>
+    </div>
+  );
+};
+
```
DSil, line 56: Instead of this story, this could be just a visual test. I don't think it's that interesting for a visual story, as it serves solely for us to be sure it does not wrap. Non-blocking, but I would remove the story and instead add a visual test. WDYT?

mainframev: Idk, there are a bunch of other tests that needed to be added then, but can be solved in the future

PR #4199, opened 2023-12-18
fix(ButtonPrimitive): width prop now works as expected

DSil:
Width prop now works as expected on ButtonPrimitive.

On the Stepper example, it may still not be 44px but that's because it's missing `flex: none`, added in #4196.

Also adjusted the `justify-content` styles, that were off.

[FEPLT-1843](https://kiwicom.atlassian.net/browse/FEPLT-1843)


DSil:
>  @DSil LGTM :) seems like just snapshots needed to be updated

I can't update snapshots. I am still unable to execute playwright locally

mainframev:
> > @DSil LGTM :) seems like just snapshots needed to be updated
> 
> I can't update snapshots. I am still unable to execute playwright locally

Hm, can you describe? maybe I can help. Meanwhile I can update them in this branch, if you do not mind

DSil:
@mainframev it's an old issue when the visual tests were introduced, I was never able to run them. It fails on installing the deps and then it fails on the test script.

Please update the snapshots yourself on this branch

mainframev (approved):

PR #4200, opened 2023-12-19
fix(tailwind): set disablePreflight to false by default

mainframev:
As stated questioned [here](https://skypicker.slack.com/archives/C05FDP4UM3N/p1702999264540639) and stated [here](https://github.com/kiwicom/orbit/blob/master/packages/orbit-tailwind-preset/README.md), it should be by default `false` 
 Storybook: https://orbit-mainframev-fix-tailwind-preflight-def.surge.sh

### DIFF
--- a/packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
+++ b/packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
@@ -3,14 +3,19 @@ import resolveConfig from "tailwindcss/resolveConfig";
 import orbitPreset from "../..";
 
 describe("orbitPreset", () => {
-  it("should match snapshot", () => {
-    const cfg = resolveConfig(orbitPreset());
+  it("should have preflight disabled", () => {
+    const cfg = resolveConfig(orbitPreset({ disablePreflight: true }));
     expect(cfg).toMatchSnapshot();
     expect(cfg.corePlugins).not.toContain("preflight");
   });
 
   it("should have preflight enabled", () => {
     const config = resolveConfig(orbitPreset({ disablePreflight: false }));
+    expect(config.corePlugins).toContain("preflight");
+  });
+
+  it("should have enabled by default", () => {
+    const config = resolveConfig(orbitPreset());
     expect(config.corePlugins).toContain("preflight");
   });
 });
--- a/packages/orbit-tailwind-preset/src/presets/components/index.ts
+++ b/packages/orbit-tailwind-preset/src/presets/components/index.ts
@@ -76,7 +76,7 @@ const getBackgroundColors = (tokens: typeof defaultTokens) => {
 };
 
 const cfg = (options?: Options): Config => {
-  const { disablePreflight = true } = options || {};
+  const { disablePreflight = false } = options || {};
   // make palette colors (foundation colors) defined as css vars and make components tokens inherit it
   const tokens = getTokens(cssVarsFoundation);
   const componentTokens = getComponentLevelTokens(tokens);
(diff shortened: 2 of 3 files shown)


DSil (commented):
This should be, at least, a fix commit, imo. It could even be considered a BC, because we're actually changing the default value. But because we had it documented like this and implemented differently, I'd just call it a fix.

packages/orbit-tailwind-preset/src/presets/components/index.ts
```diff
@@ -76,7 +76,7 @@ const getBackgroundColors = (tokens: typeof defaultTokens) => {
 };
 
 const cfg = (options?: Options): Config => {
-  const { disablePreflight = true } = options || {};
+  const { disablePreflight = false } = options || {};
```
mvidalgarcia, line 79: I ran this test in this PR's branch:

```js
> require("@kiwicom/orbit-tailwind-preset")().corePlugins.preflight
undefined
> require("@kiwicom/orbit-tailwind-preset")({disablePreflight: false}).corePlugins.preflight
undefined
> require("@kiwicom/orbit-tailwind-preset")({disablePreflight: true}).corePlugins.preflight
false
```

I assume Tailwind injects the preflight by default so `undefined` means preflight enabled? Is this correct?

We can also be more explicit on line 87 and always set a boolean:

```js
corePlugins: {
  preflight: !disablePreflight,
},
```

What do you think? This way we don't need to know that preflight is true by default.

packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
```diff
@@ -4,7 +4,7 @@ import orbitPreset from "../..";
 
 describe("orbitPreset", () => {
   it("should match snapshot", () => {
-    const cfg = resolveConfig(orbitPreset());
+    const cfg = resolveConfig(orbitPreset({ disablePreflight: true }));
```
mvidalgarcia, line 7: Can we have another test with `orbitPreset()` as we had before?

packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
```diff
@@ -4,7 +4,7 @@ import orbitPreset from "../..";
...
     expect(cfg).toMatchSnapshot();
     expect(cfg.corePlugins).not.toContain("preflight");
```
mvidalgarcia, line 9: This test only checks if the config value is present, doesn't it? Can we check the value when is present (test below)?

packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
mainframev: why? it's tested as with and without

mainframev: ok, we can, only for sake to have tested default behaviour

packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
mainframev: it's checking opposite, if it's not present, sorry, not sure if I understand what you mean

packages/orbit-tailwind-preset/src/presets/components/index.ts
mainframev: agree, updated

packages/orbit-tailwind-preset/src/presets/__tests__/configs.test.ts
mvidalgarcia: Sorry, my bad. I didn't realize `resolveConfig` was present. I see that it converts it into an array and only has `preflight` when is enabled. Looks good then! Resolving

mvidalgarcia (approved):

PR #4202, opened 2023-12-20
fix(TimelineStep): move subLabel below timeline

mainframev:
Related to [this](https://skypicker.slack.com/archives/GSGN9BN6Q/p1703068328514249) and [this](https://skypicker.slack.com/archives/C7T7QG7M5/p1701691187175999) 

Although it was fixed, in case there is no `subLabel` provided to any `TimelineStep`, there is additional spacing that is not desired 

 Storybook: https://orbit-mainframev-fix-timeline-placement-sublabewl.surge.sh

### DIFF
--- a/packages/orbit-components/src/Timeline/TimelineStep/components/TimelineStepDesktop.tsx
+++ b/packages/orbit-components/src/Timeline/TimelineStep/components/TimelineStepDesktop.tsx
@@ -9,6 +9,7 @@ import Text from "../../../Text";
 import Stack from "../../../Stack";
 import defaultTheme from "../../../defaultTheme";
 import type { Props as StepProps, Type } from "../types";
+import useTheme from "../../../hooks/useTheme";
 
 const StyledDescription = styled.div<{ active?: boolean }>`
   ${({ theme, active }) => css`
@@ -28,6 +29,7 @@ interface Props extends StepProps {
   last: boolean;
   prevType: Type;
   nextType: Type;
+  hasSubLabelMargin?: boolean;
 }
 
 const TimelineStepDesktop = ({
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/Timeline/TimelineStep/components/TimelineStepDesktop.tsx
```diff
@@ -59,6 +56,13 @@ const TimelineStepDesktop = ({
         />
       </StyledRelative>
       <Stack flex align="center" spacing="XSmall" direction="column">
+        {hasSubLabelMargin && (
+          <StyledText>
+            <Text align="center" size="small">
+              {subLabel}
+            </Text>
+          </StyledText>
+        )}
```
DSil, line 65: This will always render a `div` with a `p` inside even if `subLabel` is undefined. That doesn't seem right to me. Since we are at it, we could make it better. Can't we do something like:
Check if it has subLabel: 
- If it does, render the text (we don't need the StyledText wrapper with the min-height because the text already has a line-height bigger than that anyway)
- If it does not but `hasSubLabelMargin === true` then we can apply a top margin to the label text to add the spacing.

Something like:
```
{subLabel && (
  <StyledText>
    <Text align="center" size="small">
      {subLabel}
    </Text>
  </StyledText>
)}
<StyledText active={active || (last && type === "success")}>
  <Text
    align="center"
    weight="bold"
    margin={{ top: !subLabel && hasSubLabelMargin ? "24px" : "0" }}
  >
    {label}
  </Text>
</StyledText>
```

Note that:
- StyledText here is only needed for the color of the Text. I find this very wrong. Our Text component should just support more colors if we need them. We have the same problem in Itinerary component, where we need to have a `TemporaryText` to support more colors. If we need more colors, let's just support them. I'll bring this topic to discuss on the team in our next meeting.
- The hardcoded 24px top margin can be replaced by a token.

mainframev: The `StyledText` is out of the scope of this PR, it should be discussed with designers and so on, so it's completely different story ;)

mainframev: Thanks for the suggestion, it's better and I made it quite quick so didn't consider it! :) 

The discussion of extending Text colors make sense and let's make it in new year 👍🏻

DSil (approved):

PR #4203, opened 2023-12-22
Remove flow types

DSil:
Flow Types were deprecated in v9 and are now removed from Orbit.
 Storybook: https://orbit-mainframev-remove-flow-types.surge.sh


mainframev:
@DSil 

LGTM. Could you also remove "@noflow" from the files? there are multiple files containing that, we do not need that anymore :) 

As far as I remember, only MMB is still using flow on frontend/frontend, so we need to make sure that they are aware of this change when it is released.

DSil:
AFAIK, MMB removed flow types in frontend/frontend some time ago. I don't think anyone else is still using them

mainframev (approved):

PR #4207, opened 2024-01-02
fix(Layout): get back negative margins on Mobile in LayoutColumn

mainframev:
We used to have negative margins in the first Card inside LayoutColumn, it was applied for Mobile, to remove the Card padding and have content "full-width" on Mobile, without paddings on the sides. I guess it was an accidental breaking change and better will be to get it back, maybe it was removed during the Card migration to TW 

The Slack-related [thread](https://skypicker.slack.com/archives/C7T7QG7M5/p1701769129228689)
 Storybook: https://orbit-mainframev-fix-layout-padding.surge.sh

### DIFF
--- a/packages/orbit-components/src/Card/index.tsx
+++ b/packages/orbit-components/src/Card/index.tsx
@@ -46,7 +46,7 @@ export default function Card({
     <div
       id={id}
       className={cx(
-        "font-base bg-white-normal [&>*]:border-elevation-flat-border-color [&>*:first-child]:rounded-t-normal [&>*:last-child]:rounded-b-normal w-full [&>*:first-child]:border-t",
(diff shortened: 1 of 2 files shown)


DSil (approved):
All good, but I'd rewrite the commit message to identify that the problem was specific to Card. It was introduced [here](https://github.com/kiwicom/orbit/pull/3975/commits/76e192892cc7bdbf368db274dd21ae74e1030ca5), during the Card migration and the bug only affects Card. So I'd rewrite the message to something less technical and more straight to the point. Something like

> fix(Layout): get back correct spacing for Card on Mobile

Just a suggestion, and non-blocking

DSil (approved):

PR #4209, opened 2024-01-03
fix(Popover): fix max-height when content has actions

mvidalgarcia:
Max height was not set at all on small devices when the content had actions,
preventing the scroll from working properly.

https://skypicker.slack.com/archives/C7T7QG7M5/p1704291076964889

 Storybook: https://orbit-mainframev-marco-fix-popover.surge.sh


mainframev (approved):
LGTM 👍🏻

PR #4210, opened 2024-01-04
fix(BadgePrimitive): fix min height

mvidalgarcia:
This was having a side-effect on NotificationBagde, making it not circular.

https://skypicker.slack.com/archives/C7T7QG7M5/p1704360137684679


DSil:
This was introduced during the migration of BadgePrimitive and was caused by a mis-definition of the replacement for the deprecated token `heightBadge`. This token was used in the past to define the min-height and had a value of 24px. But when the new token structure was defined and this token was deprecated, the replacement was defined as `iconMediumSize` (20px), instead of the expected `iconLargeSize` (24px) that is being added now.

Now this raises me two questions:
- Should we update here the value of the replacement? (IMO, yes, on this PR)
- How safe are we that we did not introduce a wrong token somewhere else because of this blind trust in the replacement token value? (I did a quick search on the codebase for Orbit v8.0.0 and it seems the token was not used anywhere else, so…)

mvidalgarcia:
@DSil Thanks for the context. I updated the value of the replacement. Let me know if this is what you had in mind.

> How safe are we that we did not introduce a wrong token somewhere else because of this blind trust in the replacement token value?

I guess we would need to rely on visual tests more. This case would have been easily detected if we had a proper setup.

PR #4212, opened 2024-01-05
fix(ButtonPrimitive): add missing flex-auto

mainframev:
Reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1704445127541719)


packages/orbit-components/src/Button/Button.stories.tsx
```diff
@@ -51,6 +51,23 @@ BasicButtons.story = {
   },
 };
 
+export const MultipleFullWidthButtonsInContainer = () => {
```
DSil, line 54: The story doesn't seem to add much documentation value to the component. We should have a visual test for this.

DSil (approved):

packages/orbit-components/src/Button/Button.stories.tsx
DSil: > especially that they are in Chromatic and have to rewritten to Playwright. So maybe keep that story until we will rewrite the visual tests.

Alright, agreed.

> Tbh, I do not think we should care much about how it looks, if it's ugly or not, it does what it should.

Hmm a story has a purpose of documentation (especially now that we are not relying on stories for visual tests). This story tells nothing to the user. It doesn't showcase any prop or anything in particular… 

So, for now, I'm ok with keeping it as is 👍

PR #4218, opened 2024-01-10
Playwright: add maxDiffPixels to playwright and --ignore-scripts flag

DSil:
Flag added to avoid warnings on install.

`maxDiffPixels` added to avoid problems with icons.
 Storybook: https://orbit-mainframev-playwright-addMaxDiffPixels.surge.sh


mainframev (approved):
👍🏻

PR #4220, opened 2024-01-11
fix(Modal): check if e.target.className exists

mainframev:
Reported [here](https://skypicker.slack.com/archives/C7B5YNYHW/p1704968537274009). Not sure tbh, if this is happening somewhere not in Cypress, but added check for safety. 
 Storybook: https://orbit-mainframev-fix-modal.surge.sh

### DIFF
--- a/packages/orbit-components/src/Modal/index.tsx
+++ b/packages/orbit-components/src/Modal/index.tsx
@@ -204,6 +204,7 @@ const Modal = React.forwardRef<Instance, Props>(
         modalContent.current &&
         event.target instanceof Element &&
         !modalContent.current.contains(event.target) &&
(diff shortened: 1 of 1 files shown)


DSil (approved):

DSil:
Not sure about the commit message being relevant for a changelog if this was only breaking on cypress. And the message itself doesn't say much out of context. I'd just chore it. But not blocking because of that.

mainframev:
> Not sure about the commit message being relevant for a changelog if this was only breaking on cypress. And the message itself doesn't say much out of context. I'd just chore it. But not blocking because of that.

I do not mind it to be a chore, although it might be just not yet spotted on FE

PR #4221, opened 2024-01-11
fix(Loading): remove redundant padding and get back icon 20px size

mainframev:
Reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1704981489011759) 

also noticed, that the button loader is way too big compared to the [previous](https://orbit-mainframev-chore-nx.surge.sh/?path=/story/loading--button-loading) one 

 Storybook: https://orbit-mainframev-fix-loading-custom-size.surge.sh

### DIFF
--- a/packages/orbit-components/src/Loading/index.tsx
+++ b/packages/orbit-components/src/Loading/index.tsx
@@ -56,7 +56,11 @@ const Loader = ({ type, customSize }) => {
   }
 
   return (
-    <svg viewBox="0 0 40 40" className="orbit-loading-spinner animate-spinner h-xxl w-xxl">
+    <svg
+      viewBox="0 0 40 40"
+      className="orbit-loading-spinner animate-spinner h-xxl w-xxl"
+      stroke={type === TYPE_OPTIONS.BUTTON_LOADER ? "currentColor" : theme.orbit.paletteCloudDark}
+    >
       <circle
         cx="50%"
         cy="50%"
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Loading/index.tsx
```diff
@@ -91,11 +91,10 @@ const Loading = ({
...
             "box-border",
             type === TYPE_OPTIONS.BUTTON_LOADER &&
-              "absolute start-0 top-0 h-full w-full justify-center",
+              "[&_.orbit-loading-spinner]:h-icon-medium [&_.orbit-loading-spinner]:w-icon-medium absolute start-0 top-0 h-full w-full justify-center",
             type === TYPE_OPTIONS.SEARCH_LOADER && "h-[40px] justify-start",
-            type === TYPE_OPTIONS.INLINE_LOADER
-              ? "inline-flex min-h-[19px] justify-center"
-              : "p-sm flex",
+            type === TYPE_OPTIONS.INLINE_LOADER && "inline-flex min-h-[19px] justify-center",
+            type !== TYPE_OPTIONS.INLINE_LOADER && "flex",
```
DSil, line 97:
```suggestion
            type === TYPE_OPTIONS.INLINE_LOADER
              ? "inline-flex min-h-[19px] justify-center",
              : "flex",
```

Why can't it just be?

DSil (changes requested):
I have noticed two problems while looking at this:

On the playground story for Loading component, if one selects the type `buttonLoader` nothing is visible. It is just behind the code snippet. This is an old bug, it's happening for a long time, even with StyledComponents. **Can the story be adjusted so it is visible?**

Now, the second one comes since Tailwind: the svg is missing `stroke: currentColor` style, that it used to have, so nothing is visible even after the step above is done… So **we should add the class `stroke-current` to the svg.**

Finally, you mention that the icon is way too big compared to the previous one, but the previous one was the same size (40px). Which story were you looking at to notice such difference?

packages/orbit-components/src/Loading/index.tsx
mainframev: It can be done either this or that, I think in this case it's more readable as when it's separated, without ternary :)

mainframev:
> Finally, you mention that the icon is way too big compared to the previous one, but the previous one was the same size (40px). Which story were you looking at to notice such difference? 

@DSil link to the previous is in the description :)

PR #4224, opened 2024-01-12
docs: fix Popover example with missing OrbitProvider

DSil:
Reported [here](https://skypicker.slack.com/archives/C7T7QG7M5/p1705071967841679). The OrbitProvider was missing,
 Storybook: https://orbit-mainframev-docs-fix-popover-example-provider.surge.sh


mvidalgarcia (approved):
Working on the review app.

PR #4236, opened 2024-01-23
feat(FeatureIcon): migrate to Tailwind

mvidalgarcia:
Closes [FEPLT-1721](https://kiwicom.atlassian.net/browse/FEPLT-1721) 

### For tailwind migrations:

- [ ] Visual regression test suite has been added and run before the migration.

### For new components:

 Storybook: https://orbit-mainframev-tw-feature-icon.surge.sh

[FEPLT-1721]: https://kiwicom.atlassian.net/browse/FEPLT-1721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-components/src/FeatureIcon/index.tsx
+++ b/packages/orbit-components/src/FeatureIcon/index.tsx
@@ -1,13 +1,12 @@
 "use client";
 
 import * as React from "react";
-import styled from "styled-components";
 
-import defaultTheme from "../defaultTheme";
 import { baseURL } from "./consts";
 import type { Props } from "./types";
 
 const baseSize = 52;
+const heightClass = `h-[${baseSize}px]`;
 
 const getURL =
   (name: string) =>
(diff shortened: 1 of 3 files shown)


packages/orbit-components/src/FeatureIcon/index.tsx
```diff
@@ -24,20 +22,14 @@ const generateURL = (name: string) => {
...
-`;
-
-StyledFeatureIcon.defaultProps = {
-  theme: defaultTheme,
-};
-
 const FeatureIcon = ({ alt = "", name, dataTest, id }: Props) => (
-  <StyledFeatureIcon alt={alt} name={name} dataTest={dataTest} id={id} />
+  <img
+    className="h-[52px] w-auto bg-transparent"
```
DSil, line 27: I don't know if this height class is necessary 🤔 but if anything, we can also probably use the constant directly. Defining a variable above, something like `const heightClass = `h-[${baseSize}px]`;` and then use it here with template string.
I am not sure if it is the best practice in Tailwind, but it uses a constant that won't change in runtime, so it can be built in build time and we can rely on it. We're also safe for possible changes on the `baseSize` constant value.

But, again, I am not sure if it's a good practice and I won't enforce it if you're against it, so it's up to you.

DSil (commented):
Please add visual regression tests for this 🙏

DSil (approved):

packages/orbit-components/src/FeatureIcon/index.tsx
mvidalgarcia: Even if I don't like so much interpolating variables into Tailwind classes, I believe it's safer to reuse `baseSize` to cover us from side-effects when changing this value. My bad that I didn't realize and hardcoded the value instead.
I'll follow your suggestion.

PR #4237, opened 2024-01-24
Update tailwind 3.14

DSil:
Updated Tailwind version to 3.14.

This includes support to new properties like `size-` and the `*` selector. The linter plugins were also updated and actually helped to perform the migration.

⚠️ Some visual tests required update because of the way text is rendered when it doesn't have any font-family style

However, I do believe some more optimizations can still be made, through the objects that we use for classes. But that is not problematic for now.
 Storybook: https://orbit-mainframev-update-tailwind-3-14.surge.sh


mainframev:
@DSil LGTM, great changes :)

mainframev (approved):

PR #4239, opened 2024-01-25
Tw/migrate tag

DSil:
This is built on top of #4237 because it uses the latest tailwind version so please review the other one first (should be quick, and should be merged before this one).

Visual tests were created before the migration and executed (with no diffs) after the migration.

[FEPLT-1709](https://kiwicom.atlassian.net/browse/FEPLT-1709)

### For Tailwind migrations:

- [x] Visual regression test suite has been added if does not exist.

[FEPLT-1709]: https://kiwicom.atlassian.net/browse/FEPLT-1709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 Storybook: https://orbit-mainframev-tw-migrate-tag.surge.sh


packages/orbit-components/src/Tag/Tag.ct-story.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export function TestTag(props: Props) {
+  return (
+    <div className="p-xxs inline-block">
+      <Tag {...props} />
+    </div>
+  );
+}
+
+export function DefaultTest() {
```
mainframev, line 16:
```suggestion
export function DefaultTestStory() {
```

packages/orbit-components/src/Tag/index.tsx
```diff
@@ -233,48 +37,76 @@ const Tag = React.forwardRef<HTMLDivElement, Props>(
...
+        className={cx(
+          "orbit-tag",
+          "font-base rounded-large p-xs box-border inline-flex items-center justify-center font-medium",
+          "duration-fast transition-[color,_background-color,_box-shadow] ease-in-out",
+          "tb:rounded-normal",
+          size === SIZES.SMALL && "text-small leading-small",
+          size === SIZES.NORMAL && "text-normal leading-normal",
+          !!(onClick || onRemove) &&
+            "cursor-pointer [&_.orbit-tag-close-container]:active:opacity-100",
+          selected && [
```
mainframev, line 50: found a visual inconsistency, tbh, not sure right now what is more correct, it's not like in the current master, but Figma missing the examples selected, probably would need to ask the designers 

Screenshot 2024-01-25 at 11 20 07
Screenshot 2024-01-25 at 11 19 59

packages/orbit-components/src/Tag/index.tsx
```diff
@@ -233,48 +37,76 @@ const Tag = React.forwardRef<HTMLDivElement, Props>(
...
+              "focus:opacity-100",
+              !selected &&
+                (type === TYPES.NEUTRAL
+                  ? "text-ink-normal active:text-ink-dark"
+                  : "text-blue-darker"),
+              selected ? "text-white-normal opacity-100" : "opacity-50",
+            )}
+            tabIndex={0}
+            aria-label="close"
+            role="button"
```
mainframev, line 96: cursor pointer is present even when onClick is not applied, I believe this happens because of `role="Button"`, can we not have a role="button" when onClick is not applied? just a regular div, without tabIndex and role

mainframev:
Visual tests are good, but it seems that selected missing a state when the circle is also focused, that have additional ring on focus, that need to be tested also 🙏🏻

mainframev (changes requested):

packages/orbit-components/src/Tag/index.tsx
DSil: Yes, I noticed, but it's a bug we had before. Looking at the code, the new one is the expected. We were targeting the wrapper to apply the style but it was not applying

mainframev (approved):

PR #4246, opened 2024-01-31
chore: change storybook staging URL

DSil:
The staging URL now changes


packages/orbit-components/config/deploymentUtils.mts
```diff
@@ -1,29 +0,0 @@
-import { Octokit } from "@octokit/rest";
```
RobinCsl, line 1: Is this library still used somewhere else?

If not, let's please clean that up, as well as possible env variables related to it.

DSil: It is, I checked it. It is used to fetch the contributors to display on the documentation.

RobinCsl: Alright, thanks for checking! :)

RobinCsl (approved):

PR #4257, opened 2024-02-06
chore: format generated tokens file

DSil:
Our tokens cronjob [was failing](https://github.com/kiwicom/orbit/actions/runs/7801192554/job/21275717181) because it was always trying to commit files that, in the end, were only modified by removing the empty line at EOF.

This empty line was added because the `fetch:colors` script generates the palette files every time, without formatting it with prettier rules, hence the absence of an empty line at EOF.

Now we use prettier to format the document, avoiding false changes

### DIFF
--- a/packages/orbit-design-tokens/scripts/fetchBaseTokens.mjs
+++ b/packages/orbit-design-tokens/scripts/fetchBaseTokens.mjs
@@ -1,4 +1,5 @@
 import { $, fetch, fs, path } from "zx";
+import prettier from "prettier";
 import dotenv from "dotenv-safe";
 import dedent from "dedent";
 import { solidPaintToWebRgb, toHex } from "figx";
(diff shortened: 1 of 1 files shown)


packages/orbit-design-tokens/scripts/fetchBaseTokens.mjs
```diff
@@ -150,7 +151,10 @@ async function getColorTokensFromFigma() {
           } else {
             await fs.writeFile(
               path.join(OUTPUT_PATH, `${palette}.json`),
-              JSON.stringify(saveTokens(palette, tokens), null, 2),
+              prettier.format(JSON.stringify(saveTokens(palette, tokens), null, 2), {
+                ...prettier.resolveConfig.sync(""),
```
mvidalgarcia, line 155: Would it make sense to use [this function](https://github.com/kiwicom/orbit/blob/189f7c1cf6260333a663e329d7348f77d9502aef/packages/orbit-design-tokens/src/dictionary/utils/format.ts#L9)?

DSil: That would require the function to be exported and imported here. But importing something from a `.ts` file on an `.mjs` is not straightforward, as the typescript needs to be converted to js before… So I don't think it's worth the effort. WDYT?

mvidalgarcia: Ok, I get it.

This `.ts` vs `.mjs` vs `.mts` is a bit messy. I struggle to understand it and I faced some problems with `consts` files already, forcing me to duplicate them.

mvidalgarcia (approved):

PR #4258, opened 2024-02-06
fix(Modal): fix border radius on isMobileFullPage

mvidalgarcia:
https://skypicker.slack.com/archives/C7T7QG7M5/p1707228421814789

Closes FEPLT-1947


packages/orbit-components/src/Modal/ModalFooter/index.tsx
```diff
@@ -74,7 +74,7 @@ const ModalFooter = ({ dataTest, children, flex = "0 1 auto" }: Props) => {
         "duration-fast transition-shadow ease-in-out",
         "sm:max-lm:[&_.orbit-button-primitive]:h-form-box-normal sm:max-lm:[&_.orbit-button-primitive]:text-button-normal",
         childrenLength > 1 ? "lm:justify-between" : "lm:justify-end",
-        !isMobileFullPage && "lm:rounded-b-modal",
+        !isMobileFullPage ? "rounded-b-modal-mobile" : "lm:rounded-b-modal-mobile",
```
DSil, line 77: Why is it changing from `rounded-modal` (9px) to `rounded-modal-mobile` (12px)? **EDIT:** It seems to be the value in Figma anyway, so this is not a question anymore, just the below one.

Also, since this is the footer, why does the `isMobileFullPage` affect the border-radius of the bottom border? It is always straight corners anyway, no? 🤔 

image

packages/orbit-components/src/Modal/index.tsx
```diff
@@ -440,15 +440,13 @@ const Modal = React.forwardRef<Instance, Props>(
               "lm:relative lm:bottom-auto lm:pb-0 lm:overflow-visible",
               "lm:[&_.orbit-modal-section:last-of-type]:pb-xxl lm:[&_.orbit-modal-section:last-of-type:after]:content-none lm:[&_.orbit-modal-section:last-of-type]:mb-[var(--orbit-modal-footer-height,0px)]",
               "lm:[&_.orbit-modal-mobile-header]:w-[calc(var(--orbit-modal-width)-48px-theme(spacing.xxl))]",
+              "lm:rounded-modal",
```
DSil, line 443: I am now wondering why we have this and not `lm:rounded-modal-mobile` 🤔 Figma only seems to have one… maybe we should check with designers

packages/orbit-components/src/Modal/index.tsx
```diff
@@ -440,15 +440,13 @@ const Modal = React.forwardRef<Instance, Props>(
...
               !hasModalSection &&
                 "[&_.orbit-modal-header-container]:mb-xl lm:[&_.orbit-modal-header-container]:mb-[var(--orbit-modal-footer-height,0px)]",
+              !isMobileFullPage && "rounded-t-modal-mobile",
```
DSil, line 446: This can be on the second part of the ternary below, no?

DSil:
Maybe we should have visual tests that assert this 😶‍🌫️

packages/orbit-components/src/Modal/ModalFooter/index.tsx
mvidalgarcia: Yeah, I got confused.

I guess that just `lm:rounded-b-modal` would be enough? Regardless of `isMobileFullPage`?

packages/orbit-components/src/Modal/index.tsx
mvidalgarcia: I could find this in Figma but I'm not sure I'm looking at the right place.

**Desktop library**

Corner radius : 9px

https://www.figma.com/file/nkZ37lByVmAGOqim5PhbZT/%5BOrbit-Design-System%5D-Desktop-Web-Components-(Community)?type=design&node-id=2886-5088&mode=design&t=yYpucR259fCAlBqO-0

**Responsive library**

Corner radius (only top): 12px

https://www.figma.com/file/eDQ3YAPCK9K6bpGxFsu6J8/%5BOrbit%5D-Responsive-Web-Components-(Community)?type=design&node-id=3729-57&mode=design&t=hU6SKQzdiNdS4T3i-0

DSil: To increase the entropy, I saw this as the desktop library:
https://www.figma.com/file/Y79I18H9w8HJZVxPdAkAHb/%5BOrbit%5D-Desktop-Components?type=design&node-id=7757-12757&mode=design&t=jTpTgtdVycsPmmeR-0

packages/orbit-components/src/Modal/ModalFooter/index.tsx
DSil: Seem so. But is it `lm:rounded-b-modal` or `lm:rounded-b-modal-mobile`?

mvidalgarcia: Let's wait for designers' answer. [This](https://github.com/kiwicom/orbit/pull/4258#discussion_r1487960071) was my assumption.

mvidalgarcia: After consulting with the Orbit designers, we confirmed that all the rounded corners must be 12px.

packages/orbit-components/src/Modal/index.tsx
mvidalgarcia: After consulting with the Orbit designers, we confirmed that all the rounded corners must be 12px.

packages/orbit-design-tokens/output/docs-tokens.json
```diff
@@ -14164,28 +14164,6 @@
       "value": "rgb(186, 199, 213)"
     }
   },
-  "modalBorderRadiusMobile": {
```
DSil, line 14167: Unlike the tailwind package, a token deletion **is a breaking change**, even if it is just an internal token. I would just mark this as deprecated instead of removing it. We can (and should) do a cleanup of all the tokens soon™.

The tailwind class removal can be kept, as it is undocumented 👍

mvidalgarcia: Agree, updated.

DSil (approved):
LGTM. Let's just hold the merge for after the demo 👍

PR #4261, opened 2024-02-07
Deploy playwright reports

DSil:
Playwright reports are deployed with surge.
On PR close, they are deleted.

[FEPLT-1942](https://kiwicom.atlassian.net/browse/FEPLT-1942)


.github/workflows/test.yml
```diff
@@ -144,14 +145,53 @@ jobs:
...
-      - name: Upload Results
-        uses: actions/upload-artifact@v4
+        run: yarn components test-ct --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }} --reporter=blob
+      - name: Upload blob report artifacts
         if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: blob-report-${{ matrix.shardIndex }}
+          path: packages/orbit-components/blob-report
+          retention-days: 2
```
mvidalgarcia, line 155: If the PR lives more than 2 days, the report will be gone?

package.json
```diff
@@ -30,7 +30,8 @@
     "tokens": "yarn workspace @kiwicom/orbit-design-tokens",
     "tracking": "yarn workspace @kiwicom/orbit-tracking",
     "update-supported-browsers": "markdown --path .github/contribution/testing-conventions.md && git add .github/contribution/testing-conventions.md",
-    "docker:reset": "rm -rf packages/orbit-components/playwright/.cache && yarn install --frozen-lockfile --ignore-scripts"
+    "docker:reset": "rm -rf packages/orbit-components/playwright/.cache && yarn install --frozen-lockfile --ignore-scripts",
+    "deploy:report": "yarn surge playwright-report/"
```
mvidalgarcia, line 34: Couple of questions to try to understand better the workflow:
  1) Is the `playwright-report/` folder generated automatically by `playwright merge-reports`?
  2) Is the `all-blob-reports/` inside this folder?

DSil: 1. [Yes](https://playwright.dev/docs/test-sharding#:~:text=This%20will%20produce%20a%20standard%20HTML%20report%20into%20playwright%2Dreport%20directory.)
2. No. Inside this folder there will be an `index.html` file and a `data/` folder that are ready to be published on the web. The `all-blob-reports/` folder is generated by the download-artifacts action that downloads the 4 artifacts and merges them all on that single folder. The `playwright merge-reports` expects that folders with multiple blobs inside to merge them all in one.

.github/workflows/test.yml
mvidalgarcia: Then we don't even need 2 days, right? What's the criterion for selecting this number?

.github/workflows/test.yml
```diff
@@ -144,14 +145,53 @@ jobs:
...
+          merge-multiple: true
+      - name: Merge into HTML Report
+        env:
+          ARTIFACTS_PATH: ${{ steps.download.outputs.download-path }}
+        run: |
+          if [ -z "$(ls -A ${{ env.ARTIFACTS_PATH }})" ]; then
+            echo "No reports found. Exiting."
+            exit 0
+          else
+            npx playwright merge-reports --reporter html ./all-blob-reports
```
mvidalgarcia, line 184: I know it's documented but it seems a bit obscure that the output folder for this command is `playwright-report/`. Would it make sense to add a comment?

DSil: I had an echo saying that. But then I noticed that the command itself already echoes it too, so I deleted it.
I will add a comment so it is clear for one that reads the code.

.github/workflows/test.yml
DSil: They actually document for 1 day. I went for 2 days because as I was experimenting thins on this PR I might've wanted to go to old artifacts and recover old reports. But yeah, I'm ok with reducing it to 1 day

.github/workflows/test.yml
```diff
@@ -144,14 +145,54 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: blob-report-${{ matrix.shardIndex }}
+          path: packages/orbit-components/blob-report
+          retention-days: 1
+
+  test-ct-report:
+    name: Visual Regression Tests Report
+    runs-on: ubuntu-latest
+    if: always()
```
mvidalgarcia, line 160: I wonder if there's a _cleaner_ why to run this job **only** when `test-ct` fails instead of always and do the checks inside.

Have you considered something like this? (untested)

```yml
if: ${{ needs.test-ct.result == 'failure' }}
```

Disregard this comment if you find it so cumbersome, we can tackle this in future iterations.

.github/workflows/test.yml
mvidalgarcia: No strong opinion, just trying to understand the logic. I don't think it will hurt to keep 2 days, as you like.

.github/workflows/delete-test-reports.yml
```diff
@@ -0,0 +1,26 @@
+name: Delete Test Report
```
mvidalgarcia, line 1: nitpick: either use Test**s** Report as we have in the creation job or keep it as is and adapt the other one.

.github/workflows/delete-test-reports.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  delete-report:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Node
+        uses: ./.github/actions/node
```
mvidalgarcia, line 18: This seems a bit overkill but I guess there's no other way to get the proper surge version from our package.json 😬

.github/workflows/delete-test-reports.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Delete Report
+        run: |
+          echo "Deleting report for branch $BRANCH_NAME"
+          BRANCH_URL=$(echo ${BRANCH_NAME} | sed -e 's/[^a-zA-Z0-9]/-/g')
+          DOMAIN=https://kiwicom-orbit-test-report-${BRANCH_URL}.surge.sh
+          yarn surge teardown $DOMAIN --token ${{ secrets.SURGE_TOKEN }}
+        continue-on-error: true
```
mvidalgarcia, line 26: Why do we need this?

.github/workflows/test.yml
DSil: On the official docs, they recommend using `always()`. I have tried several other approaches and they all failed in some case. I did not try that one. I can give it a quick try

.github/workflows/delete-test-reports.yml
DSil: Going for `Delete Tests Report`

.github/workflows/delete-test-reports.yml
DSil: I didn't want the pipeline to be red if it fails on deleting (because it doesn't exist, for instance).
Imagine a PR that never fails the tests. There won't be reports. But this job will still execute and try to delete… nothing. The command above will fail, but the pipeline won't.

I found it easier to do than checking if the report exists before deleting it.

.github/workflows/delete-test-reports.yml
DSil: Yes 😕

.github/workflows/delete-test-reports.yml
mvidalgarcia: Oh I see, I hadn't considered that case.

mvidalgarcia (approved):
I think it's more than enough for a first approach, approving 👍 

Feel free to postpone other improvements for the future.

.github/workflows/test.yml
DSil: So it works for when all tests pass. Will now test with failing tests

mvidalgarcia: This is cool, let's see 🤞

DSil: As discussed IRL, `if: ${{ always() && needs.test-ct.result == 'failure' }}` does the trick. So we will go with that and the simplification of the `run` part, removing the verification for the artifacts existance

PR #4263, opened 2024-02-08
ci: do not print dry-run changelog from script

mvidalgarcia:
The output is misleading when we execute this job in dry-run mode
as it fetches the diff content of the latest tag. Since on dry-run
we don't create any tags on lerna calls, we were fetching the diff
of older tags, thus the results are not what we expect.

It is more sensible to rely on the git diff output which contains
the changes in the changelog and package.json files.


DSil (approved):
🧹

PR #4277, opened 2024-02-19
feat(Accordion): migrate to tailwind

vbulant:
Closes [FEPLT-1699](https://kiwicom.atlassian.net/browse/FEPLT-1699)

### For Tailwind migrations:

~~- [ ] Visual regression test suite has been added if does not exist.~~ (already exists)

### For new components:

[FEPLT-1699]: https://kiwicom.atlassian.net/browse/FEPLT-1699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
```diff
@@ -1,48 +1,15 @@
...
-  margin: ${({ theme }) => theme.orbit.spaceXSmall} 0;
-`;
-
-StyledAccordionWrapper.defaultProps = {
-  theme: defaultTheme,
-};
-
-const AccordionWrapper = ({ dataTest, initialExpanded, ...props }: Props) => (
-  <StyledAccordionWrapper
+const AccordionWrapper = ({ dataTest, ...props }: Props) => (
```
DSil, line 6: While reviewing this, I noticed that:
- This component is only used internally, in `AccordionSection`. It is not exported.
- No props are _ever_ passed to this component, except for `dataTest`.

With this in mind, I believe we can simplify this component. The `Props` can be simplified to just have `children` and `dataTest`. And the `expanded` condition is no longer needed, as I couldn't reproduce the addition of the said box-shadow.

Do you know any use case that contradicts what I said above?

packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
```diff
@@ -1,48 +1,15 @@
...
+  <div
+    className={cx(
+      "border-card rounded-normal duration-fast my-xs bg-white-normal relative w-full border border-solid transition-[margin] ease-in-out",
```
DSil, line 9: The `border-card` class comes from a token that is deprecated. The `border-elevation-flat-border-color` class should be used instead.

DSil, line 9: The background class should be `bg-elevation-flat`, as it is the referenced replacement for the deprecated `backgroundCard` token.

The same could be said about the border width class. The replacement token is `elevationFlatBorderSize`. However, I also prefer to keep the native `border` class. The above one makes sense because we are talking about colors, that can be mapped in a different theme. But, in this case, I don't think border widths should be configurable per theme, so I'd leave it as is.

DSil, line 9: I also don't see the margins changing 🤔 (are they?). If they don't change, I'd say we don't need those transitions, do we?

DSil (commented):
Overall great, just some questions

packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
vbulant: Great catch, thanks!

I’ve checked the history and this part of the component has been like this since its inception so safe to simplify 👍

packages/orbit-components/src/Accordion/components/AccordionWrapper/index.tsx
vbulant: 🤦‍♂️ I’ve just blindly converted it to tw. They indeed don’t change, removed 👍

packages/orbit-components/src/Accordion/AccordionSection/index.tsx
```diff
@@ -29,7 +28,10 @@ const AccordionSection = ({
   const [{ height }, ref] = useBoundingRect<HTMLDivElement>({ height: isExpanded ? null : 0 });
 
   return (
-    <AccordionWrapper dataTest={dataTest}>
+    <div
+      className="border-elevation-flat-border-color rounded-normal duration-fast my-xs bg-elevation-flat relative w-full border border-solid ease-in-out"
```
DSil, line 32: The `duration-fast` and `ease-in-out` were here because of the transition. Since the transition was removed, they can also be removed

vbulant: Sorry for the sloppiness here and thank you for being thorough. Removed 👍

packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.tsx
```diff
@@ -1,42 +1,8 @@
...
-  expanded: boolean;
-  "data-test"?: string | boolean;
-}>`
-  ${({ theme, expanded }) => css`
-    display: flex;
-    padding: ${theme.orbit.spaceLarge};
-    background-color: ${theme.orbit.paletteWhite};
-    align-items: center;
-    min-height: ${expanded ? "19px" : "44px"};
-    transition: max-height ${theme.orbit.durationFast} ease-in-out;
```
vbulant, line 18: Turned out this transition was also useless. In the end all the transitions are handled only by the `<Slide` inside the `AccordionSection`.

DSil (approved):

PR #4279, opened 2024-02-20
fix(Modal): fix bottom border radius when isMobileFullPage

mvidalgarcia:
The issue occurred because the bottom corners of a footerless modal
didn't round off when the isMobileFullPage props was enabled.

Slack: https://go.kiwi.com/8otDOYbt

Closes FEPLT-1952


packages/orbit-components/src/Modal/ModalSection/index.tsx
```diff
@@ -43,7 +43,7 @@ export const ModalSectionWrapper = ({
             ],
         !isMobileFullPage && [
           "first-of-type:rounded-t-modal lm:first-of-type:rounded-t-modal",
-          "last-of-type:rounded-b-modal lm:last-of-type:rounded-t-modal lm:[&_~_.orbit-modal-footer]:last-of-type:mt-0",
```
RobinCsl, line 46: Just one question: we removed `rounded-t-modal` and replaced it with `rounded-b-modal`; was the issue all along that we just didn't have the correct CSS class?

Looking at the previous state, it doesn't seem to make sense and was probably a copy-paste-forgot to update the class kind of situation.

packages/orbit-components/src/Modal/ModalSection/index.tsx
```diff
@@ -29,7 +29,7 @@ export const ModalSectionWrapper = ({
         "orbit-modal-section",
         "py-lg px-md lm:p-xl box-border w-full",
         "border-b-elevation-flat-border-color border-b border-solid",
-        "last-of-type:[&:not(:last-child)]:rounded-b-none",
+        "lm:last-of-type:rounded-b-modal last-of-type:[&:not(:last-child)]:rounded-b-none",
```
RobinCsl, line 32: `lm:last-of-type:rounded-b-modal last-of-type:[&:not(:last-child)]:rounded-b-none`

So that means that:
- on large mobile and last of type, we get rounded-b-modal
- otherwise, last of type and not last child, we get rounded-b-none

because of the order of the classes, that means there would still be situations when the corners wouldn't be rounded on desktop, right?

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: Actually, the top border-radius is also needed when there's no header. I'll adjust it and add visual tests.

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: The order of classes is changed automatically in Tailwind (eslint? prettier?) so I'm not sure how they are applied.

I guess specificity plays a role in these classes, and the last-child rule takes precedence, but I'm not fully sure.

packages/orbit-components/src/Modal/ModalSection/index.tsx
```diff
@@ -29,7 +29,7 @@ export const ModalSectionWrapper = ({
         "orbit-modal-section",
         "py-lg px-md lm:p-xl box-border w-full",
         "border-b-elevation-flat-border-color border-b border-solid",
-        "last-of-type:[&:not(:last-child)]:rounded-b-none",
+        "lm:last-of-type:rounded-modal last-of-type:[&:not(:last-child)]:rounded-b-none",
```
RobinCsl, line 32: So on large mobile, we have all the corners rounded, regardless of whether it's first- or last-of type like previously.

And then, if `!isMobileFullPage`, we ensure that the first-of-type on large mobile has rounded corners at the top, the last-of-type has rounded corners at the bottom.

I'm a bit confused 😅 there's some overlap between them, right?

RobinCsl: Thanks for the visual tests and screenshots, at least it looks like the behaviour should be correct now, I just don't have a good intuition for it when looking at the code.

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: This has been done.

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: I can confirm that the `:not(:last-child)` has precedence over only `last-of-type`.

As an example, this is the last section in a modal with a footer:

image

As one can see, the class with `:not(:last-child)` overrides the other one, generating the desired result: removing the bottom borders when the last section is not the last child of the block.

RobinCsl: I wonder if it's just because of how Tailwind writes the CSS file 🤔 

For example, the class `lm:last-of-type:rounded-b-modal` is contained in a media query, so maybe Tailwind writes media query-related classes later in the CSS file compared to "normal" classes, and that's why we're lucky in this instance.

I wonder how "class-specificity" works here 😅 🙈 🤔

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: I had to revert this. The modal sections need to be handled with siblings selector because there might be an overlay div sibling for the close icon.

```
"[&_~_.orbit-modal-section:first-of-type]:rounded-t-none"
```

packages/orbit-components/src/Modal/ModalSection/index.tsx
mvidalgarcia: I think this proves my theory, I don't know what else I can try 😅 

[Playground](https://codi.link/PHNlY3Rpb24+CiAgPHNwYW4+c3Bhbjwvc3Bhbj4KICA8ZGl2Pm9uZTwvZGl2PgogIDxkaXY+dHdvPC9kaXY+CiAgPHNwYW4+c3Bhbjwvc3Bhbj4KICA8ZGl2PnRocmVlPC9kaXY+CiAgPGRpdj5mb3VyPC9kaXY+Cjwvc2VjdGlvbj4=%7CZGl2Om5vdCg6Zmlyc3QtY2hpbGQpOmZpcnN0LW9mLXR5cGUgewogIGNvbG9yOiB0b21hdG8KfQoKZGl2OmZpcnN0LW9mLXR5cGUgewogIGNvbG9yOiBwdXJwbGUKfQoK%7C)

mvidalgarcia:
As discussed in Slack, now all the rounded corners are handled from the wrapper content element.

RobinCsl (approved):

PR #4293, opened 2024-03-01
feat(CallOutBanner): migrate to tailwind

vbulant:
Closes [FEPLT-1762](https://kiwicom.atlassian.net/browse/FEPLT-1762)

### For Tailwind migrations:

~~- [ ] Visual regression test suite has been added if does not exist.~~

### For new components:

[FEPLT-1762]: https://kiwicom.atlassian.net/browse/FEPLT-1762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/orbit-components/src/CallOutBanner/index.tsx
```diff
@@ -67,25 +18,36 @@ const CallOutBanner = ({
...
-    id={id}
-  >
-    {illustration && <StyledIllustration>{illustration}</StyledIllustration>}
-    <Stack spacing="medium">
-      <Stack spacing="XSmall">
-        <Stack spacing="XXSmall">
-          {title && <Heading type="title3">{title}</Heading>}
-          {description && <Text>{description}</Text>}
+}: Props) => {
+  const Element = onClick ? "button" : "div";
```
DSil, line 22: This is a good accessibility thing and the actual correct thing to do, but the button does not grow like a div. Do you think adding `w-full` is harmless here? It stretched the element to the full width but we did not have this property before, so I don't really know if it might break existing layouts 🤔 

**This branch:**
image

**Master:**
image

vbulant: Good catch! I’m not 100% sure it won’t break any existing layout. I think it will be better to get back to `div` for now to prevent any breaking changes. I’ll update the code.

vbulant: Updated 👍

DSil (approved):

PR #4300, opened 2024-03-12
chore: add missing tailwind migrated components

mvidalgarcia:
- chore: add missing tailwind migrated components

- feat: deprecate toggleUp and toggleDown utilities
These utilities are not being used by anyone.

image

- chore: ignore snapshots folders in commitlint

### DIFF
--- a/packages/orbit-components/src/utils/toggleDown/index.ts
+++ b/packages/orbit-components/src/utils/toggleDown/index.ts
@@ -1,5 +1,9 @@
 import { keyframes } from "styled-components";
 
+/**
+ * @deprecated This utility is deprecated and will be removed in future versions.
+ * Please use an alternative solution.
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/utils/toggleDown/index.ts
DSil, line 1: For context: it these utils seem to have been added for the Tile expandable but the animation is now achieved by CSS tailwind classes. Good cleanup 👍

DSil (approved):

packages/orbit-components/src/utils/toggleDown/index.ts
mvidalgarcia: Didn't know. I just ran global searches and saw no usage. Good point 👍

PR #4303, opened 2024-03-13
feat: delete BaggageStepper component

mvidalgarcia:
BREAKING CHANGE: remove BaggageStepper component

The BaggageStepper component has been removed as it was not being used.
Please use the Stepper component instead.

image


DSil (approved):
👋

PR #4306, opened 2024-03-14
Remove unnecessary GH_TOKEN from publish job

DSil:
The PAT token is enough to have the push rights.

The branch protection rules for the master branch were also adjusted. We had @RobinCsl and @mvidalgarcia as exceptions but that is not needed since they're already admins.

[FEPLT-1948](https://kiwicom.atlassian.net/browse/FEPLT-1948)

[FEPLT-1948]: https://kiwicom.atlassian.net/browse/FEPLT-1948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -58,7 +58,6 @@ jobs:
         if: ${{ github.event.inputs.dryrun == 'false' }}
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }} # must be of type Automation to create releases
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_CHANGELOG_WEBHOOK_URL: ${{ secrets.SLACK_CHANGELOG_WEBHOOK_URL }}
         run: |
           echo "scope=@kiwicom" > ~/.npmrc


.github/workflows/publish.yml
```diff
@@ -58,7 +58,6 @@ jobs:
         if: ${{ github.event.inputs.dryrun == 'false' }}
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }} # must be of type Automation to create releases
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
RobinCsl, line 61: So `secrets.PAT` is used for the Checkout step in this file, but for the Publish step, there won't be any GitHub token set.

Is that because we have the `permissions.contents = write` at the beginning? In that case, is `secrets.PAT` really necessary?

RobinCsl: Also, `GH_TOKEN` is a Lerna specific env variable, IIRC. So if we don't have it defined anymore, it's not going to work. Right?

https://github.com/lerna/lerna/tree/main/libs/commands/version#--create-release-type

RobinCsl: > The PAT used for checkout persists through the whole job. It is the bot's PAT so it has the push rights needed on this step.

The CI job's PAT is actually available under `secrets.GITHUB_TOKEN`, no? I don't think it's available under `PAT`?

DSil: Yes, but the commit and push must be authored by the bot. They can't rely on the CI job's PAT. So we basically need the bot's PAT for the commit and push and the GITHUB_TOKEN (provided by the job itself) to be used by Lerna. That is my understanding

DSil:
As discussed IRL, nothing to do here. Closing.

PR #4312, opened 2024-03-19
Trigger error on custom class 

DSil:
Update versions for typescript-lint and es-lint deps and change from warning to error the no-custom-classname rule for tailwind.

The new rule is only applied on orbit-components, so it does not affect the docs package or any other package.

[FEPLT-1760](https://kiwicom.atlassian.net/browse/FEPLT-1760)

[FEPLT-1760]: https://kiwicom.atlassian.net/browse/FEPLT-1760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mvidalgarcia:
The second commit reads:

```
chore: change warn to err tailwindcss/no-custom-classname
```

However, I don't see any replacement of a warn rule, it's essentially adding a new rule.

DSil:
@mvidalgarcia On our root eslint config, we have it defined as a warn. On the orbit-components package, I override that definition to be an error (on that package only). This way we can keep the warning on the other packages but we have a failing pipeline on the orbit-components package. WDYT?

packages/orbit-components/.eslintrc.cjs
```diff
@@ -31,6 +31,7 @@ module.exports = {
       files: ["src/**/*.ts?(x)"],
       rules: {
         "@typescript-eslint/consistent-type-imports": "warn",
+        "tailwindcss/no-custom-classname": [2, { whitelist: ["(orbit\\-).*"] }],
```
mvidalgarcia, line 34:
```suggestion
        "tailwindcss/no-custom-classname": ["error", { whitelist: ["(orbit\\-).*"] }],
```

mvidalgarcia:
> @mvidalgarcia On our root eslint config, we have it defined as a warn. On the orbit-components package, I override that definition to be an error (on that package only). This way we can keep the warning on the other packages but we have a failing pipeline on the orbit-components package. WDYT?

Ok ok, I didn't know this. Perhaps we can add this information in the commit body.

mvidalgarcia (approved):
🏎️

PR #4313, opened 2024-03-21
Remove utils/media query

DSil:
Opened on top of #4285 — to be merged only after that.

Removes the `mediaQueries` utility function, to be used within styled components.

As it is still widely used in the docs package, the implementation was moved there temporarily.

[FEPLT-1922](https://kiwicom.atlassian.net/browse/FEPLT-1922)

[FEPLT-1922]: https://kiwicom.atlassian.net/browse/FEPLT-1922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/orbit-components/src/utils/mediaQuery/README.md
```diff
@@ -55,44 +12,50 @@ You can use the following media queries in your project:
...
+| :--------- | :------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| breakpoint | `keyof typeof TOKEN` | The name of the breakpoint.                                                                                                                                                       |
+| theme      | `Theme`              | The theme object.                                                                                                                                                                 |
+| pure       | `boolean`            | Optional. If `true`, the function will return the value as number. If false, it will return a string `"(min-width: ${VAL}px)"`, where `VAL` is the value of the breakpoint width. |
+
+To test your components with Enzyme, especially styles, you can also use the `getBreakpointWidth` function.
```
mvidalgarcia, line 61: Replace Enzyme with Jest.

DSil: Deleted, as testing is already mentioned above. This was a leftover from the previous docs. Thanks

docs/src/documentation/05-development/04-migration-guides/01-v15.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: v15
+description: How to migrate to Orbit 15.0.0
+redirect_from:
+  - /migration-guides/v15/
+---
+
+# Orbit Migration Guide v15
+
+This migration guide focuses on the process of migrating from Orbit v13 to v15.0, as some breaking changes were introduced.
```
mvidalgarcia, line 10:
```suggestion
This migration guide focuses on the process of migrating from Orbit v14 to v15.0, as some breaking changes were introduced.
```

docs/src/documentation/05-development/04-migration-guides/01-v15.mdx
```diff
@@ -0,0 +1,19 @@
...
+With this guide, we aim to walk through all the breaking changes and how they can be addressed, allowing the migration to be smoother and effortlessly.
+
+## Breaking changes
+
+### mediaQueries util function removed
+
+The `mediaQueries` function is no longer available. This was a styled-components util function that is now no longer expected to be used, as we move away from StyledComponents.
+If needed, the tokens and breakpoint values are still accessible via the `getBreakpointWidth` function and the `QUERIES` and `TOKEN` consts.
```
mvidalgarcia, line 18:
```suggestion
If needed, the tokens and breakpoint values are still accessible via the `getBreakpointWidth` function and the `QUERIES` and `TOKEN` constants.
```

packages/orbit-components/src/utils/mediaQuery/index.ts
```diff
@@ -1,36 +1,31 @@
...
   tablet: "widthBreakpointTablet",
   desktop: "widthBreakpointDesktop",
   largeDesktop: "widthBreakpointLargeDesktop",
-};
+} as const;
+
+export interface GetBreakpointWidth {
+  (name: keyof typeof TOKEN, theme: Theme): string;
+  (name: keyof typeof TOKEN, theme: Theme, pure: false): string;
+  (name: keyof typeof TOKEN, theme: Theme, pure: true): number;
```
RobinCsl, line 22: Nice!

packages/orbit-components/src/utils/mediaQuery/index.ts
```diff
@@ -1,36 +1,31 @@
...
+}
 
-export const getBreakpointWidth = (
-  name: keyof typeof TOKEN,
+export const getBreakpointWidth: GetBreakpointWidth = (
+  name: string,
```
RobinCsl, line 26: Why are we relaxing the requirement here?

packages/orbit-components/src/utils/mediaQuery/types.d.ts
```diff
@@ -1,24 +1,7 @@
-// Type definitions for @kiwicom/orbit-components
-// Project: http://github.com/kiwicom/orbit
-import type { Interpolation } from "styled-components";
-
-import type { TOKEN } from ".";
-
 export type Devices =
```
RobinCsl, line 1: Is this file even useful if we have the `mediaQuery/index.ts` above? 🤔

packages/babel-plugin-orbit-components/index.js
```diff
@@ -71,6 +70,9 @@ const pathOverwrites = {
   useRandomId: "hooks/useRandomId",
   useRandomIdSeed: { path: "hooks/useRandomId", named: true },
   useFocusTrap: "hooks/useFocusTrap",
+  getBreakpointWidth: "utils/mediaQuery",
```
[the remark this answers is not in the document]
DSil: No, it is now needed because now we are actually exporting it in our index file. In the past, only the default export was being exported on the index file.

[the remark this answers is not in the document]
RobinCsl: I see. But these adjustments to the `index` file are done in the next commit, right?

packages/orbit-components/src/utils/mediaQuery/types.d.ts
DSil: No. Moved the export of the type to the index

packages/orbit-components/src/utils/mediaQuery/index.ts
DSil: It is an interesting thing, actually. So, according to the interface defined above, the return type of this function is `string | number` but if I defined this prop to be `keyof typeof TOKEN` it will consider `theme.orbit[TOKEN[name]]` even though **it will** return a number, because `TOKEN[name]` **always** refers to a key that corresponds to a number in `theme.orbit`.

However, relaxing the requirement here does not seem to be problematic, as I tested passing a random string to this function on [this file](https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/hooks/useMediaQuery/index.ts#L54) and Typescript complained, saying that it was not one of the breakpoints. So I would say we are safe here… WDYT?

RobinCsl: Do you even need to type the arguments if you're typing the function already? 🤔

DSil: I guess 😞 that was my first thought was well!

Screenshot 2024-03-28 at 17 09 25

mvidalgarcia (approved):

packages/orbit-components/src/utils/mediaQuery/index.ts
RobinCsl: So this didn't work out in the end?

DSil: Thanks for raising. Solved in #4327

PR #4317, opened 2024-03-25
Remove Styled Components

DSil:
WIP


packages/orbit-components/src/common/consts.ts
```diff
@@ -15,3 +15,13 @@ export enum PLACEMENTS {
   LEFT_START = "left-start",
   LEFT_END = "left-end",
 }
+
+export enum SPACINGS_AFTER {
```
RobinCsl, line 19: There was no need to move the `tokens` from the old file because we already have them in Tailwind, right?

packages/orbit-components/src/Icon/IconList.tsx
```diff
@@ -1,62 +1,36 @@
...
       .filter(n => n !== "__namedExportsOrder")
       .map(icon => {
         const Icon = Icons[icon];
 
         const iconName = `${icon}`;
         return (
-          <Container key={icon}>
+          <div
+            className="gap-lg bg-white-normal mb-lg rounded-large border-cloud-normal px-lg flex w-full flex-row content-center items-center justify-start border border-solid"
+            style={{ minHeight: "80px" }}
```
RobinCsl, line 16: Why not `min-h-[80px]`?

packages/orbit-components/src/Icon/IconList.tsx
```diff
@@ -1,62 +1,36 @@
...
     {Object.keys(Icons)
       .filter(n => n !== "__namedExportsOrder")
       .map(icon => {
         const Icon = Icons[icon];
 
         const iconName = `${icon}`;
         return (
-          <Container key={icon}>
+          <div
+            className="gap-lg bg-white-normal mb-lg rounded-large border-cloud-normal px-lg flex w-full flex-row content-center items-center justify-start border border-solid"
```
RobinCsl, line 15: We don't have specific classes for `borderWidthCard` and `borderStyleCard` tokens, right?

Also, is the class order somehow automatic? It'd be nice to have all the border related things in the same area.

One idea: in CubeCSS methodology, they separate concerns with a `|` character (and of course, the `|` class does nothing)

packages/orbit-components/src/Icon/IconList.tsx
```diff
@@ -1,62 +1,36 @@
...
+            key={icon}
+          >
             <Icon size="large" color="primary" />
-            <IconImport>
+            <div
+              className="leading-small text-ink-dark bg-cloud-light border-cloud-normal py-xxs ps-xs border border-solid"
+              style={{
+                fontFamily:
+                  '"SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace',
+                fontSize: "12px",
```
RobinCsl, line 25: This could also be a class, though I agree that the font family is probably easier like that?

packages/orbit-components/src/utils/Grid/README.md
```diff
@@ -75,53 +73,13 @@ All of the object for these properties have the same properties and none is requ
...
-Also, if you want to support IE 10+, you have to use only the old specification, which has a lot of limitations and does not provide the flexibility of modern specification.
-
 The main focus of this component is to provide an API that solves the most common issues and gives you a quick how to start using CSS grids in your project without worries.
 
 By default, this Grid component handles:
 
 - Compiling the `repeat` function to the old specification
 - Auto-placement of its children
 
-## Known template limitations
```
RobinCsl, line 76: Will that allow us to do things more efficiently in the codebase? i.e. are there some further improvements we should keep track of in Jira?

packages/orbit-components/src/utils/rtl/README.md
```diff
@@ -12,133 +12,10 @@ import { OrbitProvider, defaultTheme } from "@kiwicom/orbit-components";
...
 
-It's also important to pass some `orbit` object, in our case `defaultTheme`.
-With this approach, all `orbit-components` will be ready to use with some right to left language.
+In addition, you also need to set the `dir` property on the `html` element to have value `rtl`.
 
-> In addition, you need to set up `direction` on `html` to have value `rtl`.
-
-## Utilities
-
-`styled-components` doesn't know automatically how to switch `margin`, `padding` and i.e. `left` properties to the other direction. So it's necessary to use some `utilities` inside your styles.
```
RobinCsl, line 22: How does Tailwind do it? Does it rely on logical properties?

docs/src/documentation/05-development/04-migration-guides/01-v15.mdx
```diff
@@ -10,10 +10,41 @@ redirect_from:
...
+
 ### mediaQueries util function removed
 
 The `mediaQueries` function is no longer available. This was a styled-components util function that is now no longer expected to be used, as we move away from StyledComponents.
 If needed, the tokens and breakpoint values are still accessible via the `getBreakpointWidth` function and the `QUERIES` and `TOKEN` constants.
 Check [the documentation](/development/utilities/media-queries/) for more information on how to use these.
+
+### rtl utils removed
+
+The `rtl` utility functions have been removed. They were used to handle RTL styles in the `styled-components` version of Orbit.
```
RobinCsl, line 48: Maybe add a comment as to how it works with Tailwind/why it's okay for this to be removed.

Also, if folks were relying on them, would it make sense to provide a code snippet replacement?

(same question for `mediaQueries` above actually?)

packages/orbit-components/src/common/consts.ts
DSil: Yes. On the previous file, they were used to retrieve the value of the token, so it could then be used to apply `margin-bottom`. With Tailwind, we don't need it because we just use an `mb-` class with the correct suffix that matches a valid spacing value (defined by tokens, on the `tailwind-preset-package`.

The `tailwind/spaceAfter.ts` file is responsible for matching these SPACINGS_AFTER consts  (used by the components that have `spaceAfter` prop) to the correct tailwind class — so no tokens are needed in this process.

packages/orbit-components/src/Icon/IconList.tsx
DSil: I wanted to avoid that custom class to be built in the final CSS file. Since this component is to be used only in Storybook, I prefer the style to be inline, so that class is not unnecessarily created by the Tailwind compiler

packages/orbit-components/src/Icon/IconList.tsx
DSil: Those tokens are deprecated, since they've proved to be useless. The width is just 1px (class `border`) and the style is just "solid" (class `border-solid`). So we don't need them.

We have Prettier for the order, so the only way to actually control the order is if we use the `cx` function and separate into multiple strings… But we haven't been paying much attention to that, to be honest

packages/orbit-components/src/Icon/IconList.tsx
DSil: Yes, for font-family it was just for convenience. And for font-size it's the same reason as above

packages/orbit-components/src/utils/Grid/README.md
DSil: I have removed these because they referred to IE limitations. I don't think they're a reality nowadays and we don't support IE officially, so I wouldn't say it makes sense to mention limitations of one component that are exclusive to IE

packages/orbit-components/src/utils/rtl/README.md
DSil: Yes. Tailwind has builtin classes that handle it, eg: `ms-` will apply `margin-inline-start`, that browsers translate to left margin in LTR and right margin in RTL.

One can also use the `rtl:` prefix in classes to apply styles that are only applicable in RTL

docs/src/documentation/05-development/04-migration-guides/01-v15.mdx
DSil: I did some research and did not find relevant usages for those, hence why I didn't put that much effort into writing alternatives.

I did not write anything about how to achieve results in Tailwind because teams that are not yet in Tailwind and might need something will have to decide on how to handle these (I don't know if there is any, tbh).

I can add a few examples in the style of before and after scenarios, but I am afraid they will either be too incomplete (I believe it's _too much_ to cover all scenarios) or too specific, so I am not entirely sure about what to add.

packages/orbit-components/src/utils/Grid/README.md
RobinCsl: Okay, so there's no more IE-related workaround that's still in the codebase to your knowledge?

packages/orbit-components/src/utils/rtl/README.md
RobinCsl: Would it make sense to include this info in the migration guide, just to point people somewhere? I'm sure I wouldn't be the only one asking that question :)

docs/src/documentation/05-development/04-migration-guides/01-v15.mdx
RobinCsl: Okay, got it.

Maybe it still make sense to just point out to Tailwind docs for RTL and media queries
Maybe we could still offer a permalink to the previous RTL utility functions for folks to copy them over in their codebases in case they'd need them?

packages/orbit-components/src/utils/Grid/README.md
DSil: AFAIK, no

mvidalgarcia (commented):
Minor nitpicks 💅

.github/contribution/component-design.md
```diff
@@ -5,49 +5,20 @@
...
-
-Set up `defaultProps` by assigning default values to the props. Destructure them and finally return the Component.
-
-```jsx
-const Component = (props: Props) => {
-  const { size = SIZE_OPTIONS.LARGE } = props;
-  return <StyledComponent size={size}>...</StyledComponent>;
-};
-```
+We use `tailwind` for styling. Custom tailwind classes can be used but should be avoided. All design tokens from `orbit-design-tokens` have classes that apply their value to the different possible CSS properties.
```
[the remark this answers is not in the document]
DSil: I think it's ok to use `Tailwind` (or `Tailwind`, but consistently) as a short for `TailwindCSS`. For this particular sentence, I rephrased it to:
> We use `TailwindCSS` for styling. Custom Tailwind classes can be used but should be avoided.

What do you think?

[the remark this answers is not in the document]
mvidalgarcia: On the official site, there's a space between "Tailwind" and "CSS". I don't really mind the form as long as it's consistent.

image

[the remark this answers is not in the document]
DSil: Yeah, we're using both forms across all docs 🤦‍♂️  I'll unify on a separate commit

mvidalgarcia (approved):
:shipit:

PR #4319, opened 2024-03-28
fix(Icon): icon components are now correctly typed

DSil:
Icon components now have their props correctly typed.

[FEPLT-1920](https://kiwicom.atlassian.net/browse/FEPLT-1920)

[FEPLT-1920]: https://kiwicom.atlassian.net/browse/FEPLT-1920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/src/components/Navbar/Links.tsx
+++ b/docs/src/components/Navbar/Links.tsx
@@ -32,7 +32,7 @@ const LinkIcon = ({ name }: { name?: string }) => {
     case "Accessibility":
       return <IrisScanIcon {...iconStyles} />;
     case "Playground":
-      return <Code {...iconStyles} />;
+      return <Code size="medium" />;
     default:
       return <ScriptIcon {...iconStyles} />;
   }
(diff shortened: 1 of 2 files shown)


docs/src/components/Navbar/Links.tsx
```diff
@@ -32,7 +32,7 @@ const LinkIcon = ({ name }: { name?: string }) => {
     case "Accessibility":
       return <IrisScanIcon {...iconStyles} />;
     case "Playground":
-      return <Code {...iconStyles} />;
+      return <Code size="medium" />;
```
DSil, line 35: This size applies the same 20px size and the `stroke` prop is actually the default color value for an Icon. This was not marked as type error because props were not typed, but it was working properly because the size and colors used are the default ones 😅

RobinCsl: Good think we added types then :)

DSil:
> Did you verify that the generated built <icon>.d.ts files have proper typings?

Yes, it does 👍

PR #4326, opened 2024-04-04
chore(Pagination): add rtl story

vbulant:
Closes [FEPLT-1963](https://kiwicom.atlassian.net/browse/FEPLT-1963)

### For Tailwind migrations:

- [ ] Visual regression test suite has been added if does not exist.
- [ ] `tailwind-migration-status.yaml` file has been updated with the migrated component(s).

### For new components:


mvidalgarcia:
@vbulant You just blew my mind 🤯 

After [playing around a bit](https://codi.link/PGRpdiBkaXI9Imx0ciI+CiAgMyBvZiA2CjwvZGl2PgoKPGRpdiBkaXI9InJ0bCI+CiAgMyBvZiA2CjwvZGl2PgoKPGRpdiBkaXI9InJ0bCI+CiAgMyDYp9mO2YTZkti52Y7YsdmO2KjZkNmK2Y7Zkdip2Y8gNgo8L2Rpdj4=%7C%7C), I understand it better. I think it's a good point and a section should be added to the link you pointed out. I can write it if you're busy, let me know.

vbulant:
@mvidalgarcia I would appreciate if you wrote it. I’m quite busy 🙏

mvidalgarcia:
> @mvidalgarcia I would appreciate if you wrote it. I’m quite busy 🙏

Sure I'll push a commit on top of yours. Thanks so much!

mvidalgarcia (approved):

packages/orbit-components/src/utils/rtl/README.md
```diff
@@ -19,3 +19,30 @@ In addition, you also need to set the `dir` property on the `html` element to ha
...
+
+### Handling Bidirectional Text
+
+To handle bidirectional text correctly:
+
+1. Use fully RTL text to ensure proper rendering.
+2. If using bidirectional text is necessary, wrap the LTR part with an element that has `dir="ltr"` to enforce the correct rendering order.
+
+```html
+<span dir="rtl">שלום <span dir="ltr">world!</span></span>
```
RobinCsl, line 45:
```
<div dir="rtl">
<span dir="rtl">שלום <span dir="ltr">world!</span></span>
</div>
```
puts the exclamation mark in the wrong place, so it probably should be
```
<div dir="rtl">
<span dir="rtl">שלום <span dir="ltr">world</span>!</span>
</div>
```
to be more correct?

Also, what about the `<bdi>` and `<bdo>` tags? They might be more idiomatic than `<span dir="...">`. WDYT?

RobinCsl:
Is it expected the code sample is also RTL and has misplaced double quotes? 😅 
Screenshot 2024-04-05 at 12 31 06

packages/orbit-components/src/Pagination/Pagination.stories.tsx
```diff
@@ -99,3 +100,58 @@ export const Playground = () => {
...
+  const hideLabels = boolean("hideLabels", false);
+  const size = select("size", Object.values(SIZES), SIZES.NORMAL);
+  const labelProgress = text("labelProgress", "2 מתוך 6");
+
+  return (
+    <RenderInRtl>
+      <Pagination
+        labelPrev="Prev"
+        labelNext="Next"
+        labelProgress={labelProgress}
```
RobinCsl, line 116: Would it make sense to make it 
```suggestion
        labelProgress={<div dangerouslySetInnerHTML={{ __html: labelProgress}} />}
```
as it's supposed to accept a React node?

I guess it's not great from security point of view (but since Storybook is hosted on external domains, it should be okay). Otherwise, we could just make a note to run this locally to try with additional markup elements to specify the direction of the text. 🤔

packages/orbit-components/src/utils/rtl/README.md
vbulant: Re. exclamation mark – not sure what’s correct from the linguistic perspective. Maybe it would be better to remove the mark altogether.

I think `bdo` is for quite specific use cases (as noted in https://www.w3.org/International/articles/inline-bidi-markup/) but `bdi` is a good point. It’s also recommended in the latter doc for some cases. But for the cases when the directionality is known they recommend to use a span with dir attr without any further reasoining 🤔

vbulant:
> Is it expected the code sample is also RTL and has misplaced double quotes? 😅 Screenshot 2024-04-05 at 12 31 06

🤣  Well other components with RTL stories suffer from the same… Maybe we should change it so that code samples stay in LTR.

packages/orbit-components/src/Pagination/Pagination.stories.tsx
vbulant: I think this is also an issue across multiple components that accept `ReactNode` so maybe it would make sense to think about it in general. If we go the `dangerouslySetInnerHTML` way we’ll also need to disable html escaping for knobs I guess. Not sure if it’s possible to disable it per knob.

Another option could be to provide several predefined examples.

mvidalgarcia:
> > Is it expected the code sample is also RTL and has misplaced double quotes? 😅
> 
> 🤣 Well other components with RTL stories suffer from the same… Maybe we should change it so that code samples stay in LTR.

Good catch, addressed [here](https://github.com/kiwicom/orbit/pull/4326/commits/34fcb2a785079123abdceabd0c111767aaa10ce0).

packages/orbit-components/src/utils/rtl/README.md
mvidalgarcia: I added a third bullet point explaining the use case of `bdi` as it seems relevant.

packages/orbit-components/src/Pagination/Pagination.stories.tsx
mvidalgarcia: I think it would be simpler either to leave it as it is (the story represents RTL and the default value proves it) or provide a list of 2 or 3 predefined examples as @vbulant mentioned. I wouldn't complicate things by allowing you to type raw HTML in a Storybook.

WDYT?

PR #4332, opened 2024-04-16
fix(Modal): overflow now works correctly on large mobile and bigger

DSil:
Adds the missing class to make overflow visible on bigger devices.

By adding `lm:overflow-visible` to the Modal, the rounded corners were being missed. Therefore, more rounded corner classes needed to be added. To simplify, and for further documentation, I'd leave the comments inline on this PR.

[FEPLT-1979](https://kiwicom.atlassian.net/browse/FEPLT-1979)

[FEPLT-1979]: https://kiwicom.atlassian.net/browse/FEPLT-1979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-components/src/Modal/ModalFooter/index.tsx
+++ b/packages/orbit-components/src/Modal/ModalFooter/index.tsx
@@ -73,6 +73,7 @@ const ModalFooter = ({ dataTest, children, flex = "0 1 auto" }: Props) => {
         "z-overlay bg-white-normal px-md pb-md box-border flex w-full pt-0",
         "duration-fast transition-shadow ease-in-out",
         "sm:max-lm:[&_.orbit-button-primitive]:h-form-box-normal sm:max-lm:[&_.orbit-button-primitive]:text-button-normal",
+        "lm:rounded-b-modal",
         childrenLength > 1 ? "lm:justify-between" : "lm:justify-end",
         "[&_.orbit-modal-footer-child:last-of-type]:p-0",
       )}
--- a/packages/orbit-components/src/Modal/ModalHeader/index.tsx
+++ b/packages/orbit-components/src/Modal/ModalHeader/index.tsx
@@ -24,7 +24,8 @@ export const ModalHeaderWrapper = ({
       className={cx(
         className,
         "orbit-modal-header-container",
-        "box-border block w-full",
+        "rounded-t-modal box-border block w-full",
+        "lm:[&_~_.orbit-modal-section]:rounded-t-none",
         suppressed
           ? [
               "bg-cloud-light py-xl px-md lm:p-xl",
--- a/packages/orbit-components/src/Modal/ModalSection/index.tsx
+++ b/packages/orbit-components/src/Modal/ModalSection/index.tsx
@@ -29,6 +29,7 @@ export const ModalSectionWrapper = ({
         "orbit-modal-section",
         "py-lg px-md lm:p-xl box-border w-full",
         "border-b-elevation-flat-border-color border-b border-solid",
+        "lm:last-of-type:rounded-b-modal lm:first-of-type:rounded-t-modal",
         suppressed
           ? [
               "bg-cloud-light",
--- a/packages/orbit-components/src/Modal/index.tsx
+++ b/packages/orbit-components/src/Modal/index.tsx
@@ -435,11 +435,12 @@ const Modal = React.forwardRef<Instance, Props>(
           <div
             className={cx(
               "orbit-modal-wrapper-content",
-              "lm:rounded-modal overflow-y-auto overflow-x-hidden",
+              "lm:rounded-modal lm:overflow-visible overflow-y-auto overflow-x-hidden",
               "font-base bg-elevation-flat shadow-overlay absolute box-border w-full",
               "lm:relative lm:bottom-auto lm:pb-0",
               "lm:[&_.orbit-modal-section:last-of-type]:pb-xxl lm:[&_.orbit-modal-section:last-of-type:after]:content-none lm:[&_.orbit-modal-section:last-of-type]:mb-[var(--orbit-modal-footer-height,0px)]",
               "lm:[&_.orbit-modal-mobile-header]:w-[calc(var(--orbit-modal-width)-48px-theme(spacing.xxl))]",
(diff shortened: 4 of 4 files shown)


packages/orbit-components/src/Modal/ModalFooter/index.tsx
```diff
@@ -73,6 +73,7 @@ const ModalFooter = ({ dataTest, children, flex = "0 1 auto" }: Props) => {
         "z-overlay bg-white-normal px-md pb-md box-border flex w-full pt-0",
         "duration-fast transition-shadow ease-in-out",
         "sm:max-lm:[&_.orbit-button-primitive]:h-form-box-normal sm:max-lm:[&_.orbit-button-primitive]:text-button-normal",
+        "lm:rounded-b-modal",
```
DSil, line 76: This is to apply the bottom rounded corners on the Modal, when it has a footer

packages/orbit-components/src/Modal/ModalHeader/index.tsx
```diff
@@ -24,7 +24,8 @@ export const ModalHeaderWrapper = ({
       className={cx(
         className,
         "orbit-modal-header-container",
-        "box-border block w-full",
+        "rounded-t-modal box-border block w-full",
```
DSil, line 27: This is to apply the top rounded corners on the Modal, when it has a header

packages/orbit-components/src/Modal/ModalHeader/index.tsx
```diff
@@ -24,7 +24,8 @@ export const ModalHeaderWrapper = ({
...
+        "lm:[&_~_.orbit-modal-section]:rounded-t-none",
```
DSil, line 28: Be default, a modal-section has rounded top corners. But when the Modal has a header, those corners can't be rounded. This class prevents that.

packages/orbit-components/src/Modal/ModalSection/index.tsx
```diff
@@ -29,6 +29,7 @@ export const ModalSectionWrapper = ({
         "orbit-modal-section",
         "py-lg px-md lm:p-xl box-border w-full",
         "border-b-elevation-flat-border-color border-b border-solid",
+        "lm:last-of-type:rounded-b-modal lm:first-of-type:rounded-t-modal",
```
DSil, line 32: The first section should have top rounded corners. The last section should have bottom rounded corners.

packages/orbit-components/src/Modal/index.tsx
```diff
@@ -435,11 +435,12 @@ const Modal = React.forwardRef<Instance, Props>(
           <div
             className={cx(
               "orbit-modal-wrapper-content",
-              "lm:rounded-modal overflow-y-auto overflow-x-hidden",
+              "lm:rounded-modal lm:overflow-visible overflow-y-auto overflow-x-hidden",
               "font-base bg-elevation-flat shadow-overlay absolute box-border w-full",
               "lm:relative lm:bottom-auto lm:pb-0",
               "lm:[&_.orbit-modal-section:last-of-type]:pb-xxl lm:[&_.orbit-modal-section:last-of-type:after]:content-none lm:[&_.orbit-modal-section:last-of-type]:mb-[var(--orbit-modal-footer-height,0px)]",
               "lm:[&_.orbit-modal-mobile-header]:w-[calc(var(--orbit-modal-width)-48px-theme(spacing.xxl))]",
+              footerHeight && "lm:[&_.orbit-modal-section]:rounded-b-none",
```
DSil, line 443: If the Modal has a footer, the modal-section does not need to have rounded bottom corners (as the footer applies them)

packages/orbit-components/src/Modal/index.tsx
```diff
@@ -456,7 +457,7 @@ const Modal = React.forwardRef<Instance, Props>(
                   "[&_.orbit-modal-footer]:duration-fast [&_.orbit-modal-footer]:transition-shadow [&_.orbit-modal-footer]:ease-in-out",
                   fullyScrolled
                     ? "[&_.orbit-modal-footer]:shadow-modal-scrolled"
-                    : "[&_.orbit-modal-footer]:shadow-modal",
+                    : "[&_.orbit-modal-footer]:shadow-modal lm:[&_.orbit-modal-footer]:rounded-b-none",
```
DSil, line 460: If the footer is fixed and the user is scrolling, the footer can't have rounded corners, otherwise the content of the model would be visible from the edges of the rounded corners.

mvidalgarcia:
How I tested it.

I added this to one storybook:

```diff
diff --git a/packages/orbit-components/src/Modal/Modal.stories.tsx b/packages/orbit-components/src/Modal/Modal.stories.tsx
index 7cc5efeda..5fbf5e98f 100644
--- a/packages/orbit-components/src/Modal/Modal.stories.tsx
+++ b/packages/orbit-components/src/Modal/Modal.stories.tsx
@@ -511,6 +511,9 @@ export const FullPreview = () => {
             </Box>
           </ModalFooter>
         )}
+        <div style={{ zIndex: 1000, position: "absolute", top: -25 }}>
+          <Text type="critical">Foo</Text>
+        </div>
       </Modal>
     </Container>
   );
```

And then compared the output.

Master: Not visible
Current PR:
image

Is this the right approach?

DSil:
@mvidalgarcia interesting. I also noticed during the development of this task that it doesn't matter if one puts the ModalFooter before the ModalContent — it will be rendered in the correct place.

Honestly, I don't think this is a problem we have to deal with. The correct usage of the Modal is being violated in that snippet (one should rely on Modal subcomponents only). Ultimately, it's a consequence of the way the Modal is implemented, I guess. Do you see it as being a critical problem? Or can we be safe that the API is not being respected, thus it is not our responsibility?

mvidalgarcia:
@DSil Sorry I wasn't very clear in my message 😅 
My point was that elements overflowing the modal are visible (such as the date picker reported by Vojtech), so I guess it fixes the issue at hand. I couldn't find any other way to test a similar use case to the reported one.

Said that, and given the visual tests are passing, I think it solves the problem, so I'll approve.

mvidalgarcia (approved):

PR #4336, opened 2024-04-19
fix(InputSelect): it now filters options when they change via props

DSil:
InputSelect component now applies the filter to the options when they change via props. In the past, it was applying only when the input changed. If the options received via props changed as well, the filtering was applied before the change, so it was not applied on the correct set of options.

[FEPLT-1985](https://kiwicom.atlassian.net/browse/FEPLT-1985)

[FEPLT-1985]: https://kiwicom.atlassian.net/browse/FEPLT-1985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/orbit-components/src/InputSelect/index.tsx
+++ b/packages/orbit-components/src/InputSelect/index.tsx
@@ -61,7 +61,10 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
 
     const [isOpened, setIsOpened] = React.useState(false);
     const [inputValue, setInputValue] = React.useState(
-      defaultSelected ? options.find(opt => opt.value === defaultSelected.value)?.title : "",
+      defaultSelected
+        ? options.find(opt => opt.value === defaultSelected.value)?.title ||
+            String(defaultSelected.title)
+        : "",
     );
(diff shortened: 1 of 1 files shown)


mvidalgarcia:
There's a strange behavior in the `Playground` story after clearing (deleting all chars) the input. It seems that it still performs a filtering based on the last char, even though the input is empty and should show all the results. You can check it in this recording.


https://github.com/kiwicom/orbit/assets/8863238/a27dd226-d4dc-4f8c-ae05-e015c1c3e5d5

DSil:
@mvidalgarcia well spotted. It was on the condition to perform the filtering. I have fixed it, thanks 👍 

Further explanation: I was checking `if (inputValue)` and it was evaluating to `false` when you cleared the input.
I was doing such verification because `inputValue` could be undefined, in the very weird case that we have a `defaultSelected` defined but we can't find that option in the options array (that would lead to bigger problems anyway…).
So I added a fallback for that scenario, and now `inputValue` is always defined (even if it is just `""`) and so we don't need an extra verification.

mvidalgarcia (approved):

PR #4346, opened 2024-05-08
fix(Itinerary): clickable area is now more accurate

DSil:
The onClick call is now associated with the wrapper of the parts of the segment, and not with the whole segmented, that used to include separators, spacings and banners.

Clickable banners are also now focusable and accessible via keyboard navigation.

[FEPLT-1871](https://kiwicom.atlassian.net/browse/FEPLT-1871)


mvidalgarcia:
Good work, it fixes most of the problems.

However, I'm not sure if it fixes what was reported:

> We have a clickable itinerary segment banner, however the space just below the separator, and just above the bottom of the segment is not clickable. Could you rearrange the margins/paddings so that the entire banner is clickable?

Now it's definitely more clear where to click, as the cursor changes on hover. But it still doesn't cover the whole banner area. I believe this is because the separator has some margins impeding it. Do you think it can be fixed? Otherwise, I'm OK with this solution. I'm only worried about mobile devices, as those don't have a cursor.

DSil:
@mvidalgarcia good point. The problem is that what applies the spacing is the separator, and not the banner itself. I will investigate if we can rework into having the spacing applied by the banner itself, I just don't know the consequences it might have on already existing usages of the component. I'll get back to this soon

mvidalgarcia (approved):
As discussed IRL, it already improves the current solution so we can go for it and iterate in the future if needed.

PR #4359, opened 2024-05-22
chore: escape chars in icon fetching script

DSil:
Parenthesis chars are being used and they need to be escaped

This will fix the problem observed in [#4358](https://github.com/kiwicom/orbit/pull/4358)


mvidalgarcia (approved):

mvidalgarcia:
LGTM, [capturing group works as expected](https://regex101.com/r/3xWHYq/1).

PR #4363, opened 2024-05-23
Update playwright and improve ServiceLogo tests

DSil:
**Per commit review advised**

Remember to run `yarn playwright install` on your local machine after installing the dependencies of this branch

This updates the playwright version and the snapshots due to the browsers update.

The ServiceLogo tests are also updated to be more permissive.


packages/orbit-components/src/ServiceLogo/ServiceLogo.ct.tsx-snapshots/visual-ServiceLogo-default-1-Desktop-darwin.png
RobinCsl, line 1: Is it expected some logos seem to be there more than once? e.g. Alipay, American Express (same size twice), etc.?

DSil: Yes, so:
- The first two are a result of an iteration between the possible values of the prop `grayScale` (`true` or `false`), all with the same value on the prop `name`: `Alipay`.
- The second three are a result of an iteration between the possible values of the prop `size` (`small`, `medium` or `large`), all with the same value on the prop `name`: `Amex`.
- From there on we iterate over all possible names. So `Alipay` and `Amex` are repeated there just to avoid filtering them again. The `size` and `grayScale` props are undefined, so eventually they will be repeated with one of the cases above

packages/orbit-components/src/ServiceLogo/ServiceLogo.ct.tsx
```diff
@@ -2,12 +2,27 @@ import * as React from "react";
...
+      Object.values(SIZE_OPTIONS).length +
+      GRAYSCALE_OPTIONS_SIZE
+    })`;
+
+    await page.waitForSelector(selector);
+
+    await expect(component).toHaveScreenshot({
+      threshold: 0.3,
+      maxDiffPixelRatio: 0.02,
+      maxDiffPixels: undefined,
```
mvidalgarcia, line 25: How does it behave when this value is undefined?

I see that [this value is unset by default](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1-option-max-diff-pixels), so I assume it would be the same as not setting it at all?

packages/orbit-components/src/ServiceLogo/ServiceLogo.ct.tsx
```diff
@@ -2,12 +2,27 @@ import * as React from "react";
...
+      Object.values(NAME_OPTIONS).length +
+      Object.values(SIZE_OPTIONS).length +
+      GRAYSCALE_OPTIONS_SIZE
+    })`;
+
+    await page.waitForSelector(selector);
+
+    await expect(component).toHaveScreenshot({
+      threshold: 0.3,
+      maxDiffPixelRatio: 0.02,
```
mvidalgarcia, line 24: These values seem a bit arbitrary, is there a specific reason behind them? Perhaps we can add a comment.

mvidalgarcia (approved):

RobinCsl:
Updating the Playwright version also seems to require an additional step for the Action to update snapshots. Will take a quick look.

.github/workflows/test.yml
```diff
@@ -126,7 +126,7 @@ jobs:
     name: Visual Regression Tests
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.39.0-jammy
+      image: mcr.microsoft.com/playwright:v1.44.1-jammy
```
RobinCsl, line 129: We should have also updated `.github/workflows/update-visual-tests.yml`. PR here: https://github.com/kiwicom/orbit/pull/4373

packages/orbit-components/src/ServiceLogo/ServiceLogo.ct.tsx
DSil: This is `undefined` here so this overrides the [global config](https://github.com/kiwicom/orbit/blob/23532b846c399467f722c85aff8f7c048f815db6/packages/orbit-components/playwright-ct.config.ts#L19). We don't want this to be used on this test.

packages/orbit-components/src/ServiceLogo/ServiceLogo.ct.tsx
DSil: I confess they were a bit trial and error, so not science at all. Since we're comparing images I just tried the minimum values that would allow the tests to pass, trying multiple times until achieving consistent results

PR #4370, opened 2024-05-30
fix(ItinerarySegmentDetail): ensure text wraps instead of being trunc…

RobinCsl:
…ated

There was a report that some amount of key information was hidden in smaller viewports for the ItinerarySegmentDetail component. The content was previously truncated (with the Truncate component) and we now instead prefer to break words and stretch the lines as necessary.

Note that this was the only component where Truncate was used, so we might want to consider deleting that component in case it's no longer necessary and not used by our consumers.

Resolves FEPLT-2013

### For new components:

- [x] Unit and visual regression tests have been added/adjusted for my new feature


mvidalgarcia:
> Note that this was the only component where Truncate was used, so we might want to consider deleting that component in case it's no longer necessary and not used by our consumers.

However, it's still used across FE modules.

image

mvidalgarcia:
I wonder whether we should merge [this PR](https://github.com/kiwicom/orbit/pull/4363) first, to avoid conflicts and extra work from Daniel. What do you think?

RobinCsl:
> 💅 This won't look very nice
> 
> image
> 
> Perhaps we can keep just "Itinerary"?

You mean that the changelog will display the truncated version? I can adjust it, for sure.

RobinCsl:
> I wonder whether we should merge [this PR](https://github.com/kiwicom/orbit/pull/4363) first, to avoid conflicts and extra work from Daniel. What do you think?
> 
> We can always document the playwright config values later, and perhaps it would solve your screenshots commit.

We can do, if the only blocker was missing documentation, then we can do and then I'd rebase mine.

mvidalgarcia:
> > 💅 This won't look very nice
> > image
> > Perhaps we can keep just "Itinerary"?
> 
> You mean that the changelog will display the truncated version? I can adjust it, for sure.

Yes! sorry for not being explicit.

RobinCsl:
> > > > I wonder whether we should merge [this PR](https://github.com/kiwicom/orbit/pull/4363) first, to avoid conflicts and extra work from Daniel. What do you think?
> > > > We can always document the playwright config values later, and perhaps it would solve your screenshots commit.
> > > 
> > > 
> > > We can do, if the only blocker was missing documentation, then we can do and then I'd rebase mine.
> > 
> > 
> > I created a [ticket](https://kiwicom.atlassian.net/browse/FEPLT-2014) to remember, once Daniel comes back.
> > I'll merge it.
> 
> Done, you can recreate the last commit. Hopefully, it will work fine 🤞

Alright, I dropped the ServiceLogo snapshot updates and only kept the ItinerarySegment one.
Also updated the commit message so `truncated` shouldn't be truncated anymore... 😅

mvidalgarcia (approved):
Looks good, doesn't it? 😃 

Approved!

PR #4372, opened 2024-05-31
feat(Radio): update component visually to match latest Figma

RobinCsl:
The task was to only update the 'selected' state (both in 'normal' and 'disabled' states), but I noticed additional inconsistencies between Figma and the current state.

Resolves FEPLT-2012

### For new components:


mvidalgarcia:
We need to update [this picture](https://orbit.kiwi/components/input/radio/#content-structure) as well.

mvidalgarcia:
>  but I noticed additional inconsistencies between Figma and the current state.

Could you elaborate on this?

I think adding the Figma link in the PR description would help as well.

RobinCsl:
> > but I noticed additional inconsistencies between Figma and the current state.
> 
> Could you elaborate on this?
> 
> I think adding the Figma link in the PR description would help as well.

It's in the Jira ticket so didn't think it'd be necessary to add the info here as well.

(Actually, do you know why the Jira ticket wasn't picked up?)

RobinCsl:
> > but I noticed additional inconsistencies between Figma and the current state.
> 
> Could you elaborate on this?
> 
> I think adding the Figma link in the PR description would help as well.

Basically, the default border should be 1px-wide and only in some other cases it is 2px-wide.
Similarly, the outline was only 1px-wide when it should be 2px wide.
The color of the outline was also incorrect for the error state.

packages/orbit-components/src/Radio/index.tsx
```diff
@@ -64,18 +64,21 @@ const Radio = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
             "flex flex-none items-center justify-center",
             "size-icon-medium rounded-full",
             "duration-fast scale-100 transition-all ease-in-out",
-            "lm:border border-[2px] border-solid",
+            "lm:border border-solid",
+            checked ? "border-[2px]" : "border-[1px]",
```
mvidalgarcia, line 68: On large mobile or greater and checked, `lm:border` takes preference over `border-[2px]` (mq specificity I assume), resulting in a 1px border when it should be 2px.

image

mvidalgarcia:
> > We need to update [this picture](https://orbit.kiwi/components/input/radio/#content-structure) as well.
> 
> I ran the CI job to update snapshots but it seems to be taking a long time.

How is this related to snapshots? I meant this image in orbit.kiwi

https://orbit.kiwi/static/fdcbdae8f39964dc436131a8c0d06128/0d390/160%253A1219.png

mvidalgarcia:
> Basically, the default border should be 1px-wide and only in some other cases it is 2px-wide. Similarly, the outline was only 1px-wide when it should be 2px wide. The color of the outline was also incorrect for the error state.

Yeah, also the focus state is currently greyish when it should be blue

image 
image

packages/orbit-components/src/Radio/index.tsx
DSil: Well spotted. There does not seem to be a responsive difference now, so I have adjusted the classes. I have also used the native `border` and `border-2` tailwind classes, avoiding custom ones 🙂

packages/orbit-components/src/Itinerary/Itinerary.ct.tsx-snapshots/visual-Itinerary-ItinerarySegment-open-1-Small-Mobile-darwin.png
DSil, line 1: Leftover from other PR, updated here for convenience

mvidalgarcia:
> > Basically, the default border should be 1px-wide and only in some other cases it is 2px-wide. Similarly, the outline was only 1px-wide when it should be 2px wide. The color of the outline was also incorrect for the error state.
> 
> Yeah, also the focus state is currently greyish when it should be blue
> 
> image image

This still applies, I believe is the only thing left.

mvidalgarcia (approved):
:shipit:

PR #4392, opened 2024-06-21
fix(Itinerary): add type check for key prop in ItinerarySegmentDetail

domihustinova:
Adding type check for `title`, `name` and `value` in order to fix the `[object Object]` in the key.

Closes FEPLT-1999

### DIFF
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
@@ -97,7 +97,7 @@ const ItinerarySegmentDetail = ({
             >
               {content.map(({ title, items }, idx) => {
                 return (
-                  <React.Fragment key={`${title}`}>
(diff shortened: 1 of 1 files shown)


DSil (commented):
LGTM, I would just tweak the commit message. I am not sure if it should be a fix, but I am ok with it. However, the scope should be just Itinerary. We tend to use the main component only for scope. If you want, you can mention the subcomponent on the message. Something like:

> fix(Itinerary): add type check for key prop in ItinerarySegmentDetail

domihustinova:
@DSil 

> I am not sure if it should be a fix

I see, you are correct, the change is not really fixing a bug. Reading the usage, I think `refactor` would be more suitable in this case, WDYT?

> We tend to use the main component only for scope.

Thanks! I'll remember that and amend the commit.

DSil:
We don't really use refactor for commit type. Only feat, chore or fix. Chore is not mentioned on the changelog, that's why I'm ok keeping fix

DSil (approved):
🚀

PR #4396, opened 2024-06-25
fix(ItinerarySegmentStop): use ref to fix rendering glitch on Safari

RobinCsl:
There is a rendering glitch in Safari where the vertical bar is not rendered at the right place.
It seemed odd to use a 'setState' like function to create a ref, so we use a proper ref instead.
This seems to force Safari to finish the drawing of the component until the vertical bar is in the right place, though, depending on the speed of rendering, we still can see a transient state where the vertical bar is not where it should be.

Slack: https://skypicker.slack.com/archives/C7T7QG7M5/p1719321209379529


packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
```diff
@@ -396,14 +396,16 @@ export const Stop = () => {
...
       <Heading type="title2">Regular stop</Heading>
-      <ItinerarySegmentStop
-        city={city}
-        station={station}
-        hidden={hidden}
-        date={date}
-        type={type}
-        time={time}
-      />
+      <Itinerary>
```
RobinCsl, line 399: This was causing a bug in Storybook where the minimal width was 0 (because no context was available for this isolated component).

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
```diff
@@ -27,15 +27,21 @@ const ItinerarySegmentStop = ({
...
+  const { calculatedWidth, setWidth } = useWidth();
   const { last, isBanner, index } = usePart();
   const [dateWidth, setDateWidth] = React.useState<HTMLDivElement | null>(null);
 
   const textType = type === "neutral" ? "primary" : type;
 
   React.useEffect(() => {
-    setWidths(prev => (dateWidth?.clientWidth ? [...prev, minWidth, dateWidth.clientWidth] : prev));
-  }, [setWidths, dateWidth, minWidth]);
+    setWidth(minWidth);
```
RobinCsl, line 37: We still want to ensure the preference for the min width is respected, so we call it once and whenever that value changes.

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentStop/index.tsx
```diff
@@ -27,15 +27,21 @@ const ItinerarySegmentStop = ({
...
+  }, [setWidth, minWidth]);
+
+  React.useEffect(() => {
+    if (dateWidth?.clientWidth) {
+      setWidth(dateWidth.clientWidth);
```
RobinCsl, line 42: We only try to update the calculated offset when it makes sense.

packages/orbit-components/src/Itinerary/context.tsx
```diff
@@ -15,19 +15,21 @@ export const ItineraryProvider = ({
...
   children: React.ReactNode;
 }): React.ReactElement<Context> => {
-  const [widths, setWidths] = React.useState([70]);
-  const [calculatedWidth, setCalculatedWidth] = React.useState(0);
+  const [calculatedWidth, setCalculatedWidth] = React.useState(70);
 
-  React.useEffect(() => {
-    setCalculatedWidth(Math.max(...widths));
-  }, [widths]);
+  const setWidth = React.useCallback(
```
DSil, line 20: To clarify, this `setWidth`, instead of storing all the widths and then have the `useEffect` for `widths` triggered to calculate the max value, it indeed calculates the max value immediately and stores it in the same variable (`calculatedWidth`). With this, we avoid having a state storing an array for just one operation and we get the same desired result.

You confirm?

packages/orbit-components/src/Itinerary/__tests__/index.test.tsx
```diff
@@ -4,15 +4,30 @@ import { renderHook, act } from "../../test-utils";
...
+    expect(result.current.calculatedWidth).toBe(70);
+  });
   it("should have calculated width", () => {
     const wrapper = ({ children }) => <ItineraryProvider>{children}</ItineraryProvider>;
 
     const { result } = renderHook(() => useWidth(), { wrapper });
 
     act(() => {
-      result.current.setWidths([30, 40, 50]);
+      result.current.setWidth(60);
```
DSil, line 26: Are these just random numbers to check that it correctly picks the maximum?

DSil (commented):
Just minor questions for my own understanding, and for future reference if we ever need to come back to this PR 🙏

packages/orbit-components/src/Itinerary/context.tsx
RobinCsl: Yes, exactly.
AFAIK, this context is reused for all the SegmentStop components (so there likely are quite a few call to this, meaning lots of unnecessary operations on this array got removed as a result of this change).

We're only interested in the max value.
So doing Math.max(all the possible values) or Math.max(val1, val2) over each call amounts to the same.

> between the width of the date and the minWidth?

Regarding `minWidth`, it's just an initialization thing, so you just call it once with the minWidth, and you're sure that it'll be at least `minWidth`.

packages/orbit-components/src/Itinerary/__tests__/index.test.tsx
RobinCsl: Yes, they're random. I just wanted to make sure one would be higher than the default value 70 in the context.

It's a bit of a magic value and I would prefer for it to be a constant (since it's also used as a default value elsewhere). But I'm not sure about the reasoning behind that value so didn't want to propose any bigger change for now.

As I mentioned below, if we agree this is a good approach, I think there would still be one more step/commit to rename things better for them to be more descriptive as to what they really do, including that magic 70 value.

DSil: Agreed. To me they can be renamed

RobinCsl: Alright, added my proposal to rename them :)

DSil (approved):

domihustinova:
Hi @RobinCsl , how did you make the connection that the whole width issue could be causing the glitch?

RobinCsl:
> Hi @RobinCsl , how did you make the connection that the whole width issue could be causing the glitch?

It was a very strange bug since it only occurred in Safari. It seemed like the columns weren't drawn in the appropriate place, but whenever I made any modification to the CSS properties in the dev tools, the right output would be shown on the screen.

I then inquired whether some value was improperly updated and found this context governing it.

When playing around with it and trying to make it more predictable in when it would "finish" running, I realised it could be simplified a little bit and made more performant too.

It solved the weird re-drawing issues in Safari and made the rendering snappy when Fast Refresh was happening.

So, it's just been an exploration and some intuition, but there was no "aha" moment as such.

Hope that helps :)

PR #4398, opened 2024-06-27
fix(Itinerary): add onExpand and onCollapse props into ItinerarySegment

domihustinova:
The issue with usage of `ItinerarySegmentDetails` is that even if there is some `onClick` handler provided in the `ItinerarySegment`, this is not being executed when user clicks anywhere in the expanded `ItinerarySegmentDetails` due to stopPropagation in the `Slide` component used in the `ItinerarySegmentDetail` when content is provided. It fires only when user clicks anywhere around the element (within the `ItinerarySegment`).

- I added `onExpand` and `onCollapse` props to be fired when expand/collapse occurs
- I added all three props to https://orbit.kiwi/components/structure/itinerary/react/#itinerarysegment

Closes FEPLT-2043

### DIFF
--- a/packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
+++ b/packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
@@ -168,7 +168,11 @@ export const Segment = () => {
   return (
     <div className="space-y-xl">
       <Itinerary>
-        <ItinerarySegment>
+        <ItinerarySegment
+          onClick={action("clicked")}
+          onCollapse={action("collapsed")}
+          onExpand={action("expanded")}
+        >
           <ItinerarySegmentStop
             city="Prague"
             station="Václav Havel Airport Prague (PRG)"
@@ -514,7 +518,12 @@ export const Stop = () => {
 
 export const Detail = () => {
   return (
-    <ItinerarySegment noElevation>
+    <ItinerarySegment
+      noElevation
+      onClick={action("clicked")}
+      onCollapse={action("collapsed")}
+      onExpand={action("expanded")}
+    >
       <ItinerarySegmentDetail duration="2h 30m" summary={<BadgeGroup />} content={content} />
     </ItinerarySegment>
   );
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
@@ -87,7 +87,7 @@ const ItinerarySegmentDetail = ({
             ref={slideRef}
             tabIndex={-1}
             role="button"
-            onClick={toggleOpened}
+            onClick={ev => toggleOpened(ev)}
             onKeyDown={ev => {
               ev.stopPropagation();
             }}
--- a/packages/orbit-components/src/Itinerary/ItinerarySegment/context.tsx
+++ b/packages/orbit-components/src/Itinerary/ItinerarySegment/context.tsx
@@ -8,7 +8,7 @@ interface Context {
   isBanner: boolean;
   index: number;
   opened: boolean;
-  toggleOpened: () => void;
+  toggleOpened: (ev: React.SyntheticEvent<HTMLDivElement>) => void;
   last: boolean;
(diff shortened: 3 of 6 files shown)


DSil (changes requested):
If one selects text on the details part, the onClick callback is called, but it did not close nor expand. On the request, the use case is for expand / collapse callbacks, so we should have two new props `onExpand` and `onCollapse` called when the expand / collapse actually occurs.

Yes, all props should be documented. If they are not, please add them 🙏

domihustinova:
>If one selects text on the details part, the onClick callback is called, but it did not close nor expand.

1. Okay, I see what you mean. If I select text outside of details part, text is selected and card is expanded/collapsed. So for the details, just to confirm - we want to keep the current behaviour of not collapsing the details on text selection and I will remove the callback calling, correct?

> we should have two new props onExpand and onCollapse

2. I wasn’t sure if we want to have two new props. If yes, in that case, do we also keep the onClick prop? To me it seems redundant and conflicting as onClick encapsulate both onExpand and onCollapse but maybe I am missing something.

3. Is my approach with context correct or I should go different direction? I tried different routes but this was the only working solution. So if it’s not good approach, I would need a hint :sweat_smile:

DSil:
> we want to keep the current behaviour of not collapsing the details on text selection and I will remove the callback calling, correct?

Yes, no change on behavior.

> To me it seems redundant and conflicting as onClick encapsulate both onExpand and onCollapse but maybe I am missing something.

I can understand your question. In this case, we want to avoid the breaking change. We can keep onClick for whatever the users are using it for and add the two new requested ones. This makes the component more flexible and does not break functionality 🙂 

> Is my approach with context correct or I should go different direction?

I would say it is ok, giving the current implementation of the component 🙂

packages/orbit-components/src/Itinerary/ItinerarySegment/ItinerarySegmentDetail/index.tsx
```diff
@@ -87,7 +87,19 @@ const ItinerarySegmentDetail = ({
...
+
+              const selection = window.getSelection();
+              const isTextSelected = selection?.toString() !== "";
+
+              if (!isTextSelected && ev.detail === 1) {
+                if (onCollapse) onCollapse(ev);
+                if (onClick) onClick(ev);
+              }
+
+              toggleOpened();
```
DSil, line 101: I think we can avoid this logic here? 🤔 

On `ItinerarySegment` we already have the `toggleOpened` method definition. Can't we call `onCollapse` and `onExpand` there? We could even avoid the drill of the props here, no?

I didn't try this approach, so feel free to correct me if I suggested it wrong

domihustinova: I will take a look 👀

domihustinova: It looks like it's working ✨ I feel stupid for going in circles here, it was just right in front of my eyes 😅

DSil (commented):
LGTM 👍 just two notes:

- Since we don't have tests for this component (😞) please add at least some action callbacks to the Segment story so we can see it in action. Something like
```jsx
<ItinerarySegment
  onClick={action("clicked")}
  onCollapse={action("collapsed")}
  onExpand={action("expanded")}
>
```
- Please arrange the commits. Changes to the context API are not worth a mention on the changelog, since it's something the user does not see. We only want to document the two new props (and I would say everything can be on the same commit)

packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
```diff
@@ -514,7 +514,12 @@ export const Stop = () => {
 
 export const Detail = () => {
```
DSil, line 515: This is the Detail story, but the props actually exist on the Segment sub-component, so I'd say it makes sense to be on the Segment story. You can still keep this one, bue please also add to the Segment one 🙏

domihustinova: Sure, it made sense to me to put it here but you are correct that it makes sense to have it on both.

PR #4402, opened 2024-07-02
chore: update TailwindCSS dependency

sarkaaa:
This PR includes changes based on the latest version of [TailwindCSS (3.4.4)](https://tailwindcss.com/blog/tailwindcss-v3-4).

Most of the changes were covered in the past, so this PR covers following:
- Extended `min-width`, `max-width`, and `min-height` scales

Other changes (like `size-XY` classes, `has-` pseudoclasses, etc..) were covered in the previous dependency update.
Grid properties are part of styled-components, so I didn't implement grid-related properties.

I used`not-last:...` as it's defined in `orbit-tailwind-preset/src/index.ts` 

**Jira ticket:** https://kiwicom.atlassian.net/browse/FEPLT-2046


packages/orbit-components/src/AirportIllustration/AirportIllustration.ct-story.tsx
```diff
@@ -9,7 +9,8 @@ export default function AirportIllustrationStory() {
   return (
     <div className="space-y-sm flex flex-col">
       {NAMES.map(illustration => (
-        <div className="bg-white-normal min-h-[80px] w-full">
+        /* eslint-disable-next-line tailwindcss/no-custom-classname */
```
sarkaaa, line 12: Hello, I had to add these disabling comments, because the ESLint is shouting that these `min-h...`, `min-w-..`,  `max-h...`, `max-w-..` classes are not TailwindCSS classes (but that's not true).
I guess it's not implemented in the Eslint Tailwind plugin 🤷‍♀️ .


**I found these "solutions":**
- add disabling rule comments everytime it's needed
- add this rule to the whitelist for `tailwindcss/no-custom-classname` in `eslintrc.cjs` file

WDYT? Or do you suggest any other solution? @DSil

sarkaaa:
Hello @DSil, I wanted to implement the functionality of [styling children with * variant](https://tailwindcss.com/blog/tailwindcss-v3-4#style-children-with-the-variant), I tried multiple implementations, but none of them was working.

Soo many classes are starting with `[&_ ...]`, but I haven't figured out how this functionality is working.

packages/orbit-components/src/AirportIllustration/AirportIllustration.ct-story.tsx
DSil: > classes are not TailwindCSS classes (but that's not true)

It is true. We want to only accept the sizes defined in the [tailwind config](https://github.com/kiwicom/orbit/blob/master/packages/orbit-tailwind-preset/src/index.ts), and the linter is prepared for that. We should **not** rely on the default tailwind sizes, at all. I am sorry we did not introduce you to that, but you will have to revert those, unfortunately. Custom sizes are meant to be kept as custom sizes (or have tokens generated for them).

packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -81,7 +82,7 @@ const ToastMessage = ({
         }}
       >
         <div
-          className="rounded-large bg-white-normal absolute left-0 top-0 z-[1] size-full animate-[toast-light_var(--toast-message-duration)_linear] opacity-10 will-change-transform rtl:animate-[toast-light-rtl_var(--toast-message-duration)_linear]"
+          className="rounded-large bg-white-normal z-default absolute left-0 top-0 size-full animate-[toast-light_var(--toast-message-duration)_linear] opacity-10 will-change-transform rtl:animate-[toast-light-rtl_var(--toast-message-duration)_linear]"
```
DSil, line 85: This one is actually a good catch, nice!

packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -56,9 +56,10 @@ const ToastMessage = ({
       }
     >
       <div
+        /* eslint-disable-next-line tailwindcss/no-custom-classname */
         className={cx(
           "rounded-large bg-ink-dark p-xs relative w-full overflow-hidden will-change-transform",
-          "lm:max-w-[360px] lm:w-auto lm:p-sm [&_svg]:min-h-[20px]",
+          "lm:max-w-modal-extra-small lm:w-auto lm:p-sm [&_svg]:min-h-5",
```
DSil, line 62: We should not use component based tokens outside of the scope of it component. In this case, it's a Modal class used on Toast. I know they have the same value, so we should probably discuss unifying so we could reuse it. But, for now, this should be kept as a custom value. Nice catch, though.

yarn.lock
DSil, line 1: Why are there changes on the lock and none on the package.json?

DSil (changes requested):

DSil:
> Soo many classes are starting with [&_ ...], but I haven't figured out how this functionality is working.

The `*` selector should only replace those that are `[&>*]` since the `*` is a new selector that targets all direct children, which is the same as doing `>*`

packages/orbit-components/src/AirportIllustration/AirportIllustration.ct-story.tsx
sarkaaa: Ok, my theory was based on the default Tailwind sizes as it's defined in the official doc. In that case, I'll revert it. 😢

DSil: We're actually missing the size classes in [our documentation](https://orbit.kiwi/foundation/tailwind/classnames/).
I didn't check, but maybe they're the same as spacing 🤔 — if they are, it should be clear. If they're not, they should be documented.

mvidalgarcia:
This PR is linked to the wrong Jira task. Please be careful when tagging.

packages/orbit-components/src/Card/index.tsx
```diff
@@ -46,7 +46,7 @@ export default function Card({
     <div
       id={id}
       className={cx(
-        "orbit-card font-base bg-white-normal [&>*]:border-elevation-flat-border-color lm:[&>*:first-child]:rounded-t-normal lm:[&>*:last-child]:rounded-b-normal w-full [&>*:first-child]:border-t",
+        "orbit-card font-base bg-white-normal *:border-elevation-flat-border-color lm:[&>*:first-child]:rounded-t-normal lm:[&>*:last-child]:rounded-b-normal w-full [&>*:first-child]:border-t",
```
DSil, line 49: Can't the other `[&>*]` selectors be replaced as well?

sarkaaa: yes, sorry, I wasn't sure about it, bcs changing pseudoclasses to `*:first-child:...` didn't work, but I have found that's it's working in this format: `*-[:first-child]...` 🤗

It should be changed in this commit (tested with `Card` component in SB). https://github.com/kiwicom/orbit/pull/4402/commits/798c9382e9843c81703db029834b69468ea639cf

@DSil

sarkaaa: @DSil 
Nope,
I tried these variants:
- `*:first-child:...` ⛔ 
image

- `[*:first-child]:...` ⛔ 
image

- `*-[:first-child]:...` ✅ 
image

- `*[:first-child]:...` ⛔ 

image

packages/orbit-components/src/Toast/ToastMessage.tsx
```diff
@@ -58,7 +58,7 @@ const ToastMessage = ({
       <div
         className={cx(
           "rounded-large bg-ink-dark p-xs relative w-full overflow-hidden will-change-transform",
-          "lm:max-w-[360px] lm:w-auto lm:p-sm [&_svg]:min-h-[20px]",
+          "lm:max-w-modal-extra-small lm:w-auto lm:p-sm [&_svg]:min-h-icon-medium",
```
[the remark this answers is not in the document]
sarkaaa: yes, I removed it, it was rewritten automatically during the commit phase. 😖 I'll fix it.

DSil (dismissed):

DSil (approved):

PR #4403, opened 2024-07-03
fix(ButtonLink): handle hover state on mobile

domihustinova:
The reported issue was caused by the "sticky hover" issue on mobile devices. We weren't handling it for ButtonLink so I changed hover styles only when hover is applicable.

Closes FEPLT-2045

### DIFF
--- a/packages/orbit-components/src/ButtonLink/index.tsx
+++ b/packages/orbit-components/src/ButtonLink/index.tsx
@@ -16,20 +16,20 @@ import type { Props, Type } from "./types";
 
 const typeStyles: Record<Type, string> = {
   primary:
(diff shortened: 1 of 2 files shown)


DSil:
I believe we should be doing [this](https://github.com/tailwindlabs/tailwindcss/discussions/1739#discussioncomment-3630717) instead.

domihustinova:
Yes, I found this flag too but I wasn't sure if we want to use it as it's global setting so I used the solution we are already using in FilterWrapper. I can use the flag then 👌

mvidalgarcia:
As we talked about offline, we'll keep the Tailwind classes with arbitrary values and add the Tailwind config commented out as a reference for the future.

PR #4406, opened 2024-07-03
chore: run workflow once the PR is created by job

sarkaaa:
This PR includes news settings of GitHub actions workflow - once the once the PR is (automatically) created then its respective GH actions are triggered.

Jira ticket: https://kiwicom.atlassian.net/browse/FEPLT-2003

### DIFF
--- a/.github/workflows/cron-update-icons.yml
+++ b/.github/workflows/cron-update-icons.yml
@@ -46,6 +46,7 @@ jobs:
         uses: actions/github-script@v6
         if: env.new_files == 'true'
         with:
+          github-token: ${{ secrets.PAT }}
           script: |
             github.rest.pulls.create({
               owner: context.repo.owner,


.github/workflows/cron-update-icons.yml
```diff
@@ -42,17 +42,25 @@ jobs:
         env:
           FIGMA_TOKEN: ${{ secrets.FIGMA_TOKEN }}
 
+      - uses: actions/checkout@v4
+
       - name: Create PR
-        uses: actions/github-script@v6
+        uses: peter-evans/create-pull-request@v6
```
DSil, line 48: We used to have peter-evans but we changed to the official github actions one. I think we should try to keep it, if it is possible 🤔

sarkaaa:
Hi, I was searching for how to achieve our goal and it seems [it's not doable to run GH workflow from another GH workflow automatically](https://github.com/peter-evans/create-pull-request/issues/48#issuecomment-536184102), I've found this solution by using GitHub apps, but I am not sure if it's the way how we want to solve this issue.

Also, I have found some solutions using Personal Access Token, but from my POV it's not ideal either. 🤷‍♀️ 

btw, an interesting discussion about this problem is here: [https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens) 

I'll try to do deeper research and in case I still won't be successful, can we discuss it next week, please?
@DSil

DSil:
> Also, I have found some solutions using Personal Access Token

That might be an option. We are using PAT for other things, like updating visual tests… So we can use the @kiwicom-orbit-bot PAT and make it the creator of the PRs, if that's what's necessary to have the pipeline run automatically.

mvidalgarcia:
This PR is not linked to any Jira task. Please bear in mind that the Kanban board acts as the source of truth. If it's not visible there it might go unnoticed.

sarkaaa:
> That might be an option. We are using PAT for other things, like updating visual tests… So we can use the @kiwicom-orbit-bot PAT and make it the creator of the PRs, if that's what's necessary to have the pipeline run automatically.

Thank you for the reply. I'll take a look at it deeper on Monday.

.github/workflows/cron-update-icons.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         uses: actions/github-script@v6
         if: env.new_files == 'true'
         with:
+          github-token: ${{ secrets.PAT}}
```
DSil, line 49:
```suggestion
          github-token: ${{ secrets.PAT }}
```

DSil (commented):
Just a minor 💅 LGTM 👍

.github/workflows/cron-update-icons.yml
sarkaaa: I wanted to change the commit message to the `chore` type, but I got this error and I don't know what to choose. Nothing seems to be suitable for GH actions. 🤔 😄  @DSil 
image

DSil (approved):

PR #4407, opened 2024-07-03
chore: update Storybook version

domihustinova:
https://kiwicom.atlassian.net/browse/FEPLT-2048


packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -38,7 +38,7 @@ const OrbitDecorator = (storyFn, context) => {
         </Text>
         {children}
         <div style={{ marginTop: 20 }} dir="ltr">
-          <Source code={jsxToString(children, options)} language="jsx" format={false} />
+          {/*<Source code={jsxToString(children, options)} language="jsx" format={false} />*/}
```
domihustinova, line 41: There is an issue with this Source component. I tried to use this workaround but I wasn't able to make it work :/ https://github.com/storybookjs/storybook/issues/22088#issuecomment-2076906244

The issue started with v8, in 7.6.17 everything was working normally.

Snímek obrazovky 2024-07-05 v 12 12 46

packages/orbit-components/package.json
```diff
@@ -90,16 +90,16 @@
...
-    "@storybook/addon-knobs": "^7.0.2",
-    "@storybook/addon-styling": "^1.3.1",
-    "@storybook/addons": "^7.0.23",
-    "@storybook/blocks": "^7.0.23",
-    "@storybook/react": "^7.0.23",
-    "@storybook/react-webpack5": "^7.0.23",
+    "@storybook/addon-a11y": "^8.1.11",
+    "@storybook/addon-essentials": "^8.1.11",
+    "@storybook/addon-knobs": "8.0.1",
+    "@storybook/addon-styling": "^1.3.7",
```
domihustinova, line 96: These two addons are deprecated, should we replaced them with their replacements?

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -38,7 +38,7 @@ const OrbitDecorator = (storyFn, context) => {
         </Text>
         {children}
         <div style={{ marginTop: 20 }} dir="ltr">
-          <Source code={jsxToString(children, options)} language="jsx" format={false} />
+          {/* <Source code={jsxToString(children, options)} language="jsx" format={false} /> */}
```
domihustinova, line 41: There is an issue with this Source component. I tried to use this workaround but I wasn't able to make it work :/ https://github.com/storybookjs/storybook/issues/22088#issuecomment-2076906244

The issue started with v8, in 7.6.17 everything was working normally.

Snímek obrazovky 2024-07-05 v 12 12 46

packages/orbit-components/.storybook/main.ts
```diff
@@ -1,28 +1,34 @@
...
     {
-      name: "@storybook/addon-essentials",
+      name: getAbsolutePath("@storybook/addon-essentials"),
       options: {
         controls: false,
       },
     },
     {
-      name: "@storybook/addon-styling",
+      name: getAbsolutePath("@storybook/addon-styling"),
```
domihustinova, line 25: I've run into another issue - `getAbsolutePath` was added by upgrade guide (but only to the kbos, a11y and pseudo states). I added it to both addon-styling and addon-essentials because ESLint was complaining ("The @storybook/addon-styling is not installed in orbit/. Did you forget to install it or is your package.json in a different location?")

But if it's kept like this, I cannot run storybook on localhost 🤔 But there seems to be no issue in staging.

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -1,13 +1,16 @@
...
+import { Highlight, themes } from "prism-react-renderer";
 
 import defaultTheme from "../src/defaultTheme";
 import OrbitProvider from "../src/OrbitProvider";
 import Heading from "../src/Heading";
 import Text from "../src/Text";
+import Tile from "../src/Tile";
 
 const OrbitDecorator = (storyFn, context) => {
+  const [isCopied, setIsCopied] = React.useState(false);
```
domihustinova, line 12: I added this simple solution for the copying functionality. Prism renderer [doesn't offer any wrapper](https://github.com/FormidableLabs/prism-react-renderer/issues/181) and for our purposes this seems enough compared to adding extra package for handling copying. And Prism renderer seems to be the best package for the syntax highlight so this seemed reasonable to me.

domihustinova:
Error in console is already reported https://github.com/storybookjs/storybook/issues/24283

packages/orbit-components/src/Box/Box.stories.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from "react";
-import { text, select, number } from "@storybook/addon-knobs";
 
 import RenderInRtl from "../utils/rtl/RenderInRtl";
 import type { SpacingToken, ColorTokens } from "./types";
 
 import Box from ".";
 
+Box.displayName = "Box";
```
domihustinova, line 8: I added explicit displayName because `react-element-to-jsx-string` has [issue with components using forwardRef](https://github.com/algolia/react-element-to-jsx-string/issues/859), it displays "No" instead of component name. It started from SB v8 and setting displayName solves it.

domihustinova, line 8: I added explicit displayName because `react-element-to-jsx-string` has issue with components using forwardRef, it displays "No" instead of component name. It started from SB v8 and setting displayName solves it.

packages/orbit-components/src/Box/Box.stories.tsx
```diff
@@ -250,10 +270,24 @@ export const Positions = () => {
   );
 };
 
-export const PaddingMargin = () => {
-  const margin = select("margin", Object.values(SPACINGS), SPACINGS.SMALL);
-  const padding = select("margin", Object.values(SPACINGS), SPACINGS.SMALL);
```
domihustinova, line 255: This was broken in SB as it was not possible to set padding, should be fixed now.

packages/orbit-components/src/Wizard/Wizard.stories.tsx
```diff
@@ -113,3 +118,43 @@ export const Playground = () => {
 Playground.story = {
   parameters: { knobs: { escapeHTML: false } },
```
domihustinova, line 119: I am stuck on this one, controls doesn't have this param yet - it was suggested to be added but the issue is still open after 4 years. Any suggestions how to approach this?

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -16,6 +19,16 @@ const OrbitDecorator = (storyFn, context) => {
...
   }, []);
 
+  const handleCopyCode = codeToCopy => {
+    navigator.clipboard.writeText(codeToCopy);
+
+    setIsCopied(true);
+
+    setTimeout(() => {
+      setIsCopied(false);
+    }, 2000);
```
sarkaaa, line 29: If I understand correctly, you set the timeout value to have a timeframe between showing to user a "tooltip" (or any other form of showing a statement) that the code was copied (?)
@domihustinova

domihustinova: Yes, to inform user that the code was copied for a given time before the original text is displayed again.

packages/orbit-components/.storybook/main.ts
```diff
@@ -1,25 +1,47 @@
 /* eslint-disable no-param-reassign */
-import path from "path";
+import path, { dirname, join } from "path";
 import type { StorybookConfig } from "@storybook/react-webpack5";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function getAbsolutePath(value: string): any {
```
DSil, line 6: We can supress the warning. This function returns string.

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -16,6 +19,16 @@ const OrbitDecorator = (storyFn, context) => {
     };
   }, []);
 
+  const handleCopyCode = codeToCopy => {
+    navigator.clipboard.writeText(codeToCopy);
```
DSil, line 23: We have a type warning here that we can avoid if we define `codeToCopy` to be of type `string`

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -1,13 +1,16 @@
...
-import { Source } from "@storybook/blocks";
+import { Highlight, themes } from "prism-react-renderer";
 
 import defaultTheme from "../src/defaultTheme";
 import OrbitProvider from "../src/OrbitProvider";
 import Heading from "../src/Heading";
 import Text from "../src/Text";
+import Tile from "../src/Tile";
 
 const OrbitDecorator = (storyFn, context) => {
```
DSil, line 11: Can we get the types of these args from storybook and avoid the warnings? :thinking:

packages/orbit-components/.storybook/preview.tsx
DSil, line 1: What is the reason for the rename? 🤔

packages/orbit-components/src/AirportIllustration/AirportIllustration.stories.tsx
```diff
@@ -34,3 +28,32 @@ Playground.story = {
     info: "Explore our new set of Airportillustrations for Kiwi.com.",
   },
 };
+
+Playground.args = {
+  size: SIZE_OPTIONS.MEDIUM,
+  name: "BGYFastTrack" as Name,
```
DSil, line 34: I don't think we need this cast here, nor the type import above

packages/orbit-components/src/Alert/Alert.stories.tsx
```diff
@@ -161,9 +186,30 @@ InlineActions.story = {
...
 };
 
-export const WithTextLink = () => {
-  const type = select("Type", Object.values(TYPE_OPTIONS), "info");
+InlineActions.args = {
+  type: TYPE_OPTIONS.INFO,
+  title: "You can change the title by changing the Title knob",
+  button: "I am a link",
+  closable: false,
+  icon: "Airplane",
```
DSil, line 194: This seems to not be working :disappointed:. I see it's broken on master anyway. Can you take a look, please? (Same for playground)

DSil (changes requested):
I have stopped the review at the Card component because I realized there is a global issue that needs to be addressed that will affect all stories, requiring global review again.

The current setup is not doing any type validation to the props we pass to the components. We are losing the Typescript features entirely, which is something we should not do.

Take as example the ButtonLink story. You'll see I left a comment stating that the `width` prop should be string and not number. If we had the stories setup with typescript this would have been spotted earlier.

What we have to have is something like:

```ts
export const Playground: StoryObj<typeof ButtonLink>
```

This will ensure the types we define in the `args` object match the correct types of the props.

Did you try this approach and disregarded for some reason? Or is it something you didn't try? If not, I think this is the way to go 🙏

packages/orbit-components/.storybook/preview.tsx
domihustinova: Oh, I changed it when I was importing Doc Blocs for the autopage. Changed back ✅

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -37,8 +51,33 @@ const OrbitDecorator = (storyFn, context) => {
...
         </Text>
         {children}
-        <div style={{ marginTop: 20 }} dir="ltr">
-          <Source code={jsxToString(children, options)} language="jsx" format={false} />
+        <div dir="ltr" className="mt-lg relative">
+          <Tile>
+            <Highlight theme={themes.vsLight} code={code} language="tsx">
+              {({ style, tokens, getLineProps, getTokenProps }) => (
+                <pre className="text-small overflow-x-scroll" style={style}>
+                  {tokens.map((line, i) => (
```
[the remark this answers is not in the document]
domihustinova: Without it, the text was spilling over. Hm, there is no difference for me between `overflow-x-scroll` and `overflow-x-auto` 🤔 But yes, `overflow-x-auto` sounds better, changed ✅ 

Snímek obrazovky 2024-07-24 v 18 22 57

packages/orbit-components/.storybook/main.ts
```diff
@@ -1,22 +1,29 @@
...
+  return dirname(require.resolve(join(value, "package.json")));
+}
+
 const config: StorybookConfig = {
   staticDirs: [path.resolve(__dirname, "../static")],
   stories: ["../src/**/*.stories.*"],
-  framework: "@storybook/react-webpack5",
+  framework: getAbsolutePath("@storybook/react-webpack5"),
```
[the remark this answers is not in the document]
mvidalgarcia: I see, thanks!

packages/orbit-components/package.json
```diff
@@ -78,6 +78,7 @@
     "@popperjs/core": "^2.9.2",
     "clsx": "^2.0.0",
     "color2k": "^2.0.3",
+    "prism-react-renderer": "^2.3.1",
```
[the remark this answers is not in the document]
domihustinova: I followed the manual in the package https://www.npmjs.com/package/prism-react-renderer#installation But I guess   devDep is OK here, changed.

packages/orbit-components/.storybook/orbitDecorator.tsx
domihustinova: I found `StoryFn` and `StoryContext` types in `@storybook/react` but tbh I am not 100% sure here.

If I add them here

```
const OrbitDecorator = (storyFn: StoryFn, context: StoryContext) => {
```

I need to change line 33 to
```
  const children = storyFn(null, context);
```
as `storyFn` requires two arguments.

I mean, there are no warnings anymore, everything works, but I didn't find usage of this "storyFn" function we have in the SB docs :( Any ideas?

domihustinova: Ok, actually, importing `Decorator` type from `@storybook/react` should be solving this without any additional changed 🤔 

```
const OrbitDecorator: Decorator = (storyFn, context) => {
```

packages/orbit-components/package.json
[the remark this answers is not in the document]
DSil: Yes, devDep is enough here, as our users don't need to install this 🙂

packages/orbit-components/src/Wizard/Wizard.stories.tsx
domihustinova: There should be no difference, removing then 👍

DSil (commented):
Just two suggestions. I feel some stories could be deleted or rewritten, but I agree with moving forward with this and tackling them later. So, for now, I'd just consider the two comments. Good work on this!

packages/orbit-components/src/Accordion/Accordion.stories.tsx
```diff
@@ -23,238 +24,237 @@ const Footer = () => {
...
-            <Text size="small">This is a header label</Text>
-          </Stack>
-        }
-      >
-        <Text type="primary">This is a content text</Text>
-      </AccordionSection>
-    </Accordion>
-  );
-};
+export const Default: Story = {
```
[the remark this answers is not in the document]
domihustinova: Agree, it indeed looks better. I will add it to control-less stories 👍

packages/orbit-components/src/Accordion/Accordion.stories.tsx
```diff
@@ -23,238 +24,237 @@ const Footer = () => {
...
+  render: () => {
+    const [expandedSection, setExpandedSection] = React.useState("0X1");
```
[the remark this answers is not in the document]
mvidalgarcia: What arrow function do you mean? Was this fixed already?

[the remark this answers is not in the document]
DSil: Yes, it's fixed. It used to be

```js
render: () => {…}
```

but if it had hooks inside (like `React.useState`) it would trigger that warning. Therefore, it needs to be 

```js
render: function Render() {…}
```

mvidalgarcia (approved):

PR #4427, opened 2024-07-25
fix: add readOnly prop

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2061


packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -49,6 +49,7 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
       maxWidth,
       onKeyDown,
       spaceAfter,
+      readOnly = false,
```
sarkaaa, line 52: I've added an optioan prop `readOnly`, default value `false`.

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -391,16 +393,19 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
         emptyState
       );
 
+    const isNotReadOnly = isLargeMobile || !readOnly;
```
sarkaaa, line 396: I added this constant to prevent `isLargeMobile || !readOnly` duplication within the whole file.

packages/orbit-components/src/InputSelect/InputSelect.ct-story.tsx
```diff
@@ -60,55 +60,73 @@ const advancedOptions = [
...
   return (
     <div className="gap-md p-md flex flex-col">
-      <InputSelect options={simpleOptions} />
-      <InputSelect options={simpleOptions} label="Custom label" placeholder="Custom placeholder" />
+      <InputSelect options={simpleOptions} readOnly />
+      <InputSelect
+        options={simpleOptions}
+        label="Custom label"
+        placeholder="Custom placeholder"
+        readOnly
```
sarkaaa, line 68: I added `readOnly` prop to all of the stories and tests, because it has `false` value by default (even for smaller devices).

packages/orbit-components/src/InputSelect/index.tsx
sarkaaa: In case the viewport is small/medium mobile, it's not `readOnly` by default (this readOnly functionality was depended only on `isLargeMobile` value

sarkaaa:
In general, I have added readOnly prop to `InputSelect` component. This prop is `false` by default and in this case, the user can write a text in it.

---

In case it's `false` by default, I had to add the `true` prop to all of the related tests that we already have. Now I'm considering to set the default to `true` to prevent BC and prevent informing all teams that already use this component. WDYT?

Or would you suggest solving this issue in any other way? 🤔

DSil (changes requested):
The described issue is precisely the `InputSelect` being `readOnly` on mobile. That is not expected, but I see it still happens here. I also don't understand why forcing it to be `readOnly` in all stories. The idea is that it is `readOnly` **only** when the `readOnly` prop is actually passed to the component.

sarkaaa:
> The described issue is precisely the InputSelect being readOnly on mobile. That is not expected, but I see it still happens here.

I don't understand, the `readOnly` prop is set (with `false` default value), so the component shouldn't be `readOnly=true` for mobile views by default now.


> I also don't understand why forcing it to be readOnly in all stories. The idea is that it is readOnly only when the readOnly prop is actually passed to the component.

I set the readOnly prop as a knob.

---

- I updated visual tests, because screenshots have been changed by forcing `readOnly=true`.
- I still have to update unit test, it's failing in case that `readOnly` value is not set.

packages/orbit-components/src/InputSelect/index.tsx
DSil, line 52: We don't need a default value if the prop is boolean and optional

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -204,27 +205,33 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
         error={isLargeMobile && error}
         label={isLargeMobile && label}
         disabled={disabled}
-        onFocus={handleFocus}
+        onFocus={!readOnly ? handleFocus : undefined}
         onBlur={handleBlur}
-        onChange={handleInputChange}
+        onChange={!readOnly ? handleInputChange : undefined}
```
DSil, line 210: Do we need this? A readOnly input should not have `onChange` 🤔 could you try it?

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -204,27 +205,33 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
...
         id={inputId}
         placeholder={placeholder}
-        autoFocus={!isLargeMobile}
+        autoFocus={!readOnly && !isLargeMobile}
```
DSil, line 213: Why this condition here?

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -204,27 +205,33 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
         error={isLargeMobile && error}
         label={isLargeMobile && label}
         disabled={disabled}
-        onFocus={handleFocus}
+        onFocus={!readOnly ? handleFocus : undefined}
```
DSil, line 208: Thinking about this, a `readOnly` input is focusable. So we still need to call `handleFocus` here, because there might be some `onFocus` defined. What we need to do is prevent the options to be shown.

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -436,8 +443,8 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
               label={label}
               help={help}
               error={error}
-              onFocus={() => setIsOpened(true)}
-              readOnly
+              onFocus={!readOnly ? () => setIsOpened(true) : undefined}
```
DSil, line 446: This is correct, but we still need to call `onFocus` if it exists.
Basically, by calling `setIsOpened` it would open the modal and the modal would autofocus the inner input, calling `onFocus` if it exists. But in this scenario we don't open the modal. However, we still need to call the `onFocus` if it's defined.

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -436,8 +443,8 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
...
+              readOnly={readOnly}
```
DSil, line 447: As we discussed IRL, on mobile, this is **ALWAYS** `readOnly`, because it is not really an input, but a trigger to the modal that has the actual input.

packages/orbit-components/src/InputField/index.tsx
```diff
@@ -246,6 +242,7 @@ const InputField = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
             "[&::placeholder]:text-form-element-foreground",
             "[&::-ms-input-placeholder]:text-form-element-foreground",
             "[&::-ms-clear]:hidden [&::-ms-reveal]:hidden",
+            readOnly && "cursor-default",
```
DSil, line 245: Why this?

packages/orbit-components/src/InputField/index.tsx
DSil, line 1: When we write commit messages, we need to be mindful of what the users will read. In this case, we are editing the TW classes that we use, but that is not what is important for the user. What is important for the user is that the `InputField` styles will be different (in this case fixed) for when it is `readOnly`.

My suggestion for this commit message would be:
```
fix(InputField): adjust readOnly styles
```

or something similar

packages/orbit-components/src/InputSelect/index.tsx
DSil, line 1: When we have a `fix` commit, we typically don't start with `add`. They're usually associated with a `feat`. In this specific case, the `readOnly` prop was already accepted, it just barely did anything. And we are fixing it. So this commit should, indeed, be a fix, but the message should be slightly different. I'd suggest something like

```
fix(InputSelect): adjust style and behavior when readOnly is true
```

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
     expect(screen.getAllByRole("option")).toHaveLength(1 + 2 + 1 + 4); // (1 previously selected, 2 asian, 1 american, 4 all)
   });
 
+  it("when readOnly is true", async () => {
```
DSil, line 219: `when` statements are usually associated with `describe`, and not `it`. To avoid more nesting, this can be simply

```
it("can be readOnly",
```

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
...
+    render(
```
DSil, line 220: Please add an onFocus call and assert it's still called even when readOnly

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
...
+    );
+
+    await user.tab();
+
+    const input = screen.getByRole("textbox");
+    const dropdown = screen.queryByRole("listbox");
+
+    expect(input).toHaveValue(jetLiOption.title);
+
+    // Check if dropdown is not shown on a click
```
DSil, line 238: Before this, you did a `user.tab()`, so the input would be focused, the open even to be triggered already. So I would assert that as well, before clicking

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
...
+        options={options}
+        name={name}
+        defaultSelected={jetLiOption}
+        readOnly
+      />,
+    );
+
+    await user.tab();
+
+    const input = screen.getByRole("textbox");
```
DSil, line 233: I feel this should be done before the tab

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
...
+    fireEvent.click(input);
+    expect(dropdown).not.toBeInTheDocument();
+
+    // Check if dropdown is not shown on a keyDown
+    fireEvent.keyDown(input, { key: "ArrowDown" });
+    expect(dropdown).not.toBeInTheDocument();
+
+    // Check if dropdown is not shown on a tab
+    fireEvent.keyDown(input, { key: "Tab" });
+    expect(dropdown).not.toBeInTheDocument();
```
DSil, line 248: We can assert this at the beginning and remove it here.

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,50 @@ describe("InputSelect", () => {
...
+
+    // Check if dropdown is not shown on a focus
+    fireEvent.focus(input);
+    expect(dropdown).not.toBeInTheDocument();
+
+    // Check if dropdown is not shown on an enter
+    fireEvent.keyDown(input, { key: "Enter" });
+    expect(dropdown).not.toBeInTheDocument();
+
```
DSil, line 257: This one does not seem necessary 🤔

packages/orbit-components/src/InputSelect/InputSelect.ct-story.tsx
```diff
@@ -68,6 +68,12 @@ export function InputSelectStory() {
         placeholder="Custom Placeholder"
         disabled
       />
+      <InputSelect
+        options={simpleOptions}
+        label="Custom label (readOnly)"
+        placeholder="Custom Placeholder (readOnly)"
```
DSil, line 74: Instead of a placeholder, please add a `defaultSelected` so we can assert the correct logic of rendering the `X` button

packages/orbit-components/src/InputSelect/index.tsx
sarkaaa: I will check it and remove if doable.

packages/orbit-components/src/InputSelect/InputSelect.ct-story.tsx
```diff
@@ -68,6 +68,12 @@ export function InputSelectStory() {
         placeholder="Custom Placeholder"
         disabled
       />
+      <InputSelect
+        options={simpleOptions}
+        label="Custom label (readOnly)"
+        defaultSelected={simpleOptions[0]}
```
sarkaaa, line 74: I set `defaultSelected` value instead of a `placeholder`. Now it makes more sense related to `readOnly` prop (not disabled).  🙂

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,46 @@ describe("InputSelect", () => {
...
+      />,
+    );
+
+    const input = screen.getByRole("textbox");
+    const dropdown = screen.queryByRole("listbox");
+
+    expect(input).toHaveValue(jetLiOption.title);
+
+    await user.tab();
+    expect(onFocus).toHaveBeenCalled();
```
sarkaaa, line 240: I call this assertion for every of event (fireEvent, user.tab, user.type)

packages/orbit-components/src/InputSelect/index.tsx
```diff
@@ -212,19 +215,25 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
         autoFocus={!isLargeMobile}
         role="combobox"
         value={inputValue}
-        onKeyDown={ev => {
-          if (onKeyDown) onKeyDown(ev);
-          handleDropdownKey(ev);
-        }}
+        onKeyDown={
```
sarkaaa, line 218: I was thinking for a while if it's not the same flow as we have with `onFocus`. If the `onFocus` prop is passed to `InputSelect` component, then we pass it to `InputField`. But that's applicable only when `readOnly` is false, it doesn't make sense to have `onKeyDown` available in `readOnly` mode. Am I correct?

@DSil

sarkaaa:
Hi @DSil 
I made some changes, I hope I haven't forgotten anything. Please check it out. 🤞

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
sarkaaa: I removed tests for events "Enter" and "onKeyDown". I think they can't be triggered from the mobile view (?).

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
```diff
@@ -216,6 +216,46 @@ describe("InputSelect", () => {
...
+    expect(dropdown).not.toBeInTheDocument();
+
+    // Check if dropdown is not shown on a focus
+    fireEvent.focus(input);
+    expect(onFocus).toHaveBeenCalled();
+    expect(dropdown).not.toBeInTheDocument();
+
+    // Assert that input field still includes the same value
+    await user.type(input, "Arnold");
+    expect(onFocus).toHaveBeenCalled();
```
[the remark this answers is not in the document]
sarkaaa: I went with 2nd option, I would like to ensure that typing in readOnly prop is not possible and the value is still the same.

DSil (approved):

packages/orbit-components/src/InputSelect/__tests__/index.test.tsx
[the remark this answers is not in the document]
DSil: That is ok, but

>I would like to ensure that typing in readOnly prop is not possible

This is not our responsibility in the first place. We should not be testing native behaviors 😄

PR #4401, opened 2024-06-28
Add new tokens for border radius and Heading

RobinCsl:
FEPLT-2050

## What remains to be done:

- [x] enriching the transforms/tokens-v16.mjs script to capture all possible cases for the corresponding classnames (e.g. rounded-tl-normal is missing from the list, etc.)
- [x] fix (maybe local only?) issue around new `rounded-50` classes not being part of the `packages/orbit-components/lib/tailwind.css` generated file for some reason (might be a step I missed);
- [x] improving/augmenting/completing the migration guide
- [x] updating the visual snapshots

## Commits:

**docs: add changelog draft for v16**

---

**feat: update Tailwind classes for new border radius tokens**

through the new transform.

Changes generated by running: 'node ../transforms/tokens-v16.mjs' from
the 'src' folder in packages/orbit-components.

BREAKING CHANGE: All the classes using normal border radius will now see
a 1px larger border radius through the Orbit Design Token update and
mapping between old tokens and new.

---

**feat(tokens): add new tokens for border-radius**

Slack: https://skypicker.slack.com/archives/C7T7QG7M5/p1719493641111859

---

**chore(tokens): allow numbers as keys in tokens**

The change ensures we're correctly transforming the dot notation to access
object properties when numbers are used as keys:
'foundations.borderRadius.50' is incorrect JavaScript;
'foundations.borderRadius["50"]' is valid JavaScript.

---

PR #4428, opened 2024-07-25
Limit prose action to 300 changed files

DSil:
There is a limitation on the action that can't be executed when the PR has more than 300 files. Limiting this so the job does not fail in bigger PRs.

### DIFF
--- a/.github/workflows/lint-prose.yml
+++ b/.github/workflows/lint-prose.yml
@@ -16,6 +16,8 @@ on:
 jobs:
   prose:
     runs-on: ubuntu-latest
+    # https://github.com/reviewdog/reviewdog/issues/1696
+    if: ${{ github.event.pull_request.changed_files < 301 }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4


.github/workflows/lint-prose.yml
```diff
@@ -16,6 +16,7 @@ on:
 jobs:
   prose:
     runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.changed_files < 301 }}
```
sarkaaa, line 19: Why do we want to limit this action? Is it because of saving the number of action executions? Or does it take too long to run this job with bigger amount of changed files?

DSil: We want to limit because Vale uses Reviewdog in a version that can't handle more than 300 changed filed ([issue](https://github.com/reviewdog/reviewdog/issues/1696)).

sarkaaa: Ok, I get it know. Thx.

mvidalgarcia: Could we add a comment or something? Without context, it would be hard to understand why we're doing this.

mvidalgarcia: Do we have any examples of this? How did you find out?

DSil: I found out because #4401 is failing so I investigated and came across multiple people reporting this. I will add a comment

mvidalgarcia (approved):

PR #4430, opened 2024-07-29
Add & Update Typography tokens

DSil:
Opened against #4401 so should be merged into it or to `master` after the parent is merged.

This updates the values as described in Figma and adds the new type.

The new `title0` type corresponds to the previously existing `title1` styles, that were updated.

This includes two breaking changes: deprecated heading tokens are removed and on the Heading component the `type="title1"` should now be `type="title0"`.

Linux visual tests will be updated once the PR is ready to merge

FEPLT-2062


docs/src/documentation/02-foundation/10-typography/01-general.mdx
```diff
@@ -21,22 +21,24 @@ To stand out from the page, we use <InlineToken name="colorHeading" /> as the **
...
+| Title 5          | <InlineToken name="headingTitle5FontSize" />          |
+| Title 6          | <InlineToken name="headingTitle6FontSize" />          |
 
 ### Basic guidelines for headings
 
 <Guideline type="do" title="Be respectful of the content hierarchy">
 
 Try not to skip heading levels.
-If you've used Title 1, the next level used, if intentionally smaller, should be Title 2.
+If you've used level 1, the next level used, if intentionally smaller, should be level 2.
```
mvidalgarcia, line 41: Not sure if we should use level 0 instead. I guess one can get the point in any case.

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
```diff
@@ -12,13 +12,49 @@ With this guide, we aim to walk through all the breaking changes and how they ca
...
-## Token and tailwind classes
+## Breaking changes
+
+### Heading component
+
+A new heading token was added: `title0`. This token is similar to the previously existing `title1` token. Therefore, `title1` token was updated to have new font size and line height values.
+
+With that, the `Heading` component now accepts a new value for the `type` prop: `"title0"`.
+
+Because of that, all occurrences of `Heading` with `type="title1"` must be replaced with `type="title0"`, to avoid the visual change.
```
mvidalgarcia, line 23: WDYT about making **must be replaced** in bold and include "visual breaking change" in the end?

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
```diff
@@ -12,13 +12,49 @@ With this guide, we aim to walk through all the breaking changes and how they ca
...
+#### Heading
+
+Some deprecated Heading tokens were removed. They have been deprecated since the first Tailwind release and were now finally removed. The equivalent new tokens are:
+
+- `fontSizeHeading<TYPE>` -> `heading<TYPE>FontSize`
+- `lineHeightHeading<TYPE>` -> `heading<TYPE>LineHeight`
+- `fontWeightHeading<TYPE>` -> `heading<TYPE>FontWeight`
 
-Four border radius tokens were removed and six new ones were added.
+Where `<TYPE>` can be `Display`, `DisplaySubtitle`, `Title1`, `Title2`, `Title3`, `Title4`, or `Title5`.
```
mvidalgarcia, line 49: Also `Title0`, right?

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -11,11 +11,41 @@ const transformations = [
   { from: "borderRadiusNormal", to: "borderRadius100" },
   { from: "borderRadiusLarge", to: "borderRadius150" },
   { from: "borderRadiusCircle", to: "borderRadiusFull" },
-  // Add more transformations as needed
 ];
 
+const headingTransformations = [
+  { from: "fontSizeHeadingDisplay", to: "headingDisplayFontSize" },
+  { from: "fontSizeHeadingDisplaySubtitle", to: "headingDisplaySubtitleFontSize" },
+  { from: "fontSizeHeadingTitle1", to: "headingTitle1FontSize" },
```
mvidalgarcia, line 19: I wonder if this should use Title0 as the replacement already? 🤔

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
sarkaaa: and also `Title6` (?)

packages/orbit-components/src/Heading/Heading.stories.tsx
```diff
@@ -56,6 +56,22 @@ TitleDisplaySubtitle.story = {
   },
```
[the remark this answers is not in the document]
DSil: Well spotted. Thanks

packages/orbit-design-tokens/src/js/createTokens.ts
```diff
@@ -681,6 +654,8 @@ export interface Tokens {
   tabBundleTopBackground: string;
   tabBundleTopBackgroundHover: string;
   tabBundleTopForeground: string;
+  fontWeightTableHead: string;
+  fontWeightLinks: string;
```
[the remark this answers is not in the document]
DSil: Yes, I only touched Heading related tokens. But I honestly can't wait to get rid of all of these 🫣

docs/src/documentation/02-foundation/10-typography/01-general.mdx
DSil: I think `level` is actually the correct terminology. We can have any type we want, in any order, but the level (h1 > h2 > h3) is the one that should be respected. While the prop `type` refers to style only, the prop `level` refers to the actual heading level, and that is where the order matters, imo

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
DSil: Agreed. Done

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
DSil: `Title6` is missing, yes. Regarding `Title0`… it did not exist before, that's why I did not mention it. And on the section we already have the mention to the new title1 => title 0 relation.

So I will add `Title6`, but do you think I should add another mention to `Title1` now being `Title0`?

packages/orbit-components/transforms/tokens-v16.mjs
DSil: We should. Thank you

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-font-size.json
```diff
@@ -13,16 +13,22 @@
...
+      "title-0": {
         "font-size": {
           "type": "size",
           "value": 28
         }
       },
+      "title-1": {
+        "font-size": {
+          "type": "size",
+          "value": 24
```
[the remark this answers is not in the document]
DSil: You mean because values change? I don't know, but I don't think it is. I feel the tokens are the abstraction behind values and a change to those values should not result in a breaking change (if we don't consider visual breaking changes). In the past we've been changing values of already existing tokens and not mark them as BC. Or am I missing a detail here?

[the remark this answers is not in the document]
mvidalgarcia: Yes, I was referring to the value change uniquely. But if you already had this practice in the past I guess it's OK. We have other related BCs in any case.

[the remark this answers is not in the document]
mvidalgarcia: > if we don't consider visual breaking changes

I guess that's the key.

Nothing will break with the value change but there might be unexpected visual changes.

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
mvidalgarcia: Ok, I see now you're only talking about equivalences to the old tokens. So we can omit `Title0` then.

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
```diff
@@ -34,6 +34,20 @@ Because of that, all occurrences of `Heading` with `type="title1"` **must be rep
 <Heading type="title0">Title 0</Heading>
 ```
 
+The Tailwind classes should also be updated. So if you are using tailwind classes that refer to `title1` in your components, be sure to update them to use `title0` in order to keep the same styles.
+
+**Before:**
+
+```jsx
+<p className="leading-heading-title1 text-heading-title1 font-heading-title1">Title 1</p>
```
mvidalgarcia, line 42: Are we considering these on the codemod? And should we? 💭

packages/orbit-design-tokens/src/dictionary/definitions/component/heading/heading-font-size.json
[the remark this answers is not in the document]
DSil: Yes, we don't have a consistent procedure for visual breaking changes… In this case, like you say, we will thoroughly mention this in the consequent breaking changes and affected components, so I feel it's ok to leave it as is

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
DSil: We're not, but I guess we can 🤔 will add

DSil: It even detected one place where the change was needed, so great! Thank you

PR #4433, opened 2024-07-29
Fix SegmentedSwitch border radius in RTL

domihustinova:
- replaced setting values for border radius as top/bottom/left/right radius with values setting start/end radius
- added RTL story
- updated snapshots

Closes https://kiwicom.atlassian.net/browse/FEPLT-2067


DSil (approved):

sarkaaa:
just my curiosity, maybe I didn't catch something.. 
I see that the selected button has a radius on the "inner side" of the component. Shouldn't there be `none` radius? 🤔 
https://www.figma.com/design/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=7798-3163&t=nvN7EQoKe8y05fxt-0

PR #4434, opened 2024-07-30
Fix padding on Tag in RTL

domihustinova:
- replaced right padding with padding end
- added RTL visual tests as there were none
- added scenario with closable tag with icon as it was missing

Closes https://kiwicom.atlassian.net/browse/FEPLT-2069


packages/orbit-components/src/Tag/index.tsx
```diff
@@ -89,7 +89,7 @@ const Tag = React.forwardRef<HTMLDivElement, Props>(
         onKeyDown={ev => buttonClickEmulation(ev, onClick)}
       >
         {iconLeft && (
-          <div className="pr-xs [&_svg]:size-icon-small flex flex-row items-center justify-center">
+          <div className="pe-xs [&_svg]:size-icon-small flex flex-row items-center justify-center">
```
sarkaaa, line 92: 👍

sarkaaa (approved):

mvidalgarcia:
As this only affected the RTL version visually, I'd adjust the commit title to reflect this so that it would be more accurate for the consumers.

mvidalgarcia (approved):

PR #4435, opened 2024-07-31
fix(Alert): adjust styles when closable is true and inlineActions is set

sarkaaa:

### For new components:

- [x] Unit and visual regression tests have been added/adjusted for my new feature
- [x] New Components have `d.ts` files and are exported in `index.d.ts`

Closes https://kiwicom.atlassian.net/browse/FEPLT-2068


sarkaaa:
I was also thinking about extending this PR by adding visual tests? I've checked respective `ct-story` file and it looks it covers only `AlertButton` component. WDYT? 🤔

packages/orbit-components/src/Alert/index.tsx
```diff
@@ -159,6 +159,7 @@ const Alert = (props: Props) => {
           title && inlineActions ? "flex-row" : "flex-col",
           !title && "items-center",
           inlineActions && "justify-between",
+          inlineActions && closable && "mr-md",
```
DSil, line 162: This won't work well in RTL. Also, I'm wondering if we can't refactor the current _hack_ we are using for positioning the button (it being absolute with translates is 🤯 )… Can you take a look at that?

And yes, feel free to add more visual tests, for sure!

sarkaaa: > This won't work well in RTL.

Should be fixed now.

> Also, I'm wondering if we can't refactor the current hack we are using for positioning the button (it being absolute with translates is 🤯 )… Can you take a look at that?

Sure, I have some ideas. On it. 💻 


> And yes, feel free to add more visual tests, for sure!

OK. 👍

sarkaaa: Change with `absolute` position removal - https://github.com/kiwicom/orbit/pull/4435/commits/deba6bde63c532427ec4635c77c430f67e2735f9
Please, ignore the `min-h-icon-medium` classname changes.

image

- checked with RTL version ✅ 

image

- checked without button and multiline content ✅ 
image

- mobile view ✅ 
image

DSil:
**Before:**

image

**Now:**

image (1)

This is not expected. Both the icon and the close button need to stay on top, aligned with the first line of the title

sarkaaa:
I have fixed the position of icon & cancel button.
image


For `inlineAlert`:
- the icon (robot) is centered by default (on master branch). I can change the position to the top, and for cancel button, I would center this button together with the inline button. My suggestion:
image

WDYT?

(before - master branch):
image




**Additional note:**
I have found that the button is shrinked when the title/text is multiline. I'll include the fix in this MR as well.
image

packages/orbit-components/src/Alert/index.tsx
```diff
@@ -145,7 +145,7 @@ const Alert = (props: Props) => {
         <div
           className={cx(
             "me-xs m-0 shrink-0 leading-none",
-            inlineActions && "flex items-center",
+            inlineActions && "lm:mt-[6px] flex items-center self-baseline", // TODO: [6px] can be replaced by space tokens
```
sarkaaa, line 148: I added a comment for the future steps once space tokens are implemented.

packages/orbit-components/src/Alert/index.tsx
```diff
@@ -164,8 +164,9 @@ const Alert = (props: Props) => {
         {title && (
           <div
             className={cx(
-              "text-ink-dark flex min-h-[20px] items-center font-bold",
+              "text-ink-dark min-h-icon-medium flex items-center font-bold",
```
DSil, line 167: Please revert this icon class 🥲

packages/orbit-components/src/Alert/index.tsx
```diff
@@ -74,7 +74,7 @@ const ContentWrapper = ({
   return (
     <div
       className={cx(
-        "flex min-h-[20px] items-center",
+        "min-h-icon-medium flex items-center",
```
DSil, line 77: Same as above

DSil (commented):
LGTM. You can squash the last two commits into one. And please change the mentioned classes 🙏

packages/orbit-components/src/Alert/index.tsx
sarkaaa: Fixed.

packages/orbit-components/src/Alert/index.tsx
sarkaaa: Fixed.

DSil (approved):

PR #4437, opened 2024-08-02
chore: migrate Checkbox story from CSF2 to CSF3

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/jira/software/c/projects/FEPLT/boards/404?selectedIssue=FEPLT-2072

### DIFF
--- a/packages/orbit-components/src/Box/Box.stories.tsx
+++ b/packages/orbit-components/src/Box/Box.stories.tsx
@@ -1,11 +1,10 @@
 import * as React from "react";
+import type { Meta, StoryObj } from "@storybook/react/";
 
 import RenderInRtl from "../utils/rtl/RenderInRtl";
-import type { SpacingToken, ColorTokens } from "./types";
+import type { ColorTokens } from "./types";
 
 import Box from ".";
-
-Box.displayName = "Box";
 
 const DEFAULT_COLOR = "blueDark";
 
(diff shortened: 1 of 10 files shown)


packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -30,21 +56,17 @@ export const Default = ({ label }) => {
...
     </Collapse>
-  );
-};
+  ),
 
-Default.story = {
   parameters: {
-    info: "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
+    controls: {
+      exclude: ["initialExpanded", "expanded"],
```
domihustinova, line 63: Is there a reason why these are excluded in all the stories? Can't we have them in controls? There is no playground 🤷

packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -85,25 +113,20 @@ export const OpenedByDefault = ({ label }) => {
...
   },
 };
 
-OpenedByDefault.args = {
-  label: "Duration",
-};
-
-export const WithActions = ({ label }) => {
-  return (
+export const WithActions: Story = {
```
domihustinova, line 126: There is `<No Display Name` in the source instead of real component name, can you fix it, please?

domihustinova:
I started reviewing the changes via commits but as I went, the issues started repeating so I didn't added any comments to some files as suggestions are covered by the "overall suggestions" I wrote.

sarkaaa:
I have several questions that I am not sure about:

1.  Is there any way how to make repeated stories simpler? Example: I have three stories with different args and with `onChange` prop:
```
export const Default: Story = {
  render: args => <Checkbox {...args} onChange={action("onChange")} />,
  args: { ... }
 ...
```
and I would like to change it to something like this:
```
export const Default: Story = {
  args: { ... }
 ...
```
is it possible in case I have this `onChange` prop? 🤔  [related file](https://github.com/kiwicom/orbit/blob/4220e0c8e1735e60552aa26c760c81c0874eab4c/packages/orbit-components/src/Checkbox/Checkbox.stories.tsx)

2. [Some Stories for Box component](https://github.com/kiwicom/orbit/blob/ef5019831761d75b93b0aff9573d4f385c0e66ee/packages/orbit-components/src/Box/Box.stories.tsx) includes inner static text (no `children`), but `children` prop is available. Can I exclude this `children` prop? I understand that we would like to include all of the args that can be set to specific component, but I am not sure about this specific `children` prop. Also, some of the stories look almost the same as we allow to set all of the arga (props) that can be set to this component. I'm getting little bit confused. :D 

3. Card stories include three stories (Default, Accessbility, Playground) look same. I don't get the reason to have all of them available. Isn't just `Playground` story enough? [File here](https://github.com/kiwicom/orbit/blob/8f30bb84c457f74bb3dd55d6a9d24d1a3c23505f/packages/orbit-components/src/Card/Card.stories.tsx).

4. I have noticed that some of the components have much more props that we set in Storybook (my changes are based on the previous stories settings). Should I include all of the props to all of the stories that I'm working on?

5. For example, Checkbox component has `onChange` function but in storybook nothing happens once the component is clicked (action is called, but the visual change is not done). Is it expected this on Storybook or should I "fix" it to the state that the checkbox component changes to "checked" state also visually?

In general, I think that smaller components like `CarrierLogo`, `ButtonGroup`, `ChoiceGroup`, `CountryFlag`, `Checkbox` could be ready for review.

packages/orbit-components/src/CarrierLogo/CarrierLogo.stories.tsx
```diff
@@ -1,162 +1,104 @@
...
-  return <CarrierLogo carriers={carriers} />;
-};
 
-NonExistingCarriers.story = {
-  name: "Non-existing carriers",
-
-  parameters: {
-    info: "Carrier logo displays one or more logos of transport carriers. It is usually together with the travel itinerary. Visit Orbit.Kiwi for more detailed guidelines.",
+  args: {
+    ...FourCarriers.args,
```
domihustinova, line 62: Nice 👍 I was hoping we could take advantage of this somewhere.

packages/orbit-components/src/ChoiceGroup/ChoiceGroup.stories.tsx
```diff
@@ -10,103 +11,94 @@ import ChoiceGroup from ".";
...
-
-Default.args = {
-  label: "What was the reason for your cancellation?",
+  ),
 };
 
-export const Multiple = ({ label }) => {
-  return (
-    <ChoiceGroup label={label} onChange={action("onChange")}>
+export const Multiple: Story = {
```
domihustinova, line 61: Hm, when I set `filter = true`, I still don't see `onlySelectionText`, same for WithError story. In the docs it's written it's for using `onOnlySelection` callback which is not in those stories, so maybe it can be excluded? Or callback added :D

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -8,66 +9,56 @@ import Tooltip from "../Tooltip";
...
-  checked: false,
+  args: {
+    value: "value",
+    label: "Label",
+    checked: false,
+    hasError: false,
+    disabled: false,
+    readOnly: false,
+    name: "name",
+    onChange: action("onChange"),
```
domihustinova, line 28: Same as in ChoiceGroup.

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -78,108 +69,77 @@ export const WithTextLinkInLabel = ({ checked }) => {
...
-      label="Direct"
-      value="one"
-      disabled
-      tooltip={
-        <Tooltip content="There are no results available with this option" placement="top" />
-      }
-    />
-  );
-};
+export const WithTooltip: Story = {
```
domihustinova, line 87: When I checked this one with tooltip, the ✔️ is not there, do we have a bug here? 🙀

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -78,108 +69,77 @@ export const WithTextLinkInLabel = ({ checked }) => {
...
-  label: "Label",
-  value: "value",
-  checked: false,
-  disabled: false,
-  hasError: false,
-  readOnly: false,
-  info: "Additional information for this choice",
-  dataTest: "test",
-  name: "name",
+export const Playground: Story = {
```
domihustinova, line 122: `info` arg is missing in this story.

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -78,108 +69,77 @@ export const WithTextLinkInLabel = ({ checked }) => {
...
+export const ReadOnly: Story = {
+  args: {
+    readOnly: true,
+  },
 
-Playground.story = {
   parameters: {
-    info: "Playground of Checkbox",
+    control: {
+      exclude: ["onChange"],
```
domihustinova, line 117: Hm, it's strange that you excluded this but it's still in the controls there 😕

domihustinova (commented):
I reviewed only those 5 stories you mentioned: CarrierLogo, ButtonGroup, ChoiceGroup, CountryFlag, Checkbox.

domihustinova:
Regarding your questions:
3. It’s true that if we include controls everywhere, the stories are almost identical :thinking: To me it makes sense to include the Default story with just the bare minimum, when there is only few required props in the component so it really is „default“ and you need to set only those required ones. And maybe even remove other controls from there. And in the other stories we can show off different options (one at a time) and keep the controls there to be changed. And in the Playground have all the props and show off multiple things at once. I hope it make sense :sweat_smile:
4. It would make sense to me to include all available props at least in the Playground :thinking:

DSil:
Regarding the questions:
2. Basically what Domi said. If it's too obvious that stories are repeated, we can suppress some. But the main scope of these tasks is to do a direct migration. So I'd say to not spend so much time considering the options. For obvious scenarios, feel free to change. Regarding `Box`, the children prop sometimes is useful to test how the component reacts to bigger content (does it grow, does it wrap, etc…). But in this case you either add the dynamic children prop on the story or you remove it from controls. Up to you 🙂 
3. Accessibility can be removed. Default should be default basic version. Playground should have more controls than Default.
4. What Domi said.
5. No need to worry about it, as that is not an out of the box behavior of the component.

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
sarkaaa: Typo, it should be `controls` instead of `control`. 🙈

sarkaaa:
Changes are done. For Default stories, I set the bare minimum, Playground stories show the all available options for respective component.

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -78,110 +81,80 @@ export const WithTextLinkInLabel = ({ checked }) => {
...
-);
-
-Rtl.story = {
-  name: "RTL",
+export const Rtl: Story = {
+  render: args => (
+    <RenderInRtl>
+      <Checkbox
+        {...args}
+        label={
```
domihustinova, line 140: Hm, is the same here with label as I previously mentioned in different story. Maybe let's disable it also here?

packages/orbit-components/src/Card/Card.stories.tsx
```diff
@@ -12,113 +13,118 @@ import { ELEMENT_OPTIONS } from "../Heading/consts";
...
+
+export const Default: Story = {
+  render: args => <Card {...args} />,
+  args: {
+    sectionTitle: undefined,
+    sectionDescription: undefined,
+  },
+  parameters: {
+    controls: {
+      exclude: ["labelClose", "sectionTitle", "sectionDescription"],
```
domihustinova, line 67: Hm, shouldn't we also exclude expanded, initialExpanded, dataA11ySection and loading? I don't think they should be in the Default story, no?

packages/orbit-components/src/Card/Card.stories.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
```
domihustinova, line 1: Can you please take a look at all stories in this file? 🙏 I don't think we need `dataA11ySection` control in every story (maybe only in Playground?), there are also `sectionTitle` and `sectionDescription` in stories without sections, there is also `expanded` and `initialExpanded` on non-expandable stories and missing on the expandable ones, etc.

packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -228,17 +216,31 @@ export const Rtl = ({ label }) => {
...
+  ),
+
+  args: {
+    ...WithActions.args,
+  },
 };
 
-Rtl.story = {
-  name: "RTL",
+export const Playground: Story = {
```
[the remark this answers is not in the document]
sarkaaa: Yes, the reason why I added `Playground` story is to have consistency.

customLabel - I'll add ✅

packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -11,13 +12,32 @@ import RenderInRtl from "../utils/rtl/RenderInRtl";
 
 import Collapse from ".";
 
-export default {
+type CollapsePropsAndCustomArgs = React.ComponentProps<typeof Collapse> & {
+  label?: string;
+  expanded?: boolean;
```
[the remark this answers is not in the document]
sarkaaa: good point. 👍

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
sarkaaa: good point. 👍

domihustinova:
LGTM now, great job 💪 🚀

packages/orbit-components/src/CarrierLogo/CarrierLogo.stories.tsx
```diff
@@ -1,162 +1,104 @@
...
-  size: SIZE_OPTIONS.LARGE,
-  dataTest: "test",
-  carriers: [{ code: "FR", name: "Ryanair" }],
-};
-
-OneCarrier.argTypes = {
-  size: {
-    options: Object.values(SIZE_OPTIONS),
-    control: { type: "select" },
+  args: {
```
mvidalgarcia, line 17: Cannot make them `rounded`?

packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -116,25 +127,17 @@ export const WithActions = ({ label }) => {
...
   },
 };
 
-WithActions.args = {
-  label: "Transportation",
-};
-
-export const MultipleCollapses = ({ label }) => {
-  return (
+export const MultipleCollapses: Story = {
```
mvidalgarcia, line 137: I can't interact with this story in the same way I do with this [one](https://kiwicom.github.io/orbit/?path=/story/collapse--multiple-collapses).

packages/orbit-components/src/Box/Box.stories.tsx
```diff
@@ -174,565 +177,242 @@ enum SPACINGS {
...
-    >
-      <div style={{ height: 50, width: 50, background: "#ccc" }} />
-      <div style={{ height: 50, width: 50, background: "#ccc" }} />
+export const Flex: Story = {
+  render: args => (
+    <Box {...args}>
+      <div style={{ height: 50, width: 50, background: "#14b732" }} />
+      <div style={{ height: 50, width: 50, background: "#ff04c0" }} />
+      <div style={{ height: 50, width: 50, background: "#f2ff04" }} />
     </Box>
```
mvidalgarcia, line 387: Nice! 💅

packages/orbit-components/src/Collapse/Collapse.stories.tsx
sarkaaa: The interaction is "blocked" by added switches in controls. From my observation, it's not possible to have switches for expanding and also have `expanded` function available by clicking on the arrow button.

packages/orbit-components/src/CarrierLogo/CarrierLogo.stories.tsx
sarkaaa: Good point, thx. It should be fixed now. ✅

packages/orbit-components/src/Collapse/Collapse.stories.tsx
mvidalgarcia: I believe this can be achieved by using `@storybook/preview-api`, could you take a look at [these docs](https://storybook.js.org/docs/writing-stories/args#setting-args-from-within-a-story)?

mvidalgarcia: There is perhaps a 4th option. What do you think about removing the arg for this particular case and relying on the UI user interaction? In the end, that's the regular use case (uncontrolled component).

sarkaaa: Do you mean removing `expanded` arg for multipleCollapse at all? yes, we can go this way, I just wanted to include as much args as possible, but it also makes sense to exclude that arg from this story. 🤔

sarkaaa: Should be changed. I used `api-preview` at least for single collapse components. :)

packages/orbit-components/src/Card/Card.stories.tsx
```diff
@@ -12,113 +13,120 @@ import { ELEMENT_OPTIONS } from "../Heading/consts";
...
+  },
 
-Default.argTypes = {
-  titleAs: {
-    options: Object.values(ELEMENT_OPTIONS),
-    control: {
-      type: "select",
+  parameters: {
+    controls: {
+      exclude: ["labelClose"],
```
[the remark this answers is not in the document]
mvidalgarcia: what was the conclusion here? I see it's still on every story.

packages/orbit-components/src/Box/Box.stories.tsx
```diff
@@ -67,6 +66,10 @@ enum ELEVATION {
   LEVEL_3 = "level3",
   LEVEL_3REVERSE = "level3Reverse",
   LEVEL_4 = "level4",
+  OVERLAY = "overlay",
+  RAISED = "raised",
+  RAISEDREVERSE = "raisedReverse",
+  ACTION = "action",
```
mvidalgarcia, line 72: these were deprecated recently, ima remove them.

mvidalgarcia:
@domihustinova can you approve this? 🙏

domihustinova (approved):

PR #4439, opened 2024-08-02
New shadow tokens

mvidalgarcia:
https://skypicker.slack.com/archives/C7T7QG7M5/p1721117646848839?thread_ts=1720790040.455089&cid=C7T7QG7M5

Closes FEPLT-2064


packages/orbit-design-tokens/output/deprecated-tokens.json
DSil, line 1: We have this file called `SCHEMA.md` under `docs` in this package that should be updated (and also for the other tokens 🫣)

DSil (commented):
Besides the comments made, we need to update the migration guide and the occurrences of the tokens in the docs package

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
```diff
@@ -166,7 +166,7 @@ const PopoverContentWrapper = ({
           "h-auto w-full",
           "z-[1000]",
           "box-border",
-          "shadow-raised-reverse",
+          "shadow-level3-reverse",
           "bg-elevation-raised-reverse",
```
DSil, line 170: How are we addressing these? This shadow will disappear but the bg-color remains? Should we deprecate it as well and just use the color directly? In this case, it would be just bg-white-normal, no?

packages/orbit-design-tokens/output/deprecated-tokens.json
mvidalgarcia: I updated it (only shadows)

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
mvidalgarcia: Yeah, backgrounds are still missing.. I need to tackle this and probably just use the replacement (white normal as you suggested).

packages/orbit-design-tokens/output/deprecated-tokens.json
```diff
@@ -1947,6 +1947,56 @@
     "category": "Breakpoints",
     "version": "6.0.0"
   },
+  "elevationActionBackground": {
+    "replacement": "paletteWhiteNormal",
+    "category": "Colors",
+    "version": "9.0.0"
+  },
+  "elevationActionBoxShadow": {
```
DSil, line 1955: I have just realized we now have two levels of deprecations 🤦‍♂️ 
The token `boxShadowAction` is deprecated and suggests `elevationActionBoxShadow` as replacement. But now `elevationActionBoxShadow` is also deprecated and suggests `elevationLevel1BoxShadow` as replacement.

I am not entirely sure about how this will impact the breaking change, but for instance, I see the `boxShadowAction` (the _double deprecated_ one) is currently being used [here](https://github.com/kiwicom/orbit/blob/marco/shadow-tokens/packages/orbit-tailwind-preset/src/index.ts#L256)… 😬  The same for `modal-scrolled` and `modal` defined right below: they're also using doubly deprecated tokens.

So, yes, I honestly don't know what to do with these. But I'd recommend going through all the tokens (even the older ones) and replace them for the new ones. For example, in that case, replace that `switch` definition by the correct `elevationLevel1BoxShadow` token (although I'm also not sure about the consequences of such change…)

I did the quick exercise of completely deleting the tokens and the only place where the build failed was the file mentioned above, but I am not 100% sure this is all the confidence we need.

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
DSil: I see you kept the tokens or renamed them accordingly. I'm ok with that, but I had suggested to completely get rid of those tokens and use the palette directly.

In this case we would use `bg-white-normal` instead of `bg-elevation-raised-reverse` that has the value of white.

And btw, raised reverse is deprecated, so this needs to be updated to `bg-level3-reverse`, right?

packages/orbit-design-tokens/src/dictionary/definitions/global/elevations.json
```diff
@@ -186,6 +300,44 @@
         }
       },
       "raised-reverse": {
+        "background": {
+          "type": "color",
+          "category": "elevation",
+          "value": "{foundation.palette.white.normal}",
+          "deprecated": true,
+          "deprecated-replace": "{foundation.palette.white.normal}",
+          "deprecated-version": "9.0.0"
```
DSil, line 309: Following my previous comment…

I see this is being deprecated with the replacement pointing to `palette.white.normal`. But then the `level3-reverse` background color is also being created, with the value of `palette.white.normal`.

So I think we are not going through the correct approach. We either:
- Use `level3-reverse` as a replacement here and use `palette.white.normal` as a value there

**OR**

- Use `palette.white.normal` as a replacement here but do not define `level3-reverse` at all

I personally prefer the second option, as it will remove the overdefinition of tokens entirely. But I am open for discussing (and accepting) the first option as well

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
```diff
@@ -90,26 +90,48 @@ And their respective tailwind classes:
...
 
 ```jsx
-<Box borderRadius="small">"small" borderRadius</Box>
-<Box borderRadius="normal">"normal" borderRadius</Box>
-<Box borderRadius="large">"large" borderRadius</Box>
-<Box borderRadius="circle">"circle" borderRadius</Box>
+<Box borderRadius="small" elevation="action">"small" borderRadius</Box>
+<Box borderRadius="normal" elevation="raised">"normal" borderRadius</Box>
+<Box borderRadius="large" elevation="raisedReverse">"large" borderRadius</Box>
+<Box borderRadius="circle" elevation="overlay">"circle" borderRadius</Box>
```
DSil, line 125: Given that the content is no longer just about border radius…

```suggestion
<Box borderRadius="small" elevation="action">…</Box>
<Box borderRadius="normal" elevation="raised">…</Box>
<Box borderRadius="large" elevation="raisedReverse">…</Box>
<Box borderRadius="circle" elevation="overlay">…</Box>
```

And same for the example below

DSil:
I have merged the base branch to master and now I see there are some conflicts. Let me know if you wanna pair so we are sure nothing is missed 🤞

packages/orbit-design-tokens/src/dictionary/definitions/global/elevations.json
mvidalgarcia: Ok, gotcha, I'll remove the background tokens and adapt things accordingly. For some reason, I thought we wanted to keep backgrounds as well but I see they are barely used, so there's no point.

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
mvidalgarcia: I'm removing tokens and using `bg-white-normal`

DSil (approved):
Good work 💪

PR #4442, opened 2024-08-07
FEPLT-2073: batch2 - migrate from CSF2 to CSF3

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2073

### DIFF
--- a/packages/orbit-components/src/Dialog/Dialog.stories.tsx
+++ b/packages/orbit-components/src/Dialog/Dialog.stories.tsx
@@ -1,81 +1,124 @@
 import * as React from "react";
 import { action } from "@storybook/addon-actions";
+import type { Meta, StoryObj } from "@storybook/react/";
 
 import Button from "../Button";
 import ButtonLink from "../ButtonLink";
 import Illustration from "../Illustration";
 import { NAMES } from "../Illustration/consts.mts";
+import type { Name } from "../Illustration/types";
 
 import Dialog from ".";
 
-export default {
+const meta: Meta<typeof Dialog> = {
   title: "Dialog",
-};
+  component: Dialog,
 
-export const Default = ({ title, description, dataTest }) => {
-  return (
-    <Dialog
-      title={title}
-      description={description}
-      dataTest={dataTest}
-      primaryAction={<Button type="critical">Log out</Button>}
-      renderInPortal={false}
-    />
-  );
-};
-
-Default.story = {
   parameters: {
     info: "Visit Orbit.Kiwi for more detailed guidelines.",
   },
-};
 
-Default.args = {
-  title: "Log out",
-  description: "Are you sure you want to log out now?",
-  dataTest: "test",
-};
-
-export const Playground = ({ title, maxWidth, description, illustration, dataTest }) => {
-  return (
-    <Dialog
-      title={title}
-      maxWidth={maxWidth}
-      description={description}
-      illustration={<Illustration name={illustration} size="small" />}
-      dataTest={dataTest}
-      onClose={action("onClose")}
-      primaryAction={<Button type="critical">Log out</Button>}
-      secondaryAction={<ButtonLink type="secondary">Cancel</ButtonLink>}
-      renderInPortal={false}
-    />
-  );
-};
-
-Playground.story = {
-  parameters: {
-    info: "Visit Orbit.Kiwi for more detailed guidelines.",
+  args: {
+    title: "Log out",
+    description: "Are you sure you want to log out now?",
+    illustration: "NoNotification",
+    maxWidth: NaN,
+    renderInPortal: false,
+    lockScrolling: false,
+    onClose: action("onClose"),
   },
-};
 
-Playground.args = {
-  title: "Log out",
-  description: "Are you sure you want to log out now?",
-  illustration: "NoNotification",
-  dataTest: "test",
-  maxWidth: NaN,
-};
-
-Playground.argTypes = {
-  illustration: {
-    options: NAMES,
-    control: {
-      type: "select",
+  argTypes: {
+    illustration: {
+      options: NAMES,
+      control: {
+        type: "select",
+      },
+    },
+    maxWidth: {
+      control: {
+        type: "number",
+      },
     },
   },
-  maxWidth: {
-    control: {
-      type: "number",
+};
+
+export default meta;
+type Story = StoryObj<typeof Dialog>;
+
+function useDialog() {
+  const [open, toggle] = React.useState(true);
+  return {
+    Container: ({ children }) => (
+      <>
+        {open && <div>{children}</div>}
+        <Button onClick={() => toggle(true)}>Open</Button>
+      </>
+    ),
+    onClose: () => {
+      toggle(false);
+      action("onClose")();
+    },
+  };
+}
+
+export const Default: Story = {
+  render: args => {
+    const { Container, onClose } = useDialog();
+
+    return (
+      <Container>
+        <Dialog
+          {...args}
+          primaryAction={<Button type="critical">Log out</Button>}
+          onClose={onClose}
+        />
+      </Container>
+    );
+  },
+
+  args: {
+    illustration: undefined,
+    maxWidth: undefined,
+    secondaryAction: undefined,
+    lockScrolling: undefined,
+  },
+
+  parameters: {
+    controls: {
+      exclude: [
+        "illustration",
+        "onClose",
+        "renderInPortal",
+        "maxWidth",
+        "lockScrolling",
+        "primaryAction",
+        "secondaryAction",
+      ],
+    },
+  },
+};
+
+export const Playground: Story = {
+  render: ({ illustration, ...args }) => {
+    const { Container, onClose } = useDialog();
+
+    return (
+      <Container>
+        <Dialog
+          {...args}
+          primaryAction={<Button type="critical">Log out</Button>}
+          secondaryAction={<ButtonLink type="secondary">Cancel</ButtonLink>}
+          illustration={<Illustration name={illustration as Name} size="small" />}
+          onClose={onClose}
+        />
+      </Container>
+    );
+  },
+
+  parameters: {
+    controls: {
+      exclude: ["onClose", "primaryAction", "secondaryAction"],
     },
   },
 };
--- a/packages/orbit-components/src/Drawer/Drawer.stories.tsx
+++ b/packages/orbit-components/src/Drawer/Drawer.stories.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import type { Meta, StoryObj } from "@storybook/react/";
 import { action } from "@storybook/addon-actions";
 
 import Stack from "../Stack";
@@ -19,19 +20,47 @@ import LinkList from "../LinkList";
 
 import Drawer from ".";
 
-export default {
+const meta: Meta<typeof Drawer> = {
   title: "Drawer",
+  component: Drawer,
+
+  parameters: {
+    info: "Visit Orbit.Kiwi for more detailed guidelines.",
+    controls: {
+      exclude: ["onClose"],
+    },
+  },
+
+  args: {
+    title: "Drawer title",
+    shown: true,
+    position: POSITIONS.RIGHT,
+    width: "480px",
+    suppressed: false,
+    fixedHeader: false,
+    noPadding: false,
+    onClose: action("onClose"),
+    labelHide: "Hide",
+    lockScrolling: true,
+  },
+
+  argTypes: {
+    position: {
+      options: Object.values(POSITIONS),
+      control: {
+        type: "select",
+      },
+    },
+  },
 };
 
-export const SideNavigation = ({ shown, dataTest, width, fixedHeader, noPadding }) => {
-  return (
+export default meta;
+type Story = StoryObj<typeof Drawer>;
+
+export const SideNavigation: Story = {
+  render: args => (
     <Drawer
-      noPadding={noPadding}
-      dataTest={dataTest}
-      width={width}
-      shown={shown}
-      fixedHeader={fixedHeader}
-      onClose={action("onClose")}
+      {...args}
       actions={
         <Stack direction="row" justify="between" spacing="200">
           <Button type="secondary" size="small">
@@ -70,33 +99,22 @@ export const SideNavigation = ({ shown, dataTest, width, fixedHeader, noPadding 
         <TextLink type="secondary">Sign out</TextLink>
       </LinkList>
     </Drawer>
-  );
-};
+  ),
 
-SideNavigation.story = {
+  args: {
+    title: undefined,
+  },
+
   parameters: {
-    info: "Visit Orbit.Kiwi for more detailed guidelines.",
+    controls: {
+      exclude: ["title", "onClose"],
+    },
   },
 };
 
-SideNavigation.args = {
-  shown: true,
-  dataTest: "test",
-  width: "320px",
-  fixedHeader: false,
-  noPadding: false,
-};
-
-export const SmartFaq = ({ shown, dataTest, position, width, fixedHeader }) => {
-  return (
-    <Drawer
-      shown={shown}
-      width={width}
-      position={position}
-      fixedHeader={fixedHeader}
-      dataTest={dataTest}
-      onClose={action("onClose")}
-    >
+export const SmartFaq: Story = {
+  render: args => (
+    <Drawer {...args}>
       <Stack>
         <Illustration name="Accommodation" />
         <Heading as="h2">Need help?</Heading>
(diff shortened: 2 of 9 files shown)


packages/orbit-components/src/primitives/Primitives.stories.tsx
```diff
@@ -9,247 +10,168 @@ import IllustrationPrimitiveComponent, { SIZE_OPTIONS } from "./IllustrationPrim
...
+  React.ComponentProps<typeof IllustrationPrimitiveComponent> &
+  React.ComponentProps<typeof ButtonPrimitiveComponent>;
+
+const meta: Meta<MixedPropsAndArgs> = {
   title: "Primitives",
 };
 
-export const BadgePrimitive = ({ icon, className }) => {
-  const Icon = getIcon(icon);
+export default meta;
```
sarkaaa, line 21: For this story file, I haven't specified args (not mutual) as we're presenting three different components.

packages/orbit-components/src/ErrorForms.stories.tsx
```diff
@@ -27,12 +28,32 @@ const objectOptions = [
...
+    help: "Is the spelling correct?",
+    error: "Something went wrong.",
+    value: "",
+    placeholder: "Placeholder",
+    prefix: "$",
+    required: true,
+    helpClosable: true,
+  },
+
+  tags: ["dev", "!autodocs", "!test"],
```
sarkaaa, line 48: That's all that I have found about hiding/showing stories in respective modes. There are 3 tags, but enabling `dev` mode still shows the ErrorForm story in [PR's SB link](https://kiwicom-orbit-sarka-storybook-migration-b2.surge.sh/). In case I disabled the dev mode, the story is not visible when running localhost:6007 either. 🙈

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,98 @@
...
+type Story = StoryObj<IconPropsAndCustomArgs>;
 
-  return <Icon size={size} customColor={customColor} />;
-};
-
-CustomColor.story = {
-  name: "Custom Color",
-  parameters: {
-    info: "We use icons to draw attention to specific actions in our products. Visit Orbit.Kiwi for more detailed guidelines.",
+export const Default: Story = {
```
domihustinova, line 57: I would maybe not set `customColor` in the Default story as there is separate story CustomColor so they are the same now. Here I would set one of the `ICON_COLORS`, WDYT?

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,98 @@
...
+  render: ({ source, ...args }) => {
+    const Icon = getIcon(source);
 
-ReversedOnRtl.story = {
-  name: "Reversed on RTL",
-  parameters: {
-    info: "We use icons to draw attention to specific actions in our products. Visit Orbit.Kiwi for more detailed guidelines.",
+    return (
+      <RenderInRtl>
+        <Icon reverseOnRtl {...args} />
```
domihustinova, line 82: You added args here but you disabled controls, is that on purpose? I think at least the icon could be in controls, no? 🤔

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,98 @@
...
-    control: {
-      type: "select",
+  parameters: {
+    controls: {
+      disable: true,
     },
   },
 };
 
 export const ListOfAllIcons = () => <IconList />;
```
domihustinova, line 98: Maybe let's disable controls here?

packages/orbit-components/src/ErrorForms.stories.tsx
```diff
@@ -27,12 +28,32 @@ const objectOptions = [
   { value: 3, label: "Third item" },
 ];
 
-export default {
+type ErrorFormsPropsAndCustomArgs = React.ComponentProps<typeof InputField> &
+  React.ComponentProps<typeof Select> & { showMore?: boolean; value: string | undefined };
```
domihustinova, line 32: Can't this be `value?: string`? 🤔

packages/orbit-components/src/ErrorForms.stories.tsx
```diff
@@ -376,28 +380,18 @@ export const RtlError = ({ label, error, value, placeholder, prefix, required })
...
-  value: "",
-  placeholder: "Placeholder",
-  prefix: "$",
-  required: true,
-};
-
-export const withModal = ({ label, error, showMore }) => {
-  return (
+export const WithModal: Story = {
+  render: ({ label, error, showMore }) => (
```
domihustinova, line 394: Value, placeholder, prefix and required controls are not used here, let's disable them then? Or maybe I'm blind 😅

packages/orbit-components/src/Icon/Icon.stories.tsx
sarkaaa: I am considering disabling `customColor` from `Default` story. When having this prop excluded we can set and see the colors from `color` prop. There is also story `Custom Color` that allows us to set custom hex color.

WDYT? @domihustinova

packages/orbit-components/src/Drawer/Drawer.stories.tsx
```diff
@@ -70,33 +98,18 @@ export const SideNavigation = ({ shown, dataTest, width, fixedHeader, noPadding
...
     </Drawer>
-  );
-};
+  ),
 
-SideNavigation.story = {
   parameters: {
-    info: "Visit Orbit.Kiwi for more detailed guidelines.",
+    controls: {
+      exclude: "onClose",
```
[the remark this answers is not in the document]
sarkaaa: good point. I remember that I removed it, but I probably didn't save the file, ha. 🙈

packages/orbit-components/src/Dialog/Dialog.stories.tsx
```diff
@@ -1,81 +1,125 @@
...
+      },
+    },
+  },
 };
 
-Playground.story = {
+export default meta;
+type Story = StoryObj<typeof Dialog>;
+
+function useDialog() {
```
sarkaaa, line 49: I used `useDialog` to allow user close the dialog and copy the code. :D

packages/orbit-components/src/Dialog/Dialog.stories.tsx
```diff
@@ -1,81 +1,85 @@
...
 
-Playground.story = {
+export default meta;
+type Story = StoryObj<typeof Dialog>;
+
+export const Default: Story = {
+  args: {
+    illustration: undefined,
+    maxWidth: undefined,
+    primaryAction: undefined,
```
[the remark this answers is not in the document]
sarkaaa: I think that `title` was in Default :D . For primaryAction, I set it to a static part of the code as it's ReactNode type and not simple to edit in SB controls. 🤔

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,121 @@
...
 
 import IconList from "./IconList";
 import * as Icons from "../icons";
 import { ICON_SIZES, ICON_COLORS } from "./consts";
 import RenderInRtl from "../utils/rtl/RenderInRtl";
 
-export default {
-  title: "Icon",
-};
+import type OrbitIcon from ".";
```
sarkaaa, line 9: For this Story, I imported type `OrbitIcon` as we are using these props.

packages/orbit-components/src/Icon/Icon.stories.tsx
domihustinova: Agree 👍

packages/orbit-components/src/Drawer/Drawer.stories.tsx
```diff
@@ -19,19 +20,46 @@ import LinkList from "../LinkList";
...
+  args: {
+    shown: true,
+    position: POSITIONS.RIGHT,
+    width: "480px",
+    suppressed: false,
+    fixedHeader: false,
+    noPadding: false,
+    onClose: action("onClose"),
+    labelHide: "Hide",
+    lockScrolling: true,
```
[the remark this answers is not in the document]
sarkaaa: Good point, fixed. ✅

domihustinova (commented):
LGTM, good job! ✨

packages/orbit-components/src/Drawer/Drawer.stories.tsx
```diff
@@ -509,31 +396,32 @@ export const SmartFaqSearchInRtl = ({
...
 
-SmartFaqSearchInRtl.args = {
-  shown: true,
-  dataTest: "test",
-  position: POSITIONS.RIGHT,
-  width: "480px",
-  suppressed: true,
-  fixedHeader: false,
-};
+export const Playground: Story = {
```
mvidalgarcia, line 407: Do we need this story? How is it different from e.g. "With Title"?

packages/orbit-components/src/FeatureIcon/FeatureIcon.stories.tsx
```diff
@@ -1,33 +1,33 @@
...
 
-Playground.story = {
   parameters: {
     info: "Visit Orbit.Kiwi for more detailed guidelines.",
   },
-};
 
-Playground.args = {
-  name: NAME_OPTIONS.TICKETFLEXI,
-  dataTest: "test",
```
mvidalgarcia, line 23: Is there any reason why we don't keep the `dataTest` anymore?

packages/orbit-components/src/Hide/Hide.stories.tsx
```diff
@@ -1,70 +1,87 @@
...
+export const WithSeparator: Story = {
+  render: ({ largeMobile, ...args }) => {
+    const on = [largeMobile && "largeMobile"] as Devices[];
+
+    return (
+      <Hide {...args} on={on}>
+        <Separator />
+      </Hide>
+    );
+  },
```
mvidalgarcia, line 53: In this story we have a bunch of controls but they don't seem to work.

image


Perhaps we should exclude them and only keep them for the Playground?

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,121 @@
...
 
-export const ReversedOnRtl = ({ source }) => {
-  const Icon = Icons[source];
-  return (
-    <RenderInRtl>
-      <Icon reverseOnRtl />
-    </RenderInRtl>
-  );
-};
+export const ReversedOnRtl: Story = {
```
mvidalgarcia, line 91: There should be a control for the `source`.

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.tsx
```diff
@@ -96,86 +84,147 @@ export const NoScroll = ({ spacing }) => {
...
-          }}
-        >
-          <Text size="large" weight="bold" as="div">
-            <div
-              style={{
-                height: "50px",
-                display: "flex",
-                alignItems: "center",
-              }}
+export const NoScroll = {
```
mvidalgarcia, line 106: What does NoScroll mean in this context?

I can scroll it on small devices. What am I missing?

packages/orbit-components/src/FeatureIcon/FeatureIcon.stories.tsx
sarkaaa: I got information that [we won't include](https://github.com/kiwicom/orbit/pull/4437#pullrequestreview-2220818766) `dataTest` attribute to Storybook.

packages/orbit-components/src/Hide/Hide.stories.tsx
sarkaaa: Yes, it totally makes sense to keep only `Playground` story.
Also, I have found that `Separator` component is not the best representative for this type of visualization, so I'll replace it with `div` with background color to have the actual state of component more highlighted. :D 

something like this:
image

WDYT?

packages/orbit-components/src/Drawer/Drawer.stories.tsx
sarkaaa: I just wanted to keep `Playground` story for most of the components. But it does not make sense to have it for this specific component. 💯

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.tsx
sarkaaa: It's a "garbage" from the old storybook.

I am considering to remove "with scroll snap" as well and keep only `Default` and `Playground` stories, because "With scroll snap" and "Playground" will be the same and this component is "simple" in general. 🤔

packages/orbit-components/src/FeatureIcon/FeatureIcon.stories.tsx
mvidalgarcia: Ok, thanks for the info.

packages/orbit-components/src/Hide/Hide.stories.tsx
mvidalgarcia: Ok, it makes sense.

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.tsx
mvidalgarcia: Not sure if "garbage" is the best definition. I'm sure this story was added for a specific purpose, I was just wondering what it was.

sarkaaa: yeah, maybe "garbage" is too strong. 🙈  in the [old storybook](https://kiwicom-orbit-sarka-storybook-migration-b4.surge.sh/?path=/story/horizontalscroll--no-scroll ) there are only a few items in the component and it demonstrates that scrolling is available only when the width of all items is bigger than parent contaner's width.

packages/orbit-components/src/Dialog/Dialog.stories.tsx
```diff
@@ -1,81 +1,124 @@
...
+    },
+  },
 };
 
-Playground.story = {
+export default meta;
+type Story = StoryObj<typeof Dialog>;
+
+function useDialog() {
+  const [open, toggle] = React.useReducer((_, next) => next, true);
```
mvidalgarcia, line 50: Could you please explain the reason behind using `useReducer` here?

sarkaaa: I just used the same function as it was used in [Modal.stories.tsx](https://github.com/kiwicom/orbit/blob/5c87463e700a0bd7879259cc478197d4293fe3a0/packages/orbit-components/src/Modal/Modal.stories.tsx#L35) file, but I think using `useState` should be enough for this case. 🤔  WDYT?

sarkaaa: yes, that's true. going to change it ASAP. :)

sarkaaa: Changed. I've also changed:
```
<div style={{ display: open ? "block" : "none" }}>{children}</div>
```

to

```
{open && <div>{children}</div>
```

to make the code little bit simpler.

packages/orbit-components/src/Hide/Hide.stories.tsx
```diff
@@ -1,70 +1,65 @@
...
+      desktop && "desktop",
+      tablet && "tablet",
+      largeMobile && "largeMobile",
+      mediumMobile && "mediumMobile",
+      smallMobile && "smallMobile",
+    ].filter(item => typeof item === "string") as Devices[];
+
+    return (
+      <Hide block={block} on={on}>
+        <div style={{ height: 20, width: 300, background: "darkViolet" }} />
```
mvidalgarcia, line 57: Could we use a percentage width here? It doesn't look very nice on `smallMobile`

image

packages/orbit-components/src/Icon/Icon.stories.tsx
```diff
@@ -1,129 +1,123 @@
...
-  ariaLabel: "label",
-  ariaHidden: true,
-};
+  args: {
+    size: ICON_SIZES.MEDIUM,
+    color: ICON_COLORS.PRIMARY,
+    source: "Airplane",
+    ariaLabel: "label",
+    ariaHidden: true,
+    customColor: "#ABCDEF",
```
mvidalgarcia, line 28: What's this value for? I believe you're overriding it below with `customColor: "#ff0000"`.

packages/orbit-components/src/Hide/Hide.stories.tsx
sarkaaa: I wanted to set width in `%`, but the parent wrapper component (`<Hide>`) doesn't have set width, so percentage value for width is not working when `display: inline-block` is set for Hide component.

image

I can set `width: 100%` and `min-width: 100px` at least to have "something" for "inline-block" display property. 🤔

packages/orbit-components/src/Icon/Icon.stories.tsx
sarkaaa: good catch, thx. I will keep the "default" customColor. 👍

packages/orbit-components/src/HorizontalScroll/HorizontalScroll.stories.tsx
```diff
@@ -96,86 +83,61 @@ export const NoScroll = ({ spacing }) => {
...
+                  display: "flex",
+                  alignItems: "center",
+                }}
+              >
+                {key}
+              </div>
+            </Text>
+          </Box>
+        ))
+      }
```
mvidalgarcia, line 136: Could you wrap this into a component to avoid repetition?

mvidalgarcia (approved):
:shipit:

PR #4445, opened 2024-08-08
Optimize email icons generator

DSil:
Sometimes we need to generate PNG from icons and add them to images.kiwi. The script to generate them was optimized and now generates the whole directory to be copied and pasted there.

We should also have this process better documented, but I was unsure about where to document this. Something to be discussed soon™.

FEPLT-2093


packages/orbit-components/package.json
```diff
@@ -23,7 +23,9 @@
     "fetch:icons": "node --loader ts-node/esm config/fetchIcons.mts",
     "check:icons": "node --loader ts-node/esm config/checkIconsCli.mts",
     "check:types": "tsc --p tsconfig.json",
-    "build:iconsPng": "node --loader ts-node/esm config/generatePngIcons.mts",
+    "build:icons": "node --loader ts-node/esm config/build/buildIcons.mts",
+    "build:iconsFont": "node --loader ts-node/esm config/createSVGFont.mts",
+    "build:emailIcons": "yarn build:icons && yarn build:iconsFont && node --loader ts-node/esm config/generateEmailIcons.mts",
     "build:storybook": "node --loader ts-node/esm  config/build/buildIcons.mts && storybook build -c .storybook -o .out",
```
mvidalgarcia, line 29:
```suggestion
    "build:storybook": "yarn build:icons && storybook build -c .storybook -o .out",
```

mvidalgarcia (approved):

PR #4446, opened 2024-08-09
FEPLT-2074: Migrate Stories to CSF3 - Batch 3

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2074

### DIFF
--- a/packages/orbit-components/src/Heading/Heading.stories.tsx
+++ b/packages/orbit-components/src/Heading/Heading.stories.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
 
 import { ELEMENT_OPTIONS, TYPE_OPTIONS, ALIGN } from "./consts";
 import { SPACINGS_AFTER } from "../common/consts";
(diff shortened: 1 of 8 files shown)


packages/orbit-components/src/Heading/Heading.stories.tsx
```diff
@@ -1,442 +1,244 @@
...
-    <Heading type="title2" as={as}>
-      {customTitle}
-    </Heading>
-  );
-};
-
-Title2.story = {
-  parameters: {
-    info: "Headings are used for showing content hierarchy and are important for improving the reading experience for our users. Visit Orbit.Kiwi for more detailed guidelines.",
+export const Title0: Story = {
```
sarkaaa, line 91: I am thinking if it does make sense to have 7 almost same stories. I am facing same "issue" in `Box` component - same stories with different property name in Story name. What do you think of "squashing" them into 1 story? I don't know what's the best practice, tbh. 🥲

packages/orbit-components/src/Inline/Inline.stories.tsx
```diff
@@ -25,106 +24,91 @@ const TestChildren = () => {
...
-      type: "select",
-    },
+  args: {
+    as: "div",
+    spacing: SPACING.XXXSMALL,
+    align: ALIGN.START,
+    justify: JUSTIFY.START,
+    className: "",
+    id: "",
+    mediumMobile: {},
```
sarkaaa, line 89: I would consider to add `example` values. Now these "query" keys are empty objects and it's not clear what to fill at the first moment.

image

For example, I would change it to:
```
mediumMobile: {
  spacing: SPACING.XXXSMALL,
  align: ALIGN.START,
  justify: JUSTIFY.START,
}
```

WDYT? 🤔

packages/orbit-components/src/InputFile/InputFile.stories.tsx
```diff
@@ -1,208 +1,111 @@
...
-};
-
-WithError.story = {
-  name: "With error",
-
+export const WithHelp: Story = {
   parameters: {
-    info: "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
+    controls: {
+      exclude: ["onChange", "onFocus", "onBlur", "onRemoveFile", "help"],
```
sarkaaa, line 43: It's just about consistency, but for `Error` story I can see that the input field in SB is available, here it's excluded.
I can see that default values for `help` is ReactNode type, for `error` it's string type, although both of the props are `ReactNode` type.
I would align it to make consincency for both `Error` and `Help` stories. :D

sarkaaa: and also, in Playground story there is Help available.

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -280,149 +283,106 @@ export const OnChangeBehavior = ({ inputValue }) => {
...
-
-Playground.story = {
-  parameters: {
-    info: "Some description about this type of InputGroup in general.",
+  ),
+
+  args: {
+    label: "Phone number",
+    flex: ["1 0 200px", "1 1 100%", "1 0 150px", "0 1 50%"],
+    size: SIZE_OPTIONS.NORMAL,
```
sarkaaa, line 341: In most of the stories, we have this arg `size`, but I can't see any change when changing from `small` to `normal` or vice versa. Does it make sense to exclude it from controls (in case it really does nothing, maybe I'm just blind :D :D )

packages/orbit-components/src/Inline/Inline.stories.tsx
domihustinova: I was contemplating adding the examples as you suggest for the same reason but then I didn't because then the other controls (`spacing`, `align` and `justify`) wouldn't work so you would need to change those objects to see the change which seemed too complicated to me 😄 Maybe adding just for one or two and leave the rest empty?

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -10,236 +11,236 @@ import ServiceLogo from "../ServiceLogo";
...
-      type="number"
-      value={value}
-      label={label}
-      placeholder={placeholder}
-      maxValue={maxValue}
-      minValue={minValue}
-      onChange={action("change")}
-    />
-  );
+  args: {
```
sarkaaa, line 35: What about adding arg `helpclosable` to show that user can have this type of help tooltip as well?

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -10,236 +11,236 @@ import ServiceLogo from "../ServiceLogo";
...
-  error: "",
+  args: {
+    inlineLabel: true,
+  },
 };
 
-export const CompactInputWithTags = ({ value, label, placeholder, required, error }) => {
-  return (
+export const CompactInputWithTags: Story = {
+  render: args => (
```
sarkaaa, line 242: When having `required=true`, `inlineLabel=true` and help input filled in, I don't see the help icon in the input. Is it expected for this type of input field or is it a bug (it is not happening for other stories)?

image

number input story:
image

packages/orbit-components/src/InputSelect/InputSelect.stories.tsx
```diff
@@ -1,267 +1,225 @@
...
 };
 
-export const Grouped = ({ options, showAll, readOnly }) => {
-  return (
+export default meta;
+type Story = StoryObj<typeof InputSelect>;
+
+export const Grouped: Story = {
+  render: args => (
     <div className="min-h-[1000px]">
```
sarkaaa, line 150: Is it needed to have the `InputSelect` wrapped by `div` element?

I think it can be rewritten to:
```
export const Grouped: Story = {
  parameters: {
  ...
}
```

but maybe I haven't caught anything. 🤔 :D

packages/orbit-components/src/InputSelect/InputSelect.stories.tsx
```diff
@@ -1,267 +1,225 @@
...
-  showAll: true,
-  readOnly: false,
+  },
 };
 
-export const PreviouslySelected = ({ options, showAll, readOnly, prevSelectedLabel }) => {
-  return (
+export const PreviouslySelected: Story = {
+  render: args => (
     <div className="min-h-[1000px]">
```
sarkaaa, line 176: Same as previous comment.

packages/orbit-components/src/Inline/Inline.stories.tsx
sarkaaa: yes, I would add at least one/two to demonstrate how it can be filled in. :D

packages/orbit-components/src/InputSelect/InputSelect.stories.tsx
domihustinova: Yes, it's there so when you open the select, it doesn't open over the source code. I think it makes sense to have it there but I guess 1000px is a bit of an overkill. Since `maxHeight` is set to 400px, I changed this min-h to 500px.

packages/orbit-components/src/InputFile/InputFile.stories.tsx
domihustinova: Yes, there was `help` defined as ReactNode type so I left it there as such, but if it's being changed via controls, it breaks the whole StoryBook 🤕 so I disabled the control as it's currently in production. To be consistent, I can add `error` as ReactNode to Error story and disable control as well and in the Playground both will be string and enabled.

packages/orbit-components/src/Heading/Heading.stories.tsx
domihustinova: I thought about this and it makes sense but then I re-read [this comment](https://github.com/kiwicom/orbit/pull/4437#issuecomment-2276154525) where we discussed it and considering this, I wouldn't squash them to one story. On one hand yes, it's almost the same 7 stories, but if I imagine wanting to use simple Heading component, when I find it in SB, I can just go over all different heading sizes without changing two props. I think it's the same with Badge, there we could also squash them into one story 🤔 And it should be direct migration so I would just keep it like this, WDYT?

sarkaaa: yeah, it makes sense to have a specific story for a specific heading type. 🤔 I was comparing it with "my" Box component as I squash multiple stories, but I guess that it makes more sense for Box component. 😄

packages/orbit-components/src/InputField/InputField.stories.tsx
sarkaaa: Let's wait for the information in [Slack thread](https://skypicker.slack.com/archives/C05FDP4UM3N/p1723638350535459) and then I think PR is ok. 😃

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -111,32 +121,22 @@ export const PhoneNumber = ({
...
-  selectValue: {
-    options: selectOptionsPhoneNumber.map(opt => opt.value),
-    control: {
-      type: "select",
+  argTypes: {
+    selectValue: {
+      options: selectOptionsPhoneNumber.map(opt => opt.value),
+      control: {
+        type: "select",
+      },
```
mvidalgarcia, line 139: Switching this value does not do anything to the UI. Shouldn't it change the number prefix?

image

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -155,53 +155,41 @@ const selectOptionsError = [
...
+export const Error: Story = {
+  render: ({ error, help, selectValue, ...args }) => (
+    <InputGroup {...args}>
       <InputField placeholder="DD" error={error} help={help} />
       <Select
         options={selectOptionsError}
         value={selectValue}
         placeholder="Month"
-        error="Something went wrong on month field"
+        error={error}
```
mvidalgarcia, line 167: Is this change expected?

Same for the one below on InputField (year)

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -215,7 +203,7 @@ export const ValidationApproaches = ({
         <InputField placeholder="b" maxLength={11} value={inputValue} />
       </InputGroup>
 
-      <InputGroup label="Events andd states on children" error={errorGroup}>
+      <InputGroup label="Events and states on children" error={errorGroup}>
```
mvidalgarcia, line 206: Shouldn't this have `helpGroup`?

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -280,149 +276,105 @@ export const OnChangeBehavior = ({ inputValue }) => {
...
-  );
-};
-
-Playground.story = {
-  parameters: {
-    info: "Some description about this type of InputGroup in general.",
+  ),
+
+  args: {
+    label: "Phone number",
```
mvidalgarcia, line 332: I'm aware it was like this, but this is not a phone number 😅 

Could you please edit the label?

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -280,149 +276,105 @@ export const OnChangeBehavior = ({ inputValue }) => {
...
 
-Rtl.args = {
-  flex: ["0 0 60px", "1 1 100%", "0 0 90px"],
-  selectValue: selectOptionsMonths[0].value,
-  error: "",
-  help: "",
-  label: "Phone number",
-};
+  args: {
+    label: "Phone number",
```
mvidalgarcia, line 367: Same here.

mvidalgarcia (commented):
I went through all the commits one by one and I found some things. Probably some of them were already wrong in the past but I took the opportunity to leave feedback to fix them proactively.

Great effort 👏🏻

packages/orbit-components/src/Heading/Heading.stories.tsx
```diff
@@ -1,442 +1,244 @@
...
 import { ELEMENT_OPTIONS, TYPE_OPTIONS, ALIGN } from "./consts";
 import { SPACINGS_AFTER } from "../common/consts";
 import RenderInRtl from "../utils/rtl/RenderInRtl";
 
 import Heading from ".";
 
-export default {
-  title: "Heading",
+type HeadingPropsAndCustomArgs = React.ComponentProps<typeof Heading> & {
+  customTitle: string;
```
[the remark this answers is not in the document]
domihustinova: No we don't. It was used before instead of children but I changed the stories to use children instead. Will remove it.

packages/orbit-components/src/Inline/Inline.stories.tsx
```diff
@@ -25,106 +24,91 @@ const TestChildren = () => {
...
-    control: {
-      type: "select",
-    },
-  },
-  spacing: {
-    options: Object.values(SPACING),
-    control: {
-      type: "select",
-    },
+  args: {
```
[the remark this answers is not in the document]
domihustinova: We have been debating this with Sarka previously https://github.com/kiwicom/orbit/pull/4446#discussion_r1715501396

With the current implementation, you need to change the values in `mediumMobile` and `desktop` for the controls to work properly which I agree is not the most comfortable :D I will make it as current master.

packages/orbit-components/src/Layout/Layout.stories.tsx
```diff
@@ -81,67 +94,63 @@ export const Mmb = () => {
...
-          <CustomDiv />
-        </CardSection>
-      </Card>
-    </LayoutColumn>
-  </Layout>
-);
-
-export const Rtl = () => (
-  <RenderInRtl>
+export const Customized: Story = {
```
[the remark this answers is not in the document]
domihustinova: It looks like with the `type="Search"`, the component of LayoutColumn is `aside` which is changed to `span` in the Customized story. It makes sense to my to show that it can be changed but it also feels redundant. Would you rather remove this whole story?

packages/orbit-components/src/InputFile/InputFile.stories.tsx
```diff
@@ -1,208 +1,121 @@
...
-};
-
-WithHelp.story = {
-  name: "With help",
-
+export const WithHelp: Story = {
   parameters: {
-    info: "You can try all possible configurations of this component. However, check Orbit.Kiwi for more detailed design guidelines.",
+    controls: {
+      exclude: ["onChange", "onFocus", "onBlur", "onRemoveFile", "help"],
```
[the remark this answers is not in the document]
domihustinova: Unfortunately, yes 😞  It's not very clear in the SB docs, but I found the response on the SB Discord that it will always override the meta excluded values and we need to provide a complete value for each override, there are two ways.

We can do it either like I did or we can spread the meta values but with disabling ESlint as well (or we could add  there`|| {}` to fix the error)

```
  parameters: {
    controls: {
      // eslint-disable-next-line no-unsafe-optional-chaining
      exclude: [...meta.parameters?.controls.exclude, "help"],
    },
  },
```

Which one do you prefer? To me the least complicated one is to just enumerate it again. But with the spread it at least provides some context that we are aware of the meta excludes and we need to extend it.

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -10,236 +11,238 @@ import ServiceLogo from "../ServiceLogo";
...
+    required: false,
+    inlineLabel: false,
+    type: TYPE_OPTIONS.TEXT,
+  },
 };
 
-PasswordInput.story = {
-  name: "Password input",
-
+export const NumberInput: Story = {
```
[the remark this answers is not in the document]
domihustinova: I rechecked our reports and this was not part of it. As it it related to Label, maybe it could be fixed as part of this task? https://kiwicom.atlassian.net/browse/FEPLT-2098

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -10,236 +11,238 @@ import ServiceLogo from "../ServiceLogo";
...
-      onChange={action("change")}
-    />
-  );
+  args: {
+    help: "",
+    error: "",
+    required: false,
+    inlineLabel: false,
+    type: TYPE_OPTIONS.TEXT,
+  },
```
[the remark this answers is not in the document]
domihustinova: True. I think I have it here before I set the same values on meta (and also excluded those) and I forgot to delete them.

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -10,236 +11,238 @@ import ServiceLogo from "../ServiceLogo";
...
 };
 
-PasswordInput.story = {
-  name: "Password input",
-
+export const NumberInput: Story = {
   parameters: {
-    info: "Some description about this type of InputField in general.",
+    controls: {
+      exclude: ["onChange", "prefix", "suffix"],
```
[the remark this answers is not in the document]
domihustinova: You are correct, I will fix it in all places.

packages/orbit-components/src/InputField/InputField.stories.tsx
```diff
@@ -266,594 +269,275 @@ export const CompactInputWithTags = ({ value, label, placeholder, required, erro
...
-};
-
-WithButtonLinkSuffix.story = {
-  name: "With ButtonLink suffix",
+    return <InputField prefix={Prefix && <Prefix />} {...args} />;
+  },
 
   parameters: {
-    info: "Some description about this type of InputField in general.",
+    info: "Compact input with FormLabel as prefix",
```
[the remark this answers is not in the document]
domihustinova: You are correct 🤦‍♀️ Will fix it.

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
domihustinova: It is changing the number prefix, but it' being overwritten by the `customValueText` value 🙈 As it didn't really make sense to have it there, I removed it. Also the CountryFlag code value was hardcoded, so I added it into `selectOptionsPhoneNumber`. Changing the value now should be changing the number prefix as well as the flag.

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
domihustinova: Yes, it was expected, I made the change intentionally. Before the error message was hardcoded so changing the value in controls wouldn't have any effect.

But thinking about it now, as I disabled type controls in the InputField stories so selected type couldn't be changed, I will remove the "error" control from this Error story so it cannot be changed. And with that, it can be hardcoded like it was previously.

packages/orbit-components/src/InputField/InputField.stories.tsx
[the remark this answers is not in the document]
mvidalgarcia: Ok, please add it to the Jira ticket.

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -155,16 +158,11 @@ const selectOptionsError = [
...
-      onChange={action("onChange")}
-      onFocus={action("onFocus")}
-      onBlur={action("onBlur")}
-    >
-      <InputField placeholder="DD" error={error} help={help} />
+
+export const Error: Story = {
+  render: ({ help, selectValue, ...args }) => (
+    <InputGroup {...args}>
+      <InputField placeholder="DD" help={help} error="Something went wrong on day field" />
```
mvidalgarcia, line 165: I would actually make the `error` here controllable, so we can change the text or even remove it to display the error on the Select.

Same as the [current](https://kiwicom.github.io/orbit/?path=/story/inputgroup--error&args=error:) master.

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
```diff
@@ -174,34 +172,32 @@ export const Error = ({ label, flex, error, help, selectValue }) => {
...
-Error.args = {
-  label: "Label",
-  flex: ["0 0 60px", "1 1 100%", "0 0 90px"],
-  error: "Something went wrong on day field",
-  help: "",
-  selectValue: selectOptionsError[0].value,
-};
+  parameters: {
+    controls: {
+      exclude: ["onChange", "onFocus", "onBlur", "onBlurGroup", "size", "error"],
```
mvidalgarcia, line 179: I'd also exclude `help` here, as there's always an error and it will never be visible.

packages/orbit-components/src/InputField/InputField.stories.tsx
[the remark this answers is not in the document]
domihustinova: Added to the ticket ✅

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
domihustinova: Fixed ✅

packages/orbit-components/src/InputGroup/InputGroup.stories.tsx
domihustinova: Excluded, as well as `helpClosable`. And I removed `error` from the list.

mvidalgarcia (approved):
🚀 Ship it!

PR #4453, opened 2024-08-15
fix(InputSelect): use option value as key

domihustinova:
We were using title as key in`InputSelectOption` for the dropdown which was causing duplicate keys if there were options with the same title (and different value). I changed the keys to be created from title-value combination to secure uniqueness. Once this was done, filtering started to work properly and options were not duplicated.

Closes https://kiwicom.atlassian.net/browse/FEPLT-2096

### DIFF
--- a/packages/orbit-components/src/InputSelect/index.tsx
+++ b/packages/orbit-components/src/InputSelect/index.tsx
@@ -262,7 +262,7 @@ const InputSelect = React.forwardRef<HTMLInputElement, Props>(
       if (results.groups.length === 0) {
         return results.all.map((option, idx) => {
           const { title, description, prefix, value: optValue } = option;
-          const optionId = randomId(title);
+          const optionId = randomId(`${title}_${optValue}`);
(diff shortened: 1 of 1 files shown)


sarkaaa:
LGTM and I haven't found any issue when checking component locally. You are the hero! <3 🎉

mvidalgarcia:
> EDIT: I am also thinking about the possibility that somebody will fill array with two (or more) objects with the same `value`. 🤔


In our documentation, we already include a warning that the value should be unique.
https://orbit.kiwi/components/input/inputselect/react/#option

mvidalgarcia:
The solution LGTM, good work! ✨ 

However, I'm not sure about the commit title. It tells us what was done, but doesn't really explain how it helps users or what problem it fixes. Could you please adapt it?

mvidalgarcia (approved):

mvidalgarcia:
Feel free to merge it and inform the reporter on Slack!

PR #4458, opened 2024-08-22
FEPLT-2100 update codemod file for v16

sarkaaa:
This commit adds word boundaries for border  radius values, tailwind theme access for space keyword.

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2100


packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -128,15 +128,15 @@ const spacingTransformations = [
...
-  { from: /theme\(spacing.xxxs\)/, to: "theme(spacing.50)" },
-  { from: /theme\(spacing.xxs\)/, to: "theme(spacing.100)" },
-  { from: /theme\(spacing.xs\)/, to: "theme(spacing.200)" },
-  { from: /theme\(spacing.sm\)/, to: "theme(spacing.300)" },
-  { from: /theme\(spacing.md\)/, to: "theme(spacing.400)" },
-  { from: /theme\(spacing.lg\)/, to: "theme(spacing.600)" },
-  { from: /theme\(spacing.xl\)/, to: "theme(spacing.800)" },
-  { from: /theme\(spacing.xxl\)/, to: "theme(spacing.1000)" },
-  { from: /theme\(spacing.xxxl\)/, to: "52px" },
+  { from: /theme\((spacing|space)\.xxxs\)/, to: "theme(spacing.50)" },
```
mvidalgarcia, line 131: @DSil Do you know the difference between `spacing` and `space` here? In our codebase we seem to have references only to `spacing`. Does this match with the [Tailwind theme config](https://github.com/kiwicom/orbit/blob/db70be98696b3a330d1cecc7116f2629351bb704/packages/orbit-tailwind-preset/src/foundation/getTailwindTheme.ts#L129)?

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -5,10 +5,10 @@ import { join } from "path";
 
 const borderRadiusTransformations = [
   // tailwind classes
-  { from: /rounded(-[a-z]+)?-small/, to: "rounded$1-50" },
-  { from: /rounded(-[a-z]+)?-normal/, to: "rounded$1-100" },
-  { from: /rounded(-[a-z]+)?-large/, to: "rounded$1-150" },
-  { from: /rounded(-[a-z]+)?-circle/, to: "rounded$1-full" },
+  { from: /rounded(-[a-z]+)?-small\b/, to: "rounded$1-50" },
+  { from: /rounded(-[a-z]+)?-normal\b/, to: "rounded$1-100" },
+  { from: /rounded(-[a-z]+)?-large\b/, to: "rounded$1-150" },
```
mvidalgarcia, line 10: Works well, nice!

https://regex101.com/r/bnyJlh/1

I'm thinking that maybe we should add the word boundary at the beginning to be even safer. 

And also, WDYT about adding boundaries to other replacements, not only border-radius?

DSil: > I'm thinking that maybe we should add the word boundary at the beginning to be even safer.

I thought that too, but I am concerned about things like `lm:rounded-small`? 🤔

packages/orbit-components/transforms/tokens-v16.mjs
DSil: Official docs for [spacing](https://tailwindcss.com/docs/customizing-spacing) and [theme function](https://tailwindcss.com/docs/functions-and-directives#theme) mention `spacing` and that is what we use in Orbit. I don't know why others refer to it as `space` even if their config does not mention it 🤷‍♂️ apparently it works. But I don't know if it's some legacy naming that Tailwind kept for retrocompatibility… But I would prefer to just use spacing, tbh

packages/orbit-components/transforms/tokens-v16.mjs
sarkaaa: When I was "doing a research" I also got idea to set a boundary `\b` to the beginning, but then I had thoughts about "media queries" as Daniel mentions, so I added it only to the end. 🙈

sarkaaa: I also added the "end" boundaries to the other `from` regex expression within the codemod file (discussed with Daniel in PM).

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -80,63 +93,63 @@ const elevationTransformations = [
...
-  { from: /(top|bottom|left|right|start|end)-xxxl/, to: "$1-[52px]" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xxxs\b/, to: "$1$2-50" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xxs\b/, to: "$1$2-100" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xs\b/, to: "$1$2-200" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-sm\b/, to: "$1$2-300" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-md\b/, to: "$1$2-400" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-lg\b/, to: "$1$2-600" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xl\b/, to: "$1$2-800" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xxl\b/, to: "$1$2-1000" },
+  { from: /(p|m)([s|e|t|b|l|r|x|y])?-xxxl\b/, to: "$1$2-[52px]" },
```
DSil, line 114: Btw, while here, I got confirmation from design that `xxxl` should be replaced by `-1200` so please update the instances accordingly 🙏

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -51,6 +63,7 @@ const headingTransformations = [
   { from: "fontWeightHeadingTitle6", to: "headingTitle6FontWeight" },
   // keep type and tailwind classes to avoid visual breaking changes
   { from: 'type="title1"', to: 'type="title0"' },
+  { from: 'type: "title1"', to: 'type: "title0"' },
```
sarkaaa, line 66: Not sure if I understood this specific request correctly, but it should cover [this statement](https://skypicker.slack.com/archives/C7T7QG7M5/p1724156159852009?thread_ts=1724147511.514349&cid=C7T7QG7M5):
```
And neither does <Heading as="h2" type="title2" tablet={{ type: "title1" }}>
```

packages/orbit-components/transforms/tokens-v16.mjs
sarkaaa: Done. ✅

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -51,7 +51,9 @@ const headingTransformations = [
...
   { from: "leading-heading-title1", to: "leading-heading-title0" },
+  { from: "theme(lineHeight.heading-title1)", to: "theme(lineHeight.heading-title0)" },
```
sarkaaa, line 56: This sure covers Vojta's note:
```
theme(lineHeight.heading-title1) doesn’t get caught by the codemod.
```

but I am not sure if the change is 100% correct (based on my understanding). 🙈  Please, check it out. 🙏

packages/orbit-components/transforms/tokens-v16.mjs
mvidalgarcia: > I don't know why others refer to it as space even if their config does not mention it 🤷‍♂️ apparently it works.

Wow, I'm surprised about this. Maybe we should ask them about it.

packages/orbit-components/transforms/tokens-v16.mjs
DSil: That is the same as "what if someone uses a different notation"?

This won't work either, in both cases:

```jsx
<Heading
  type={"title1"}
  tablet={{ type: `title0` }}
>...
```

I don't think we need to cover that. Double quote is the linting standard

packages/orbit-components/transforms/tokens-v16.mjs
DSil: Yes, that is what is expected

packages/orbit-components/transforms/tokens-v16.mjs
```diff
@@ -155,27 +157,27 @@ const spacingTransformations = [
...
-  { from: /(spacing|padding|margin|sideOffset)="XXXLarge"/, to: '$1="1200"' }, // visual breaking change
+  { from: /(spacing|padding|margin)="XXXSmall"\b/, to: '$1="50"' },
+  { from: /(spacing|padding|margin)="XXSmall"\b/, to: '$1="100"' },
+  { from: /(spacing|padding|margin)="XSmall"\b/, to: '$1="200"' },
+  { from: /(spacing|padding|margin|sideOffset)="small"\b/, to: '$1="300"' },
+  { from: /(spacing|padding|margin|sideOffset)="medium"\b/, to: '$1="400"' },
+  { from: /(spacing|padding|margin|sideOffset)="large"\b/, to: '$1="600"' },
+  { from: /(spacing|padding|margin|sideOffset)="XLarge"\b/, to: '$1="800"' },
+  { from: /(spacing|padding|margin|sideOffset)="XXLarge"\b/, to: '$1="1000"' },
+  { from: /(spacing|padding|margin|sideOffset)="XXXLarge"\b/, to: '$1="1200"' }, // visual breaking change
```
DSil, line 170: We don't need `\b` if we already have `"`

sarkaaa: yeah, sure. fixed. thank you.

DSil (approved):

PR #4465, opened 2024-08-27
Domi/fix required label color

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2098


DSil:
I was actually noticing and I don't know why the label turns to Ink Normal when filled. That should not happen, according to Figma. Can you please sync with designers?

Apparently, Ink Normal should only be used for inline label. Curious enough, we're not using it 😄 I think it can be further fixed in this PR

packages/orbit-components/src/FormLabel/index.tsx
```diff
@@ -45,10 +45,7 @@ const FormLabel = ({
 
     {required && !error && (
       <span
-        className={cx(
-          "text-normal align-top font-bold",
-          filled ? "text-form-element-label-filled-foreground" : "text-critical-foreground",
-        )}
+        className={cx("text-normal align-top font-bold", "text-critical-foreground")}
```
sarkaaa, line 48: 👍

sarkaaa:
besides Daniel's comment, it LGTM

packages/orbit-components/src/FormLabel/index.tsx
mvidalgarcia, line 48: Do we need `cx` now?

domihustinova: Good point, will remove 👍

domihustinova:
I synced with the designers and indeed, normal Label should always be Ink Dark, inline label should always be Ink Normal. For both filled and empty value. I will do the changes.

DSil (approved):
LGTM, good job 👍 Just concerned about the first commit message. It appears cropped to me. And I believe we can simplify to just

```
fix(FormLabel): ensure correct color
```

(Yes, color instead of colour, as we use American English 🤭)

PR #4466, opened 2024-08-27
docs: upgrade migration guide with new border radius value

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2101


docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
DSil, line 1: We need to be careful with this change. This token **does not** exist in version 16. If anything, it'll happen in 16.1.0 or something similar… 

I think it is fair to add it here but we should mention it is only available from version 16.X

Also because once this is merged the migration guide is updated and visible for users but the token is only available on the next release

DSil (commented):
The commit messages **should** mention the new value, so they are more descriptive and clear

docs/src/documentation/05-development/04-migration-guides/01-v16.mdx
sarkaaa: Ok, I see. What do you think about these changes:


- `borderRadius200` (new value of 8px, **available from version 16.X**)
- `rounded-200` (new value of 8px, **available from version 16.X**)

Or should I be even more descriptive in this migration guide?

sarkaaa:
> The commit messages **should** mention the new value, so they are more descriptive and clear

Ok, so my suggestion is:

1. docs: upgrade migration guide with new border radius 200 value
2. feat(tokens): add new borderRadius200 token
3. feat(Box): add new value for borderRadius 200 prop
4. feat(tailwind): add new rounded-200 class

WDYT?

DSil (approved):

PR #4467, opened 2024-08-28
fix(Checkbox): make checkmark visible when tooltip is on

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2097


domihustinova:
Have you consider adding this scenario to visual tests?

sarkaaa:
> Have you consider adding this scenario to visual tests?

actually, that's a good point! I'll cover it, thx.

packages/orbit-components/src/primitives/TooltipPrimitive/index.tsx
```diff
@@ -20,7 +20,7 @@ export const TooltipWrapper = React.forwardRef<
       className={cx(
         "orbit-tooltip-wrapper",
         "max-w-full cursor-auto",
-        "focus:outline-none active:outline-none [&_:disabled]:pointer-events-none",
+        "focus:outline-none active:outline-none [&_:disabled]:pointer-events-none peer-checked:[&_svg]:visible",
```
domihustinova, line 23: Is it a good idea to add this to Tooltip directly? Wouldn't it be better to fix it in the `Checkbox` component?

I assume this part `[&>svg]:invisible peer-checked:[&>svg]:visible` on the `"orbit-checkbox-icon-container"` (line 80) element doesn't work because because element with `peer` class (the input) is not a sibling of it but sibling of its parent.

Maybe we could use the `checked` value and set it on `"orbit-checkbox-icon-container"` like this `checked ? "[&>svg]:visible" : "[&>svg]:invisible"` instead of the current one? WDYT?

sarkaaa: I wanted to keep the "peer-checked" classes, but it makes sense to add a `checked` condition as well. With this approach, we can be sure that the checked icon will be visible even if the component in `tooltip` prop won't be `<Tooltip>`. 👍

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,22 @@
...
 
 test.describe("visual Checkbox", () => {
   test("Checkbox", async ({ mount }) => {
     const component = await mount(<CheckboxStory />);
 
+    component.getByRole("button").hover();
+    await expect(component).toHaveScreenshot();
+  });
+
+  test("Checkbox - active tooltip on mobile", async ({ mount, browserName }) => {
```
sarkaaa, line 14: I added this test to distinguish click on tooltip for mobile devices (hover doesn't work for mobile as the tooltip is like "modal window")

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,22 @@
 import * as React from "react";
 import { test, expect } from "@playwright/experimental-ct-react";
 
-import CheckboxStory from "./Checkbox.ct-story";
+import { CheckboxStory, CheckboxWithTooltipStory } from "./Checkbox.ct-story";
 
 test.describe("visual Checkbox", () => {
   test("Checkbox", async ({ mount }) => {
```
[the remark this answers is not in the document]
sarkaaa: I was also considering this, but I wanted to avoid a huge number of screenshots. But it makes more sense from the point of the code "readability". 🤔 👍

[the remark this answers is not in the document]
sarkaaa: and what do you think about the `test.skip` command? I haven't found any better way how to solve this problem.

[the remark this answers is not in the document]
domihustinova: I like the `test.skip`, I was at first confused as I didn't know you can skip the test based on the condition 👍 TIL

But I am not sure about the condition in this case. I see we are setting different devices/viewports in the config, isn't there a way to access this info inside the test and act on that info?

sarkaaa:
As we discussed during the day, I've added tests for various types of checkboxes + tooltip (`disabled`, `checked`, `hasError`, `readOnly`) to cover as much as possible.
cc @domihustinova

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,91 @@
...
+    await expect(component).toHaveScreenshot();
+  });
+
+  test("Checkbox - hasError, active tooltip on mobile", async ({ mount, browserName }) => {
+    test.skip(browserName !== "webkit", "This feature is mobile and tablet only");
+    const component = await mount(<CheckboxWithTooltipStory hasError />);
 
+    await component.getByRole("button").click();
     await expect(component).toHaveScreenshot();
   });
```
domihustinova, line 90: Would it make sense to add one more scenario with checked and disabled states?

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
domihustinova: Is this behaviour for darwin/linux random?

You are mentioning that `tablet` and `largeMobile` work differently, however there are several more discrepancies for other viewports as well. E.g. 

- `readOnly` has missing also `smallMobile` for linux (next to `tablet` and `largeMobile`)
- `hasError` has missing snapshots for both linux and darwin for `smallMobile` and weirdly on darwin for `mediumMobile` (it's there for linux)
- `checked` has it missing also for `mediumMobile` for darwin and `smallMobile` for linux (next to `tablet` and `largeMobile`)
- `disabled` has all snapshots correct

This randomness is worrying :(

sarkaaa:
it took some time, but in the end, I replaced `browserName` condition with `viewport.width`. With that, we can distinguish small/medium mobile and largeMobile/tablet, which have different type of a tooltip component.

Then, I wrapped checkbox click command & check of visibility of tooltip component to `expect` function. I've seen this approach in tutorial, this allow us to repeat assertions without using wait command with fixed value.

```
await expect(async () => {
      await component.getByRole("button").click();
      await expect(page.getByRole("tooltip")).toBeVisible();
    }).toPass();
```

checkbox with checked & disabled state is also covered. ✅

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
...
+      "This feature is largeMobile, tablet and desktop-only",
+    );
+
+    const component = await mount(<CheckboxWithTooltipStory hasError />);
+
+    await component.getByRole("button").hover();
+    await expect(component).toHaveScreenshot();
+  });
+
+  test("Checkbox - checked, disabled tooltip on desktop", async ({ mount, viewport }) => {
```
domihustinova, line 69: Shouldn't this be `checked, disabled, active tooltip on desktop`? Same below.

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
...
+  const largeMobileBreakpoint = defaultTokens.breakpointLargeMobile;
+
   test("Checkbox", async ({ mount }) => {
     const component = await mount(<CheckboxStory />);
+    await expect(component).toHaveScreenshot();
+  });
+
+  test("Checkbox - disabled, active tooltip on desktop", async ({ mount, viewport }) => {
+    /**
+     * This command skips the test if the viewport width is bigger than largeMobile.
```
domihustinova, line 17: Smaller?

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
...
+    const component = await mount(<CheckboxWithTooltipStory hasError />);
+    const tooltip = await page.getByRole("tooltip");
+
+    await component.getByRole("button").click();
+
+    await expect(tooltip).toBeVisible();
+    await expect(component).toHaveScreenshot();
+  });
+
+  test("Checkbox - checked, disbled tooltip on mobile", async ({ mount, viewport, page }) => {
```
domihustinova, line 143: You have a typo here in `disabled` :D

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
...
+
+  test("Checkbox - disabled, active tooltip on desktop", async ({ mount, viewport }) => {
+    /**
+     * This command skips the test if the viewport width is bigger than largeMobile.
+     * Condition inside is based on the viewport width,
+     * which is specified for each of the device in playwright-ct.config.ts file.
+     * Using that we can distinguish mobileSmall, mobileMedium devices and larger devices.
+     */
+    test.skip(
+      viewport !== null && viewport.width < largeMobileBreakpoint,
```
domihustinova, line 23: So thanks to changing this it's not random anymore and it works as expected?

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
...
+      "This feature is largeMobile, tablet and desktop-only",
+    );
+
+    const component = await mount(<CheckboxWithTooltipStory disabled />);
+
+    await component.getByRole("button").hover();
+    await expect(component).toHaveScreenshot();
```
domihustinova, line 30: You said you added assertion to check the visibility of tooltip but you added it only to mobile scenarios, it's not necessary here? Why it's necessary for the mobile?

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,157 @@
 import * as React from "react";
 import { test, expect } from "@playwright/experimental-ct-react";
+import { defaultTokens } from "@kiwicom/orbit-design-tokens";
 
-import CheckboxStory from "./Checkbox.ct-story";
+import { CheckboxStory, CheckboxWithTooltipStory } from "./Checkbox.ct-story";
 
 test.describe("visual Checkbox", () => {
+  const largeMobileBreakpoint = defaultTokens.breakpointLargeMobile;
```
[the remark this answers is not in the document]
sarkaaa: yeah, I think it could work, I was a little bit tired that evening. 😄  🤯

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
sarkaaa: ok, so let me explain how it works further.

we have two types of how the tooltip can be shown:
- "usual" design of tooltip - just classic tooltip component (shown on `desktop`, `tablet`, `largeMobile`)
image

- tooltip shown in the bottom part of the viewport, visible on click - this is implemented for `smallMobile` and `mediumMobile`

image



**Why is this assertion set up only for `smallMobile` and `mediumMobile`?**
as I mentioned above, the tooltip for these display widths is shown after click, and sometimes the click is slower. Without that assertion (which waits for the visibility of respective component) the screenshot command could be called before the tooltip was really visible. In the case of display > `largeMobile`, we have tooltip shown on hover and there is not needed to wait as we are sure, that the tooltip is shown immediately.

And the reason why I replaced `browserName` condition by `viewport.width < largeMobileBreakpoint` is that `browserName` included `largeMobile` and `tablet` (these widths have a tooltip shown on hover). This way of displaying the tooltip causes there is attribute `role="tooltip"` twice and then the mentioned assertion fails, which is not the desired state (of course :D).

If there is something that's not still clear, feel free to ask.

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
sarkaaa: fixed (and also typos above).

domihustinova (approved):
LGTM 👏

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,161 @@
...
+    /**
+     * This command skips the test if the viewport width is smaller than largeMobile.
+     * Condition inside is based on the viewport width,
+     * which is specified for each of the device in playwright-ct.config.ts file.
+     * Using that we can distinguish mobileSmall, mobileMedium devices and larger devices.
+     */
+    test.skip(
+      viewport !== null && viewport.width < breakpointLargeMobile,
+      "This feature is largeMobile, tablet and desktop-only",
+    );
```
mvidalgarcia, line 25: What do you think about keeping something like this at the beggining of the tests and reuse it across?

```ts

const skipIfViewportSmallerThan = (viewport, width, message) => {
  test.skip(
    viewport !== null && viewport.width < width,
    message,
  );
};

const skipIfViewportLargerThanOrEqual = (viewport, width, message) => {
  test.skip(
    viewport !== null && viewport.width >= width,
    message,
  );
};
  ```

File: packages/orbit-components/src/Checkbox/Checkbox.ct-story.tsx [hunk 12]
```diff
@@ -1,19 +1,43 @@
...
+  readOnly?: boolean;
+}) {
+  return (
+    <div className="space-y-200 lm:pt-1200 lm:min-h-800 flex min-h-[700px] flex-col">
+      <Checkbox
+        label="Check this box"
+        checked={checked}
+        hasError={hasError}
+        disabled={disabled}
+        readOnly={readOnly}
```
mvidalgarcia, line 38: Do we need these?

sarkaaa: removed.

packages/orbit-components/src/Checkbox/Checkbox.ct.tsx
```diff
@@ -1,12 +1,63 @@
...
+    /**
+     * This command skips the test if the viewport width is smaller than largeMobile.
+     * Condition inside is based on the viewport width,
+     * which is specified for each of the device in playwright-ct.config.ts file.
+     * Using that we can distinguish mobileSmall, mobileMedium devices and larger devices.
+     */
+    skipIfViewportSmallerThan(
+      viewport,
+      breakpointLargeMobile,
+      "This feature is largeMobile, tablet and desktop-only",
```
mvidalgarcia, line 33:
```suggestion
      "This feature is largeMobile, tablet and desktop only",
```

mvidalgarcia (approved):
💯

PR #4469, opened 2024-08-30
Update icons unicode

DSil:
Some icons were re-fetched from Figma to remove the duplicated ids.
Repeated ids can only appear in different Figma groups. We currently only have thre Figma groups: colored, "normal" and "light" icons. The first is not really changed often, so we don't expect any id clash any time soon. If there is and ID clash at some point, we can manually change any ID.

This should not have any impact on users

This change was requested by mobile team, because unicode chars need to be unique

FEPLT-2085


mvidalgarcia (approved):

mvidalgarcia:
@DSil Found [this](https://kiwicom.atlassian.net/browse/FEPLT-2004) while scrolling through our backlog, is it related?

PR #4472, opened 2024-09-05
Migrate Stories to CSF3 - Batch 6

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2077


packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -18,129 +15,113 @@ enum PRESETS {
...
-      rows={rows}
-      spaceAfter={spaceAfter}
-      viewBox={viewBox}
-      width={width}
-    />
-  );
-};
-
-Playground.story = {
+export const Playground: Story = {
```
sarkaaa, line 37: I would move `Playground` and `RTL` as the last stories as it's in the previous batches we did before.

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -37,231 +38,233 @@ const Content = (
...
-                <Text size="small" type="secondary">
-                  Sat, Mar 31 Trip length: 1h55m
-                </Text>
-                <Stack direction="row" spacing="100" align="center">
-                  <Text weight="bold">London LHR</Text>
-                  <FlightDirect size="small" />
-                  <Text weight="bold">Prague PRG</Text>
+export const WithinModal: Story = {
+  render: () => (
+    <Modal onClose={action("onClose")} fixedFooter>
```
sarkaaa, line 82: I would add a function to close/open modal to be able to copy the code. :D

packages/orbit-components/src/SmartPassIllustrations/SmartPassIllustrations.stories.tsx
```diff
@@ -10,63 +14,51 @@ const SIZE_OPTIONS = {
   DISPLAY: "display",
 };
 
-export const Default = () => {
-  return <SmartPassIllustrations.SmartPassV1 />;
+type SmartPassPropsAndCustomArgs = React.ComponentProps<typeof SmartPassV1> & {
```
sarkaaa, line 17: I'm thinking if it wouldn't be (maybe) better to replace `React.ComponentProps<typeof SmartPassV1>` by general `Props` imported from `./types`:
```
type SmartPassPropsAndCustomArgs = Props & {  illustration: string; };
```

The functionality is still the same, in this story, we're switching between 5 similar components (`SmartPassV1,...SmartPassV5`), it's just a visual thing within the file to make the code a little bit simpler (?).

Point to the discussion. :D

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
sarkaaa, line 37: `Preset` arg is missing in Playground story.

domihustinova: I didn't include it there, as it would rewrote the other args so it would be as in the other stories when setting `mediumMobile`, `tablet` etc. that it overwrote the individual properties. There is a separate story for `Preset`.

packages/orbit-components/src/Slider/Slider.stories.tsx
```diff
@@ -1,476 +1,278 @@
...
-  step,
-}) => {
-  const [times, setTimes] = React.useState<Value>(defaultValue);
-  const valueDescription = `${times[0]}:00 - ${times[1]}:00`;
-  return (
-    <div style={{ backgroundColor: "#f1f5f7", padding: "24px" }}>
+    return (
       <Slider
+        {...args}
+        onChangeAfter={action("onChangeAfter")}
```
sarkaaa, line 65: why is not `onChangeAfter` in the args and defined "statically" in the code?

packages/orbit-components/src/Slider/Slider.stories.tsx
```diff
@@ -1,476 +1,278 @@
...
-        ariaLabel={ariaLabel}
-        ariaValueText={ariaValueText}
-        dataTest={dataTest}
-        defaultValue={defaultValue}
-        label={label}
-        maxValue={maxValue}
-        minValue={minValue}
-        step={step}
-        onChange={action("onChange")}
         onChangeBefore={action("onChangeBefore")}
```
sarkaaa, line 158: same as the comment before here with `onChangeBefore` and `onChangeAfter`.

packages/orbit-components/src/Stack/Stack.stories.tsx
```diff
@@ -37,845 +38,355 @@ import ChoiceGroup from "../ChoiceGroup";
...
-    basis,
-    align,
-    justify,
-    spacing,
-    spaceAfter,
-  };
-
-  return (
-    <Stack largeMobile={largeMobile}>
+export const MediumMobileProperties: Story = {
```
sarkaaa, line 131: Have you considered to squash these 'viewport' stories?
The code is redundant, I was thinking about the solution, and I found something like this:

```
export const MediumMobileProperties: Story &
  StoryObj<{
    largeDesktop: boolean;
    desktop: boolean;
    tablet: boolean;
    largeMobile: boolean;
    mediumMobile: boolean;
  }> = {
  render: ({ largeDesktop, desktop, tablet, largeMobile, mediumMobile, ...args }) => {

    return (
      <Stack
        largeDesktop={largeDesktop && { ...args }}
        desktop={desktop && { ...args }}
        tablet={tablet && { ...args }}
        mediumMobile={mediumMobile && { ...args }}
        largeMobile={largeMobile && { ...args }}
        {...args}
      >
        ...
      </Stack>
    );
  },

  args: {
    direction: DIRECTIONS.COLUMN,
    largeDesktop: false,
    desktop: false,
    tablet: false,
    largeMobile: false,
    mediumMobile: false,
  },
};
```

but the "final" code in SB might look messy for the user. :(

packages/orbit-components/src/SmartPassIllustrations/SmartPassIllustrations.stories.tsx
domihustinova: Good idea.

I also removed `component: SmartPassV1,` as per docs, _The default export must contain a title property that can be read statically or a component property from which an automatic title can be computed._, so I believe in this case the title property is enough and we don't have to then import `SmartPassV1` at all.

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
domihustinova: Good point, added the same solution as in Modal stories 👍 (we could later extract it so we are not creating the same thing in multiple stories 🎗️ ).

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -37,231 +38,233 @@ const Content = (
...
+      <SkipLink
+        links={[
```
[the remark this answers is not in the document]
domihustinova: Because here the controls are disabled and it was like this 😅 But it's true that elsewhere where we are disabling controls, we have it in args anyway, moved.

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
domihustinova: I moved those two stories down to keep consistency. But I am not really sure about reusing the args and argTypes, it's not saving anything, it's just three values 😅

packages/orbit-components/src/Stack/Stack.stories.tsx
domihustinova: No I haven't consider squashing them. If we do it like this, then you cannot set those values separately for `mediumMobile`, `tablet` etc. as they all are using the same value as the default mobile properties so the only added value is that you can see it in the code source. And for this reason we didn't include those `mediumMobile`, `tablet` etc. in the Playground stories in some other batch.

Maybe it's redundant to have all those stories and the Mobile one would be sufficient, tbh I don't know anymore 😅

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
sarkaaa: I understand your reason, I was pointing to a consistency, as we have the same approach in other stories (if it is not forgotten :D ), where we also "save" just 2-3 args. :D

packages/orbit-components/src/Stack/Stack.stories.tsx
sarkaaa: yes, yes. I was also thinking about it and I haven't found any reasonable solution.
At first, my thoughts went to `Hide` component stories, but there is slightly different approach.

I would keep it as it is. 👍

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -18,129 +15,113 @@ enum PRESETS {
...
-      rowOffset={rowOffset}
-      rows={rows}
-      spaceAfter={spaceAfter}
-      viewBox={viewBox}
-      width={width}
-    />
-  );
-};
+export const Default: Story = {
+  render: () => <Skeleton height="150px" width="500px" />,
```
sarkaaa, line 27: We don't have to define `width` and `height` at all as these params are optional and the component is visible without setting them up (and if yes, these props could be set as args and controls could be enabled from my POV).

packages/orbit-components/src/Slider/Slider.stories.tsx
```diff
@@ -1,476 +1,271 @@
...
-  label: "Depart from Prague",
-  maxValue: 24,
-  minValue: 1,
-  step: 1,
-  valueDescription: "From midnight to 12:00 pm",
+  args: {
+    ...SliderWithLoadingHistogram.args,
+    ariaLabel: "Label of handle",
+    ariaValueText: "Text alternative of actual value",
+    valueDescription: "From midnight to 12:00 pm",
```
sarkaaa, line 173: prop `id` is missing there.

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
sarkaaa: - https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Collapse/Collapse.stories.tsx
- https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/CarrierLogo/CarrierLogo.stories.tsx

reusing Playground stories (yes, it's 4 args, not 3 :D ):
- https://github.com/kiwicom/orbit/blob/master/packages/orbit-components/src/Inline/Inline.stories.tsx

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
domihustinova: Good point 👍

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
sarkaaa: Maybe I'm missing something, but is there any reason why `preset` is not included in Playground story?

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -37,231 +38,257 @@ const Content = (
...
-      name: "Reguest a refund",
-      onClick: action("Reguest a refund"),
-    },
-  ],
-  buttonLabel:
-    "Description for screen readers about where the user currently is or describing a more complex component.",
-  dataTest: "test",
-  id: "ID",
-};
+export const WithinModal: Story = {
```
sarkaaa, line 65: Story is not scrollable (can't get the code).

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -37,231 +38,257 @@ const Content = (
...
+                  }
+                >
+                  <Stack direction="row" align="center" justify="between" spacing="200">
+                    <CarrierLogo carriers={[{ code: "FR", name: "Ryanair" }]} size="large" />
+                    <Stack spacing="100">
+                      <Text size="small" type="secondary">
+                        Sat, Mar 31 Trip length: 1h55m
+                      </Text>
+                      <Stack direction="row" spacing="100" align="center">
+                        <Text weight="bold">London LHR</Text>
```
sarkaaa, line 145: Where is this whole block of code in modal? :D
It's not visible there at all (maybe there's something I didn't catch, but I am confused :D :D)

image

packages/orbit-components/src/SkipNavigation/SkipNavigation.stories.tsx
```diff
@@ -55,55 +56,60 @@ const Content = (
...
-          {
-            name: "Request refund",
-            onClick: action("Request refund"),
-          },
-        ]}
-        feedbackUrl="#"
-        feedbackLabel={feedbackLabel}
-      />
-
+export const Playground: Story = {
```
sarkaaa, line 83: `id` arg is missing here. :D

packages/orbit-components/src/Skeleton/Skeleton.stories.tsx
sarkaaa: Solved, my bad. :D

packages/orbit-components/src/Stack/Stack.stories.tsx
```diff
@@ -37,845 +38,355 @@ import ChoiceGroup from "../ChoiceGroup";
...
-      spacing={spacing}
-      legacy={legacy}
-      spaceAfter={spaceAfter}
-      mediumMobile={mediumMobile}
-      largeMobile={largeMobile}
-      tablet={tablet}
-      desktop={desktop}
-      largeDesktop={largeDesktop}
-    >
+export const Playground: Story = {
```
sarkaaa, line 218: `id` is also missing here.
also, `as` args is not included either, I am considering if we would like to included this arg to our controls (if I remember correctly, we didn't include for some component, bcs there was not visible change (Button, I guess (??))

packages/orbit-components/src/SkipLink/SkipLink.stories.tsx
domihustinova: Fixed ✅

packages/orbit-components/src/Stack/Stack.stories.tsx
domihustinova: Yeah, I guess we can include it 🤔 ✅

sarkaaa:
LGTM. 🎉

sarkaaa (approved):

PR #4479, opened 2024-09-17
FEPLT-2113: Fix Tab color when not active

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2113


packages/orbit-components/src/Tabs/Tabs.ct.tsx
```diff
@@ -49,4 +49,11 @@ test.describe("visual Tabs", () => {
     const component = await mount(<TabsComponentStory fullWidth />);
     await expect(component).toHaveScreenshot();
   });
+
+  test("screenshot for default with one selected", async ({ mount }) => {
```
DSil, line 53: I'd say we can avoid extra screenshots and add a new Tab to the already existing story. I'd also say we can skip the click test. WDYT?

DSil (commented):
Just a question about the tests. Also, don't forget to update the linux screenshots 🙂

DSil (approved):

PR #4484, opened 2024-09-24
Migrate Stories to CSF3 - Batch 8

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2079

### DIFF
--- a/packages/orbit-components/src/Tile/Tile.stories.tsx
+++ b/packages/orbit-components/src/Tile/Tile.stories.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { action } from "@storybook/addon-actions";
+import type { Meta, StoryObj } from "@storybook/react";
 
-import TileGroup from "../TileGroup";
 import * as Icons from "../icons";
 import Stack from "../Stack";
 import Text from "../Text";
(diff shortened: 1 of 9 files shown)


packages/orbit-components/src/Wizard/Wizard.stories.tsx
```diff
@@ -1,156 +1,140 @@
...
-      <WizardStep title="Ticket fare" />
-      <WizardStep title="Kiwi.com guarantee" isCompleted />
-      <WizardStep title="Seating" />
-      <WizardStep title="Overview & Payment" />
+export default meta;
+type Story = StoryObj<WizardPropsAndCustomArgs>;
+
+export const Default: Story = {
+  render: ({ completedSteps, steps, ...args }) => (
+    <Wizard {...args} activeStep={completedSteps} completedSteps={completedSteps}>
```
domihustinova, line 62: I excluded `activeStep` from controls/args as before we have been setting argType for `activeStep` dynamically based on the value of `completedSteps` and `steps`. Unfortunately, this is not possible with CSF3 (and it seems to me that it's not working currently on master either 🤔).

I also hardcoded `steps` because we have also been setting value of `completedSteps` dynamically based on the length of `steps`. With hardcoded value of steps, we can just set argType of `completedSteps`. And then we can calculate value of `activeStep` which can be `<= completedSteps` but I think it's fine to always have `= completedSteps`.

packages/orbit-components/src/Tile/Tile.stories.tsx
```diff
@@ -14,284 +14,209 @@ import Tile from ".";
...
-  icon: {
-    options: Object.keys(Icons),
-    control: {
-      type: "select",
-    },
+  args: {
+    href: "https://www.kiwi.com/",
+    external: false,
+    id: "ID",
+    icon: "Airplane",
```
sarkaaa, line 180: What do you think about inheriting icon arg from the story above?
Also, I would add `undefined` option to the list of icons to give the users the possibility to show `Tile` component without any icon in the Playground.

packages/orbit-components/src/Tile/Tile.stories.tsx
```diff
@@ -14,284 +14,209 @@ import Tile from ".";
...
+    title: "Tile with title",
+    expanded: false,
+    description:
+      "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco.",
+    header: "",
+    children: "",
+    noPadding: false,
```
sarkaaa, line 187: in the Playground, `noPadding` is applied only for the "children" element, but it could also be applicable for "parent". I couldn't set it for the parent, but it should be working (I added `noPadding` for the first default story):

image

Can you take a look at it and include it to the Playground as well? 🙏

packages/orbit-components/src/Tile/Tile.stories.tsx
```diff
@@ -14,284 +14,209 @@ import Tile from ".";
...
-    options: Object.keys(Icons),
-    control: {
-      type: "select",
-    },
+  args: {
+    href: "https://www.kiwi.com/",
+    external: false,
+    id: "ID",
+    icon: "Airplane",
+    title: "Tile with title",
```
sarkaaa, line 181: What do you think about spreading `DefaultWithHeaderPropsAsHref` here?

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
```diff
@@ -14,272 +15,265 @@ import TileGroup from ".";
...
-      <Tile
-        onClick={action("clicked")}
-        icon={Icon && <Icon />}
-        title={title}
-        description={description}
-      />
-    </TileGroup>
-  );
-};
+export const DefaultWithHeaderProps: Story = {
```
sarkaaa, line 50: I would add the `noHeaderIcon` arg as it's related to the "header props".

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
```diff
@@ -14,272 +15,265 @@ import TileGroup from ".";
...
+  render: ({ title, description, icon }) => {
+    const Icon = getIcon(icon);
 
-DefaultWithHeaderProps.story = {
-  name: "Default - with header props",
+    return (
+      <TileGroup>
+        <Tile
+          onClick={action("clicked")}
```
sarkaaa, line 57: We can use `...args` here.

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
```diff
@@ -14,272 +15,265 @@ import TileGroup from ".";
...
-export const DefaultJustWrapper = ({ dataTest, as, content }) => {
-  return (
-    <TileGroup as={as} dataTest={dataTest}>
+export default meta;
+type Story = StoryObj<TileGroupPropsAndCustomArgs>;
+
+export const DefaultJustWrapper: Story = {
+  render: ({ content }) => (
+    <TileGroup>
       <Tile onClick={action("clicked")}>{content}</Tile>
```
sarkaaa, line 34: I would set `onClick={actions("clicked"))}` as a general `args` as it's used multiple times within the file.

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
```diff
@@ -14,272 +15,265 @@ import TileGroup from ".";
...
-              <Stack direction="row" align="center" spacing="200">
-                <Heading type="title4" as="h4">
-                  Mr. Hot potato
-                </Heading>
-                <CountryFlag code="cz" />
+          }
+        >
+          This is example of expanded content
+        </Tile>
+        <Tile
```
sarkaaa, line 157: There's code duplication of `Tile` component in each story. I would use `Array.from(...)` and get rid of redundant code.

packages/orbit-components/src/Timeline/Timeline.stories.tsx
```diff
@@ -116,107 +114,49 @@ export const WithWarning = ({ label, time, type, children }) => {
...
-
-WithCritical.args = {
-  label: "Non refundable",
-  time: "7th May 10:30",
-  type: Types.Critical,
-  children: "The carrier rejected the refund",
-};
+  args: {
+    children: "We’ve applied for a refund from the carrier(s).",
+    label: "Waiting for the carrier",
```
sarkaaa, line 125: I would add `{ table: { category: xxx } }` to distinguish args for `Timeline` and `TimelineStep` components. I know that we were discussing this during the process on the last batched, we can include it to the `final fixes` ticket.

packages/orbit-components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -188,127 +151,133 @@ export const WithImageInside = ({ size, placement }) => {
...
+    content:
+      "Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. Sed ac dolor sit amet purus malesuada congue. Sed vel lectus.",
+    id: "ID",
+    labelClose: "Close",
+    lockScrolling: false,
+    renderInPortal: true,
+    stopPropagation: false,
+    removeUnderlinedText: false,
+    block: false,
+    noFlip: false,
```
sarkaaa, line 235: I looks that `noFlip` and `offset` props are not used in `Tooltip` component at all.

packages/orbit-components/src/primitives/MobileDialogPrimitive/MobileDialogPrimitive.stories.tsx
```diff
@@ -1,49 +1,43 @@
...
-  stopPropagation: true,
-  removeUnderlinedText: false,
-  labelClose: "Close",
+export default meta;
+type Story = StoryObj<typeof MobileDialogPrimitive>;
+
+export const Playground: Story = {
+  render: args => (
+    <MobileDialogPrimitive {...args}>
+      <Button>Open Mobile Dialog</Button>
```
sarkaaa, line 40: I'm thinking if using `Text` component instead of the `Button` wouldn't be more suitable to demonstrate how `removeUnderlinedText` arg works (?). With this current setup, this arg in control panel does nothing.

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
```diff
@@ -1,100 +1,75 @@
...
-  desktop: {
-    columns: "repeat(2, minmax(100px, 1fr))",
-    rows: "repeat(4, 40px)",
-    gap: "40px",
+  args: {
+    divsCount: 8,
+    className: "",
+    as: "div",
+    width: "100%",
+    columns: "",
```
sarkaaa, line 19: in case I fill any value(eg. `repeat(2, 1fr)`), columns are not set and there's blank space.

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
```diff
@@ -1,100 +1,75 @@
...
-  },
-  tablet: {
-    columnGap: "40px",
-  },
-  desktop: {
-    columns: "repeat(2, minmax(100px, 1fr))",
-    rows: "repeat(4, 40px)",
-    gap: "40px",
+  args: {
+    divsCount: 8,
```
sarkaaa, line 15: `divsCount` are not visible in case it's higher then number of rows on `row`. I understand that it can be tricky, in that way I would add a note to `info` and inform the user. WDYT?

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
```diff
@@ -1,100 +1,75 @@
...
+    },
+    largeMobile: {
+      columns: "repeat(4, 1fr)",
+      rows: "repeat(2, 40px)",
+      gap: "20px",
+    },
+    tablet: {
+      columnGap: "40px",
+    },
+    desktop: {
```
sarkaaa, line 38: If I understand correctly, you don't fill all props (columns, columnGap, rows, etc..) to all of the breakpoints, just an example. Maybe I would add other props (maxWidth, width, inline) to one of these breakpoint views at least (?). WDYT?

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
```diff
@@ -1,100 +1,75 @@
...
+  args: {
+    divsCount: 8,
+    className: "",
+    as: "div",
+    width: "100%",
+    columns: "",
+    rows: "repeat(8, 40px)",
+    columnGap: "",
+    rowGap: "10px",
+    gap: "",
```
sarkaaa, line 23: Maybe I didn't catch something, but `gap` is not working at all (even if I remove values for `columnGap`, `rowGap`).

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
```diff
@@ -1,100 +1,75 @@
 import * as React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
 
 import { SPACINGS_AFTER } from "../../common/consts";
 
 import Grid from ".";
 
-export default {
```
sarkaaa, line 7: Overall, SB for this component is a little bit difficult to understand. 😢  Have you tried different ways how to achieve the goal? 🤔

packages/orbit-components/src/Tile/Tile.stories.tsx
domihustinova: Above? And only icon? Not sure which one do you mean but I inherited it from `DefaultWithHeaderPropsAsHref` as it has already 6 args in total that could be replaced. `undefined` option also added ✅

packages/orbit-components/src/Tile/Tile.stories.tsx
domihustinova: No, it cannot be included. In your example you don't have header for the tile (= title, description, icon, header) so `noPadding` is applied to the same element as in the Playground which is `TileContent` component.

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
domihustinova: I didn't want to add any more of `Tile` args as, in the end, it's `TileGroup` story. But as agreed, I kept only the Playground and RTL story and since `header` is hardcoded there as ReactNode, I kept only `children` arg from `Tile` component. I hope it makes sense like this.

packages/orbit-components/src/TileGroup/TileGroup.stories.tsx
domihustinova: I did this for both Playground and RTL stories.

packages/orbit-components/src/Tooltip/Tooltip.stories.tsx
domihustinova: I got carried away 😅

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
domihustinova: I added more info about this, I hope it's sufficient like this :)

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
domihustinova: Yes, but I didn't make any changes to the previous setup. I removed these props from almost all breakpoints and put everything into `largeDesktop` so it's easier to edit something, WDYT?

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
domihustinova: It was because of this `vars["--grid-gap"] != null && "gap-[var(--grid-gap)]"` (and similar for other props) part of the code - we were setting either value or, if the value was deleted from controls, empty string, which is not null and therefore applied as empty string. I added condition to either pass the value or `undefined`. Now all args should work, please check 🙏

packages/orbit-components/src/utils/Grid/Grid.stories.tsx
domihustinova: I think now with added default values, extra info and actually working controls it's like any other story, what do you think?

sarkaaa:
I think the `Grid` story looks much better now. Just one nitpick, maybe I would add another "info" note for the user, something like:
```
Be aware of each viewport breakpoint and its screen min width value.
```
soo:
_You can try all possible configurations of this component. When changing the value of rows, columns or divsCount, the sum of allowed rows and columns needs to be higher than divsCount. Be aware of each viewport breakpoint and its screen min width value. Visit Orbit.Kiwi for more detailed guidelines._

I'm not sure if this sentence is 100% suitable, but the point overall is to inform the user that there are multiple viewports (mediumMobile...largeDesktop) and the changes in SB values don't have to be present as the user has switched to different device view. :D

sarkaaa (approved):

PR #4488, opened 2024-10-02
FEPLT-2123: fix(Accordion): spacing with no actions was wrong

DSil:
When it had no actions and was not expandable, we were still giving space for actions, leaving the spacing inconsistent

### DIFF
--- a/packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.tsx
+++ b/packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.tsx
@@ -23,19 +23,18 @@ const AccordionSectionHeader = ({
   <div
     className={cx(
       "p-600 bg-white-normal flex items-center",
-      expanded ? "min-h-[19px]" : "min-h-[44px]",
+      expanded ? "min-h-[19px]" : "min-h-form-box-normal",
     )}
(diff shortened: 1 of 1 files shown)


packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.tsx
```diff
@@ -23,19 +23,18 @@ const AccordionSectionHeader = ({
   <div
     className={cx(
       "p-600 bg-white-normal flex items-center",
-      expanded ? "min-h-[19px]" : "min-h-[44px]",
+      expanded ? "min-h-[19px]" : "min-h-form-box-normal",
```
RobinCsl, line 26: It's just replacing a custom value with an already existing token (and semantically it makes sense). Right?

RobinCsl (approved):

packages/orbit-components/src/Accordion/AccordionSection/components/SectionHeader.tsx
DSil: Although the linter suggests so, we want to refrain from using component specific tokens outside of their context. In this case, a token for Forms should not be used in Accordion

PR #4489, opened 2024-10-03
FEPLT-2124: add X icon and deprecate Twitter icon

DSil:
The `Twitter` icon is now deprecated and the `X` icon was added.


sarkaaa:
I have a question related to `BREAKING CHANGE`, as I am not sure about it. The commit message doesn't include BC message, because it's deprecation, not removal, so the "customer's" packages won't be broken by the upgrade, they will just use the deprecated icon, right?

docs/src/documentation/05-development/04-migration-guides/01-v17.mdx
```diff
@@ -14,4 +14,10 @@ With this guide, we aim to walk through all the breaking changes and how they ca
 
 ### ColoredTwitter icon removed
 
-The `ColoredTwitter` icon was removed. The `Twitter` icon should be used instead.
+The `ColoredTwitter` icon was removed. The `X` icon should be used instead.
```
RobinCsl, line 17: Given this, I'm wondering if we should just bundle the removal with the `Twitter` icon too.

packages/orbit-components/src/icons/svg/twitter.svg
```diff
@@ -1,4 +1,4 @@
-<!--character:1007-->
+<!--character:0007-->
```
RobinCsl, line 1: Curious about this change, do you know what caused this?

RobinCsl (approved):

docs/src/documentation/05-development/04-migration-guides/01-v17.mdx
DSil: I thought the same… We can… I have nothing against it. And we avoid the deprecation entirely?

packages/orbit-components/src/icons/svg/twitter.svg
DSil: I manually changed this 🙈 we can't have repeated `character` ids, but they are not really "scientific". And since we were gonna remove it soon, I just changed this to an unexisting number

docs/src/documentation/05-development/04-migration-guides/01-v17.mdx
DSil: Agreed IRL that we will avoid the deprecation and just remove the icon in this release as well

packages/orbit-components/src/icons/svg/twitter.svg
RobinCsl: Just for me to understand the relevance: how/where is this information used? :D

RobinCsl (approved):

PR #4492, opened 2024-10-09
Remove safari 14.0 support

DSil:
Given the end of support for Safari <14.1 (14.5 on iOS), we can remove some of the extra styles and logic we were maintaining in Orbit.

The `legacy` prop in `Stack` and `LinkList` components is deprecated and should be replaced by a much more explicit `useMargin` prop. The `legacy` prop should be removed soon.


.github/contribution/testing-conventions.md
```diff
@@ -235,32 +235,42 @@ This part of testing is often abandoned by people, but we are aware of how impor
 
 <!-- AUTO-GENERATED-CONTENT:START (SUPPORTED_BROWSERS) -->
 
-- Chrome for Android 114
-- Chrome 114
-- Chrome 113
-- Chrome 112
-- Chrome 111
+- Chrome for Android 128
```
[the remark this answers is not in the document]
DSil: (I didn't confirm but my best guess is) because these MD files are generated on build, so when it was generated there the inheritance from browserslistrc generated a different output 🤷

packages/orbit-components/src/Slider/components/Handle/index.tsx
```diff
@@ -87,8 +86,7 @@ const Handle = ({
         onTop && "z-20",
       )}
       style={{
-        // TODO use inset-inline-start once we don’t support ios_safari 14
-        [leftRight({ theme })]: `calc(${left}% - ${theme.orbit.space300})`,
+        insetInlineStart: `calc(${left}% - ${theme.orbit.space300})`,
```
sarkaaa, line 89: I'm not sure if my question is not out of the scope of this PR, but I noticed that `leftRight({ theme })` is also used in `Bar/index.tsx` file. I tried to replace it with `insetInlineStart` locally and it seems that it looks the same. Why was not changed this property in that file as well?

DSil: Well spotted. Can be removed indeed. Thanks

sarkaaa (approved):

PR #4494, opened 2024-10-15
FEPLT-2129: Replace react-popper by floating-ui

DSil:
Dependency from react-popper removed and replaced by floating-ui


domihustinova (commented):
I've noticed that for some placements (left-start, right-start) the arrow starts at the start of the tooltip and it doesn't look good 🙈 I've seen it happening in Tooltip Playground only, in TooltipPrimitive it was okay. Can you take a look at that, please?

Snímek obrazovky 2024-10-16 v 19 53 21

packages/orbit-components/src/Popover/Popover.ct.tsx
```diff
@@ -9,7 +9,7 @@ import {
   PopoverOverlapped,
   PopoverLongContent,
 } from "./Popover.ct-story";
-import { PLACEMENTS } from "../common/consts";
+import { PLACEMENTS } from "../common/placements";
```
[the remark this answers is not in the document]
DSil: Yes, well spotted

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
```diff
@@ -64,39 +77,39 @@ const PopoverContentWrapper = ({
...
     ],
   });
 
-  const { popper } = styles;
-
   React.useEffect(() => {
     const timer = setTimeout(() => {
-      if (popoverRef.current) {
-        popoverRef.current.focus();
+      if (refs.floating.current) {
```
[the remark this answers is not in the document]
DSil: I don't have access to `elements` here 🤔 I need the ref in `refs.floating` given to me by the hook

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
```diff
@@ -125,12 +138,12 @@ const PopoverContentWrapper = ({
...
-    "--popper-right": popper.right,
-    "--popper-bottom": popper.bottom,
-    "--popper-transform": popper.transform,
-    "--popper-position": popper.position,
+    "--popover-top": floatingStyles.top ?? 0,
+    "--popover-left": floatingStyles.left ?? 0,
+    "--popover-right": floatingStyles.right ?? "auto",
+    "--popover-bottom": floatingStyles.bottom ?? "auto",
+    "--popover-transform": floatingStyles.transform,
+    "--popover-position": strategy,
```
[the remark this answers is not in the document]
DSil: I guess we could, yes. I don't think it makes a difference. I will change it so I don't need to extract `strategy` from the hook

packages/orbit-components/src/Tooltip/types.d.ts
```diff
@@ -22,4 +21,7 @@ export interface Props extends Common.Globals, Popper {
   readonly removeUnderlinedText?: boolean;
   readonly block?: boolean;
   readonly lockScrolling?: boolean;
+  readonly placement?: Placement | AutoPlacement;
```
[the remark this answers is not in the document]
DSil: Thanks. Crazy how no linter or type checking warns me about that

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.tsx
```diff
@@ -1,67 +1,27 @@
...
+const ARROW_SIZE = 6;
 
-    if (help)
-      return `${placementClasses} ${commonBorderClasses} ${arrowBorderClasses.help[formattedPlacement]}`;
-
-    return `${placementClasses} ${commonBorderClasses} ${arrowBorderClasses[formattedPlacement]}`;
-  }
-
-  return "";
+const getArrowColor = ({ error, help, theme }) => {
```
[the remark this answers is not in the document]
DSil: _Move fast_ 🙈  will add

packages/orbit-components/src/ErrorFormTooltip/Tooltip/index.tsx
```diff
@@ -87,28 +60,23 @@ const ErrorFormTooltip = ({
...
-  }, [onShown, isHelp, helpClosable]);
+  }, [onShown, isHelp, helpClosable, refs.floating]);
 
-  const isVertical =
-    attrs.popper &&
-    (attrs.popper["data-popper-placement"] === "top-start" ||
-      attrs.popper["data-popper-placement"] === "top-end")
-      ? "bottom"
-      : "top";
+  const isVertical = placement === "top-start" || placement === "top-end" ? "bottom" : "top";
```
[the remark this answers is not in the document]
DSil: Yeah, I got the same question as you. I think you are right, because we don't have any auto-positioning in place, but I did not want to risk the logic here while I was implementing the migration. Giving it a second thought now, I guess we can remove and come back if some problem is noticed or reported

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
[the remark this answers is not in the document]
domihustinova: But `elements` can be returned from the `useFloating` the same as `refs`, no?

I am asking because in the docs it's written:

> For usage in Effects, prefer using `elements.floating`, since it’s not guaranteed the floating element will be available on the first pass.

So that's why I was curious why you chose the `refs.floating`.

DSil:
The Tooltip vs TooltipPrimitive difference was quite tricky to spot but I have found out that it was actually the story that misled me, by having `offset = [0, 0]`, instead of the default value of `[0, 5]`, that is the value applied to `Tooltip`, as it does not accept the `offset` prop.

With that, I noticed that I was compensating the arrow size for the main axis, but not the cross axis, that seems to be needed as well. I added this and stories all look good now. But we should have visual tests… I will create a task for it.

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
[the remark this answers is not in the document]
DSil: Alright, I completely missed that. I will follow the docs and change it. Thank you, well spotted!

package.json
```diff
@@ -57,6 +57,7 @@
     "@babel/types": "^7.21.4",
     "@commitlint/cli": "^19.4.0",
     "@commitlint/config-conventional": "^17.0.3",
+    "@floating-ui/react": "^0.26.24",
```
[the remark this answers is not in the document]
domihustinova: I see. I asked because before popper related dependencies weren't there.

packages/orbit-components/src/common/placements.ts
```diff
@@ -0,0 +1,39 @@
...
+  BOTTOM_START = "bottom-start",
+  BOTTOM_END = "bottom-end",
+  RIGHT = "right",
+  RIGHT_START = "right-start",
+  RIGHT_END = "right-end",
+  LEFT = "left",
+  LEFT_START = "left-start",
+  LEFT_END = "left-end",
+}
+export const isFixedPlacement = (placement: string): placement is Placement =>
```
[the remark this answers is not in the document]
sarkaaa: I'm still slightly confused, why string and not `Placement | AutoPlacement`? this function is used in `TooltipContent` and `ContentWrapper`, where the `placement` prop is `Placement | AutoPlacement` anyway. Is it for possible future purposes?

packages/orbit-components/src/common/consts.ts
```diff
@@ -1,21 +1,3 @@
-export enum PLACEMENTS {
```
RobinCsl, line 1: What's the reason for moving this to a different file and separating the `auto` ones from the others?

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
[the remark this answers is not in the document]
domihustinova: You forgot to replace it also in the dependency array 😅

I didn't mention it in other files because I didn't want to open the same discussion in multiple places and I assumed if you would change it here you would change it there as well, my bad 🙈 `refs.floating.current` is also used inside useEffect in `TooltipContent` and `ErrorFormTooltip`. Not sure if you want to change it there as well but I guess it makes sense.

packages/orbit-components/src/common/placements.ts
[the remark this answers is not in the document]
DSil: By having this on a _common place_, I don't expect the function to be aware of where it is called, that is why I prefer to receive a more generic type as arg

packages/orbit-components/src/common/consts.ts
DSil: This was moved to a different file because there were more types and enums to export and so I thought it would make sense to keep them in a single dedicated file.

The separation from `auto` happened because popper's type `Placement` used to include the `auto`. But that is not the case with `floating-ui`, so I needed some workarounds to keep supporting it.
Since the enums map the types, and also because of the `isFixedPlacement`, I thought it would be safer to keep them separated

packages/orbit-components/src/Popover/components/ContentWrapper.tsx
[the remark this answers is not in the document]
DSil: It does, my bad. Changed

packages/orbit-components/src/common/placements.ts
```diff
@@ -0,0 +1,39 @@
+import type { Placement as FloatingPlacement, Alignment } from "@floating-ui/react";
+
+export type AutoPlacement = "auto" | "auto-start" | "auto-end";
+export type Placement = FloatingPlacement;
```
RobinCsl, line 4: Maybe aliasing here would remove some amount of code changes elsewhere
```suggestion
export type FixedPlacement = FloatingPlacement;
export type Placement = FixedPlacement | AutoPlacement;
```

packages/orbit-components/src/common/placements.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export enum AUTO_PLACEMENTS {
+  AUTO = "auto",
+  AUTO_START = "auto-start",
+  AUTO_END = "auto-end",
+}
+
+export enum PLACEMENTS {
```
RobinCsl, line 12:
```suggestion
export enum FIXED_PLACEMENTS {
```

packages/orbit-components/src/common/placements.ts
```diff
@@ -0,0 +1,39 @@
...
+  BOTTOM = "bottom",
+  BOTTOM_START = "bottom-start",
+  BOTTOM_END = "bottom-end",
+  RIGHT = "right",
+  RIGHT_START = "right-start",
+  RIGHT_END = "right-end",
+  LEFT = "left",
+  LEFT_START = "left-start",
+  LEFT_END = "left-end",
+}
```
RobinCsl, line 25:
```suggestion
}
export enum PLACEMENT = {
   ...FIXED_PLACEMENT,
   ...AUTO_PLACEMENT,
}
```

packages/orbit-components/src/common/placements.ts
RobinCsl, line 26:
```suggestion
export const isFixedPlacement = (placement: string): placement is FixedPlacement =>
```

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.stories.tsx
```diff
@@ -12,7 +12,7 @@ import ListItem from "../../List/ListItem";
 import Heading from "../../Heading";
 import Button from "../../Button";
 import RenderInRtl from "../../utils/rtl/RenderInRtl";
-import { PLACEMENTS } from "../../common/consts";
```
RobinCsl, line 15: Now this could stay the same (maybe just changing the path if we feel strongly about it?)

packages/orbit-components/src/common/placements.ts
DSil: I don't think we can spread enums 🤔

packages/orbit-components/src/common/placements.ts
DSil: Yes, I did this one. But the enums I couldn't, because I don't think we can compose enums from other enums 🤔

packages/orbit-components/src/common/placements.ts
RobinCsl: You're right, sorry.

We could imagine instead something like: https://www.typescriptlang.org/play/?ssl=21&ssc=27&pln=1&pc=1#code/KYOwrgtgBAouEEYoG8BQUoEEoF4oHJN8AadKAIVwPJNQF9VVRJZ4AmFMgYSvy9owARXoNoNUAYwD2IAM4AXKAFlgAJwDmwACZwWeNBgB0x3YlJGT7UnSgBDWVGlz5AbkbyAngAdgytZp14ABVvXzxPHykAMz8NbVM3RgB6JKgAVVlbTUkZBSgAN1sAGzBgAC5YgNMQnwBtAGtgD2jK+ODQgF0qFTjAyENMNydZKSLgQyKpdQAKQpLgAEoXKBSCIiA

```suggestion
const _placement = {
   ...FIXED_PLACEMENT,
   ...AUTO_PLACEMENT,
} as const

export type PLACEMENT = typeof _placement
```

DSil: I also couldn't achieve a valid definition

https://www.typescriptlang.org/play/?#code/KYOwrgtgBAouEEYoG8BQUoEEoF4oHJN8AadKAIVwPJNQF9VVRJZ4AmFMgYSvy9owARXoNoNUAYwD2IAM4AXKAFlgAJwDmwACZwWeNBgB0x3YlJGT7UnSgBDWVGlz5AbkbyAngAdgytZp14ABVvXzxPHykAMz8NbVM3RgB6JKgAVVlbTUkZBSgAN1sAGzBgAC5YgNMQnwBtAGtgD2jK+ODQgF0qFTjAyENMNydZKSLgQyKpdQAKQpLgAEoXKBSCIkZgAA8vKVVFZmhMNKCAeQB9AAUAGUwuGCUYADkggGVODCPTqgAiWzB5KTfcxYY7nF5BTAAJSCPz+AIAtApbHsgWRPucnsI8L9-lJ4aAtKjxFsdnsoAcoAAxACSAA0YIJLjc7g9nm8DFBThcfgCvKiMFyzuCoTDsbzEfJkfJ+ZyThczpieVIvPiQITgeQTkFTkofgAjKTyAEQGWa7UnJRCiHQ-WG40SqWmrU6hWPLFQb4Go1SCCq9VkSHUgDiAAlRR7VABLdQAC2lwMDoaCVpFPyjsfkDpRCeDYdd7u+6bjfplVxglPD3zGUXjZDLFZTNux1czSOzdfLycVzeANZL1kYw0UZy8RVsEmAEFAin0ZCgxkMNPpjOut3uT1ewPnxnRTLXrM39DsDiHG22u0UEV8q5ZG6oV5aI7HE6nIHkg9yik2FRv6+eVCXBk91vNlDC5D9nCgDwf2ZP9K15b4ILyAB3GD9zvPAn3HSdpzAuUkNGcZJhmTYFhyORCImKZpg8Mjhko4jpmQhYgA

RobinCsl: Very well. Ignore my suggestions then 😅 thanks for trying!

packages/orbit-components/src/common/placements.ts
RobinCsl: I wonder if renaming this to FixedPlacement would still make sense and be more readable (given the name of the function)?

DSil: Yes, this was done 🙂

PR #4495, opened 2024-10-18
FEPLT-1739: Add defaultSelected to Checkbox and Radio

DSil:
Checkbox and Radio components now support `defaultSelected` prop and a full uncontrolled experience.

The style classes moved around a bit, but visual tests ensure nothing is broken (I am glad we have a good coverage on this!)


packages/orbit-components/src/Radio/README.md
```diff
@@ -16,28 +16,29 @@ After adding import into your project you can use it simply like:
...
+| ref            | `func`                     |         | Prop for forwarded ref of the Radio. [See Functional specs](#functional-specs)                            |
+| tabIndex       | `string \| number`         |         | Specifies the tab order of an element                                                                     |
+| tooltip        | `Element<Tooltip>`         |         | Optional property when you need to attach Tooltip to the Radio. [See Functional specs](#functional-specs) |
+| value          | `string`                   |         | The value of the Radio.                                                                                   |
+| readOnly       | `boolean`                  |         | If `true`, the Radio will be set up as readOnly.                                                          |
 
 ## Functional specs
 
-- The`hasError` prop will be visible only when the Radio has `checked` or `disabled` prop set on false.
+- The`hasError` prop will be visible only when the Radio is not checked or disabled.
```
sarkaaa, line 39: nitpick:
```
checked nor disabled.
```
as it's written it `Checkbox/README.md`

DSil: Thanks

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -16,7 +16,7 @@ const meta: Meta<typeof Checkbox> = {
   parameters: {
     info: "Additionally you can add tooltip to this component.",
     controls: {
-      exclude: ["onChange"],
+      exclude: ["onChange", "defaultChecked"],
```
sarkaaa, line 19: I would add a separate story for the variant with `defaultChecked` prop, it could be called `Uncontrolled`. We have this type of story in Textarea or Card. WDYT?

DSil:
Regarding the problem you reported with the tooltip, the issue is actually bigger. When we have a tooltip, the `peer-*` classes don't work! The `peer-focus` classes that we already had don't work when the checkbox has a tooltip. If you look in master, the checkbox does not show a focus style when it has a tooltip. That is clearly a bug. So I will try to look into it.

> Editing line 91 in Checkbox/index.tsx could help:

That is not true. If you define `defaultChecked` to be true, it will always be true, even if the user unchecks it. So it would always be visible. Although it is white, so you can't really see it, I don't think it's the correct way to do it. I'll fix it together with the focus issue

packages/orbit-components/src/Checkbox/Checkbox.stories.tsx
DSil: Yes, we can have uncontrolled stories here as well. Will add

sarkaaa:
> Regarding the problem you reported with the tooltip, the issue is actually bigger. When we have a tooltip, the `peer-*` classes don't work! The `peer-focus` classes that we already had don't work when the checkbox has a tooltip. If you look in master, the checkbox does not show a focus style when it has a tooltip. That is clearly a bug. So I will try to look into it.
> 
> > Editing line 91 in Checkbox/index.tsx could help:
> 
> That is not true. If you define `defaultChecked` to be true, it will always be true, even if the user unchecks it. So it would always be visible. Although it is white, so you can't really see it, I don't think it's the correct way to do it. I'll fix it together with the focus issue

Yeah, I remember that I had the `peer-*` issue when I was fixing this checkbox element before, so I handled it using `checked` in condition, but I should manage the focus state as well. 🙈

packages/orbit-components/src/Radio/Radio.stories.tsx
```diff
@@ -16,7 +16,7 @@ const meta: Meta<typeof Radio> = {
   parameters: {
     info: "Radio component. Check Orbit.Kiwi for more detailed guidelines.",
     controls: {
-      exclude: ["onChange"],
+      exclude: ["onChange", "defaultChecked"],
```
sarkaaa, line 19: Separate story here as I mentioned on Checkbox.stories.tsx. 🙏

packages/orbit-components/src/Radio/index.tsx
```diff
@@ -29,25 +30,36 @@ const Radio = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
...
-              : "[&>.orbit-radio-icon-container]:border-form-element [&>.orbit-radio-icon-container]:hover:border-form-element-hover [&>.orbit-radio-icon-container]:active:border-form-element-active"),
-          checked &&
-            "[&>.orbit-radio-icon-container]:border-form-element-focus active:border-form-element-focus [&>.orbit-radio-icon-container]:bg-white-normal",
-        ],
+        "[&_.orbit-radio-icon-container]:has-[:checked]:border-2 [&_.orbit-radio-icon-container_span]:has-[:checked]:visible",
+        "[&_.orbit-radio-icon-container]:has-[:focus]:outline-blue-normal [&_.orbit-radio-icon-container]:has-[:focus]:outline [&_.orbit-radio-icon-container]:has-[:focus]:outline-2",
+        disabled
+          ? [
+              "cursor-not-allowed",
+              "[&_.orbit-radio-icon-container]:bg-cloud-light [&_.orbit-radio-icon-container]:border-cloud-dark",
```
sarkaaa, line 38: Is there any reason why these two lines are separated? Is it just "polish" thing 💅  or any technical reason behind?

DSil: It's just a matter of readability and organization. While we don't have any standard within the team, I found it would be better to have classes that target other elements in separate lines

sarkaaa (approved):

PR #4497, opened 2024-10-24
Orbit Storybook - a11y issues

sarkaaa:

### For new components:

List of SB errors:
- https://docs.google.com/spreadsheets/d/13HDxhH017uVSmGX3q8rpXP1aJ3h_Z6CvFLrwIwi5Ooo/edit?gid=0#gid=0 

Closes: https://kiwicom.atlassian.net/browse/FEPLT-2146

### DIFF
--- a/.github/workflows/update-visual-tests.yml
+++ b/.github/workflows/update-visual-tests.yml
@@ -12,7 +12,7 @@ jobs:
   update-visual-tests:
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.44.1-jammy
+      image: mcr.microsoft.com/playwright:v1.48.2-jammy
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/packages/orbit-components/.storybook/README.md
+++ b/packages/orbit-components/.storybook/README.md
@@ -1,0 +1,27 @@
+# About
+
+Storybook included the list of Orbit components.
+
+URL: https://kiwicom.github.io/orbit/
+
+## Run
+
+**Install dependencies:**
+
+```
+yarn install
+```
+
+**Run Storybook:**
+
+```
+yarn storybook
+```
+
+**Run a11y tests:**
+
+This will run test-runner script and generates HTML files with the list of a11y issues. Be aware that the local storybook server must be running as well (`yarn storybook`).
+
+```
+yarn test-storybook
+```
--- a/packages/orbit-components/.storybook/orbitDecorator.tsx
+++ b/packages/orbit-components/.storybook/orbitDecorator.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import type { Decorator } from "@storybook/react";
 import jsxToString from "react-element-to-jsx-string";
 import { Highlight, themes } from "prism-react-renderer";
+import { addons } from "@storybook/preview-api";
 
 import defaultTheme from "../src/defaultTheme";
 import OrbitProvider from "../src/OrbitProvider";
@@ -15,8 +16,18 @@ const OrbitDecorator: Decorator = (storyFn, context) => {
   React.useEffect(() => {
     const html = document.querySelector("html");
     if (html && !html.getAttribute("dir")) html.setAttribute("dir", "ltr");
+    const componentForceRerender = () => {
+      // Invokes Storybook's addon API method event to trigger a UI refresh.
+      // Ensures that a11y addons is refreshed and shows the correct list of component accessibility issues.
+      console.log("Component is being rerendered.");
+      addons.getChannel().emit("forceReRender");
+    };
+
+    componentForceRerender();
     return () => {
-      if (html) html.removeAttribute("dir");
+      if (html) {
+        html.removeAttribute("dir");
+      }
     };
   }, []);
 
--- a/packages/orbit-components/.storybook/test-runner.ts
+++ b/packages/orbit-components/.storybook/test-runner.ts
@@ -1,0 +1,32 @@
+import type { TestRunnerConfig } from "@storybook/test-runner";
+import { injectAxe, checkA11y } from "axe-playwright";
+
+/*
+ * See https://storybook.js.org/docs/writing-tests/test-runner#test-hook-api
+ * to learn more about the test-runner hooks API.
+ */
+const config: TestRunnerConfig = {
+  async preVisit(page) {
+    await injectAxe(page);
+  },
+  async postVisit(page, story) {
+    const fileName = story.name.replace(/\s+/g, "");
+    await checkA11y(
+      page,
+      ".story-content",
+      {
+        detailedReport: true,
+        detailedReportOptions: {
+          html: true,
+        },
+      },
+      undefined,
+      "html",
+      {
+        reportFileName: `${fileName}-result.html`,
+      },
+    );
+  },
+};
+
+export default config;
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,7 +126,7 @@ jobs:
     name: Visual Regression Tests
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.44.1-jammy
+      image: mcr.microsoft.com/playwright:v1.48.2-jammy
     strategy:
       fail-fast: false
       matrix:
--- a/packages/orbit-components/.gitignore
+++ b/packages/orbit-components/.gitignore
@@ -21,3 +21,4 @@ cypress/videos
 /test-results/
 /playwright-report/
 /playwright/.cache/
+artifacts/
--- a/packages/orbit-components/package.json
+++ b/packages/orbit-components/package.json
@@ -11,6 +11,7 @@
   },
   "scripts": {
     "storybook": "storybook dev -p 6007 -c .storybook --ci",
+    "test-storybook": "test-storybook --url http://127.0.0.1:6007",
     "cy:dev": "vite cypress/integration --logLevel silent",
     "cy:run": "start-server-and-test cy:dev http-get://localhost:3000 'cypress run'",
     "cy:open": "start-server-and-test cy:dev http-get://localhost:3000 'cypress open'",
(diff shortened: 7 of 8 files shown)


packages/orbit-components/.storybook/preview.ts
```diff
@@ -29,6 +29,25 @@ const preview: Preview = {
...
+            selector: "*:not(div.text-heading-title0[role='heading'])",
+          },
+          {
+            id: "color-contrast",
+            selector: "*:not(pre .attr-name.tag.token)",
+          },
+          {
+            id: "color-contrast-enhanced",
+            selector: "*:not(pre .attr-name.tag.token)",
+          },
```
mvidalgarcia, line 47: I believe it is safer to add a custom class to the [whole code snippet div](https://github.com/kiwicom/orbit/blob/e9327f465f6cb88cc6b3ebf456fc3dd0526a5058/packages/orbit-components/.storybook/orbitDecorator.tsx#L55) and target it, WDYT?

packages/orbit-components/.storybook/preview.ts
```diff
@@ -29,6 +29,25 @@ const preview: Preview = {
   decorators: [orbitDecorator],
   parameters: {
     viewport: { viewports },
+    a11y: {
+      element: "#storybook-root",
+      config: {
+        rules: [
+          {
+            id: "aria-required-attr",
+            selector: "*:not(div.text-heading-title0[role='heading'])",
```
mvidalgarcia, line 38: Same here, we can add a custom class to the heading. For instance, this selector could target headings on the Heading story, and we want to avoid that.

packages/orbit-components/.storybook/preview.ts
```diff
@@ -29,6 +29,25 @@ const preview: Preview = {
...
+          },
+          {
+            id: "color-contrast",
```
mvidalgarcia, line 41: We also need to disable `scrollable-region-focusable` for the code snippet. Check [this story](https://kiwicom-orbit-sarka-sb-a11y-fixes.surge.sh/?path=/story/popover--default), for example.

image

sarkaaa: yes, I had it there in my first version, I need to revert it back,

packages/orbit-components/.storybook/preview.ts
```diff
@@ -29,6 +29,29 @@ const preview: Preview = {
   decorators: [orbitDecorator],
   parameters: {
     viewport: { viewports },
+    a11y: {
+      element: "#storybook-root",
```
mvidalgarcia, line 33: Actually it would be much simpler to wrap the story into a div with a `.story-content` class and set it here. So we don't need config at all :)

```diff
diff --git a/packages/orbit-components/.storybook/orbitDecorator.tsx b/packages/orbit-components/.storybook/orbitDecorator.tsx
index 5ad9607c5..34ec0a0ea 100644
--- a/packages/orbit-components/.storybook/orbitDecorator.tsx
+++ b/packages/orbit-components/.storybook/orbitDecorator.tsx
@@ -51,7 +51,7 @@ const OrbitDecorator: Decorator = (storyFn, context) => {
         <Text spaceAfter="largest" type={inverted ? "white" : "primary"}>
           {context.parameters?.info}
         </Text>
-        {children}
+        <div className="story-content">{children}</div>
         <div dir="ltr" className="mt-600 relative">
           <Tile>
             <Highlight theme={themes.vsLight} code={code} language="tsx">
diff --git a/packages/orbit-components/.storybook/preview.ts b/packages/orbit-components/.storybook/preview.ts
index 5441c320b..b7362c19d 100644
--- a/packages/orbit-components/.storybook/preview.ts
+++ b/packages/orbit-components/.storybook/preview.ts
@@ -30,27 +30,7 @@ const preview: Preview = {
   parameters: {
     viewport: { viewports },
     a11y: {
-      element: "#storybook-root",
-      config: {
-        rules: [
-          {
-            id: "aria-required-attr",
-            selector: '*:not(div.text-heading-title0[role="heading"])',
-          },
-          {
-            id: "color-contrast",
-            selector: "*:not(pre .attr-name.tag.token)",
-          },
-          {
-            id: "color-contrast-enhanced",
-            selector: "*:not(pre .attr-name.tag.token)",
-          },
-          {
-            id: "scrollable-region-focusable",
-            selector: "*not(pre.overflow-x-auto)",
-          },
-        ],
-      },
+      element: ".story-content",
     },
   },
 };
```

sarkaaa: yes, that's also a good idea to include only wrapped `children` prop. thx. 👍

packages/orbit-components/.storybook/orbitDecorator.tsx
```diff
@@ -15,8 +16,18 @@ const OrbitDecorator: Decorator = (storyFn, context) => {
   React.useEffect(() => {
     const html = document.querySelector("html");
     if (html && !html.getAttribute("dir")) html.setAttribute("dir", "ltr");
+    const componentForceRerender = () => {
+      // Invokes Storybook's addon API method event to trigger a UI refresh.
+      // Ensures that a11y addons is refreshed and shows the correct list of component accessibility issues.
+      console.log("Component is being rerendered.");
+      addons.getChannel().emit("forceReRender");
```
sarkaaa, line 23: An explanation of the function is [described here](https://gitlab.skypicker.com/frontend/core/-/merge_requests/1784#note_4983189).

packages/orbit-components/.storybook/test-runner.ts
```diff
@@ -0,0 +1,32 @@
+import type { TestRunnerConfig } from "@storybook/test-runner";
+import { injectAxe, checkA11y } from "axe-playwright";
+
+/*
+ * See https://storybook.js.org/docs/writing-tests/test-runner#test-hook-api
```
sarkaaa, line 5: Using the `test-runner` and depended packages, we are able to run a11y SB checks and generate an HTML report.

packages/orbit-components/.storybook/README.md
```diff
@@ -0,0 +1,27 @@
+# About
```
sarkaaa, line 1: I created this file to document somewhere information on how to run test-runner and that's important to run this script once the storybook local server is running as well.

.github/workflows/update-visual-tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   update-visual-tests:
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.44.1-jammy
+      image: mcr.microsoft.com/playwright:v1.48.2-jammy
```
sarkaaa, line 15: I've changed the PW version based on this error log.
https://github.com/kiwicom/orbit/actions/runs/11632651461/job/32396199861?pr=4497

sarkaaa: Also, I had to update visual tests as they were failing.

packages/orbit-components/.storybook/preview.ts
```diff
@@ -29,6 +29,9 @@ const preview: Preview = {
   decorators: [orbitDecorator],
   parameters: {
     viewport: { viewports },
+    a11y: {
```
mvidalgarcia, line 32: @sarkaaa Could you please create a separate PR with this configuration and the changes on `orbitDecorator.tsx`? 🙏  I think that part can be extracted and merged already :)

packages/orbit-components/.storybook/preview.ts
mvidalgarcia: Perfect, that's all we need for a separate PR 😄

packages/orbit-components/.storybook/preview.ts
sarkaaa: Done - https://github.com/kiwicom/orbit/pull/4510

mvidalgarcia: thanks 🫶 
resolving

packages/orbit-components/.storybook/test-runner.ts
```diff
@@ -0,0 +1,32 @@
...
+ * to learn more about the test-runner hooks API.
+ */
+const config: TestRunnerConfig = {
+  async preVisit(page) {
+    await injectAxe(page);
+  },
+  async postVisit(page, story) {
+    const fileName = story.name.replace(/\s+/g, "");
+    await checkA11y(
```
mvidalgarcia, line 14: We agreed we don't need the HTML reports for Orbit components, right?

packages/orbit-components/.storybook/orbitDecorator.tsx
mvidalgarcia, line 1: Is this issue still relevant? Do we have problems with the Orbit storybook? 

I remember Nitro's Storybook being problematic but Orbit one was working correctly, wasn't it?

sarkaaa: Yes, Orbit Storybook was not problematic at all.

sarkaaa: no, Orbit also has an issue with no-updating a11y issues when switching the stories. however, Orbit is working well after this change, but Nitro is still problematic.

packages/orbit-components/.storybook/test-runner.ts
sarkaaa: This was done before I was struggling with Nitro, and then we agreed we won't generate HTML reports. I thought that the "agreement" was only for Nitro, anyway, but I can remove it from Orbit.

packages/orbit-components/.storybook/orbitDecorator.tsx
mvidalgarcia: So then we can close this PR and focus on Nitro? Is this assumption correct? 🤔

mvidalgarcia: It's working for me now (Chrome) on master... 🤔

mvidalgarcia: Since it seems we have different behaviors on different machines and different browsers. What do you think about closing this PR for the time being? We can always re-open it if this becomes annoying again.

mvidalgarcia: Recording of my experience:

https://github.com/user-attachments/assets/b70cc8d8-797a-4635-97de-5ce8352533a5

packages/orbit-components/.storybook/test-runner.ts
mvidalgarcia: If I recall correctly, the final conclusion is that HTML reports were not bringing that much value at this period as problems will be detected with the general Accessibility audits.

I really value your proactivity on exploring this idea, great work! I believe these fine-grained reports could have more utility in the future when we have a higher accessibility score on all the Kiwi.com pages, so we can get more into the details to find problems we don't see at first sight. What do you think?

PR #4506, opened 2024-11-05
Deconsider eslint-plugin package

DSil:
This marks the repository as private, so it is not considered by lerna for the release process.

It also removes all the references in docs and usage of the plugin across the repository.

FEPLT-2083


mvidalgarcia:
What about [this](https://github.com/kiwicom/orbit/blob/dsil/deconsider-eslint-plugin/.eslintignore#L9)?

mvidalgarcia (approved):
The rest LGTM 👍

DSil:
> What about [this](https://github.com/kiwicom/orbit/blob/dsil/deconsider-eslint-plugin/.eslintignore#L9)?

If I understood correctly, that is for eslint to ignore the dist folder of that package. Since we are not removing the package yet, I understood we still need to ignore it. Otherwise it might trigger lint warnings in built files in that dist folder

PR #4508, opened 2024-11-05
Tooltip/TooltipPrimitive: add visual tests

sarkaaa:

### For new components:


sarkaaa:
Hi guys, I would like to discuss some minor things - Tooltip visual tests. I’ve done some tests (tooltip/tooltip primitive), but the number of screenshots is pretty huge. I haven’t covered all of the cases (eg. all of the possible tooltip placements,..). Currently, I’m thinking about ways how to decrease the final number of screenshots.

- I can’t include multiple tooltips in one screenshot, as the behavior is depended on hover/click events
- Exclude different placements at all and have only one placement (eg. bottom)
- Exclude some of the viewports (eg. have only one mobile device and only two desktop viewports?)
- Exclude some variants - eg. block tooltip, with/without image

Do you have any ideas/suggestions on how to reduce the amount of images?

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct.tsx
```diff
@@ -0,0 +1,50 @@
...
+import type { Placement } from "../../common/placements";
+
+test.describe("Tooltip primitive visual tests", () => {
+  const placements = ["top", "right", "bottom", "left"];
+
+  placements.forEach(placement => {
+    test(`screenshot for default - ${placement}`, async ({ mount }) => {
+      const component = await mount(<DefaultTooltipPrimitive placement={placement as Placement} />);
+
+      await component.getByRole("button").hover();
```
DSil, line 15: We can avoid having the hover action. This behavior should be tested in unit tests, not visual tests

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct-story.tsx
```diff
@@ -0,0 +1,73 @@
...
+        (placement === "left" || placement === "right") && "items-center",
+      ])}
+    >
+      <Stack justify={placement.includes("left") ? "end" : "start"}>
+        <TooltipPrimitive
+          content={tooltipContentCtStory}
+          placement={placement}
+          error={error}
+          help={help}
+          block={block}
```
DSil, line 43: If we add `tooltipShown` here, we don't need the hover

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct-story.tsx
```diff
@@ -0,0 +1,73 @@
...
+export function DefaultTooltipPrimitive({
+  placement = PLACEMENTS.BOTTOM,
+  error,
+  help,
+  block,
+}: TooltipPrimitiveProps) {
+  return (
+    <div
+      className={cx([
+        "flex min-h-[128px]",
```
DSil, line 32: We can add some padding here, so the tooltip can "breathe" better. We probably need a bigger min-height as well.

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct-story.tsx
```diff
@@ -0,0 +1,73 @@
...
+          block={block}
+        >
+          <Icons.Airplane />
+        </TooltipPrimitive>
+      </Stack>
+    </div>
+  );
+}
+
+export function TooltipPrimitiveWithImage({ size = "medium" }: { size?: Size }) {
```
DSil, line 52: I feel we don't need this test with image at all

DSil (commented):
> I can’t include multiple tooltips in one screenshot, as the behavior is dependent on hover/click events

You can 😉 no need to use actions. You can open via props.

> Exclude different placements at all and have only one placement (eg. bottom)

I think it's important to test all placements. Ideally we can have multiple placements in one screenshots (no need to have them all in one, but grouped somehow).

> Exclude some of the viewports (eg. have only one mobile device and only two desktop viewports?)

That seems reasonable to me.

> Exclude some variants - eg. block tooltip, with/without image

We don't need these. I'd say for TooltipPrimitive it's important to test: `placement`, `size`, `error`, `help` and `removeUnderlinedText` props. Same for Tooltip.

I'd also advice not to use an icon as the trigger, so we can actually test `removeUnderlinedText`.

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct-story.tsx
sarkaaa: there was set bigger height before, I wanted to avoid "big blank spaces", but side-padding will be good (especially for small mobile viewport).

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
+import { test, expect } from "@playwright/experimental-ct-react";
+
+import { TooltipPrimitiveVisualDefaultStory } from "./TooltipPrimitive.ct-story";
+
+export const resizeViewportHeight = async page => {
+  const viewportSize = page.viewportSize();
+
+  if (viewportSize) {
```
sarkaaa, line 9: This function sets a larger height of the viewport, without that, I couldn't cover all of the tooltip placements in one screen, eg: in the picture below, the last row of the tooltips are indented, because they were out of the default viewport height:
image

sarkaaa:
Thank you for the feedback, currently I covered these variants:
- all possible placements
- error & help state
- no underline text
- size (`"small" | "medium"`)

(basic) Tooltip component test covers only default tooltip. This tooltip is a wrapper of primitiveTooltip (without some props - tooltipShown, error, help,..), and that's the reason why no all of the variants mentioned above are not covered in these test file. The main difference is how the tooltip is shown on mobile devices.

DSil:
LGTM, but I think we don't need to test all the placements for all the variations. So, for small size, error, help and noUnderline I'd say we only need one tooltip. This way we can have much smaller snapshots (bigger snapshots are more likely to give false results)

sarkaaa:
> LGTM, but I think we don't need to test all the placements for all the variations. So, for small size, error, help and noUnderline I'd say we only need one tooltip. This way we can have much smaller snapshots (bigger snapshots are more likely to give false results)

Yeah, I agree with the statement that we don't need test all plasements for variations (noUderline, error, etc..). I'll exclude them, and the size of the snapshots will be smaller.

DSil (approved):

PR #4509, opened 2024-11-06
Fix checkbox+radio disabled styles

DSil:
During #4427 some visual bugs were introduced. This PR addresses them:
- Hover style while checked and disabled
- Active style while disabled

FEPLT-2148


sarkaaa (commented):
I have found this issue:
Disabled checkbox has focus state if `checked={true}` and there is (text) link in Label. I think it's out of the scope of this fix.

image

Otherwise, LGTM.

sarkaaa (approved):

DSil:
@sarkaaa that should not happen in the first place 😌 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#interactive_content

PR #4515, opened 2024-11-11
Sarka/a11y fixes tooltip

sarkaaa:
~~New props `aria-label` and `aria-labelledby` were added to tooltipPrimitive/tooltip/primitiveModalDialog components. Using these props, we are further to our goal in a11y.~~

Redundant `role="tooltip"` attribute was removed. 

---

Closes https://kiwicom.atlassian.net/browse/FEPLT-2156

### DIFF
--- a/docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
+++ b/docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
@@ -1,0 +1,49 @@
+---
+title: Accessibility
+redirect_from:
+  - /components/tooltip/accessibility/
+---
+
+## Accessibility
+
+### Tooltip
+
+The Tooltip component has been designed with accessibility in mind. It can be used with keyboard navigation and includes properties that enhance the experience for users of assistive technologies.
+
+While the `aria-label` and `aria-labelledby` attributes are not needed for the Tooltip component itself, it is important to ensure that the components passed as children can be perceived by assistive technologies. This ensures that the Tooltip component is accessible.
+
+#### Example 1
+
+The content of children component is text, so it is read by screen reader.
+
+```jsx
+<Tooltip
+  content={
+    <div>
+      <p>Lorem ipsum dolor sit amet.</p>
+    </div>
+  }
+>
+  <Text>Learn more.</Text>
+</Tooltip>
+```
+
+The screen reader will announce: "Learn more. Lorem ipsum dolor sit amet.".
+
+#### Example 2
+
+The children element is an icon component. To achieve the accessibility of the Tooltip, adding an `aria-label` to the icon component is necessary.
+
+```jsx
+<Tooltip
+  content={
+    <div>
+      <p>Lorem ipsum dolor sit amet.</p>
+    </div>
+  }
+>
+  <Airplane ariaLabel="More information" />
+</Tooltip>
+```
+
+Once the icon is focused by the screen reader, it will announce: "More information", followed by announcing the content of the tooltip: "Lorem ipsum dolor sit amet.".
--- a/packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.stories.tsx
+++ b/packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.stories.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { action } from "@storybook/addon-actions";
 import type { Meta, StoryObj } from "@storybook/react";
 
-import * as Icons from "../../icons";
+import Airplane from "../../icons/Airplane";
 import Stack from "../../Stack";
 import Alert from "../../Alert";
 import Text from "../../Text";
@@ -43,7 +43,7 @@ export const Default: Story = {
   render: args => (
     <Stack justify="center">
       <TooltipPrimitive {...args}>
-        <Icons.Airplane />
+        <Airplane ariaLabel="More information" />
       </TooltipPrimitive>
     </Stack>
   ),
@@ -147,7 +147,7 @@ export const WithImageInside: Story = {
         </Stack>
       }
     >
-      <Icons.Airplane />
+      <Airplane ariaLabel="More information" />
     </TooltipPrimitive>
   ),
 
--- a/packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.tsx
+++ b/packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.tsx
@@ -105,54 +105,55 @@ const TooltipContent = ({
     [onClose, onCloseMobile, elements.floating],
   );
 
+  const handleCombinedClick = (ev: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+    handleInnerClick(ev);
+    onClick(ev);
+  };
+
   return (
-    // Disabling because the onClick exists just to stop propagation of events
+    // Disabling because the onClick exists to close tooltip when clicking in interactive elements, which should not happen with keyboard.
     // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
-    <div className="w-full" role="tooltip" id={tooltipId} data-test={dataTest} onClick={onClick}>
-      {/* Disabling because the onClick exists to close tooltip when clicking in interactibe elements, which should not happen with keyboard */}
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
+    <div
+      className={cx(
+        "rounded-100 px-300 shadow-level3 z-[10012] box-border block w-auto overflow-visible",
+        "duration-fast transition-[visibility,_opacity] ease-in-out",
+        "[&_img]:max-w-full]",
+        contentHeight <= Math.floor(parseFloat(theme.orbit.lineHeightNormal)) ? "py-200" : "py-300",
+        shown ? "visible opacity-100" : "invisible opacity-0",
+        size === "small" && "max-w-[240px]",
+        size === "medium" && "max-w-[380px]",
+        error && "bg-red-normal",
+        !error && help && "bg-blue-normal",
+        !error && !help && "bg-ink-dark",
+      )}
+      ref={refs.setFloating}
+      role="tooltip"
+      id={tooltipId}
+      aria-hidden={!shown}
+      onMouseEnter={onEnter}
+      onMouseLeave={onClose}
+      onClick={handleCombinedClick}
+      style={floatingStyles}
+      data-test={dataTest}
+    >
       <div
         className={cx(
-          "rounded-100 px-300 shadow-level3 z-[10012] box-border block w-auto overflow-visible",
-          "duration-fast transition-[visibility,_opacity] ease-in-out",
-          "[&_img]:max-w-full]",
-          contentHeight <= Math.floor(parseFloat(theme.orbit.lineHeightNormal))
-            ? "py-200"
-            : "py-300",
-          shown ? "visible opacity-100" : "invisible opacity-0",
-          size === "small" && "max-w-[240px]",
-          size === "medium" && "max-w-[380px]",
-          error && "bg-red-normal",
-          !error && help && "bg-blue-normal",
-          !error && !help && "bg-ink-dark",
+          "font-base text-small text-white-normal mb-0 font-medium leading-normal",
+          "[&_.orbit-text]:text-small [&_.orbit-text]:text-white-normal [&_.orbit-text]:font-medium",
+          "[&_.orbit-list-item]:text-small [&_.orbit-list-item]:text-white-normal [&_.orbit-list-item]:font-medium",
+          "[&_.orbit-text-link]:text-white-normal",
         )}
-        ref={refs.setFloating}
-        role="tooltip"
-        aria-hidden={!shown}
-        onMouseEnter={onEnter}
-        onMouseLeave={onClose}
-        onClick={handleInnerClick}
-        style={floatingStyles}
+        ref={content}
       >
-        <div
-          className={cx(
-            "font-base text-small text-white-normal mb-0 font-medium leading-normal",
-            "[&_.orbit-text]:text-small [&_.orbit-text]:text-white-normal [&_.orbit-text]:font-medium",
-            "[&_.orbit-list-item]:text-small [&_.orbit-list-item]:text-white-normal [&_.orbit-list-item]:font-medium",
-            "[&_.orbit-text-link]:text-white-normal",
-          )}
-          ref={content}
-        >
-          {children}
-        </div>
-        <FloatingArrow
-          ref={arrowRef}
-          context={context}
-          height={ARROW_SIZE}
-          width={ARROW_SIZE * 2}
-          fill={arrowColor}
-        />
+        {children}
       </div>
+      <FloatingArrow
+        ref={arrowRef}
+        context={context}
+        height={ARROW_SIZE}
+        width={ARROW_SIZE * 2}
+        fill={arrowColor}
+      />
     </div>
   );
 };
--- a/packages/orbit-components/src/Tooltip/Tooltip.stories.tsx
+++ b/packages/orbit-components/src/Tooltip/Tooltip.stories.tsx
@@ -3,7 +3,7 @@ import { action } from "@storybook/addon-actions";
 import type { Meta, StoryObj } from "@storybook/react";
 
 import { PLACEMENTS, AUTO_PLACEMENTS } from "../common/placements";
-import * as Icons from "../icons";
+import Airplane from "../icons/Airplane";
 import { SIZE_OPTIONS } from "./consts";
 import Stack from "../Stack";
 import Alert from "../Alert";
@@ -47,7 +47,7 @@ export const Default: Story = {
   render: args => (
     <Stack justify="center">
       <Tooltip {...args}>
-        <Icons.Airplane />
+        <Airplane ariaLabel="More information" />
       </Tooltip>
     </Stack>
   ),
@@ -149,7 +149,7 @@ export const WithImageInside: Story = {
         </Stack>
       }
     >
-      <Icons.Airplane />
+      <Airplane ariaLabel="More information" />
     </Tooltip>
   ),
 


docs/src/__examples__/Tooltip/INDICATION.tsx
```diff
@@ -25,8 +25,10 @@ export default {
                 <Text>An open source design system for your next travel project.</Text>
               </Stack>
             }
+            aria-label="Orbit design system"
+            aria-labelledby="orbit-success"
           >
-            <Text>Orbit</Text>
+            <Text id="orbit-app">Orbit</Text>
```
sarkaaa, line 31: I am not sure about this whole file, in case it's necessary, I can revert it back.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+The Tooltip component has been designed with accessibility in mind.
+
+It can be used with keyboard navigation, and it includes the following properties that allow to improve the experience for users of assistive technologies:
+
+| Name            | Type     | Description                                                                               |
+| :-------------- | :------- | :---------------------------------------------------------------------------------------- |
+| aria-label      | `string` | Adds `aria-label` attribute to the rendered tooltip element. Announced by screen readers. |
+| aria-labelledby | `string` | Id(s) of elements that announce the component to screen readers.                          |
+
+Mentioned props are optional, but highly recommended to use to ensure the best accessibility experience for all users.
```
DSil, line 20:
```suggestion
The mentioned props are optional, but highly recommended to use to ensure the best accessibility experience for all users.
```

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The `aria-label` prop can be used to provide a label for the tooltip that is announced by screen readers.
```
DSil, line 22:
```suggestion
The `aria-label` prop can be used to provide a label for the tooltip that is announced by screen readers but is not visible on the screen.
```

Maybe we should explain the purpose of this attribute

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The `aria-labelledby` prop can reference multiple ids, separated by a space. Value of the attribute is not announced by screen reader, content of connected
```
DSil, line 24:
```suggestion
The `aria-labelledby` prop can reference multiple ids, separated by a space. The value of the attribute is not announced by the screen reader, but the content of the referred element is.
```

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+element does.
```
DSil, line 25: If you accept the suggestion above, this can be deleted.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The conjugation of these properties allows to provide a content of tooltip component to users of assistive technologies.
```
DSil, line 27: This is not exactly true. What is the purpose of an aria-label for a tooltip trigger? Is it to announce the tooltip content? Or to announce that it triggers a tooltip?
Honest question, I didn't research, but this sentence is not clear to me.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+    </div>
+  }
+  aria-label="Tooltip label"
+  aria-labelledby="tooltip-labelledby"
+>
+  <Text id="tooltip-labelledby">Hello world.</Text>
+</Tooltip>
+```
+
+Note that attribute `aria-labelledby` is connected with `id` attribute of the component that is the base for the Tooltip. Without that connection, the `aria-labelledby` attribute is ignored and not read by screen reader.
```
DSil, line 45:
```suggestion
Note that the attribute `aria-labelledby` is connected with the `id` attribute of the component that is the base for the Tooltip. Without that connection, the `aria-labelledby` attribute is ignored and not read by screen reader.
```

Yes, but it doesn't say what it's for. It should say that the content of the element with the `id` is announced.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The content of attribute `aria-labelledby` is prioritized before the content of `aria-label`. In case `aria-labelledby` is not filled in,
```
DSil, line 47:
```suggestion
The content of attribute `aria-labelledby` is prioritized over the content of `aria-label`.
```

I believe only this is necessary (the line below can also be deleted).

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+then the content of `aria-label` is read by screen reader.
+
+Be sure to include all relevant information on all of the properties that are announced by screen readers.
```
DSil, line 50: I don't understand what this sentence means 😕

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+Mentioned props are optional, but highly recommended to use to ensure the best accessibility experience for all users.
+
+The `aria-label` prop can be used to provide a label for the tooltip that is announced by screen readers.
+
+The `aria-labelledby` prop can reference multiple ids, separated by a space. Value of the attribute is not announced by screen reader, content of connected
+element does.
+
+The conjugation of these properties allows to provide a content of tooltip component to users of assistive technologies.
+
+For example, the following code snippet shows how to use these properties:
```
DSil, line 29: It would be nice to know what is announced by screen readers

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+Note that attribute `aria-labelledby` is connected with `id` attribute of the component that is the base for the Tooltip. Without that connection, the `aria-labelledby` attribute is ignored and not read by screen reader.
+
+The content of attribute `aria-labelledby` is prioritized before the content of `aria-label`. In case `aria-labelledby` is not filled in,
+then the content of `aria-label` is read by screen reader.
+
+Be sure to include all relevant information on all of the properties that are announced by screen readers.
+
+### Examples
+
+1. Prop `id` in `<Text>` component is not filled in, content of `aria-label` is read by screen reader. Then the content inside the tooltip is read by screen reader. Content of `aria-labelledby` is ignored.
```
DSil, line 54: This is misleading. The content of `aria-label` is read because there is no element with the `id` passed on `aria-labelledby`. It is not because `Text` has no `id`.

It is also important to note that the paragraph inside the content is also announced (although it shouldn't…). So this is a wrong implementation that we currently have and that we will need to fix. But until that happens I feel we should be clear about it, and adjust the documentation when the time comes.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+    </div>
+  }
+  aria-label="Tooltip label"
+  aria-labelledby="tooltip-labelledby"
+>
+  <Text>Hello world.</Text>
+</Tooltip>
+```
+
+2. Prop `id` in `<Text>` component is filled in, content of this component (`Hello world`) is read by screen reader. Then the content inside the tooltip is read by screen reader. Content of `aria-label` is not read at all.
```
DSil, line 70:
```suggestion
2. Prop `id` in `<Text>` component is filled in with the same value passed to `aria-labelledby`. The content of this component (`Hello world`) is read by the screen reader. The value of `aria-label` is not read at all.
```

DSil (commented):
I reviewed the 03-accessibility.mdx file **only**.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
domihustinova: Can you elaborate on the problem here, please? By "paragraph inside the content" you mean specifically the "Lorem ipsum dolor sit amet" text from the example? Or that the word "paragraph" as an element/role is being announced? Or what is the problem we need to fix?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Note that attribute `aria-labelledby` is connected with `id` attribute of the component that is the base for the Tooltip. Without that connection, the `aria-labelledby` attribute is ignored and not read by screen reader.
+
+The content of attribute `aria-labelledby` is prioritized before the content of `aria-label`. In case `aria-labelledby` is not filled in,
+then the content of `aria-label` is read by screen reader.
+
+Be sure to include all relevant information on all of the properties that are announced by screen readers.
+
+### Examples
```
domihustinova, line 52: Numbering in your examples is broken, both examples have `1.`, can you take a look at that, please?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
domihustinova: Is this the best example we should provide? What's confusing me is that currently the `Hello world.` would be announced which I am still not sure if that's wrong or not. And for that we don't need `aria-label` nor `aria-labelledby` so I think for the showcase of usage of `aria-labelledby` I would put `id` to something different than children inside the `<Tooltip>`, WDYT? Or maybe also the example with icon could be shown?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
domihustinova:
```suggestion
The `aria-labelledby` prop can reference multiple ids, separated by a space. The value of the attribute is not announced by the screen reader, but the content of the referred element(s) is.
```

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,84 @@
+---
+title: Accessibility
+redirect_from:
+  - /components/tooltip/accessibility/
+---
+
+# Accessibility
```
domihustinova, line 7: I will post my comment here because it is connected to multiple places. I am super confused what is the purpose of the `aria-label` and `aria-labelledby` props in this component so I have a lot of questions 🙈 

1. In the current implementation, `TooltipWrapper` has the role `button`, is that correct? Do you know why it has this role? I was trying to use screen reader and the "button" announce was always very confusing.

2. If there is a text provided as a children, there are no a11y violations - is it enough like this or should the `aria-label` (or `aria-labelledby`) be provided anyway, maybe with prefix "tooltip", that there is a tooltip? All of the violations with tooltip I saw were caused by the triggering (reference) element being icon - is this one of the examples where should the `aria-label` (or `aria-labelledby`) be provided or it should really be provided always, e.g. to avoid (double) announce of the content of reference element (I am kinda repeating myself here but you get what I am trying to ask 😅)? Should this be mentioned? The double announce happened to me on booking page with the self-transfer note, will we fix this?

3. Have you take a look at the [aria-details](https://www.w3.org/TR/wai-aria/#aria-details) property? Is it something we should also add here? If I understand correctly, it might be useful when there is more complex content inside the tooltip (e.g. lists) so the content of the element referenced by `aria-details` is not flattened to a string when presented to assistive technology users. I am asking as that seems to be a common usage of tooltips across our pages.

img

5. Two of three examples are the same, maybe one can be removed? I am not sure what should be the correct behavior with the children and what we need to fix, but if the `aria-label` (or `aria-labelledby`) are not provided and children is text, component is already accessible (or at least it doesn't violate anything) and these props don't need to be added. Should we mention that or is that just adding the expected things to the documentation (and what Daniel said is a bad precedent)? Maybe we could mention what should be passed to `aria-label` ideally or how should the element(s) being referenced by the `aria-labelledby` look like - should it sum up the content of the tooltip? Or just make introduction to tooltip content? Maybe the example can be real (as in the Seat component documentation where it mentions seat for a specific passenger) instead of just lorem ipsum text. Current examples also suggests that the `aria-labelledby` needs to be connected with the reference element but that't not true, no? Maybe we can use it with different element than children in the example?

packages/orbit-components/src/primitives/TooltipPrimitive/README.md
```diff
@@ -19,7 +19,7 @@ After adding import into your project you can use it simply like:
 Table below contains all types of the props available in the TooltipPrimitive component.
 
 | Name                 | Type                          | Default  | Description                                                                                                                |
-| :------------------- | :---------------------------- | :------- | :------------------------------------------------------------------------------------------------------------------------- |
+| :------------------- | :---------------------------- | :------- | :------------------------------------------------------------------------------------------------------------------------- | --- |
```
domihustinova, line 22: This table is broken, can you take a look at that, please?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
DSil: The `"Lorem ipsum dolor sit amet"` is also announced because the tooltip is focused when it's open (it shouldn't, but that's an issue for another moment)

domihustinova: Isn't that the desired behavior to announce the tooltip content (at least I was under the impression it is, from reading the W3 materials)? Can you share some resources on this, please?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: I'll rewrite it to:
```
Be sure to include all relevant information that is announced by screen readers.
```

or

```
Be sure to fill in all relevant attributes whose values are announced by the screen readers.
```

My point was to note that all attributes that are announced by screen readers should be filled in to reach the a11y goal.

DSil: But that's not necessarily true. If the tooltip is able to announce itself, maybe we don't need any `aria-`.

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.stories.tsx
```diff
@@ -188,7 +202,7 @@ export const WithImageInside: Story = {
 };
 
 export const Playground: Story = {
-  render: ({ children, content, ...args }) => (
+  render: ({ children, content, "aria-labelledby": ariaLabelledby, ...args }) => (
```
domihustinova, line 205: I was thinking, maybe we can fix the stories a bit (applies to all 3 files) - since it's not recommended to use interactive elements inside the tooltip, we would remove the link from there. And we could also either remove the stories with tooltip on icon or add aria- attributes there as it makes more sense to have it there than here in Playground where it's already accessible with text. WDYT?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
DSil: The content of the tooltip **must** be announced, but not the way we are doing it, I suppose. Now with the migration to floating-ui things are a bit better, but in the past they were announced because we forced focus on the element, which should not be the case…

I am just saying that: if we have the content already being announced naturally (even if the implementation isn't perfect), what is the point of the aria-label? What should one put on the `aria-label` to be announced that is not announced by the content itself?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: yes, I get the point (from your comment below).
In that case, I think we can delete this sentence at all. WDYT?

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
domihustinova: > I get the point that aria-label is not needed (?) in this case when the tooltip content itself is read anyway.

Now I am confused by _you_ 😅

From my understanding, the `aria-label`/`aria-labelledby` is not needed when reference element contains understandable text:

Snímek obrazovky 2024-11-20 v 15 39 11

but is needed when we use the icon:

Snímek obrazovky 2024-11-20 v 15 36 40

But maybe if the icon has accessible text already, `aria-label` on the Tooltip reference element is not needed at all?

sarkaaa: yes, that's what I mean, sorry for the confusion. :D

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.stories.tsx
sarkaaa: yes, I was also considering to align the stories based on the updated tooltip and our observation.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: tbh, it's not clear to me what do you mean exactly. 

I removed the `The conjugation of these...` sentence at all, it looks redundant to me in the whole doc specification.

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
DSil: Yes, I think it's a general accessibility sentence and not really tooltip focused, so we can omit it from here

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: 1. I've found that the [triggering component should be interactive](https://sarahmhigley.com/writing/tooltips-in-wcag-21/) - it was written on more websites, but I can't find it right now. I've found some sources talking about the `<a>` element, but AFAIK the `role=button` should be a general role for this. On the other side, we don't know what devs (will) put inside as a children component, so it might escalate other violation issues (?) -
Example: in case there will be a `<button>` element as children, then we can get the `Interactive controls must not be nested`.
 
2. I think it was discussed above in the comment. ⏫ 

3. No, I haven't checked this aria-details property, tbh. It's hard to say if it makes sense to add this prop, on the other side, according to the articles about tooltips, the tooltip content should be simple, without any images/links/whatever, on the other side, we have various content in our tooltips across the Kiwi pages. 🤔 
I'm confused and lost about which approach should be better, tbh.

4. Yes, the two examples are the same, as I wanted to describe the further behavior. But I can take a look at it, and polish it, ofc.

sarkaaa:
Hi guys 👋 Thank you for the exhausting CR. I made some changes based on the comments (excluding SB stories cleanup). Feel free to take a look. Thank you! ❤️ 

@DSil @domihustinova

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: Back to the first comment, I made some changes...

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
sarkaaa: Yes, that's true that the "children" text (`Cras elementum` in this case) is read without and the `aria-label` is not needed at all.

In case the children element is not readable, but it is (for example) and icon, then the `aria-label` should be read.

I removed `The conjugation of these properties allows...` in the end.

DSil: > In case the children element is not readable, but it is (for example) and icon, then the aria-label should be read.

In that case, it should be the Icon to have the aria-label

sarkaaa: Yes, you're right, I tried to set `aria-label` attribute for children component and it seems there are not any violations. I'll change the doc file.

sarkaaa:
Based on the discussion with Daniel during our 1on1, it seems that the tooltip doesn't need `aria` labels at all in the end. These labels depend on the children's elements. However, during the implementation, I found that we had two _divs_ (wrapper and inner) with the same `role="tooltip"` attribute, which is not the best practice. This PR includes the fix of this inconsistency.

Please, don't take in mind the last (revert) commit, it's temporary. Once the changes are approved, I'll do a commit cleanup.

@DSil @domihustinova

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.tsx
```diff
@@ -105,54 +105,57 @@ const TooltipContent = ({
...
+  };
+
   return (
-    // Disabling because the onClick exists just to stop propagation of events
+    // Disabling because the onClick exists to close tooltip when clicking in interactibe elements, which should not happen with keyboard
     // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
-    <div className="w-full" role="tooltip" id={tooltipId} data-test={dataTest} onClick={onClick}>
-      {/* Disabling because the onClick exists to close tooltip when clicking in interactibe elements, which should not happen with keyboard */}
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
+    <div
```
sarkaaa, line 118: In this file, I removed `<div>` wrapper and the  `onClick` event on child `div` calls `handleCombinedClick`. This function includes two functions

- `onClick` - passed from the parent
- `handleInnerClick`

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
DSil: Regarding point 1., I believe the `role="button"` is present because on mobile it is actually a clickable element, to trigger the MobileDialog. If that is the most correct approach…? I honestly don't know 😓 probably the role should be added only when it is indeed expecting a click. However, that would not prevent the raised question: what happens if one uses button as a tooltip trigger (should one even do it, ever?)

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```jsx
+<Tooltip
+  content={
+    <div>
+      <p>Lorem ipsum dolor sit amet.</p>
+    </div>
+  }
+>
+  <Icons.Airplane ariaLabel="More information" />
```
DSil, line 45: Users don't usually use their icons like this. They import the icon directly, instead of the whole Icons set.
```suggestion
  <Airplane ariaLabel="More information" />
```

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,49 @@
...
+</Tooltip>
+```
+
+The screen reader will announce: `More information`. After the user activates the icon, the screen reader will announce: `Lorem ipsum dolor sit amet.`.
```
DSil, line 49: I'd rephrase this. It's not clear to me what `user activates` means. I'd go with something simpler like

>Once the icon is focused by the screen reader, it will announce "More information", followed by the content of the tooltip: "Lorem ipsum…"

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,49 @@
...
+    <div>
+      <p>Lorem ipsum dolor sit amet.</p>
+    </div>
+  }
+>
+  <Text>Learn more.</Text>
+</Tooltip>
+```
+
+The screen reader will announce: `Learn more. Lorem ipsum dolor sit amet.`.
```
DSil, line 31: Quotes should be with proper quoting chars: `"`, and not code snippets. (The same for the example below)

DSil (commented):
Good work on the refactor 💪 just some details left

docs/src/documentation/03-components/04-overlay/tooltip/03-accessibility.mdx
```diff
@@ -0,0 +1,49 @@
...
+  - /components/tooltip/accessibility/
+---
+
+## Accessibility
+
+### Tooltip
+
+The Tooltip component has been designed with accessibility in mind. It can be used with keyboard navigation and includes properties that enhance the experience for users of assistive technologies.
+
+While the `aria-label` and `aria-labelledby` attributes are not needed for the Tooltip component itself, it is important to ensure that the components passed as children to the Tooltip have appropriate ARIA labels. This ensures that the content within the Tooltip is accessible and properly announced by screen readers.
```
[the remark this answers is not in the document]
sarkaaa: yes, thanks. when I was writing the first mentioned sentence, I had a version with `Icon` component in my head. 😅

packages/orbit-components/src/primitives/TooltipPrimitive/components/TooltipContent.tsx
```diff
@@ -105,54 +105,57 @@ const TooltipContent = ({
...
+  const handleCombinedClick = ev => {
+    handleInnerClick(ev);
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    onClick(ev);
+  };
+
   return (
-    // Disabling because the onClick exists just to stop propagation of events
+    // Disabling because the onClick exists to close tooltip when clicking in interactibe elements, which should not happen with keyboard
```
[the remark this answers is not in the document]
sarkaaa: Leftover from the past. :D Fixed.

DSil:
Changes in stories are **never** a _feat_. They're **chore**.

The removal of the incorrect role is a **fix**.

sarkaaa:
> Changes in stories are **never** a _feat_. They're **chore**.
> 
> The removal of the incorrect role is a **fix**.

Fixed and I squashed SB commits as the changes are almost the same. THx.

domihustinova (approved):

DSil (approved):

PR #4516, opened 2024-11-14
Add accessibility props to Heading

DSil:
Two new props added `role` and `level`. They're conditional props and allow to define the role and level of the rendered elements.

The component also no longer automatically renders `role="heading"` when the rendered element is a `div`.

FEPLT-2155

### DIFF
--- a/docs/src/__examples__/Heading/DEFAULT.tsx
+++ b/docs/src/__examples__/Heading/DEFAULT.tsx
@@ -2,7 +2,11 @@ import React from "react";
 import { Heading } from "@kiwicom/orbit-components";
 
 export default {
-  Example: () => <Heading>Orbit components</Heading>,
+  Example: () => (
+    <Heading role="heading" level={1}>
+      Orbit components
+    </Heading>
+  ),
   exampleKnobs: [
     {
       component: "Heading",
@@ -26,7 +30,7 @@ export default {
             "title5",
             "title6",
           ],
-          defaultValue: "div",
+          defaultValue: "title1",
         },
       ],
     },
--- a/packages/orbit-components/src/Heading/README.md
+++ b/packages/orbit-components/src/Heading/README.md
@@ -16,22 +16,24 @@ After adding import to your project you can use it simply like:
 
 The table below contains all types of props available in the Heading component.
 
-| Name            | Type                       | Default    | Description                                                                                         |
-| :-------------- | :------------------------- | :--------- | :-------------------------------------------------------------------------------------------------- |
-| as              | [`enum`](#enum)            | `"div"`    | The element used for the root node.                                                                 |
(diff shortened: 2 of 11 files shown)


packages/orbit-components/src/Heading/README.md
```diff
@@ -16,22 +16,24 @@ After adding import to your project you can use it simply like:
...
+| as              | [`enum`](#enum)            | `"div"`    | The element used for the root node.                                                                        |
+| role            | `"heading"`                |            | The role attribute of the element. Can only be defined if `as="div"`. If defined, `level` must be defined. |
+| level           | `number`                   |            | The level of the Heading. Required if `role` is defined as `"heading"`.                                    |
+| children        | `React.Node`               |            | The content of the Heading.                                                                                |
+| dataTest        | `string`                   |            | Optional prop for testing purposes.                                                                        |
+| align           | [`enum`](#enum)            | `left`     | `text-align` of `Heading` component                                                                        |
```
sarkaaa, line 26: just small nitpick - `.` is missing here

docs/src/__examples__/Heading/DEFAULT.tsx
```diff
@@ -2,7 +2,11 @@ import React from "react";
 import { Heading } from "@kiwicom/orbit-components";
 
 export default {
-  Example: () => <Heading>Orbit components</Heading>,
+  Example: () => (
```
sarkaaa, line 5: What do you think about adding `role` and `level` attrs also for [Typography](https://dsil-add-a11y-heading.orbit.pages.dev/foundation/typography/) section in Orbit docs? Does it make sense for you, although it's not sooo "important"?

DSil: I don't think it makes a difference, they're not visible. The Typography section should be more to display the different visual variations we have for text

packages/orbit-components/src/Heading/index.tsx
```diff
@@ -44,9 +46,10 @@ const Heading = ({
 
   return (
     <Component
+      aria-level={Component === "div" ? level : undefined}
       id={id}
       data-test={dataTest}
-      role={Component === "div" ? "heading" : undefined}
+      role={Component === "div" ? role : undefined}
```
sarkaaa, line 52: I would add these new props (`aria-level`, `role`) to Storybook as well (Playground story).

sarkaaa: There are conditional controls in SB, however, it's limited and we can't include two conditions for one control (show `level` when `as="div"` **and** `role="heading"`). :(

DSil: Precisely, I'd avoid unnecessary complication for something users can't see 😕

sarkaaa (approved):
LGTM.

PR #4527, opened 2024-11-25
Add label to Separator

DSil:
Some minor changes in visual tests probably due to small pixel shifts related to the styles reimplementation, to preserve original features.

FEPLT-2086


packages/orbit-components/src/Separator/Separator.stories.tsx
```diff
@@ -77,6 +77,11 @@ export const Playground: Story = {
         type: "select",
       },
     },
+    label: {
```
sarkaaa, line 80: Value for label should be added in `args` (now there is 'Set object' button in SB):
```
args: {
    ...
    label: "Your label",
  },
```

and this could be removed as it's "text" type and it's not necessary ("text" is default type in SB).

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -80,18 +80,24 @@ const Separator = ({
...
-          "mt-0 box-border h-0 border-t",
+          "absolute left-0 top-1/2 mt-0 box-border -translate-y-1/2 border-t",
           color || "border-elevation-flat-border-color",
+          getSideOffsetAmount(sideOffset, align),
           BORDER_TYPE_CLASSES[type],
-          spaceAfter && getSpaceAfterClasses(spaceAfter),
         )}
       />
+      {label && (
+        <span className="bg-white-normal px-100 -translate-x-50 absolute start-1/2 top-1/2 min-w-max -translate-y-1/2">
```
sarkaaa, line 97: Label text inside the separator is not centered:
image

packages/orbit-components/src/Separator/index.tsx
```diff
@@ -31,43 +31,43 @@ const enum Indent {
...
+      [Indent.THREE_HUNDRED]: "ms-300 w-[calc(100%_-_theme(spacing.300)*2)]",
+      [Indent.small]: "ms-300 w-[calc(100%_-_theme(spacing.300)*2)]", // deprecated
+      [Indent.FOUR_HUNDRED]: "ms-400 w-[calc(100%_-_theme(spacing.400)*2)]",
+      [Indent.medium]: "ms-400 w-[calc(100%_-_theme(spacing.400)*2)]", // deprecated
+      [Indent.SIX_HUNDRED]: "ms-600 w-[calc(100%_-_theme(spacing.600)*2)]",
+      [Indent.large]: "ms-600 w-[calc(100%_-_theme(spacing.600)*2)]", // deprecated
+      [Indent.EIGHT_HUNDRED]: "ms-800 w-[calc(100%_-_theme(spacing.800)*2)]",
+      [Indent.XLarge]: "ms-800 w-[calc(100%_-_theme(spacing.800)*2)]", // deprecated
+      [Indent.ONE_THOUSAND]: "ms-1000 w-[calc(100%_-_theme(spacing.1000)*2)]",
+      [Indent.XXLarge]: "ms-1000 w-[calc(100%_-_theme(spacing.1000)*2)]", // deprecated
```
sarkaaa, line 70: SideOffset & align for RTL version is not cooperating well:
- for `left | right` align, I get the offset on the right side

image

- for `center` align, I get doubled-offset on the right side

image

packages/orbit-components/src/Separator/Separator.stories.tsx
sarkaaa, line 1: As the values `small`, `medium`, `large`, etc.. are deprecated, I would replace these values as well in this file on line 61.

packages/orbit-components/src/Separator/Separator.stories.tsx
DSil: I don't see the Set object you refer to 🤔

packages/orbit-components/src/Separator/index.tsx
DSil: You should use the Text element, not direct string

DSil: In Storybook it won't be possible, but that's for display purposes only. The prop is documented as accepting ReactNode, so devs should be aware of the correct usage

packages/orbit-components/src/Separator/Separator.stories.tsx
sarkaaa: We were used to avoiding this button when we were doing SB tasks with domi, and we went by an empty string in `args` object.

I think it's not bad to have there `Set String` button, it was just our convention to avoid it.

packages/orbit-components/src/Separator/index.tsx
sarkaaa: ReactNode also includes `string` prop, so it's possible to add a string there as well. Also, in the [documentation](https://dsil-add-label-separator.orbit.pages.dev/components/structure/separator/react/), there is `string` type for `label` prop mentioned.

Although the text is wrapped by the `<Text>` component, it's not still centered.

DSil: Yes, documentation needs a change. And yes, a string is possible, but that does not mean we have to enforce the styles. On a Design System, one is expected to render text using typography components, no direct strings, especially if it's documented.

I see the Text centered with the `<Text>` component
image

packages/orbit-components/src/Separator/Separator.stories.tsx
DSil: Alright, I'll have empty string

packages/orbit-components/src/Separator/index.tsx
sarkaaa: If you mean vertical-align, then it's correct, I mean horizontal-align, which is still incorrect. :( 
image

DSil: Ah, got it, it was a wrong class. Fixed. Visual tests updating

DSil: Fixed

sarkaaa (approved):
LGTM. ✨

PR #4533, opened 2024-11-27
fix(HorizontalScroll): adjust arrows visibility when component is ove…

sarkaaa:
…rflowing

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-1749


sarkaaa:
I tested changes in editor on [docs page](https://sarka-horizontal-scroll-arro.orbit.pages.dev/components/layout/horizontalscroll/) and also in Playground:
[Production](https://orbit.kiwi/playground/) playground x [Branch](https://sarka-horizontal-scroll-arro.orbit.pages.dev/playground/) playground
Added this code snippet:
```
() => {
  return (
    <div style={{ backgroundColor: "lightgray", width: 200 }}>
      <Tabs>
        <HorizontalScroll
          arrows
          overflowElevation
          elevationColor="paletteWhite"
        >
          <TabList>
            <Tab>Tab 0</Tab>
            <Tab>Tab 1</Tab>
            <Tab>Tab 2</Tab>
          </TabList>
        </HorizontalScroll>
      </Tabs>
    </div>
  );
};

```

docs/src/__examples__/HorizontalScroll/default.tsx
```diff
@@ -88,7 +88,20 @@ export default {
         {
           name: "spacing",
           type: "select",
-          options: ["none", "50", "100", "200", "300", "400", "600", "800", "1000"],
+          options: [
```
sarkaaa, line 91: I added some values to have it aligned.

packages/orbit-components/src/HorizontalScroll/index.tsx
```diff
@@ -20,6 +19,35 @@ const getSnap = (scrollSnap: ScrollSnap) => {
   return scrollSnap;
 };
 
+const getTriggerOffset = (spacing: string) => {
```
sarkaaa, line 22: Tbh, I'm not sure if it's the best approach (probably not (?)), however, I needed to get the pixel value of set spacing and use this value as an integer for `TRIGGER_OFFSET` const. This value is used for computations in `handleScroll` functions, which is crucial for setting the visibility of arrow buttons.

DSil:
I have tested with the code provided on the original request and noticed no change. Can you please confirm?

packages/orbit-components/src/HorizontalScroll/index.tsx
```diff
@@ -76,6 +104,8 @@ const HorizontalScroll = React.forwardRef<HTMLDivElement, Props>(
     const theme = useTheme();
     const scrollEl = scrollWrapperRef.current;
 
+    const TRIGGER_OFFSET = getTriggerOffset(spacing);
```
DSil, line 107: All uppercase constants should be the same value every time, not calculated in runtime

packages/orbit-components/src/HorizontalScroll/index.tsx
DSil, line 22: This completely changes the logic of the component, no? 🤔 the spacing was only for the space between elements. If you add it to the arrow positioning logic, it does not work...

Yours:
image


Prod:
image

DSil (commented):
I wonder if the code provided on the bug report is correct. Shouldn't the HorizontalScroll be inside `TabList`? 🤔

packages/orbit-components/src/HorizontalScroll/index.tsx
```diff
@@ -11,8 +11,6 @@ import ChevronBackward from "../icons/ChevronBackward";
 import ChevronForward from "../icons/ChevronForward";
 import type { Props, ScrollSnap } from "./types";
 
-const TRIGGER_OFFSET = 20;
```
DSil, line 14: I'm still curious why we added this in the first place and can't see any new glitch by removing it… So I guess it will have to work for now

DSil (approved):

PR #4541, opened 2024-12-05
fix(Accordion): remove white background for Header and Footer

RobinCsl:
Reported on Slack.

### For new components:


packages/orbit-components/src/Accordion/AccordionSection/components/SectionFooter.tsx
```diff
@@ -8,7 +8,7 @@ interface Props extends Common.Globals {
 
 const AccordionSectionFooter = ({ children, dataTest }: Props) => (
   <div
-    className="p-600 bg-white-normal shadow-level2 sticky bottom-0 flex"
+    className="p-600 shadow-level2 sticky bottom-0 flex"
```
domihustinova, line 11: For the footer, the issue is still there, maybe we could add rounded corners here? 🤔 https://kiwicom-orbit-rcsl-remove-white-bg-accordion.surge.sh/?path=/story/accordion--accordion-with-sticky-footer

Snímek obrazovky 2024-12-06 v 12 49 04

RobinCsl: Thank you Domi, didn't catch that. Should have zoomed in like you did!

I don't want to add the rounded corners (because I would want that to be controlled only by the parent component). But adding `overflow: hidden` might be risky. Still strange that there's something showing there, will check what component imposed that background.

RobinCsl: Also, for it to be properly inset, we'd need a custom border-radius value.

RobinCsl: It seems to be because the footer has some drop-shadow, probably to provide some contrast/elevation with the rest of the content.

domihustinova (approved):
LGTM 👍

PR #4545, opened 2024-12-12
Tooltip now closes on ESC key press

DSil:
Also did some cleanup by moving a component to an external file.

I tried to use [useDismiss](https://floating-ui.com/docs/useDismiss) from floating-ui but there were some conflicts with accepting other events so I ended up just hardcoding an event listener.

For the tests, I used `toBeVisible` because it asserts CSS, that changes immediately to `invisible`. There is a delay on the re-render, so we would have to wait before using `toBeInTheDocument`

FEPLT-2159

### DIFF
--- a/packages/orbit-components/src/primitives/TooltipPrimitive/index.tsx
+++ b/packages/orbit-components/src/primitives/TooltipPrimitive/index.tsx
@@ -1,37 +1,11 @@
 import * as React from "react";
-import cx from "clsx";
 import { FloatingPortal } from "@floating-ui/react";
 
 import useRandomId from "../../hooks/useRandomId";
-import TooltipContent from "./components/TooltipContent";
 import useStateWithTimeout from "../../hooks/useStateWithTimeout";
+import TooltipContent from "./components/TooltipContent";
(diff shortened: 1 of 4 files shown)


packages/orbit-components/src/primitives/TooltipPrimitive/index.tsx
```diff
@@ -61,17 +35,29 @@ const TooltipPrimitive = ({
 
   const tooltipId = useRandomId();
 
+  const handleEsc = React.useCallback(
+    (ev: { key: string }) => {
+      if (ev.key === "Escape") {
+        setShown(false);
+        setRenderWithTimeout(false);
```
sarkaaa, line 42: Out of curiosity, why isn't `document.removeEventListener("keydown", handleEsc);` added also here? I can call the `handleEsc` function, although the tooltip is no longer visible.

https://github.com/user-attachments/assets/c5d5aa32-d9df-4ad6-a6f4-e18620e118c2

DSil: Well spotted. Changed

sarkaaa (approved):
LGTM.

PR #4553, opened 2024-12-18
Collapse: fix a11y violation (interactive nested elements)

sarkaaa:

### For new components:

Closes https://kiwicom.atlassian.net/browse/FEPLT-2186

<details open>

This PR fixes an accessibility violation by addressing interactive nested elements in the Collapse component.

<details>

```mermaid
sequenceDiagram
    actor User
    participant Label
    participant IconButton
    participant CollapseComponent
    participant Content

    Note over Label,IconButton: Two ways to trigger collapse
    
    alt Click Label
        User->>Label: Clicks label
        Label->>CollapseComponent: Triggers handleClick()
        CollapseComponent->>Content: Toggles visibility
    else Click Icon Button
        User->>IconButton: Clicks button
        IconButton->>CollapseComponent: Triggers handleClick()
        CollapseComponent->>Content: Toggles visibility
    end

    Note over CollapseComponent: Updates aria-expanded state
    Note over Content: Content slides in/out
```

</details>

<details>
<table>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4553/files#diff-48429084200f269c10d3ad7773176ecf7d51b14abf7f650000c9c3e35d49b6f1>packages/orbit-components/src/Collapse/Collapse.stories.tsx</a></td><td>Added parameters to exclude 'expanded' control in the story.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4553/files#diff-2f6cc33d95212addda62ad2d66b9739f6029dea17ea502341b6fe27504cc6f2a>packages/orbit-components/src/Collapse/__tests__/index.test.tsx</a></td><td>Refactored tests to simplify button handling and improve readability.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4553/files#diff-b3b6aae52e21bdcedcc41f3be8f2f10f5b00a348c26e9ee1a08e8baff3cbe046>packages/orbit-components/src/Collapse/index.tsx</a></td><td>Updated the Collapse component to improve accessibility and interaction handling.</td></tr>

</table>
</details>

</details>

Closes https://kiwicom.atlassian.net/browse/FEPLT-2186

### DIFF
--- a/packages/orbit-components/src/Collapse/Collapse.stories.tsx
+++ b/packages/orbit-components/src/Collapse/Collapse.stories.tsx
@@ -213,6 +213,12 @@ export const Uncontrolled: Story = {
   args: {
     expanded: undefined,
   },
+
+  parameters: {
+    controls: {
+      exclude: "expanded",
+    },
+  },
 };
 
 export const Rtl: Story = {
--- a/packages/orbit-components/src/Collapse/__tests__/index.test.tsx
+++ b/packages/orbit-components/src/Collapse/__tests__/index.test.tsx
@@ -4,10 +4,7 @@ import userEvent from "@testing-library/user-event";
 import { render, screen } from "../../test-utils";
 import Collapse from "..";
 
-const toggleButtons = [
-  [0, "label"],
-  [1, "icon button"],
-];
+const toggleButtons = ["label", "icon button"];
 
 describe("Collapse", () => {
   const user = userEvent.setup();
@@ -31,32 +28,41 @@ describe("Collapse", () => {
     expect(screen.getByText("customLabel")).toBeInTheDocument();
   });
 
-  it.each(toggleButtons)("should trigger click handler when clicking on %s", async buttonIndex => {
-    const onClick = jest.fn();
-    render(
-      <Collapse label="Collapse" onClick={onClick}>
-        <div>children</div>
-      </Collapse>,
-    );
-    const toggleButton = screen.getAllByRole("button", { name: "Collapse" })[buttonIndex];
-    await user.click(toggleButton);
-    expect(onClick).toHaveBeenCalled();
-  });
+  it.each(toggleButtons)(
+    "should trigger click handler when clicking on %s",
+    async triggerButton => {
+      const onClick = jest.fn();
+      render(
+        <Collapse label="Collapse" onClick={onClick}>
+          <div>children</div>
+        </Collapse>,
+      );
+
+      const toggleButton =
+        triggerButton === "label"
+          ? screen.getByText("Collapse")
+          : screen.getByRole("button", { expanded: false });
+
+      await user.click(toggleButton);
+      expect(onClick).toHaveBeenCalled();
+    },
+  );
 
   describe("uncontrolled", () => {
-    it.each(toggleButtons)("should expand/collapse when clicking on %s", async buttonIndex => {
+    it("should expand/collapse when clicking on button", async () => {
       render(
         <Collapse label="Collapse">
           <article>children</article>
         </Collapse>,
       );
-      const toggleButton = screen.getAllByRole("button", { name: "Collapse" })[buttonIndex];
-      // with ByRole we can test whether the content is visible because of aria-hidden
-      expect(screen.queryByRole("article")).not.toBeInTheDocument();
-      await user.click(toggleButton);
-      expect(screen.getByRole("article")).toBeInTheDocument();
-      await user.click(toggleButton);
-      expect(screen.queryByRole("article")).not.toBeInTheDocument();
+
+      await screen.getAllByRole("button").forEach(async toggleButton => {
+        expect(screen.queryByRole("article")).not.toBeInTheDocument();
+        await user.click(toggleButton);
+        expect(screen.getByRole("article")).toBeInTheDocument();
+        await user.click(toggleButton);
+        expect(screen.queryByRole("article")).not.toBeInTheDocument();
+      });
     });
   });
 
--- a/packages/orbit-components/src/Collapse/index.tsx
+++ b/packages/orbit-components/src/Collapse/index.tsx
@@ -64,36 +64,35 @@ const Collapse = ({
       data-test={dataTest}
       id={id}
     >
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
-      <div
-        className="block w-full cursor-pointer"
-        onClick={handleClick}
-        role="button"
-        tabIndex={-1}
-        id={labelID}
-      >
-        <Stack justify="between" align="center">
-          {label && !customLabel && <Heading type="title4">{label}</Heading>}
-          {customLabel}
-          <Stack inline grow={false} align="center" spacing="300">
-            {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
-            <div
-              className="flex items-center"
-              onClick={ev => {
-                ev.stopPropagation();
-              }}
-            >
-              {actions}
-            </div>
-            <ButtonLink
-              iconLeft={<AnimatedIcon expanded={expanded} />}
-              size="small"
-              type="secondary"
-              ariaLabelledby={labelID}
-              ariaExpanded={expanded}
-              ariaControls={slideID}
-            />
+      <div className="flex items-center justify-between">
+        <div
+          className="flex w-full self-stretch"
+          id={labelID}
+          role="button"
+          tabIndex={0}
+          onClick={handleClick}
+          onKeyDown={e => {
+            if (e.key === "Enter" || e.key === " ") {
+              handleClick(e);
+            }
+          }}
+        >
+          <Stack justify="between" align="center">
+            {label && !customLabel && <Heading type="title4">{label}</Heading>}
+            {customLabel}
           </Stack>
+        </div>
+        <Stack inline grow={false} align="center" spacing="none">
+          {actions && <div className="mx-300 flex items-center">{actions}</div>}
+          <ButtonLink
+            iconLeft={<AnimatedIcon expanded={expanded} />}
+            size="small"
+            type="secondary"
+            onClick={handleClick}
+            ariaLabelledby={labelID}
+            ariaExpanded={expanded}
+            ariaControls={slideID}
+          />
         </Stack>
       </div>
       <Slide maxHeight={height} expanded={expanded} id={slideID} ariaLabelledBy={labelID}>


packages/orbit-components/src/Collapse/index.tsx
```diff
@@ -64,14 +64,8 @@ const Collapse = ({
...
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
-      <div
-        className="block w-full cursor-pointer"
-        onClick={handleClick}
-        role="button"
-        tabIndex={-1}
-        id={labelID}
-      >
+      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
+      <div className="block w-full cursor-pointer" id={labelID} onClick={handleClick} role="group">
```
mvidalgarcia, line 68: Could you please explain why we're using the `group` role here now?
In the end, the div is acting as a button, isn't it? It even has an onClick handler.

DSil: IMO, the div should be kept as a button and the inner ButtonLink should be rendered as a div

mvidalgarcia: > the inner ButtonLink should be rendered as a div

I assume this would remove the role="button" quality, right?

DSil: Yes, because the role="button" is only applied if the element is natively a button or if it has an onClick, which it doesn't, since the onClick is on the outer div

sarkaaa: The reason why I was focused on the wrapper element (button/div) is that there is an optional prop called `actions` which can contain an element with `onClick` function could be passed as well - [example](https://kiwicom.github.io/orbit/?path=/story/collapse--with-actions).

In that case we will still get the same a11y violation.

mvidalgarcia: You're totally right, I hadn't considered we could add actions there 😅 good catch!

In that case, I believe it would be better to refactor the component a little bit to have the interactive elements as siblings instead of children.

My idea (pseudo-code) would be to have something like this:

```jsx
<div className="main-wrapper">
    {/* Main content */}
    <button 
        className="flex-grow"
        onClick={handleClick}
    />

    {/* Separate actions container */}
    <div className="actions-wrapper">        
        {actions && <button onClick={action}>Action</button>}
        <ButtonLink
              iconLeft={<AnimatedIcon expanded={expanded} />}
              size="small"
              type="secondary"
              ariaLabelledby={labelID}
              ariaExpanded={expanded}
              ariaControls={slideID}
            />
    </div>
</div>
```

Let me know what you think, I'm interested in your input as you have been checking this component for a while :)

If we go for this approach, let's make sure we have visual tests with actions before doing the refactor so we have rely on them.

packages/orbit-components/src/Collapse/index.tsx
```diff
@@ -64,36 +64,39 @@ const Collapse = ({
...
+          }}
+        >
+          <Stack justify="between" align="center">
+            {label && !customLabel && <Heading type="title4">{label}</Heading>}
+            {customLabel}
+          </Stack>
+        </div>
+        <Stack inline grow={false} align="center" spacing="none">
+          {actions && (
+            <div aria-labelledby={labelID} className="mx-300 flex items-center">
```
sarkaaa, line 87: The div is visible only when `actions !== null` (because of the set x-margin). I also set the `aria-labelledby` attribute to inform the user about the connection between the action button/link and the current collapse element. I removed the "eslint ignoring comment" as we don't need the `onClick` function on the div.

packages/orbit-components/src/Collapse/index.tsx
```diff
@@ -64,36 +64,39 @@ const Collapse = ({
...
-              ariaExpanded={expanded}
-              ariaControls={slideID}
-            />
-          </Stack>
+          )}
+          <ButtonLink
+            iconLeft={<AnimatedIcon expanded={expanded} />}
+            size="small"
+            type="secondary"
+            onClick={handleClick}
```
sarkaaa, line 95: I set `onClick` prop as this ButtonLink is not wrapped in a `div[role=button]` anymore.

packages/orbit-components/src/Collapse/index.tsx
sarkaaa: Yes, I agree. I made some changes and added comments, feel free to check it out. 🙏 Tests are not added/updated yet, but working on that. ⌨️

packages/orbit-components/src/Collapse/Collapse.ct.tsx
```diff
@@ -4,9 +4,19 @@ import { test, expect } from "@playwright/experimental-ct-react";
...
+  test("Default collapse", async ({ mount }) => {
     const component = await mount(<CollapseStory />);
 
     await expect(component).toHaveScreenshot();
   });
+
+  test("First Collapse opens on label click and last Collapse opens on arrow button click", async ({
+    mount,
+  }) => {
+    const component = await mount(<CollapseStory />);
```
sarkaaa, line 16: I included visual tests of collapsed component on click for wrapper & arrow button within one test, because I wanted to avoid the huge number of screenshots.

packages/orbit-components/src/Collapse/Collapse.ct-story.tsx
```diff
@@ -19,9 +19,15 @@ export default function CollapseStory() {
...
-      <Collapse label="Expanded" initialExpanded>
+      <Collapse label="Initial expanded" initialExpanded>
+        <p>Initial expanded content</p>
+      </Collapse>
+      <Collapse label="Expanded" expanded>
         <p>Expanded content</p>
       </Collapse>
+      <Collapse label="Collapsed">
+        <p>Content</p>
+      </Collapse>
```
mvidalgarcia, line 30: Why do we need these new Collapses?

My idea was to rely on the existing tests with actions, so we're sure nothing changed visually after the component refactor. Perhaps I'm missing some details.

packages/orbit-components/src/Collapse/index.tsx
mvidalgarcia: > The div is visible only when actions !== null (because of the set x-margin).

Makes sense! although I'm not sure I understand the "set x-margin" part.

> I also set the aria-labelledby attribute to inform the user about the connection between the action button/link and the current collapse element. 

I'm not sure if this is correct 🤔. The div with `labelID` only collapses/expands the content whereas `actions` might trigger a completely different functionality (e.g. clear some collapsed filters).

@DSil I'm interested in your opinion here.

> I removed the "eslint ignoring comment" as we don't need the onClick function on the div.

Perfect! The fewer suppressed warnings, the better.

packages/orbit-components/src/Collapse/index.tsx
mvidalgarcia: Yes! Perfect, as it's a sibling, it needs its own handler 👍

packages/orbit-components/src/Collapse/Collapse.ct-story.tsx
sarkaaa: My idea was to cover cases when `Collapse` is clicked (case for click on collapse wrapper, case for click on collapse arrow button). Before that, we had visual snapshots of expanded collapses. Do you think these cases are unnecessary?

mvidalgarcia: Hmm, I guess those cases are better covered by unit tests instead, what do you think?

If we keep visual tests as they were, we can make sure that the refactor is not affecting the component visually, which is very helpful.

sarkaaa: Ok, I agree. I'm reverting those test back. ♻️

packages/orbit-components/src/Collapse/Collapse.stories.tsx
```diff
@@ -213,6 +213,12 @@ export const Uncontrolled: Story = {
   args: {
     expanded: undefined,
   },
+
+  parameters: {
+    controls: {
+      exclude: "expanded",
+    },
+  },
```
mvidalgarcia, line 221: Do we still need this?

mvidalgarcia: Nevermind, I think what you did it's ok 👍

packages/orbit-components/src/Collapse/index.tsx
```diff
@@ -64,36 +64,39 @@ const Collapse = ({
...
-            {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
-            <div
-              className="flex items-center"
-              onClick={ev => {
-                ev.stopPropagation();
-              }}
-            >
+      <div className="flex items-center justify-between">
+        <div
+          className="flex w-full cursor-pointer self-stretch"
```
mvidalgarcia, line 69: I believe we don't need `cursor-pointer` when the role is a button, but I'm not 100% sure, could you confirm?

sarkaaa: good catch, thx. Fixed. 👍

packages/orbit-components/src/Collapse/index.tsx
mvidalgarcia: I checked with @DSil offline, and we don't need the aria-labelledby on the actions wrapper

sarkaaa: Regarding using `labelled-by` in static div: I think you're correct and we probably should get rid of it.

_NOTE 💡 : Yesterday I've received a newsletter with a link leading to [this article](https://benmyers.dev/blog/dont-use-aria-label-on-static-text-elements/?utm_source=stefanjudis&utm_medium=email&utm_campaign=web-weekly-148-scrollbar-gutter-ethical-web) about using `aria-label`, `aria-labelledby`._

packages/orbit-components/src/Collapse/__tests__/index.test.tsx
```diff
@@ -31,32 +26,35 @@ describe("Collapse", () => {
...
       </Collapse>,
     );
-    const toggleButton = screen.getAllByRole("button", { name: "Collapse" })[buttonIndex];
-    await user.click(toggleButton);
-    expect(onClick).toHaveBeenCalled();
+
+    screen.getAllByRole("button").forEach(async toggleButton => {
+      await user.click(toggleButton);
+      expect(onClick).toHaveBeenCalled();
+    });
```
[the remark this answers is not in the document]
sarkaaa: Fixed. ➡️  https://github.com/kiwicom/orbit/pull/4553/files#r1918731472

mvidalgarcia (approved):
Greeeeat 🎉

PR #4557, opened 2024-12-19
fix(Popover): popover content now closes on ESC key press

sarkaaa:
closes https://kiwicom.atlassian.net/browse/FEPLT-2185 

### For new components:

<details open>

This PR adds functionality to close the Popover component when the ESC key is pressed, enhancing user experience and accessibility.

<details>

```mermaid
sequenceDiagram
    participant User
    participant Popover
    participant Document

    Note over User,Document: New Keyboard Interactions

    alt Enter Key Press
        User->>Popover: Press Enter (when focused)
        Popover->>Popover: Show content
        Popover->>Document: Add keydown listener
    end

    alt Click Interaction
        User->>Popover: Click trigger
        Popover->>Popover: Show content
        Popover->>Document: Add keydown listener
    end

    alt Escape Key Press
        User->>Document: Press Escape
        Document->>Popover: Trigger handleEsc
        Popover->>Popover: closePopover()
        Popover->>Popover: Hide content
        Popover->>Document: Remove keydown listener
    end
```

</details>

<details>
<table>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4557/files#diff-ca29b1b6c6e3ddd66c963c9eb160b7558afe340a282948a873269ca514eda188>packages/orbit-components/src/Popover/Popover.stories.tsx</a></td><td>Updated the story to remove unnecessary state management for the Popover component.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4557/files#diff-9ca39918a965f479ab1b26f600baee10d735108736c0babb261fe247b0605444>packages/orbit-components/src/Popover/README.md</a></td><td>Updated documentation to reflect changes in the Popover component's props and behavior.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4557/files#diff-79bc5314c287684d5433a86b13bf18b4711347b950f3d806e0b0a13737fea98e>packages/orbit-components/src/Popover/__tests__/index.test.tsx</a></td><td>Added tests to verify that the Popover opens on Enter and closes on ESC key press.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4557/files#diff-1749fd3089b157f03d9a036ffcccba86319fe6c34eb3331487f8bcdf94be5c2f>packages/orbit-components/src/Popover/index.tsx</a></td><td>Implemented logic to handle ESC key press for closing the Popover.</td></tr>

</table>
</details>

*This PR includes files in programming languages that we currently do not support. We have not reviewed files with the extensions `.md`. <a href=https://docs.callstack.ai/introduction>See list of supported languages.</a>*

</details>

### DIFF
--- a/packages/orbit-components/src/Popover/Popover.stories.tsx
+++ b/packages/orbit-components/src/Popover/Popover.stories.tsx
@@ -360,7 +360,6 @@ export const LazyContentSimulated: Story = {
 export const InsideCard: Story = {
   render: args => {
     const [isOpened, setIsOpened] = React.useState(false);
-    const [isOpenedPopover, setIsOpenedPopover] = React.useState(false);
 
     return (
       <>
@@ -370,25 +369,15 @@ export const InsideCard: Story = {
               setIsOpened(true);
             }}
             onClose={() => setIsOpened(false)}
-          >
-            <Popover
-              {...args}
-              opened={isOpenedPopover}
-              onClose={() => {
-                setIsOpenedPopover(false);
-              }}
-              content={<div>Content</div>}
-            >
-              <Button
-                onClick={ev => {
-                  ev.stopPropagation();
-                  setIsOpenedPopover(true);
-                }}
-              >
-                Open Popover
-              </Button>
-            </Popover>
-          </CardSection>
+            title={<Text>Card title</Text>}
+            actions={
+              <Popover {...args} content={<div>Content</div>}>
+                <Button asComponent="div" type="secondary">
+                  Open Popover
+                </Button>
+              </Popover>
+            }
+          />
         </Card>
         {isOpened && <Modal onClose={() => setIsOpened(false)}>kek</Modal>}
       </>
--- a/packages/orbit-components/src/Popover/__tests__/index.test.tsx
+++ b/packages/orbit-components/src/Popover/__tests__/index.test.tsx
@@ -142,4 +142,36 @@ describe("Popover", () => {
     await user.click(screen.getByTestId("content-button"));
     expect(contentOnClick).toHaveBeenCalled();
   });
+
+  it("should open on Enter and close on Esc", async () => {
+    render(
+      <Popover content={<div>kek</div>}>
+        <Button dataTest="popover-trigger">Open popover</Button>
+      </Popover>,
+    );
+
+    screen.getByTestId("popover-trigger").focus();
+
+    await user.keyboard("{Enter}");
+    expect(screen.queryByText("kek")).toBeInTheDocument();
+
+    await user.keyboard("{Escape}");
+    expect(screen.queryByText("kek")).not.toBeInTheDocument();
+  });
+
+  it("should open on Click and close on Esc", async () => {
+    render(
+      <Popover content={<div>kek</div>}>
+        <Button dataTest="popover-trigger">Open popover</Button>
+      </Popover>,
+    );
+
+    const popoverTrigger = screen.getByTestId("popover-trigger");
+
+    await user.click(popoverTrigger);
+    expect(screen.queryByText("kek")).toBeInTheDocument();
+
+    await user.keyboard("{Escape}");
+    expect(screen.queryByText("kek")).not.toBeInTheDocument();
+  });
 });
--- a/packages/orbit-components/src/Popover/index.tsx
+++ b/packages/orbit-components/src/Popover/index.tsx
@@ -55,20 +55,34 @@ const Popover = ({
     [onClose, onOpen],
   );
 
-  const handleOut = React.useCallback(
-    (ev?: MouseEvent | React.SyntheticEvent<HTMLElement>) => {
-      // If open prop is present ignore custom handler
-      ev?.stopPropagation();
-      if (ev && ref.current && !ref.current.contains(ev.target as Node)) {
-        if (typeof opened === "undefined") {
-          setShown(false);
-          clearShownTimeout();
-          setRenderWithTimeout(false);
-          resolveCallback(false);
-        } else if (onClose) onClose();
+  const closePopover = React.useCallback(() => {
+    // If open prop is present ignore custom handler
+    if (typeof opened === "undefined") {
+      setShown(false);
+      clearShownTimeout();
+      resolveCallback(false);
+      setRenderWithTimeout(false);
+    } else if (onClose) onClose();
+  }, [setShown, clearShownTimeout, onClose, opened, resolveCallback, setRenderWithTimeout]);
+
+  const handleEsc = React.useCallback(
+    (ev: { key: string }) => {
+      if (ev.key === "Escape") {
+        closePopover();
       }
     },
-    [clearShownTimeout, onClose, opened, resolveCallback, setRenderWithTimeout, setShown],
+    [closePopover],
+  );
+
+  const handleOut = React.useCallback(
+    (ev?: MouseEvent | React.SyntheticEvent<HTMLElement>) => {
+      ev?.stopPropagation();
+
+      if (ev && ref.current && !ref.current.contains(ev.target as Node)) {
+        closePopover();
+      }
+    },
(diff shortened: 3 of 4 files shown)


packages/orbit-components/src/Popover/index.tsx
```diff
@@ -66,6 +66,9 @@ const Popover = ({
           setRenderWithTimeout(false);
           resolveCallback(false);
         } else if (onClose) onClose();
+      } else {
+        setShown(false);
+        resolveCallback(false);
```
domihustinova, line 71: I was checking Storybook and there are several issues now:

1. On Popover open, both `onOpen` & `onClose` events are fired (on mouse click, for keyboard only on second focus)
2. When I close the Popover with ESC, `onClose` event is firing on every ESC click even if the Popover has been closed
3. When I used `opened` prop and provided my own `onOpen` & `onClose` functions, using keyboard I was able to close the Popover with ESC but I couldn't open it afterwards on the second focus. Both `onOpen` & `onClose` events been firing though. When I go back once more (shift + tab) to the first focus, I could open it normally. It worked with mouse click but then the `onClose` event was firing twice. Some issues are present on production already, but adding ESC functionality is adding more mess 😅 

I am not sure if fixing also the current issues (double focus) is in the scope of this MR but maybe we should expand the scope 🤔

sarkaaa: Should be fixed. 🤞

packages/orbit-components/src/Popover/index.tsx
```diff
@@ -113,8 +128,10 @@ const Popover = ({
         clearShownTimeout();
         setRenderWithTimeout(false);
       }
+      document.addEventListener("keydown", handleEsc);
```
domihustinova, line 131: Why are you adding event listener also here since you already added it in `handleClick`? And you are not removing it on the outside click.

And why you just didn't add the event listener here in the effect, e.g. when Popover is shown, with cleanup function but you are doing it inside the `handleClick`? What is the advantage of doing it like this?

packages/orbit-components/src/Popover/Popover.stories.tsx
```diff
@@ -370,25 +369,24 @@ export const InsideCard: Story = {
...
-                setIsOpenedPopover(false);
-              }}
-              content={<div>Content</div>}
-            >
-              <Button
-                onClick={ev => {
-                  ev.stopPropagation();
-                  setIsOpenedPopover(true);
+            title="Card title"
+            actions={
```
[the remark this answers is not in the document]
sarkaaa: This is handled in https://github.com/kiwicom/orbit/pull/4585 . Also, the `Popover` component is moved to `actions` props, because there shouldn't be `onClick` function and interactive controls (popover button) as a `children` prop simultaneously, which will raise an a11y violation.

packages/orbit-components/src/Popover/index.tsx
sarkaaa: I added `addEventListener` to `useEffect` as well, bcs we have to attach this listener in case when Popover is opened by default (prop `opened=true`). If I checked it correctly, when clicked/enter the Popover, only `handleClick` is called and I need to attach the eventListener there.

packages/orbit-components/src/Popover/__tests__/index.test.tsx
```diff
@@ -142,4 +142,38 @@ describe("Popover", () => {
...
+    expect(screen.queryByText("kek")).toBeInTheDocument();
+
+    await user.keyboard("{Escape}");
+    expect(screen.queryByText("kek")).not.toBeInTheDocument();
+  });
+
+  it("should open on Click and close on Esc", async () => {
+    render(
+      <Popover content={<div>kek</div>}>
+        <Button asComponent="div" dataTest="popover-trigger">
```
mvidalgarcia, line 165: Is the `asComponent` required?

packages/orbit-components/src/Popover/Popover.stories.tsx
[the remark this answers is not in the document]
mvidalgarcia: Is this expected? It looks strange to me.


https://github.com/user-attachments/assets/23673881-b3dd-4af8-9f0e-cd2e2f926554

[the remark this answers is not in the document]
sarkaaa: right now, in scope of this PR, yes. once the [CardSection PR is merged](https://github.com/kiwicom/orbit/pull/4585), then it's going to work properly.

packages/orbit-components/src/Popover/__tests__/index.test.tsx
sarkaaa: not in this case, the unit test doesn't check the a11y rules. should be removed now. 👍

packages/orbit-components/src/Popover/index.tsx
```diff
@@ -55,6 +55,18 @@ const Popover = ({
...
   );
 
+  const handleEsc = React.useCallback(
+    (ev: { key: string }) => {
+      if (ev.key === "Escape") {
+        setShown(false);
+        resolveCallback(false);
+        clearShownTimeout();
+        setRenderWithTimeout(false);
+      }
```
domihustinova, line 65: Why are you not differentiating here between `opened` prop being undefined in the same way as it's in the `handleOut`?

sarkaaa: yeah, there could be this double-check. in this case there will be code duplication, so I suggest this approach:
```
  const closePopover = React.useCallback(() => {
    if (typeof opened === "undefined") {
      setShown(false);
      clearShownTimeout();
      resolveCallback(false);
      setRenderWithTimeout(false);
    } else if (onClose) onClose();
  }, [setShown, clearShownTimeout, onClose, opened, resolveCallback, setRenderWithTimeout]);

  const handleEsc = React.useCallback(
    (ev: { key: string }) => {
      if (ev.key === "Escape") {
        closePopover();
      }
    },
    [closePopover],
  );

  const handleOut = React.useCallback(
    (ev?: MouseEvent | React.SyntheticEvent<HTMLElement>) => {
      // If open prop is present ignore custom handler
      ev?.stopPropagation();
      if (ev && ref.current && !ref.current.contains(ev.target as Node)) {
        closePopover();
      }
    },
    [closePopover],
  );
```

wdyt?

packages/orbit-components/src/Popover/Popover.stories.tsx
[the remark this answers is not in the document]
mvidalgarcia: Hmm I see, I find it somewhat hard for reviewers to look at it if it depends on a different branch. What if you change the target branch to the card section PR branch so it's easier to review and add a note informing about it? Does it make sense or is there some limitation?

packages/orbit-components/src/Popover/index.tsx
domihustinova: Seems reasonable to me :+1:

packages/orbit-components/src/Popover/index.tsx
```diff
@@ -55,20 +55,34 @@ const Popover = ({
...
+      if (ev.key === "Escape") {
+        closePopover();
+      }
+    },
+    [closePopover],
+  );
+
   const handleOut = React.useCallback(
     (ev?: MouseEvent | React.SyntheticEvent<HTMLElement>) => {
       // If open prop is present ignore custom handler
```
domihustinova, line 78: Shouldn't this comment be moved to the `closePopover`?

domihustinova (approved):
It looks good to me now 👍 

The only remaining issue is that on the second focus click, both `onOpen` and `onClose` are fired, but that's also happening on production 😬 and will probably be fixed when nested controls are fixed.

I guess you will now remove the commit from the Card PR and merge this one?

PR #4560, opened 2024-12-20
WIP: remove deprecated tokens

DSil:
WIP


packages/orbit-components/src/Box/README.md
```diff
@@ -164,26 +164,18 @@ If you want to define `spacingObject`, you can define specific directions like t
...
-| `"200"`                                     |
-| `"300"`                                     |
-| `"400"`                                     |
-| `"500"`                                     |
-| `"600"`                                     |
-| `"800"`                                     |
-| `"1000"`                                    |
-| `"1200"`                                    |
-| `"1600"`                                    |
+| spacingToken |
```
mvidalgarcia, line 167: Just curious, don't we need the new spacing tokens here?

DSil: I had the same thought. The truth is that these new tokens were not design requests. They were created to be used internally, as some components were using the values of those tokens as custom tailwind classes too often…

So we can have one of two approaches here:
- Don't add them to the components' APIs and wait until they are a design requirement
- Add them and just support from now on

The second option doesn't harm and I am not against it at all. Being honest, I didn't do it because I was focused on everything else this task included 😅 but I am fine with adding it now if we agree on that

packages/orbit-tailwind-preset/src/index.ts
```diff
@@ -8,21 +8,19 @@ import getComponentLevelTokens, { ExportedComponentLevelTokens } from "./getComp
...
+  "heading",
+  "icon",
   "tab",
-  "textLink",
+  "tag",
   "text",
-  "loading",
-  "heading",
-  "countryFlag",
+  "textLink",
```
mvidalgarcia, line 23: I wonder if this is a BC

mvidalgarcia:
> Updates local visual tests (minor pixel adjusts)

Is there any explanation for these changes?

I would have expected to rely on visual tests for this kind of huge PRs

DSil:
> Is there any explanation for these changes?

They were **prior** to this PR tbh. I can remove them and the pipeline will still go all green. These are local snapshots and the disadvantages of not using docker to run the tests locally 😞

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct.tsx-snapshots/Tooltip-primitive-visual-tests-screenshot-for-default---various-placements-1-Desktop-darwin.png
mvidalgarcia, line 1: The text moves a bit in this case; is it expected?

docs/src/documentation/05-development/04-migration-guides/01-v19.mdx
```diff
@@ -16,31 +16,63 @@ With this guide, we aim to walk through all the breaking changes and how they ca
...
+
+### Removal of deprecated props and tokens
+
+Component props, tokens and tailwind classes deprecated in version 16 have been removed.
+
+Refer to the [migration guide for version 16](/migration-guides/v16/) for an extensive list of removed tokens and classes and their replacements.
+
+Besides the ones mentioned, some other tokens (and classes) were affected, namely component specific tokens and breakpoint tokens.
+
+The component specific tokens and classes were not documented and were not expected to be used.
```
mvidalgarcia, line 56:
```suggestion
Besides the ones mentioned, some other tokens (and classes) were affected, namely component-specific tokens and breakpoint tokens.

The component-specific tokens and classes were not documented and were not expected to be used.
```

docs/src/documentation/05-development/04-migration-guides/01-v19.mdx
```diff
@@ -16,31 +16,63 @@ With this guide, we aim to walk through all the breaking changes and how they ca
...
+- `widthBreakpointMediumMobile` -> `breakpointMediumMobile`
+- `widthBreakpointLargeMobile` -> `breakpointLargeMobile`
+- `widthBreakpointTablet` -> `breakpointTablet`
+- `widthBreakpointDesktop` -> `breakpointDesktop`
+- `widthBreakpointLargeDesktop` -> `breakpointLargeDesktop`
+
+#### New spacing tokens
+
+Two new spacing tokens were added. One replaces the deprecated `spaceXXXLarge` and the other introduces a new value.
+These tokens are available in the `theme` object. And the corresponding tailwind classes are also available.
```
mvidalgarcia, line 74: Nitpick, feel free to disregard

```suggestion
These tokens are available in the `theme` object and the corresponding tailwind classes are also available.
```

mvidalgarcia (approved):
Nice cleanup 😅 🧹 

Went through all the commits although not super thoroughly as I mainly rely on the integration tools and the token/class usage breaks + visual tests.

mvidalgarcia:
> > Is there any explanation for these changes?
> 
> They were **prior** to this PR tbh. I can remove them and the pipeline will still go all green. These are local snapshots and the disadvantages of not using docker to run the tests locally 😞

Ok ok, that's enough for me, didn't know that.

packages/orbit-components/src/Box/README.md
mvidalgarcia: I'm ok with 1), not sure if this should be communicated to designers though, they might have a stronger opinion.

> as some components were using the values of those tokens as custom tailwind classes too often…

🤔 ...and could those classes (I assume it's something like `p-[28px]` -> `p-700`) be avoided by adding a new prop?

packages/orbit-tailwind-preset/src/index.ts
DSil: They're marked as BC on the individual commits that correspond to the class removals. This only tells the build phase to not consider those component tokens (because they no longer exist). Plus, this refers to component-specific classes, that are undocumented (although I documented their removal as BC 🤷‍♂️), so I don't think this step in particular needs to be a BC

packages/orbit-components/src/primitives/TooltipPrimitive/TooltipPrimitive.ct.tsx-snapshots/Tooltip-primitive-visual-tests-screenshot-for-default---various-placements-1-Desktop-darwin.png
DSil: Hmm yes, I guess. I think the problem was with the underline style, that ends up shifting the text by a few pixels. Can I ask you to please run the visual tests locally on your computer **on the master branch** and then on this branch and see if the failed ones are the same? 🙏

packages/orbit-components/src/Box/README.md
DSil: Those classes are not tied with props, they're used by ourselves to style the components and are now changed (example [here](https://github.com/kiwicom/orbit/pull/4560/commits/e68eaf544d7ee0e08955a664b050ca156080ff9a#diff-4a2085b61b6cf43b04930132f7d0b3f1d1b50b915a81b175ee23a694318f0f7eR8)). The classes that are matched with props already have tokens assigned.

I will let designers know that we are adding these two tokens anyway and let them decide if they're ok with opening the prop to components that accept token values for spacing. If they say yes, I'll create a separate task that should be straightforward

packages/orbit-tailwind-preset/src/index.ts
mvidalgarcia: Oh true, this is actually not being exported.

I meant the change on [getComponentLevelTokens.ts](https://github.com/kiwicom/orbit/pull/4560/commits/b1ad36c627bc52630426d2094cf3c18ebd5062f6#diff-abb5f236013bdc7f6e61903f2493cfef57942671d9a7edf4e0220b8e4fa372a2), my bad.

In that case, an exported type is being altered. I understand it's internal and undocumented; I'm just trying to make sure this doesn't have side effects on consumers.

packages/orbit-components/src/Box/README.md
mvidalgarcia: Ok, I think I didn't explain myself properly.

> if they're ok with opening the prop to components that accept token values for spacing

This was my point.

For instance, the Inline component doesn't accept the new tokens on the spacing prop.

<details><summary>orbit.kiwi</summary>
<p>

image

</p>
</details> 

But I understand this on designers to decide 👍

DSil: Yes, I decided not to add them because the tokens were created for internal usage needs. If designers never asked for them, it's probably because they don't have those options in Figma. But I agree it wouldn't harm to have them… That being said, if agreed to be added, they'll be added on a separate task.

PR #4622, opened 2025-02-13
Heading docs adjustments

domihustinova:

### For new components:

<details open>

This PR adjusts the documentation for the Heading component, improving clarity on props and their usage, particularly regarding accessibility.

<details>

```mermaid
sequenceDiagram
    participant User
    participant Heading
    participant DOM

    User->>Heading: Render Heading Component
    alt No 'as' prop specified
        Heading->>DOM: Render as div (decorative only)
    else as="div" with role="heading"
        Heading->>DOM: Render div with role="heading"
        Note over Heading,DOM: Requires level prop (1-6)
    else as="h1" to "h6"
        Heading->>DOM: Render semantic heading element
        Note over Heading,DOM: role & level props not allowed
    end
```

</details>

<details>
<table>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4622/files#diff-13208d2cbe2ea86e1c0c97848246ccb95caeed20227a6b3379e3e71eedc23f73>docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx</a></td><td>Updated accessibility documentation for the Heading component, clarifying prop requirements and examples.</td></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4622/files#diff-377eaa24493b8483aeac915f03d6f9809ef3cbd0583f7f5a923784b2215d40bf>packages/orbit-components/src/Heading/README.md</a></td><td>Revised README documentation for the Heading component, enhancing descriptions of props and their defaults.</td></tr>

</table>
</details>

*This PR includes files in programming languages that we currently do not support. We have not reviewed files with the extensions `.mdx`, `.md`. <a href=https://docs.callstack.ai/introduction>See list of supported languages.</a>*

</details>

### DIFF
--- a/docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
+++ b/docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
@@ -13,16 +13,18 @@ The Heading component has been designed with accessibility in mind.
 The component offers flexibility in terms of the HTML element used for the root node, the `role` attribute, and the `level` attribute.
 These properties allow for the creation of semantic and accessible headings.
 
-| Name  | Type                                                    | Description                                                                                                                 |
-| :---- | :------------------------------------------------------ | :-------------------------------------------------------------------------------------------------------------------------- |
-| as    | `"h1" \| "h2" \| "h3" \| "h4" \| "h5" \| "h6" \| "div"` | Defines the HTML element to be rendered.                                                                                    |
-| role  | `"heading"`                                             | Can only be used if `as="div"`. If defined, sets the role of the element to be "heading".                                   |
-| level | `number`                                                | Can only be used if `as="div"` and "`role="heading`". Defines the `aria-level` of the rendered `div` with the heading role. |
+| Name  | Type                                                    | Description                                                                                                            |
+| :---- | :------------------------------------------------------ | :--------------------------------------------------------------------------------------------------------------------- |
+| as    | `"h1" \| "h2" \| "h3" \| "h4" \| "h5" \| "h6" \| "div"` | Defines the HTML element to be rendered.                                                                               |
+| role  | `"heading"`                                             | Can only be used if `as="div"`. If defined, sets the role of the element to be "heading".                              |
+| level | `number`                                                | Must be used if `as="div"` and `role="heading"`. Defines the `aria-level` of the rendered `div` with the heading role. |
 
 All the props above are optional by default.
 
-If they are not provided, the component will render a `div` element with no role or aria-level defined.
-It is not semantically wrong but won't tell screen readers that the element is a heading. This should be used only for decorative purposes.
+### Example 1
+
+If the `as` prop is not provided, the component will render a `div` element.
+In this case, no `role` or `level` are defined by default, and the component will render just a `div` element.
 
 ```jsx
 <Heading>Hello World!</Heading>
@@ -34,8 +36,12 @@ renders:
 <div>Hello World!</div>
 ```
 
-If the `as` prop is set to `"div"` (or undefined), the `role` prop is optional, but only accepts one possible value (if not `undefined`): `"heading"`.
-If the `role` prop is set to `"heading"`, the `level` prop must be defined as well. It will tell assistive technologies the level of the heading.
+It is not semantically wrong but won't tell screen readers that the element is a heading. This should be used only for decorative purposes.
+
+If the `as` prop is set to `"div"` (or undefined), it's possible to set the `role="heading` to indicate to assistive technologies that this element should be treated and formatted like a heading.
+This is the only accepted value for the `role` prop.
+
+In addition, if the `role` prop is set to `"heading"`, the `level` prop **must** be defined as well. It will tell assistive technologies the level of the heading.
 The `level` prop must be a number between 1 and 6 and cannot be used if the `role` prop is not set to `"heading"`.
 
 ```jsx
@@ -50,8 +56,10 @@ renders:
 <div role="heading" aria-level="1">Hello World!</div>
 ```
 
+### Example 2
+
 If the `as` prop is set to `"h1"`, `"h2"`, `"h3"`, `"h4"`, `"h5"`, or `"h6"`, the component will render the corresponding HTML element.
-In that case, the `role` and `level` props are not needed, since assistive technologies will recognize the element as a heading and its correct level automatically.
+In that case, the `role` and `level` props **cannot** be used, since assistive technologies will recognize the element as a heading and its correct level automatically.
 
 ```jsx
 <Heading as="h1">Hello World!</Heading>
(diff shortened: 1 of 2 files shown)


docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -4,25 +4,25 @@ redirect_from:
   - /components/heading/accessibility/
 ---
 
-# Accessibility
-
-## Heading
+## Accessibility
```
DSil, line 7: Ok, we need to get this consistent through all pages 😄

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -4,25 +4,25 @@ redirect_from:
...
-| as    | `"h1" \| "h2" \| "h3" \| "h4" \| "h5" \| "h6" \| "div"` | Defines the HTML element to be rendered.                                                                                    |
-| role  | `"heading"`                                             | Can only be used if `as="div"`. If defined, sets the role of the element to be "heading".                                   |
-| level | `number`                                                | Can only be used if `as="div"` and "`role="heading`". Defines the `aria-level` of the rendered `div` with the heading role. |
+| Name  | Type                                                    | Description                                                                                                                |
+| :---- | :------------------------------------------------------ | :------------------------------------------------------------------------------------------------------------------------- |
+| as    | `"h1" \| "h2" \| "h3" \| "h4" \| "h5" \| "h6" \| "div"` | Defines the HTML element to be rendered as root node.                                                                      |
```
DSil, line 16: I feel having `as root node` is more confusing that nothing. A Heading is not supposed to be a nesting of things, so I think it's pretty straightforward that it refers to the main element node

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -4,25 +4,25 @@ redirect_from:
...
+| role  | `"heading"`                                             | Can only be used if `as="div"`. If defined, sets the role of the element to be "heading".                                  |
+| level | `number`                                                | Can only be used if `as="div"` and `role="heading"`. Defines the `aria-level` of the rendered `div` with the heading role. |
```
DSil, line 18: Actually, this should be

```suggestion
| level | `number`                                                | Must used if `as="div"` and `role="heading"`. Defines the `aria-level` of the rendered `div` with the heading role. |
```

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -4,25 +4,25 @@ redirect_from:
...
 
 All the props above are optional by default.
 
-If they are not provided, the component will render a `div` element with no role or aria-level defined.
-It is not semantically wrong but won't tell screen readers that the element is a heading. This should be used only for decorative purposes.
+### Example 1
+
+If the `as` prop is not provided, the component will render a `div` element as a root node. You can optionally provide prop `as="div"` to make it explicit.
```
DSil, line 24:
```suggestion
If the `as` prop is not provided, the component will render a `div` element.
```

Per what I said above, and based on what is said below, I think this is more straight to the point. No need to be redundant and say that you can pass the default value as a value, as that is real with every prop in React 😄

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -4,25 +4,25 @@ redirect_from:
...
+In this case, no `role` or `level` is defined by default, and the component will render just a `div` element.
```
DSil, line 25:
```suggestion
In this case, no `role` or `level` are defined by default, and the component will render just a `div` element.
```

packages/orbit-components/src/Heading/README.md
```diff
@@ -52,9 +52,11 @@ The table below contains all types of props available in the Heading component.
...
 To make Heading responsive you can use props `mediumMobile`, `largeMobile`, `tablet`, `desktop` and `largeDesktop`,
-which match the [mediaQueries](https://github.com/kiwicom/orbit/tree/master/packages/orbit-components/src/utils/mediaQuery) functions and contain the following properties:
+which match
+the [mediaQueries](https://github.com/kiwicom/orbit/tree/master/packages/orbit-components/src/utils/mediaQuery)
+functions and contain the following properties:
 
 | Name       | Type            | Default    | Description                                 |
 | :--------- | :-------------- | :--------- | :------------------------------------------ |
-| **type**   | [`enum`](#enum) | `"title1"` | The size type of Heading.                   |
+| type       | [`enum`](#enum) | `"title0"` | The size type of Heading.                   |
```
DSil, line 61: Good catch 👍

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
```diff
@@ -34,8 +34,12 @@ renders:
 <div>Hello World!</div>
 ```
 
-If the `as` prop is set to `"div"` (or undefined), the `role` prop is optional, but only accepts one possible value (if not `undefined`): `"heading"`.
-If the `role` prop is set to `"heading"`, the `level` prop must be defined as well. It will tell assistive technologies the level of the heading.
+It is not semantically wrong but won't tell screen readers that the element is a heading. This should be used only for decorative purposes.
+
+Therefore, it's highly recommended to use the `role="heading` to indicate to assistive technologies that this element should be treated and formatted like a heading.
```
DSil, line 39: This is not necessarily true. Sometimes we don't want to recommend heading. The `Heading` component can be used as a decorative element. It is better to not use headings than using them with incorrect semantics.

I would refrain from giving the recommendation. We can specify what each prop does and the effects and consequences, but abstain from recommending

domihustinova: Yes, but if it's not decorative purposes, then it is recommended, no? If it's not possible to use other `as` than `div` for some reason and it's not decorative heading, then they should use it, no?

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
domihustinova: I guess it should be "Defines the HTML element to be rendered" also here then? https://orbit.kiwi/components/text/heading/react/#props

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
DSil: Yes, yes. My only concern is with the term `highly recommended`. I would personally feel pressured to find a reason to use it, and eventually end up using it wrong 😄

docs/src/documentation/03-components/09-text/heading/03-accessibility.mdx
DSil: IMO, absolutely

DSil (approved):

PR #4630, opened 2025-02-24
Tag accessibility

sarkaaa:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2269


packages/orbit-components/src/Tag/index.tsx
```diff
@@ -80,25 +79,33 @@ const Tag = React.forwardRef<HTMLDivElement, Props>(
...
+            </div>
+          )}
+          {children}
+        </div>
         {onRemove && (
           <div
             className={cx(
               "orbit-tag-close-container",
-              "ms-200 flex rounded-full",
+              "pe-200 ps-100 rounded-r-150 flex items-center justify-center",
```
sarkaaa, line 108: I added `ps-100` to make space between outline and close icon when it's focused -> visual reasons. Same for "title" part of the tag, when `removable=true`.

`ps-0`
Snímek obrazovky 2025-03-07 v 19 24 43


`ps-100`
Snímek obrazovky 2025-03-07 v 19 24 21

DSil: It is very weird that the user is able to click outside of the icon and trigger the onRemove. Can you please check how other design systems approach this? 😕

sarkaaa: My first idea was to have "clickable" only the `X` icon, but then I was thinking about it, and it could be difficult to click that area on mobile phones. also, it can easily happen that the user will click `onClick` function instead of `onRemove`, because there won't be enough space. 

 I've found some articles about "minimum" target size:
- https://dequeuniversity.com/rules/axe/4.9/target-size
- https://tetralogical.com/blog/2022/12/20/foundations-target-size/

Here is an example where the "dismiss" button is 100% height of the tag. The design is quite different, but it correspond the minimum 24px size according the accessibility rules.
https://carbondesignsystem.com/components/tag/code/

packages/orbit-components/src/Tag/Tag.ct.tsx-snapshots/visual-Tag-screenshot-Colored-DateTag-Selected-focus-1-Desktop-darwin.png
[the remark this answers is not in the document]
DSil: Yes, please, we should be consistent on these settings if we are keeping darwin screenshots on the repo

packages/orbit-components/src/Tag/index.tsx
DSil: Alright, I am sold. But I would **loudly** present this to designers so they can think of a solution

sarkaaa: I'll write a note to the _accessibility compliance_ doc.

Also, I can see that `onRemove` and `onClick` props are used in search - check [sourcebot](https://sourcebot.skypicker.com/search?query=orbit-components%2Flib%2FTag%20repo%3Afrontend%2F*%20-repo%3A%5Egithub%5C.com%2Fkiwicom%2Forbit%24), however, they put the same function in both props. so the functionality is the same for both props, they just wanted to have a cross button in the Tag component (from what I understand).

https://sourcebot.skypicker.com/search?query=orbit-components%2Flib%2FTag%20repo%3Afrontend%2F*%20-repo%3A%5Egithub%5C.com%2Fkiwicom%2Forbit%24

@DSil

packages/orbit-components/src/Tag/index.tsx
```diff
@@ -80,35 +79,80 @@ const Tag = React.forwardRef<HTMLDivElement, Props>(
...
+            ],
+          )}
+          id={id}
+          ref={ref}
+          {...((onClick || onRemove) && {
+            role: "button",
+            tabIndex: 0,
+            onClick,
+            onKeyDown: ev => buttonClickEmulation(ev, onClick),
+          })}
```
DSil, line 114: This silences the linter because it can't analyze this part of the code 😄

But actually… looking at this, I noticed a problem with our current implementation.

We are saying that any tag with `onRemove` is rendered as a button. But we clearly have the dismiss icon separated from the "content", in two separate elements…

So why is a Tag with `onRemove` but without `onClick` focusable, clickable, and all, if only the dismiss part is clickable?

You can see this on the default story, the Tag that says "removable". I don't think it is working correctly.

DSil (changes requested):
I noticed an issue with how we are structuring the component in different elements and I think it can be improved. Take a look at it, please

packages/orbit-components/src/Tag/index.tsx
DSil: > they put the same function in both props. so the functionality is the same for both props, they just wanted to have a cross button in the Tag component

Yes, but that is not an assumption we can make for all the cases

packages/orbit-components/src/Tag/index.tsx
sarkaaa: aha! good point.

sarkaaa: Fixed, I also did TW classNames cleanup. Regarding these "spread attributes", I was inspired by changes in `SectionHeader` file. @DSil

DSil (approved):

PR #4637, opened 2025-02-28
ListChoice Accessibility

DSil:
Added a new internal (for now) component: FakeCheckbox. It allows to use the Checkbox in scenarios like ListChoice. We can consider making it public…

Tweaked the ARIA attributes logic and fixed stories.

Accessibility documentation will be done on a separate task.

FEPLT-2261

### DIFF
--- a/packages/orbit-components/src/ListChoice/ListChoice.stories.tsx
+++ b/packages/orbit-components/src/ListChoice/ListChoice.stories.tsx
@@ -50,7 +50,7 @@ export const MultipleChoices: Story = {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
           Array(...Array(3)).map((_, idx) => (
             // eslint-disable-next-line react/no-array-index-key
-            <ListChoice key={idx} {...args} icon={<Icon />} />
+            <ListChoice key={idx} {...args} icon={<Icon ariaHidden />} />
           ))
         }
       </>
@@ -64,7 +64,6 @@ export const MultipleChoices: Story = {
     selected: false,
     disabled: false,
     tabIndex: 0,
-    role: "list",
     onClick: action("onClick"),
   },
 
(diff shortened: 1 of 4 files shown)


sarkaaa:
image

Checkbox has lightblue border on hover (production doesn't include this visual effect).

packages/orbit-components/src/ListChoice/ListChoice.stories.tsx
```diff
@@ -92,11 +91,11 @@ export const WithAction: Story = {
     return (
       <ListChoice
```
sarkaaa, line 92: `Default` story:
image

DSil:
Production **_has_** the same light blue border on hover 🤔

packages/orbit-components/src/ListChoice/index.tsx
```diff
@@ -22,15 +22,11 @@ const ListChoice = React.forwardRef<HTMLDivElement, Props>(
       role,
       onClick,
       tabIndex = 0,
-      selected,
+      selected = false,
```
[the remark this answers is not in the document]
DSil: The component was controlled all the time

packages/orbit-components/src/ListChoice/ListChoice.stories.tsx
```diff
@@ -64,7 +64,6 @@ export const MultipleChoices: Story = {
     selected: false,
     disabled: false,
     tabIndex: 0,
```
sarkaaa, line 66: just small nitpick (and out of your PR), could you set the range for `tabIndex` to have the possible values clear to users?

something like this:
```
argTypes: { tabIndex: { control: { type: 'number', min:-1, max:0 }}
```

DSil: To me, tabIndex shouldn't even be in the controls… Storybook controls should be more about of what we see. But won't reinvent this now. However, tabIndex can have any number users want. Although it is recommended to only use `-1` or `0`, there is no actual limitation to other numbers, so I don't feel our storybook should be enforcing something that isn't actually enforced

sarkaaa: ok, my idea was to align the information according the "best practices" of the `tabIndex`.

sarkaaa (approved):

PR #4663, opened 2025-03-10
dataTest was not applied on Drawer

DSil:
The dataTest was being applied on the overlay area. In the past, the overlay area was a wrapper for the drawer. But now, it is not.

It is more relevant to have the dataTest applied on the drawer itself. So we change this here.


sarkaaa (approved):
very nitpicking nitpick (out of your changes) 😅  - could you add empty lines between `it` blocks in the test file?
Otherwise, LGTM.

DSil:
That can be done on a different moment 😄

PR #4666, opened 2025-03-12
Wizard: hide step number from screen readers

DSil:
Small accessibility improvement for when using the component with screen readers.

Accessibility documentation will be done on a separate task.

FEPLT-2266


packages/orbit-components/src/Wizard/Wizard.stories.tsx
```diff
@@ -88,7 +87,7 @@ export const Playground: Story = {
   render: ({ activeStep, completedSteps, labelProgress, steps, ...args }) => (
     <Wizard
       {...args}
-      labelProgress={`${completedSteps} ${labelProgress} ${steps.length}`}
+      labelProgress={`${completedSteps + 1} ${labelProgress} ${steps.length}`}
```
mvidalgarcia, line 90: I can get to this situation, is it expected?

image

DSil: 🤔 I don't think so. I'll fix

mvidalgarcia:
I wonder whether we should announce the completed steps somehow. For instance, in the video, both Kiwi.com Guarantee and Seating have the same announcement, but one is completed and the other is not.

Do you know if there's some aria attribute to mark as completed?

https://github.com/user-attachments/assets/06a18140-61ec-4a4b-a300-5bc6b76590e8

packages/orbit-components/src/Wizard/Wizard.stories.tsx
DSil: It's weird because what does "X of 7" means? Does it mean I'm on step X? Or that I completed X steps (and so I'm on step X+1)? It's weird… I guess I won't change anything after all

DSil:
@mvidalgarcia It is basically the same issue that with the Timeline component… I don't think we have an answer for that right now. We can (and should) improve it on Phase 3

mvidalgarcia:
> @mvidalgarcia It is basically the same issue that with the Timeline component… I don't think we have an answer for that right now. We can (and should) improve it on Phase 3

Sure, I just wanted to raise awareness, not blocking.

mvidalgarcia (approved):

PR #4671, opened 2025-03-14
feat(BadgeList): add aria labels to list and icons

mvidalgarcia:
closes FEPLT-2257

### DIFF
--- a/packages/orbit-components/src/BadgeList/index.tsx
+++ b/packages/orbit-components/src/BadgeList/index.tsx
@@ -4,9 +4,9 @@ import * as React from "react";
 
 import type { Props } from "./types";
 
-const BadgeList = ({ children, dataTest, id }: Props) => {
+const BadgeList = ({ children, dataTest, id, ariaLabel }: Props) => {
   return (
-    <ul className="m-0 flex flex-col p-0" data-test={dataTest} id={id}>
+    <ul className="m-0 flex flex-col p-0" data-test={dataTest} id={id} aria-label={ariaLabel}>
       {children}
     </ul>
   );
--- a/packages/orbit-components/src/BadgeList/BadgeList.stories.tsx
+++ b/packages/orbit-components/src/BadgeList/BadgeList.stories.tsx
@@ -89,13 +89,14 @@ export const Sizes: Story = {
 };
 
 export const Playground: Story = {
-  render: ({ type, size, strikeThrough }) => (
-    <BadgeList>
+  render: ({ type, size, strikeThrough, ariaLabel, iconLabel }) => (
+    <BadgeList ariaLabel={ariaLabel}>
       <BadgeListItem
         icon={<Icons.AlertCircle />}
         type={type}
         strikeThrough={strikeThrough}
         size={size}
+        iconLabel={iconLabel}
       >
         You&apos;re departing from a different place
       </BadgeListItem>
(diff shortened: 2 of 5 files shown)


DSil (changes requested):
As commented on the JIRA task, we agreed to add a spacing prop to this component, like we did on the List component.
This is to avoid the necessity to have wrapper components to add gaps, which might result in violation.

packages/orbit-components/src/BadgeList/index.tsx
```diff
@@ -1,12 +1,22 @@
...
+import { getSpacingClasses } from "../common/tailwind";
 
-const BadgeList = ({ children, dataTest, id, ariaLabel }: Props) => {
+const BadgeList = ({ children, dataTest, id, ariaLabel, spacing }: Props) => {
   return (
-    <ul className="m-0 flex flex-col p-0" data-test={dataTest} id={id} aria-label={ariaLabel}>
+    <ul
+      className={cx(
+        "orbit-badge-list m-0 flex flex-col p-0",
+        spacing && getSpacingClasses(spacing, undefined, "column", false),
```
mvidalgarcia, line 14: Not sure if this is correct or if getSpacingClasses should receive different args

packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
```diff
@@ -799,22 +800,20 @@ export const MultipleBanners: Story = {
...
-                          You’ll depart from a different place in New York: John F. Kennedy
-                          International.
-                        </ItineraryBadgeListItem>
-                        <ItineraryBadgeListItem icon={<Accommodation color="secondary" />}>
-                          We won’t cover your overnight stay. Hotel coverage is only available if
-                          the disruption happens during the trip. If you want to avoid extra hotel
-                          costs please choose a different alternative or a refund.
-                        </ItineraryBadgeListItem>
-                      </Stack>
+                    <ItineraryBadgeList spacing={SPACINGS.ONE_HUNDRED}>
```
DSil, line 803: Interesting that it works nice! Could you please add an example also at "Badge List"? With a control maybe. WDYT?

At the same time, I don't know if it makes sense to document this behavior here… I would assume designs are consistent regarding the spacing between items inside an Itinerary 💭 

So I'd leave it up to you, really

DSil (approved):
Only small non-blocking questions. Approving. Feel free to consider or not the suggestions

packages/orbit-components/src/BadgeList/README.md
```diff
@@ -18,21 +18,25 @@ After adding import into your project you can use it simply like:
...
-| :------- | :----------- | :------ | :---------------------------------- |
-| children | `React.Node` |         | The content of the BadgeList.       |
-| dataTest | `string`     |         | Optional prop for testing purposes. |
-| id       | `string`     |         | Set `id` for `BadgeList`            |
+| Name      | Type                  | Default | Description                                                |
+| :-------- | :-------------------- | :------ | :--------------------------------------------------------- |
+| children  | `React.Node`          |         | The content of the BadgeList.                              |
+| dataTest  | `string`              |         | Optional prop for testing purposes.                        |
+| id        | `string`              |         | Set `id` for `BadgeList`                                   |
+| ariaLabel | `string`              |         | Adds aria-label to the BadgeList for better accessibility. |
```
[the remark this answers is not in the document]
DSil: Can be, yes. But again, that probably makes more sense in the accessibility documentation. Either way, it's fine

packages/orbit-components/src/Itinerary/Itinerary.stories.tsx
mvidalgarcia: Added! 👍

DSil (approved):

PR #4686, opened 2025-03-26
NavigationBar accessibility

domihustinova:
We currently don't have any `header` element on the page. Based on my research, I believe the NavigationBar component should have `header` tag and children rendered inside this component should be `nav` elements (when appropriate). As there shouldn't be more than one element on the page with `role="banner"` (role of `header` element), I believe `ariaLabel` prop is redundant here and should be set on the children `nav` elements so I removed it.

As `Stack` component supports `as` prop that can be used to convert current `div` elements to `nav` elements, I added `ariaLabel` prop there so it can ease the migration.

I created [new task](https://kiwicom.atlassian.net/browse/FEPLT-2396) to refactor Nitro's TopNav to use `nav` elements and add SkipNavigation.

Closes https://kiwicom.atlassian.net/browse/FEPLT-2298

<details open>
<summary>:sparkles: <i><h3>Description by Callstackai</h3></i></summary>

This PR updates the `NavigationBar` component to use a `header` element instead of a `nav` element, removes the `ariaLabel` prop, and adds the `aria-label` attribute to the `nav` elements passed as children. It also updates the `Box` and `Stack` components to support the `ariaLabel` prop for accessibility.

<details>
<summary><b>Diagrams of code changes</b></summary>

```mermaid
sequenceDiagram
    participant User
    participant NavigationBar
    participant Box
    participant Stack

    Note over NavigationBar: Changed from nav to header element
    User->>NavigationBar: Interact with component
    NavigationBar-->>User: Renders as header element
    Note over NavigationBar: Removed ariaLabel prop

    User->>Box: Use Box component
    Box-->>User: Render with new ariaLabel support
    Note over Box: Added ariaLabel prop

    User->>Stack: Use Stack component
    Stack-->>User: Render with new ariaLabel support
    Note over Stack: Added ariaLabel prop
```

</details>

<details>
<summary><b>Files Changed</b></summary>
<table>
<tr><th>File</th><th>Summary</th></tr>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4686/files#diff-ecefe6caf96517f49615628dd191b37ea0860205a0ad7e08c562a3ad3eb8f48a>docs/src/documentation/05-development/04-migration-guides/01-v23.mdx</a></td><td>Added a section about the removal of the <code>ariaLabel</code> prop in the <code>NavigationBar</code> component.</td></tr>
[...]

PR #4690, opened 2025-03-31
CountryFlag Accessibility

DSil:
The role is no longer `img` by default.
The alt is now an open prop with no fallbacks and `""` as default value. The title attribute was removed as it was redundant

The accessibility documentation will be done on a separate task.

FEPLT-2282


packages/orbit-components/src/CountryFlag/index.tsx
```diff
@@ -6,19 +6,18 @@ import cx from "clsx";
...
 
   const countryCode = countryCodeExists ? CODES[codeNormalized] : CODES.UNDEFINED;
-  const countryName = countryCode === CODES.UNDEFINED && !name ? "Undefined" : name;
-  return { code: countryCode, name: countryName };
+  return { code: countryCode };
 }
 
-const CountryFlag = ({ dataTest, size = SIZES.MEDIUM, id, role = "img", ...props }: Props) => {
-  const { code, name } = getCountryProps(props.code, props.name);
+const CountryFlag = ({ dataTest, size = SIZES.MEDIUM, id, role, name, ...props }: Props) => {
```
sarkaaa, line 19: why did you decide to set the value of `alt` in a `name` prop? it looks that `name` is used only on line 40 as a value of `alt` attribute. why didn't you keep it as a prop called `alt`?

packages/orbit-components/src/CountryFlag/README.md
```diff
@@ -16,18 +16,18 @@ After adding import into your project you can use it simply like:
...
-| role     | `React.HTMLAttributes<HTMLImageElement>["role"]` | `"img"`       | ARIA role for the flag image.        |
-| size     | [`enum`](#enum)                                  | `"medium"`    | The size of the CountryFlag.         |
+| Name     | Type                                             | Default       | Description                                |
+| :------- | :----------------------------------------------- | :------------ | :----------------------------------------- |
+| code     | [`enum`](#enum)                                  | `"undefined"` | Code for the displayed country flag.       |
+| dataTest | `string`                                         |               | Optional prop for testing purposes.        |
+| id       | `string`                                         |               | Set `id` for `CountryFlag`                 |
+| name     | `string`                                         | `""`          | Defines the `alt` attribute for the image. |
+| role     | `React.HTMLAttributes<HTMLImageElement>["role"]` |               | ARIA role for the flag image.              |
```
sarkaaa, line 25: Is there any reason why we don't set the `presentation` role as a default value? these flags are usually used as a "decorator", aren't they? 🤔

DSil: Having `name` with default value `""` is enough. Applying `role="presentation"` is redundant

packages/orbit-components/src/CountryFlag/index.tsx
DSil: Because I did not want to introduce a breaking change

packages/orbit-components/src/CountryFlag/README.md
sarkaaa: I know, but we did the same approach in the illustrations components. So I thought we want to keep same modus operandi everywhere. :D

packages/orbit-components/src/CountryFlag/index.tsx
sarkaaa: why not? we're going to have BC anyways. :D 

ie:
- https://github.com/kiwicom/orbit/pull/4686/commits/18d5e4fef119fed48cd967dcc22f4077f01bf6b6
- https://github.com/kiwicom/orbit/pull/4691/commits/9d1611e6bbd522887cec91df8304d35c1f6e990f

DSil: Because we already have a prop called "name" that was being applied to `alt` property… There's no point in complicating

packages/orbit-components/src/CountryFlag/README.md
DSil: According to [this](https://kb.daisy.org/publishing/docs/html/images-deco.html#:~:text=Authors%20may%20use%20the%20role%20presentation%20in%20place%20of%20an%20empty%20alt%20attribute%2C%20but%20this%20method%20is%20not%20well%20supported.), role is not as supported as empty alt. It's ok to have any approach, really.

packages/orbit-components/src/CountryFlag/README.md
```diff
@@ -16,18 +16,18 @@ After adding import into your project you can use it simply like:
...
-| name     | `string`                                         |               | The name for the flag.               |
-| role     | `React.HTMLAttributes<HTMLImageElement>["role"]` | `"img"`       | ARIA role for the flag image.        |
-| size     | [`enum`](#enum)                                  | `"medium"`    | The size of the CountryFlag.         |
+| Name     | Type                                             | Default       | Description                                |
+| :------- | :----------------------------------------------- | :------------ | :----------------------------------------- |
+| code     | [`enum`](#enum)                                  | `"undefined"` | Code for the displayed country flag.       |
+| dataTest | `string`                                         |               | Optional prop for testing purposes.        |
+| id       | `string`                                         |               | Set `id` for `CountryFlag`.                |
+| name     | `string`                                         | `""`          | Defines the `alt` attribute for the image. |
```
sarkaaa, line 24: you wrote that `""` is a default value for for `name`, but there is no default value for `name` set in `packages/orbit-components/src/CountryFlag/index.tsx`

packages/orbit-components/src/CountryFlag/CountryFlag.stories.tsx
```diff
@@ -32,10 +31,9 @@ const meta: Meta<typeof CountryFlag> = {
         type: "select",
       },
     },
-    role: {
-      options: ["img", "presentation", "none"],
+    name: {
```
sarkaaa, line 34: Why did you remove `role` field from SB? I understand there's no visual change for the users - just DOM change, but we included other "no-visible" fields in other stories.

DSil: To incentivize using name more than role because empty name is more recommendable than using role

sarkaaa (approved):

PR #4700, opened 2025-04-07
fix(Button): remove invalid className prop type

Ajuska:
This PR removes invalid className prop type on the button component. It's overwritten in the implementation and having it in types was misleading to users as the styles were never applied.

FEPLT-2331

<details open>

This PR removes the invalid `className` prop type from the `ButtonProps` interface, which was misleading as it was overwritten in the implementation.

<details>

```mermaid
sequenceDiagram
    participant ButtonPrimitive
    participant ButtonProps
    participant Types

    ButtonPrimitive->>Types: Exports Size type
    ButtonPrimitive->>ButtonProps: Extends ButtonCommonProps
    Note over ButtonProps: Omits className from ButtonCommonProps
    ButtonProps->>Types: Defines Type & Size properties
    Note over ButtonProps: Creates interface with<br/>type?: Type<br/>size?: Size
```

</details>

<details>
<table>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4700/files#diff-b99d0dc786f2a1199c8d70155b9a40c5f674e2368864696834bc3479a3e318e1>packages/orbit-components/src/Button/types.d.ts</a></td><td>Updated <code>ButtonProps</code> to omit the <code>className</code> prop type.</td></tr>

</table>
</details>

</details>

### DIFF
--- a/packages/orbit-components/src/Button/types.d.ts
+++ b/packages/orbit-components/src/Button/types.d.ts
@@ -3,9 +3,9 @@
 
 import type {
   ButtonCommonProps,
-  Size,
   DownloadWithHrefConditionalProps,
   FullWidthConditionalProps,
+  Size,
 } from "../primitives/ButtonPrimitive/types";
 
 export type Type =
@@ -20,8 +20,9 @@ export type Type =
   | "bundleTop";
 
 export type ButtonStates = "default" | "hover" | "active" | "focus";
+type OmittedButtonCommonProps = Omit<ButtonCommonProps, "className">;
 
-interface ButtonProps extends ButtonCommonProps {
+interface ButtonProps extends OmittedButtonCommonProps {
   readonly type?: Type;
   readonly size?: Size;
 }


packages/orbit-components/src/Button/types.d.ts
```diff
@@ -3,9 +3,9 @@
 
 import type {
   ButtonCommonProps,
-  Size,
   DownloadWithHrefConditionalProps,
   FullWidthConditionalProps,
+  Size,
```
DSil, line 8: 😅

DSil (approved):
Good work! Welcome to Orbit 💪 🚀

packages/orbit-components/src/Button/types.d.ts
Ajuska: this is a plugin that automatically reorganize imports alphabetically on save 💪  does not hurt to have

DSil: Agreed. Just concerned because not the whole team has it, so we might face these diffs in future PRs, which might increase entropy. But doesn't harm for now.

PR #4714, opened 2025-04-15
Remove role button from Tooltip on desktop + Stop supporting inner click

DSil:
This was causing many nested interactive elements violations unnecessarily.

It also does not make sense to wrap something that is not actually clickable. Therefore, the role="button" is kept only for Mobile.

The support for managing clicks inside the content is also removed, since it is no longer supported.


packages/orbit-components/src/Tooltip/index.tsx
Ajuska, line 1: Only thing I noticed is when using screenreader, the tooltip description is not announced right away as with `role="button"`. Is that ok?

https://github.com/user-attachments/assets/068388ff-3dda-4bfe-bf22-768b5cc43de8

Ajuska (commented):
just one comment and one question, otherwise looks good 🎉

packages/orbit-components/src/Tooltip/index.tsx
Ajuska: I think it's better to maintain correct semantics and work with the limitations than to use incorrect semantics just to achieve consistent behavior? The inconsistency in screen reader behavior is a known issue with aria-describedby on non-interactive element so maybe we could just highlight that It in the component documentation?

DSil:
Closing for now until we handle it later

PR #4746, opened 2025-05-07
Switch type definitions

DSil:
From now on, we no longer define types in .d.ts files. Those are to be generated automatically.
We keep a file for types definition, but it's now called only types.ts

All the build scripts were updated to reflect the change.

Some type files had unused imports / declarations that have always been… undetected. They were now detected, so those unused imports / declarations were removed.

This was tested using `yalc` in nitro.


packages/orbit-components/config/build/generateDeclarations.mts
```diff
@@ -3,11 +3,20 @@ import { $, chalk } from "zx";
...
   await $`tsc --p tsconfig-build.json`;
   await $`tsc --p tsconfig-build.json --rootDir src --outDir es --declaration --emitDeclarationOnly --moduleResolution node`;
+
+  // Clean up test-related declaration files
+  console.log(chalk.greenBright.bold("Cleaning up test-related declaration files..."));
+  await $`del "lib/**/*.ct-story.d.ts" "lib/**/*.ct-story.d.ts.map" "es/**/*.ct-story.d.ts" "es/**/*.ct-story.d.ts.map" "lib/**/*.ct.d.ts" "lib/**/*.ct.d.ts.map" "es/**/*.ct.d.ts" "es/**/*.ct.d.ts.map"`;
+
+  // Clean up generated types.js files
+  console.log(chalk.greenBright.bold("Cleaning up types.js files..."));
+  await $`del "lib/**/types.js" "es/**/types.js"`;
```
mvidalgarcia, line 21: We don't need to do this if we add `"**/types.ts",` to [COMPILE_IGNORE_PATTERNS](https://github.com/kiwicom/orbit/blob/dsil/switch-type-definition/packages/orbit-components/config/build/consts.mts#L12). Same for the `buildSize.mts` script.

mvidalgarcia (approved):

packages/orbit-components/config/build/generateDeclarations.mts
DSil: Good catch. Changed

PR #4788, opened 2025-06-13
chore(docs): extract /docs from workspace to allow different React ve…

RobinCsl:
…rsions

As we want to adopt React 19 in orbit-components, we need to adjust orbit docs to keep using React 18.

Closes https://kiwicom.atlassian.net/browse/FEPLT-2770

### For new components:


RobinCsl:
I adapted a bunch of changes from #4786, while trying to minimize amount of changes. (We could remove themer and other functionality to check links in a separate spike).

docs/.eslintignore
```diff
@@ -0,0 +1 @@
+src/documentation/06-kiwi-use/02-content/02-grammar-and-mechanics.mdx
```
RobinCsl, line 1: This file somehow always errors out our lint job. It seems counter-productive at this point to try and fix the linting setup just for this file.

docs/gatsby-node.ts
```diff
@@ -289,3 +289,12 @@ export const createSchemaCustomization: GatsbyNode["createSchemaCustomization"]
 
   actions.createTypes(typeDefs);
 };
+
+export const onCreateWebpackConfig: GatsbyNode["onCreateWebpackConfig"] = ({ actions }) => {
+  actions.setWebpackConfig({
+    resolve: {
+      // this is necessary for the symlinked packages to be resolved correctly
+      symlinks: false,
```
RobinCsl, line 297: Because of this, we lose Hot Module Replacement, meaning that changing a README.md file in the `@kiwicom/orbit-components` package won't retrigger a build. We need to kill the Gatsby dev server and restart it to see the changes.

I tried a bunch of other things in Gatsby config files to ensure Gatsby MDX would bust its cache when detecting changes in these symlinked files, but it never updated the content even though the server somehow reloaded.

We need proper rendering of the README files, HMR is more a nice-to-have on this docs website IMO.

But that's a tradeoff to be aware of.

docs/package.json
```diff
@@ -23,19 +24,35 @@
...
-    "@kiwicom/babel-plugin-orbit-components": "*",
-    "@kiwicom/orbit-components": "*",
-    "@kiwicom/orbit-design-tokens": "*",
-    "@kiwicom/orbit-tailwind-preset": "*",
-    "@kiwicom/orbit-themer": "*",
+    "@kiwicom/babel-plugin-orbit-components": "link:../packages/babel-plugin-orbit-components",
+    "@kiwicom/orbit-components": "link:../packages/orbit-components",
+    "@kiwicom/orbit-design-tokens": "link:../packages/orbit-design-tokens",
+    "@kiwicom/orbit-tailwind-preset": "link:../packages/orbit-tailwind-preset",
+    "@kiwicom/orbit-themer": "link:../packages/orbit-themer",
```
DSil, line 52: TIL that this works for deployments 🤔 did you do any setup on Cloudflare? Or the link just works seamlessly?

DSil (approved):
LGTM 👍 Seems to work fine. Just one minor 💅 and one question for my own understanding. Nice!

docs/.eslintignore
RobinCsl: Yes! Done

RobinCsl:
Thanks for the review @DSil! Applied the change you requested :)

PR #4807, opened 2025-07-03
chore(docs): fix dependencies for changelog generation

Ajuska:

### For new components:


mvidalgarcia (approved):

docs/yarn.lock
```diff
@@ -1545,10 +1545,25 @@
   version "0.0.0"
   uid ""
 
+"@kiwicom/orbit-design-tokens@^10.1.0":
```
DSil, line 1548: These should not be here. Same for tailwnd-preset

Ajuska: this gets added automatically when you run yarn install in /docs folder

DSil: Yes, it needs to be executed with the `--frozen-lockfile` flag, to ensure the link for these two packages are kept to the local implementations

PR #4808, opened 2025-07-04
React 19

DSil:
Orbit now supports and includes some React 19 features.

Some adjustments in docs were made, as well as other dependencies, all to improve compatibility with the new React version.

The feature [<Context> as a Provider](https://react.dev/blog/2024/12/05/react-19#context-as-a-provider) causes the docs build to fail, so we can't take advantage of it in Orbit for now.

FEPLT-2471


packages/orbit-components/package.json
```diff
@@ -73,8 +73,8 @@
   },
   "peerDependencies": {
     "@babel/runtime": "^7.21.0",
-    "react": ">=18.3.0",
-    "react-dom": ">=18.3.0"
+    "react": ">=19.1.0",
+    "react-dom": ">=19.1.0"
```
mvidalgarcia, line 77: Does it need to be 19.1? Not a strong opinion, just curious because in frontend/core we set ^19.0.0

packages/orbit-components/src/ButtonLink/types.ts
```diff
@@ -14,8 +13,6 @@ export type Props = {
   readonly compact?: boolean;
   readonly type?: Type;
   readonly size?: Size;
-} & Common.Globals &
-  Common.SpaceAfter &
-  ButtonCommonProps &
+} & ButtonCommonProps &
```
mvidalgarcia, line 16: good catch 🔍

docs/src/components/Navbar/index.tsx
```diff
@@ -98,6 +98,11 @@ interface Props {
   docNavigation?: React.ReactNode;
 }
 
+const PortalWrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+  // @ts-expect-error
+  <Portal>{children}</Portal>
+);
+
```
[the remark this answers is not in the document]
DSil: Yes, I did the abstraction but then forgot to remove from here. Will change

packages/orbit-components/package.json
DSil: I don't think so, no. Will change

packages/orbit-components/src/hooks/useIntersect/__tests__/index.test.tsx
```diff
@@ -17,7 +17,7 @@ const LazyImage = ({ src, placeholder, threshold }) => {
     }
   }, [src, entry]);
 
-  return <img alt="" src={source} ref={ref} />;
+  return <img alt="alt" src={source} ref={ref} />;
```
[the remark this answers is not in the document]
DSil: I honestly don't know… But it was not able to find image by role 🤷‍♂️

packages/orbit-components/src/ButtonLink/index.tsx
```diff
@@ -44,51 +44,45 @@ const typeCompactDisabledStyles: Record<Type, string> = {
...
+  disabled,
+  children,
+  iconLeft,
+  iconRight,
+  ref,
+  ...props
+}: Props) => {
+  return (
+    <ButtonPrimitive
+      ref={ref}
```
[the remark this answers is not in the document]
DSil: No, there shouldn't be any difference. I guess it was just during refactor it happened to stay like this. I don't have a strong preference, but I can keep it consistent, one way or another, if you prefer

packages/orbit-components/package.json
```diff
@@ -73,8 +73,8 @@
   },
   "peerDependencies": {
     "@babel/runtime": "^7.21.0",
-    "react": ">=19.1.0",
-    "react-dom": ">=19.1.0"
+    "react": ">=19.0.0",
```
sarkaaa, line 76: Why is there `>=` and not `^`? I see the version was `greater than or equal` before, is there any reason why we don't want to set the caret char here (and also for `react-dom` peer dep)?

DSil: Yes, the Booking example is a good one. We allow for further majors. They usually support older versions as well. So we don't block teams from being on the latest versions.

sarkaaa: ok, that makes sense in that case. 👍

PR #4810, opened 2025-07-07
fix(Stack): incorrectly rendered as flex when ariaLabel was provided

domihustinova:
Closes https://kiwicom.atlassian.net/browse/FEPLT-2907

<details open>

This PR modifies the `shouldUseFlex` function to include `ariaLabel` in the list of properties that determine whether to use flex layout.

<details>

```mermaid
sequenceDiagram
    participant Stack
    participant shouldUseFlex
    participant PropertyValidator

    Stack->>shouldUseFlex: Check if flex should be used
    shouldUseFlex->>PropertyValidator: Validate properties
    Note over PropertyValidator: Check against allowed props:<br/>spacing, spaceAfter, dataTest,<br/>children, ariaLabel
    PropertyValidator-->>shouldUseFlex: Return validation result
    shouldUseFlex-->>Stack: Return flex decision
```

</details>

<details>
<table>
<tr><td><a href=https://github.com/kiwicom/orbit/pull/4810/files#diff-30c9103b612af0d457bcc391892bf0e96e77003fb02f1f0df204308229b934e0>packages/orbit-components/src/Stack/index.tsx</a></td><td>Updated the <code>shouldUseFlex</code> function to account for <code>ariaLabel</code> in the props check.</td></tr>

</table>
</details>

</details>

### DIFF
--- a/packages/orbit-components/src/Stack/index.tsx
+++ b/packages/orbit-components/src/Stack/index.tsx
@@ -26,7 +26,7 @@ import { JUSTIFY } from "../common/tailwind/justify";
 const shouldUseFlex = (props: CommonProps & Common.SpaceAfter) =>
   props.flex ||
   Object.keys(props)
-    .map(prop => ["spacing", "spaceAfter", "dataTest", "children"].includes(prop))
+    .map(prop => ["spacing", "spaceAfter", "dataTest", "children", "ariaLabel"].includes(prop))
     .includes(false);
 
 // use margins instead of gap to work with display:block


packages/orbit-components/src/Stack/index.tsx
```diff
@@ -26,7 +26,7 @@ import { JUSTIFY } from "../common/tailwind/justify";
 const shouldUseFlex = (props: CommonProps & Common.SpaceAfter) =>
   props.flex ||
   Object.keys(props)
-    .map(prop => ["spacing", "spaceAfter", "dataTest", "children"].includes(prop))
+    .map(prop => ["spacing", "spaceAfter", "dataTest", "children", "ariaLabel"].includes(prop))
```
DSil, line 29: This puts to evidence how faulty this logic is 😞 any prop we add from now on we have to remind ourselves to come here and add it, otherwise the Stack won't use flex. Besides that, if one uses the media queries props (`desktop`, `tablet`, etc…) it will also consider it should not use flex… I wonder if that's expected 😬 

Do you think you can reconsider the logic here? I wouldn't ask you to take too much time, but give it a chance 🙏 If you feel it's too convoluted, let it be, as this fixes the immediate bug.

DSil (approved):

packages/orbit-components/src/Stack/index.tsx
domihustinova: As agreed, let's proceed with this as it is and address the bigger issue in different task after consulting with the team.

